PR #15, opened 2023-12-01
Rearrange snapcraft.yaml to support remote-builds

addyess:
Begin to try to publish snaps using github action which triggers a remote build for the snap

This requires that each snap part's `source` be tarred locally by the gh builder, then checked into an ephemeral lp repo. The lp builders can use the tar.gz as the sources for each part when it builds the snap. Much of these changes move the
`components` out of the `build-scripts` directory so that they can be put into independent sources.

The publish to edge action is largely untested, and none of the appropriate secrets are yet defined

### DIFF
--- a/build-scripts/components/.gitignore
+++ b/build-scripts/components/.gitignore
@@ -1,2 +1,0 @@
-.build
-.install


.github/workflows/snap-store-publish-to-edge.yml
```diff
@@ -0,0 +1,251 @@
...
+              issue_number: ${{ needs.create_issue.outputs.issue_number }},
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: `The following screenshots were taken during automated testing:
+
+              ![window](${{ steps.screenshots.outputs.window }})
+
+              ![screen](${{ steps.screenshots.outputs.screen }})
+              `
+            })
```
neoaggelos, line 251: Do we need this?

neoaggelos (commented):
I'd keep the `components` directory under `build-scripts` to avoid clattering the main source dir. But this also works, so I'm happy that it unblocks remote builds, let's do it :+1:

Just one minor comment

.github/workflows/snap-store-publish-to-edge.yml
addyess: No, i've not tested any of the GH action logic.  There's likely quite a bit of fluff left to remove.  Hence the Draft PR

PR #31, opened 2023-12-12
Adding the network enablement support

berkayoz:
To get the CNI working:
```bash
sudo snap install ./k8s.snap --dangerous
sudo /snap/k8s/current/k8s/connect-interfaces.sh
sudo /snap/k8s/current/k8s/network-requirements.sh
sudo k8s init
sudo k8s enable network 
```

Having `kubernetes-support` interface on `k8sd` is not ideal, however we can not check existing paths without it. Although `mount-observe` feels like it should be enough.

Passing in dynamic values to deployment mechanism is also up for discussion.

### DIFF
--- a/src/k8s/pkg/component/hook.go
+++ b/src/k8s/pkg/component/hook.go
@@ -1,0 +1,93 @@
+package component
+
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/canonical/k8s/pkg/utils"
+)
+
+type valuesHook func() (map[string]any, error)
+
+var valuesHooks = map[string]valuesHook{
+	"network": networkValues,
+}
+
+func networkValues() (map[string]any, error) {
+	bpfMnt, err := utils.GetMountPath("bpf")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get bpf mount path: %w", err)
+	}
+
+	cgrMnt, err := utils.GetMountPath("cgroup2")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get cgroup2 mount path: %w", err)
+	}
+
+	// TODO: the cluster cidr should be configurable through a common interface
+	clusterCIDRStr, err := utils.GetServiceArgument("kube-proxy", "--cluster-cidr")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get cluster cidrs from kube-proxy arguments: %w", err)
+	}
+
+	clusterCIDRs := strings.Split(clusterCIDRStr, ",")
+	if v := len(clusterCIDRs); v != 1 && v != 2 {
+		return nil, fmt.Errorf("invalid kube-proxy --cluster-cidr value: %v", clusterCIDRs)
+	}
+
+	var (
+		ipv4CIDR string
+		ipv6CIDR string
+	)
+	for _, cidr := range clusterCIDRs {
+		_, parsed, err := net.ParseCIDR(cidr)
+		switch {
+		case err != nil:
+			return nil, fmt.Errorf("failed to parse cidr: %w", err)
+		case parsed.IP.To4() != nil:
+			ipv4CIDR = cidr
+		default:
+			ipv6CIDR = cidr
+		}
+	}
+
+	values := map[string]any{
+		"cni": map[string]any{
+			"confPath": "/etc/cni/net.d",
+			"binPath":  "/opt/cni/bin",
+		},
+		"daemon": map[string]any{
+			"runPath": utils.SnapCommonPath("var", "run", "cilium"),
+		},
+		"operator": map[string]any{
+			"replicas": 1,
+		},
+		"ipam": map[string]any{
+			"operator": map[string]any{
+				"clusterPoolIPv4PodCIDRList": ipv4CIDR,
+				"clusterPoolIPv6PodCIDRList": ipv6CIDR,
+			},
+		},
+		"nodePort": map[string]any{
+			"enabled": true,
+		},
+		"bpf": map[string]any{
+			"autoMount": map[string]any{
+				"enabled": false,
+			},
+			"root": bpfMnt,
+		},
+		"cgroup": map[string]any{
+			"autoMount": map[string]any{
+				"enabled": false,
+			},
+			"hostRoot": cgrMnt,
+		},
+		"l2announcements": map[string]any{
+			"enabled": true,
+		},
+	}
+
+	return values, nil
+}
--- a/tests/e2e/lxd-profile.yaml
+++ b/tests/e2e/lxd-profile.yaml
@@ -10,7 +10,19 @@ config:
   security.nesting: "true"
   security.privileged: "true"
 devices:
-  dev_kmsg:
+  aadisable:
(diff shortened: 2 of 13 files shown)


src/k8s/pkg/component/hook.go
```diff
@@ -0,0 +1,93 @@
...
+		_, parsedCidr, err := net.ParseCIDR(cidr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse cidr: %w", err)
+		}
+		if parsedCidr.IP.To4() != nil {
+			ipv4Cidr = cidr
+		} else {
+			ipv6Cidr = cidr
+		}
+	}
```
neoaggelos, line 53: Nice work, some nit styling suggestions:

- capitalize things like CIDR
- len(CIDR) is only used in the if statement check
- line 38, we are not wrapping the error (we have already checked it's nil), rather explain what is the problem
- for multiple if-else statements, prefer a switch instead

```suggestion
	clusterCIDRStr, err := utils.GetServiceArgument("kube-proxy", "--cluster-cidr")
	if err != nil {
		return nil, fmt.Errorf("failed to read kube-proxy --cluster-cidr argument: %w", err)
	}

	clusterCIDRs := strings.Split(clusterCIDRStr, ",")
	if v := len(clusterCIDRs); v != 1 && v != 2 {
		return nil, fmt.Errorf("invalid kube-proxy --cluster-cidr value: %v", clusterCIDRs)
	}

	var (
		ipv4CIDR string
		ipv6CIDR string
	)
	for _, cidr := range clusterCIDRs {
		_, parsed, err := net.ParseCIDR(cidr)
		switch {
		case err != nil:
			return nil, fmt.Errorf("failed to parse cidr: %w", err)
		case parsed.IP.To4() != nil:
			ipv4CIDR = cidr
		case parsed.IP.To6() != nil:
			ipv6CIDR = cidr
		}
	}
```

tests/e2e/lxd-profile.yaml
```diff
@@ -10,7 +10,19 @@ config:
...
     source: /dev/kmsg
     type: unix-char
+  aadisable3:
+    path: /sys/fs/bpf
+    source: /sys/fs/bpf
+    type: disk
+  aadisable4:
+    path: /proc/sys/net/netfilter/nf_conntrack_max
+    source: /proc/sys/net/netfilter/nf_conntrack_max
+    type: disk
```
neoaggelos, line 28: newline at EOF

neoaggelos (commented):
Almost there, just some cleanup and we're good to go! Nice work!

src/k8s/pkg/component/hook.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/canonical/k8s/pkg/snap"
+)
+
+type valuesHook func() (map[string]interface{}, error)
+
+var valuesHooks = map[string]valuesHook{
+	"network": networkValues,
+}
+
+func networkValues() (map[string]interface{}, error) {
```
bschimke95, line 18: here and in a lot of other places: Use `any` instead of `interface{}`

https://google.github.io/styleguide/go/decisions#use-any

src/k8s/pkg/k8s/utils/file.go
```diff
@@ -11,8 +11,22 @@ import (
...
 )
 
+// GetMountPath returns the first mountpath for a given filesystem type.
+func GetMountPath(fsType string) (string, error) {
+	mounts, err := mountinfo.GetMounts(mountinfo.FSTypeFilter(fsType))
+	if err != nil {
+		return "", fmt.Errorf("failed to find the mount info for %s: %w", fsType, err)
+	}
+	if len(mounts) == 0 {
+		return "", fmt.Errorf("failed to find the mount info for %s", fsType)
```
bschimke95, line 24: `fmt.Error("could not find any %s filesystem mount", fsType)`

seems more accurate.

tests/e2e/tests/e2e_util/util.py
```diff
@@ -57,11 +56,120 @@ def run_with_retry(
                     raise e
 
 
+def retry_until_condition(
```
bschimke95, line 59: seems like we start to need more advanced retry logic. 
It's probably time to move to something like https://github.com/jd/tenacity

bschimke95: no need to do this in this PR - just add a TODO so that we can create follow-up cards

tests/e2e/tests/e2e_util/util.py
```diff
@@ -57,11 +56,120 @@ def run_with_retry(
...
+        check=True,
+    )
+
+
 # Installs and setups the k8s snap on the given instance and connects the interfaces.
 def setup_k8s_snap(h: harness.Harness, instance_id: str, snap_path: Path):
     LOG.info("Install snap")
     h.send_file(instance_id, config.SNAP, snap_path)
     h.exec(instance_id, ["snap", "install", snap_path, "--dangerous"])
+    h.exec(instance_id, ["snap", "install", "kubectl", "--classic"])
```
bschimke95, line 172: We still ship the kubectl binary at `/snap/k8s/current/bin/kubectl`.
No need to install it via snap.

tests/e2e/tests/e2e_util/util.py
```diff
@@ -72,25 +180,12 @@ def wait_until_k8s_ready(h: harness.Harness, instance_id):
...
-        except subprocess.CalledProcessError:
-            time.sleep(5)
-
-    if not success:
-        pytest.fail("Kubelet node did not register")
+
+    result = retry_until_condition(
+        h,
+        instance_id,
+        ["k8s", "kubectl", "get", "node", hostname, "--no-headers"],
```
bschimke95, line 187: This will break.
We do not expose `k8s kubectl` right now.

tests/e2e/tests/test_network.py
```diff
@@ -0,0 +1,131 @@
...
+            "cilium-agent",
+            "--",
+            "cilium",
+            "status",
+            "--brief",
+        ],
+        capture_output=True,
+    )
+
+    assert p.stdout.decode().strip() == "OK"
```
bschimke95, line 72: @neoaggelos WDYT about changing the default behavior of `harness.exec` to return the decoded stdout? In most of the tests that we only want to check the output. 
We could still have something like a `decode=False` parameter in case we need to inspect the Command.

tests/e2e/tests/test_smoke.py
```diff
@@ -21,6 +21,7 @@ def test_smoke(h: harness.Harness, tmp_path: Path):
 
     util.setup_k8s_snap(h, instance_id, snap_path)
     h.exec(instance_id, ["k8s", "init"])
+    util.setup_network(h, instance_id)
 
     # TODO(bschimke): The node will not report ready as the CNI is not yet implemented in the k8s snap.
```
bschimke95, line 26: This is not true anymore.

As a side-note:
For more sophisticated tests we probably will need another method later. Something like:
```
util.setup_k8s(..., components=["network", "dns", ...])
```

bschimke95 (commented):
Looks good, some final comments

bschimke95 (approved):
LGTM, great work!

PR #43, opened 2024-01-03
Added local storage via hostpath

berkayoz:
Deploys local storage via hostpath.
Existing images for microk8s is used so the chart is configured to use the `microk8s.io/hostpath` provider tag.


tests/e2e/templates/storage-test.yaml
```diff
@@ -0,0 +1,53 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: myclaim
+spec:
+  accessModes:
+    - ReadWriteMany
+  volumeMode: Filesystem
```
neoaggelos, line 8: This is ignored atm, but let's keep it

tests/e2e/tests/test_storage.py
```diff
@@ -0,0 +1,146 @@
...
+    instance_id = h.new_instance()
+
+    util.setup_k8s_snap(h, instance_id, snap_path, wait_ready=False)
+
+    out = h.exec(
+        instance_id,
+        [K8S_BINARY_PATH, "enable", "storage"],
+        capture_output=True,
+    )
+    assert "enabled" in out.stderr.decode()
```
neoaggelos, line 32: can we have `check=True` here as well? or do we need something else?

tests/e2e/tests/test_storage.py
```diff
@@ -0,0 +1,146 @@
...
+        check=True,
+    )
+
+    LOG.info("Waiting for storage to get provisioned...")
+    util.retry_until_condition(
+        h,
+        instance_id,
+        ["k8s", "kubectl", "get", "pvc", "-o", "json"],
+        condition=lambda p: "myclaim" in p.stdout.decode()
+        and "Bound" in p.stdout.decode(),
```
neoaggelos, line 107: should we instead parse json and check the correct fields?

tests/e2e/tests/test_storage.py
```diff
@@ -0,0 +1,146 @@
...
+        check=True,
+    )
+
+    p = h.exec(
+        instance_id,
+        ["k8s", "kubectl", "logs", "storage-reader-pod"],
+        capture_output=True,
+    )
+    assert "test_date" in p.stdout.decode()
+    assert str(datetime.now().year) in p.stdout.decode()
```
neoaggelos, line 146: hmm, could we have a fixed text like "LOREM IPSUM" instead of the current date? should not be a problem, but would keep the logic here simpler, as I don't have to think too much looking at this and the pod template

neoaggelos (commented):
Looks great, just a couple minor nits

neoaggelos (approved):

PR #44, opened 2024-01-03
Refactor all the bespoke utils into one

addyess:
Various `utils` modules existed sprinkled throughout the code space.  This is a quick refactor of those utils all into `github.com/canonical/k8s/pkg/utils`

Before:
image

After:
image


src/k8s/pkg/utils/file.go
```diff
@@ -9,10 +9,20 @@ import (
...
+}
+
+func DataPath(parts ...string) string {
+	return filepath.Join(append([]string{os.Getenv("SNAP_DATA")}, parts...)...)
+}
+
+func CommonPath(parts ...string) string {
+	return filepath.Join(append([]string{os.Getenv("SNAP_COMMON")}, parts...)...)
+}
+
```
neoaggelos, line 25: Let's make these `SnapPath()`, `SnapDataPath()`, `SnapCommonPath()` instead.

On the calling sites, `util.SnapCommonPath()` would be a bit more helpful than `util.CommonPath()`, without having to lookup the implementation.

src/k8s/pkg/utils/file_test.go
```diff
@@ -0,0 +1,48 @@
...
+			common := fmt.Sprintf("common/%s", tc.expected_path)
+			if parsed := Path(tc.input_path...); !reflect.DeepEqual(parsed, simple) {
+				t.Fatalf("expected path to be %v but it was %v instead", simple, parsed)
+			}
+			if parsed := DataPath(tc.input_path...); !reflect.DeepEqual(parsed, data) {
+				t.Fatalf("expected data path to be %v but it was %v instead", data, parsed)
+			}
+			if parsed := CommonPath(tc.input_path...); !reflect.DeepEqual(parsed, common) {
+				t.Fatalf("expected common path to be %v but it was %v instead", common, parsed)
+			}
```
neoaggelos, line 45: Yup, I like this test. Note that we found it gets very tedious very quickly to write tests like this, can we bring in gomega instead?

See for example https://github.com/canonical/microk8s-cluster-agent/blob/main/pkg/k8sinit/launcher_test.go

The test would then like this:

```
g := NewWithT(t)
g.Expect(Path(tc.input_path...)).To(Equal(tc.expected_path))
g.Expect(DataPath(tc.input_path...)).To(Equal(fmt.Sprintf("/data/%s", tc.expected_path))
g.Expect(CommonPath(tc.input_path...)).To(Equal(fmt.Sprintf("common/%s", tc.expected_path))
```

This also gives us much nicer output and failure messages when something breaks.

neoaggelos (commented):
Looks great, nice work @addyess, left some comments.

neoaggelos (approved):

bschimke95 (approved):
LGTM

PR #49, opened 2024-01-08
Configure k8s services for joining control-plane nodes

bschimke95:
Configures the k8s service configurations appropriately when joining a node to a cluster.

The joining node will request the configuration from the master node by querying the `/k8sd/cluster/join` endpoint. The joining node will then apply the configuration overwrites on top of the arguments defined in the `k8s/args` directory, generate the certificates and sign them with the cluster CA.
Note that the control-plane nodes directly query the CA certkey pair from the `k8sd` database, worker-nodes (follow-up PR) will get the cert via the REST endpoint.

```
ubuntu@basic:~/k8s-snap$ sudo /snap/k8s/current/bin/kubectl --kubeconfig /var/snap/k8s/common/etc/kubernetes/admin.conf get nodes -A
NAME                STATUS   ROLES    AGE   VERSION
my-happy-node       Ready    <none>   16h   v1.29.0
node-mc-nodeface    Ready    <none>   65s   v1.29.0
``` 

Also, moves the snap implementation behind the `snap` interface to allow for proper testing in the future.


src/k8s/pkg/k8s/setup/kubeconfigs.go
```diff
@@ -58,11 +67,87 @@ func InitKubeconfigs(ctx context.Context, state *state.State, ca *cert.CertKeyPa
...
 		if err != nil {
 			return fmt.Errorf("failed to generate kubeconfig for %s: %w", config.username, err)
 		}
 	}
 
 	return nil
 }
+
+// GenerateX509Kubeconfig creates a kubeconfig file with the given x509 certificate and key data.
+func GenerateX509Kubeconfig(keyPem, certPem, caCertPem []byte, path string, overwrites *KubeConfigOverwriteOpts) error {
```
bschimke95, line 80: Moved this here because of cyclic dependencies.
Also, that methods are only used in this file, so it makes sense to have it contained in one place.

src/k8s/api/v1/cluster.go
```diff
@@ -14,6 +14,23 @@ type InitClusterRequest struct{}
...
 type InitClusterResponse struct{}
 
+// JoinClusterRequest is used to request the configuration to join the k8s cluster.
+type JoinClusterRequest struct {
+	Token string `json:"token"`
+}
+
+// JoinClusterResponse is the response for "POST 1.0/k8sd/cluster/join".
+type JoinClusterResponse struct {
+	ExtraServiceArgs ExtraServiceArgs `json:"service_argument_overwrites"`
```
neoaggelos, line 24: `extraServiceArgs` or `extra_service_args`, let's pick one and follow it everywhere.
```suggestion
	ExtraServiceArgs ExtraServiceArgs `json:"extraServiceArgs"`
```

neoaggelos: Personally, I am a fan of "named" arguments, e.g. `ExtraKubeletArgs`, `Extra$$$$Args`, etc, as it sets explicit expectations about field names and supported values, even though it leads to duplication.

I am happy to keep this if you consider too much work to change it.

src/k8s/pkg/component/hook.go
```diff
@@ -26,11 +27,7 @@ func networkValues() (map[string]any, error) {
 	}
 
 	// TODO: the cluster cidr should be configurable through a common interface
-	clusterCIDRStr, err := utils.GetServiceArgument("kube-proxy", "--cluster-cidr")
-	if err != nil {
-		return nil, fmt.Errorf("failed to get cluster cidrs from kube-proxy arguments: %w", err)
-	}
-
+	clusterCIDRStr := snap.GetServiceArgument(snap.Instance(), "kube-proxy", "--cluster-cidr")
```
neoaggelos, line 30: Ideally, `snap.Instance()` would eventually be initialized once and passed to all callers as needed, but it's OK to start with this.

src/k8s/pkg/k8s/setup/certificates.go
```diff
@@ -7,15 +7,20 @@ import (
 )
 
 // InitCertificates sets up the CAs and the necessary server certificates that is used by Kubernetes.
-func InitCertificates() (*cert.CertificateManager, error) {
+// An initial CertificateAuthority can be provided. If not, a self-signed one will be generated.
```
neoaggelos, line 10: minor fix, godoc should start with the function name

src/k8s/pkg/k8s/setup/kubeconfigs.go
```diff
@@ -58,11 +67,87 @@ func InitKubeconfigs(ctx context.Context, state *state.State, ca *cert.CertKeyPa
...
+			ApiServerPort: port,
+			CertData:      base64.StdEncoding.EncodeToString(certPem),
+			KeyData:       base64.StdEncoding.EncodeToString(keyPem),
+		},
+		path,
+	)
+}
+
+// generateKubeconfig creates a kubeconfig file with the given token and CA data.
+func generateKubeconfig(token string, caCertPem []byte, path string, overwrites *ExtraKubeConfigArgs) error {
```
neoaggelos, line 104: nit, from the caller site, seeing "generate" makes me expect to get the kubeconfig as a return value, but instead we write to a file on disk. maybe change this to something like `renderKubeconfig`?

src/k8s/pkg/k8s/setup/kubeconfigs.go
```diff
@@ -2,17 +2,26 @@ package setup
...
 )
 
+type ExtraKubeConfigArgs struct {
+	ApiServerIp   *string
+	ApiServerPort *int
+}
+
 // InitKubeconfigs generates the kubeconfig files that services use to communicate with the apiserver.
-func InitKubeconfigs(ctx context.Context, state *state.State, ca *cert.CertKeyPair) error {
+func InitKubeconfigs(ctx context.Context, state *state.State, ca *cert.CertKeyPair, overwrites *ExtraKubeConfigArgs) error {
```
neoaggelos, line 24: `ExtraArgs` could also be a struct, not a pointer? simplifies the if-else logic later on?

src/k8s/pkg/k8s/setup/kubeconfigs.go
```diff
@@ -58,11 +67,87 @@ func InitKubeconfigs(ctx context.Context, state *state.State, ca *cert.CertKeyPa
...
+	}
+
+	if overwrites.ApiServerIp != nil {
+		if net.ParseIP(*overwrites.ApiServerIp) == nil {
+			return "", 0, fmt.Errorf("%s is not a valid ip", *overwrites.ApiServerIp)
+		}
+		ip = *overwrites.ApiServerIp
+	} else {
+		ip = "127.0.0.1"
+	}
```
neoaggelos, line 137: Hostnames can also work here (e.g. `my-lb.example.com`). I would skip this check. Maybe set `ip` (or `host`) to `127.0.0.1`, or use the override.

src/k8s/pkg/k8s/setup/serviceargs.go
```diff
@@ -2,19 +2,44 @@ package setup
...
+	"os"
+	"regexp"
+	"strings"
 
+	apiv1 "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/k8s/pkg/snap"
 	"github.com/canonical/k8s/pkg/utils"
 )
 
+const argumentPattern = `--?($ARGUMENT_NAME)[[:space:]]*=[[:space:]]*([^[:space:]\n]+)`
```
neoaggelos, line 14: If we are using regex:

```suggestion
var argumentPatternRe = regexp.MustCompile(`--?($ARGUMENT_NAME)[[:space:]]*=[[:space:]]*([^[:space:]\n]+)`)
```

src/k8s/pkg/k8s/setup/serviceargs.go
```diff
@@ -2,19 +2,44 @@ package setup
...
+		if ao, exists := (*overwrites)[service]; exists {
+			for argument, value := range ao {
+				pattern := strings.ReplaceAll(argumentPattern, "$ARGUMENT_NAME", argument)
+				re := regexp.MustCompile(pattern)
+				if matches := re.FindStringSubmatch(serviceArgs); matches != nil {
+					argumentWithNewValue := strings.ReplaceAll(matches[0], matches[2], value)
+					serviceArgs = strings.ReplaceAll(serviceArgs, matches[0], argumentWithNewValue)
+				}
+			}
+		}
```
neoaggelos, line 40: I see the value in having a regex for this, but if we do so, please ensure we have proper unit tests for the cases that we _at least_ expect to happen most of the time.

As reference or example, see https://github.com/canonical/microk8s-cluster-agent/blob/main/pkg/snap/services.go and https://github.com/canonical/microk8s-cluster-agent/blob/main/pkg/snap/services_test.go

Also, as a general comment, I would prefer working on structured data (e.g. arguments should be a `map[string]string` or something similar when we mutate them), and only ever render them at the end (in the format of the args file we have).

Otherwise, we mix business logic with the representation, and makes it hard to test/validate/maintain in the future.

neoaggelos (commented):
Did an initial pass for comments. Overall it's going in the right direction.

src/k8s/pkg/snap/snap.go
```diff
@@ -0,0 +1,98 @@
...
+var lock = &sync.Mutex{}
+
+// Instance returns an instance of the snap.
+//
+// Singletons have a lot of downsides but I think
+// this fits quite well here. The snap struct doesn't really
+// handle any state and with the given structure of the k8s snap
+// it is not easy to create and inject one snap instance in the
+// application root.
+func Instance() Snap {
```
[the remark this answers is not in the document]
neoaggelos: see also my other comment, it's fine to keep this as-is for now, maybe with a FIXME at the top describing what we should _eventually_ do

src/k8s/pkg/k8s/setup/serviceargs.go
```diff
@@ -2,19 +2,44 @@ package setup
...
-//   - For each service, copies the default arguments files from the snap under $SNAP_DATA/args
+//   - For each service, copies the default arguments files from the snap under $SNAP_DATA/args and apply any overwrites
 //   - Note that the `k8sd` service is already configured in the snap install hook and thus not included here
-func InitServiceArgs() error {
+func InitServiceArgs(overwrites *apiv1.ExtraServiceArgs) error {
+	if overwrites == nil {
+		overwrites = &apiv1.ExtraServiceArgs{}
+	}
+
 	for _, service := range []string{"containerd", "k8s-dqlite", "kube-apiserver", "kube-controller-manager", "kube-proxy", "kube-scheduler", "kubelet"} {
```
neoaggelos, line 24: If the reason for a single `ExtraServiceArgs` is to do this, see https://github.com/canonical/microk8s-cluster-agent/blob/main/pkg/k8sinit/apply.go#L68-L98 for a way to do it with named struct fields as well.

src/k8s/api/v1/cluster.go
bschimke95: I agree. However, in this phase we probably need to play around a bit in terms of what parameters we want to overwrite or add. Having explicit parameters would mean significant overhead for a simple test (e.g. adding it to the response struct, parsing it etc.). I added a TODO to revise this once the k8s-snap matured a bit.

src/k8s/pkg/component/hook.go
bschimke95: I agree. However, this is not yet easily possible with the microcluster framework as described in the GH comment. We'll can update this once microcluster supports something like `OnRequest` hooks (which is "in the making" according to the microcluster folks).

src/k8s/pkg/k8s/setup/certificates.go
bschimke95: Not sure if I miss anything, but it does this already, right? It's a multiline comment.

src/k8s/pkg/k8s/setup/serviceargs.go
bschimke95: `regexp.MustCompile` is called after the `$ARGUMENT_NAME` is replaced to ensure that the regexp is still valid.

src/k8s/pkg/k8s/setup/serviceargs.go
bschimke95: Very good point. I parse the argument file now into a map and overwrite the values there. Way cleaner

src/k8s/pkg/k8s/setup/serviceargs.go
bschimke95: I went for map approach so this is obsolete anyway

src/k8s/pkg/snap/snap.go
```diff
@@ -0,0 +1,102 @@
...
+		if snapInstance == nil {
+			snapInstance = NewSnap(
+				os.Getenv("SNAP"),
+				os.Getenv("SNAP_DATA"),
+				os.Getenv("SNAP_COMMON"),
+			)
+		}
+	}
+
+	return snapInstance
```
addyess, line 63: Is this semantic a go norm to double check the `inst` value?  I don't have much experience here in go so i'm just asking


I did find another pattern using `sync.Once`

```go
var once sync.Once

type single struct {
}

var singleInstance *single

func getInstance() *single {
    if singleInstance == nil {
        once.Do(
            func() {
                fmt.Println("Creating single instance now.")
                singleInstance = &single{}
            })
    } else {
        fmt.Println("Single instance already created.")
    }

    return singleInstance
}
```

addyess: make the instance a pointer instead and use `once.Do` for the instance creation

```go
var snapInstance *Snap
```


```suggestion
    once.Do(func() {
    	snapInstance = &NewSnap(
		    os.Getenv("SNAP"),
			os.Getenv("SNAP_DATA"),
			os.Getenv("SNAP_COMMON"),
		)
    }
	return snapInstance
```

src/k8s/pkg/snap/services.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/canonical/k8s/pkg/utils"
+)
+
```
addyess, line 11: what's the reason for the functions `GetServiceArgument` and `UpdateServiceArguments` to accept `Snap` as a first argument type rather than as a method on the `Snap`

`func (Snap *)GetServiceArgument(serviceName string ...) (...)`

neoaggelos: `Snap` is an interface, therefore it cannot implement methods.

src/k8s/pkg/snap/snap.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	for _, opt := range options {
+		opt(s)
+	}
+	return s
+}
+
+var snapInstance Snap
+var lock = &sync.Mutex{}
```
neoaggelos, line 37: I see why we do this here, but I don't believe there is any reason to go with a singleton object. I think at this point it's OK to initialize new instances as needed, there is not much initialization happening, just a struct being created.

The right way to approach is that we create an instance when we start then ensure to pass it around as needed, not a singleton object.

I would recommend just removing this, we can keep `snap.Instance()` as a temporary easy solution.

src/k8s/pkg/snap/snap.go
neoaggelos: `sync.Once` is indeed a way to do this, but see comment above, we don't really need this.

neoaggelos (commented):
Some more comments

src/k8s/pkg/snap/snap.go
bschimke95: The double `nil` check is because:
```
There is another nil-check after the lock is acquired. This is to ensure that if more than one goroutine bypasses the first check, only one goroutine can create the singleton instance. Otherwise, all goroutines will create their own instances of the singleton struct.
```

(shamelessly copied from here: https://refactoring.guru/design-patterns/singleton/go/example)

Your suggestion works as well (thanks for that!) but this is kind of a temporary solution as Angelos already wrote, so I'll keep my variant. Also note that all this locking is required to avoid problems with a massive amount of go routines, which we do not have.

src/k8s/pkg/snap/services.go
bschimke95: I guess Adam refers to the `snap` struct. 

If you do this, there would be no easy way to test this method. You could provide a mock implementation for other methods to use but you could not test that method in isolation. By injecting a mock `Snap` implementation this is possible.

src/k8s/pkg/snap/snap.go
bschimke95: fair - i'll change it

src/k8s/api/v1/cluster.go
```diff
@@ -22,6 +22,24 @@ type GetKubeConfigResponse struct {
...
 
+// JoinClusterRequest is used to request the configuration to join the k8s cluster.
+type JoinClusterRequest struct {
+	Token string `json:"token"`
+}
+
+// JoinClusterResponse is the response for "POST 1.0/k8sd/cluster/join".
+type JoinClusterResponse struct {
+	// TODO: use named arguments (e.g. ExtraKubeletArgs) instead.
+	ExtraServiceArgs ExtraServiceArgs `json:"extraServiceArgs"`
```
neoaggelos, line 33: Note that this is not just a matter of TODO here, since it might be hard to retrofit without an API change. Not a problem (and there are ways to embed structs and hopefully preserve backwards-compatibility in the future.

Please also add a godoc here, explaining how ExtraServiceArgs are used.

src/k8s/pkg/component/hook.go
neoaggelos, line 30: what about possible errors herer?

src/k8s/pkg/k8s/setup/kubeconfigs.go
```diff
@@ -58,11 +66,80 @@ func InitKubeconfigs(ctx context.Context, state *state.State, ca *cert.CertKeyPa
...
+			"kube-apiserver",
+			"--secure-port",
+		))
+		if err != nil {
+			return "", 0, fmt.Errorf("apiserver port is not an integer: %w", err)
+		}
+	}
+
+	return ip, port, err
+}
```
neoaggelos, line 145: nit, this is maybe just two separate functions, also no need for extra structs.

neoaggelos (commented):
Getting there, mostly minor nits.

src/k8s/pkg/snap/services.go
addyess: i think `Snap` as an interface was the reasoning i was looking for.  I don't have the go chops yet to understand the language semantics so this was just an effort to learn the language

src/k8s/pkg/component/hook.go
bschimke95: Then an empty string is returned. I copied that behavior over from the cluster-agent. IMHO there is not much that a user can do that requires to distinguish between an empty value and an error

src/k8s/pkg/component/hook.go
```diff
@@ -5,16 +5,17 @@ import (
...
 
-type valuesHook func() (map[string]any, error)
+type valuesHook func(snap.Snap) (map[string]any, error)
 
 var valuesHooks = map[string]valuesHook{
 	"network": networkValues,
 }
 
-func networkValues() (map[string]any, error) {
+func networkValues(sn snap.Snap) (map[string]any, error) {
```
neoaggelos, line 18: `sn` does not read very well. `snap` is better here, also other places. it's fine if we shadow the `snap` package name.

neoaggelos: or perhaps `s` if context is well-known

bschimke95: There is a lot of potential for collisions. 
s -> micrcluster state
Simply shadowing the snap package also is not always possible as we have `snap.MySnapFunction` and `snap := snapFromContext`

Where required, I renamed the package to `snapPkg` and shadowed the `snap`name otherwise.

src/k8s/pkg/component/hook.go
```diff
@@ -5,16 +5,17 @@ import (
...
 
-type valuesHook func() (map[string]any, error)
+type valuesHook func(s.Snap) (map[string]any, error)
 
 var valuesHooks = map[string]valuesHook{
 	"network": networkValues,
 }
 
-func networkValues() (map[string]any, error) {
+func networkValues(snap s.Snap) (map[string]any, error) {
```
neoaggelos, line 18: the comment was about making this `s snap.Snap` instead, not change the package name to `s`, sorry if this was not clear.

neoaggelos (commented):
LGTM, there are things we can iterate on as future work.

Just a minor nit to make the code a bit more readable and we are good to go!

neoaggelos (approved):

PR #50, opened 2024-01-10
Use a resolv.conf file with non-loopback addresses by default

neoaggelos:
### Summary

Patch Kubelet so that it picks a `resolv.conf` file from the host that contains valid non-loopback upstream nameserver addresses.

This is useful so that we can remove friction from pods being unable to resolve external hostnames (e.g. `canonical.com`).

### Testing

Build the snap with the applied patch, and try to resolve an external hostname:

```
kubectl run --rm -it --image alpine -- nslookup canonical.com
```

It should the nameservers configured by systemd-resolve and work without issues. The same does not happen if `/etc/resolv.conf` is used instead, since the loopback address does not work inside the pod.

### Notes

The implementation is loosely based on https://github.com/canonical/microk8s/blob/master/scripts/find-resolv-conf.py

### DIFF
--- a/build-scripts/components/kubernetes/patches/default/0001-use-a-resolv.conf-file-with-non-loopback-nameservers.patch
+++ b/build-scripts/components/kubernetes/patches/default/0001-use-a-resolv.conf-file-with-non-loopback-nameservers.patch
@@ -1,0 +1,288 @@
+From 6eadc61c35316d18b5dff105ffdcf76673bba9e4 Mon Sep 17 00:00:00 2001
+From: Angelos Kolaitis <angelos.kolaitis@canonical.com>
+Date: Wed, 10 Jan 2024 15:22:41 +0200
+Subject: [PATCH] use a resolv.conf file with non-loopback nameservers
+
+This replaces the const kubetypes.ResolvConfDefault with a function.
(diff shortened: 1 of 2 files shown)


bschimke95 (approved):
Great work, LGTM

neoaggelos:
An example of why this PR is useful:

_without this patch_, on an empty cluster:

```
root@u1:~# kubectl run --rm -it --image alpine -- sh

/ # nslookup google.com
nslookup: write to '127.0.0.53': Connection refused
;; connection timed out; no servers could be reached
```

_with the patch_, on an empty cluster:

```
root@u1:~# journalctl -f -u snap.k8s.kubelet -n 1000 | grep resolv
Jan 11 09:35:33 u1 k8s.kubelet[6214]: I0111 09:35:33.567552    6214 resolv_conf.go:34] Using /run/systemd/resolve/resolv.conf for the DNS resolver config


root@u1:~# kubectl run --rm -it --image alpine -- sh

/ # nslookup google.com
Server:		10.100.255.254
Address:	10.100.255.254:53

Non-authoritative answer:
Name:	google.com
Address: 142.251.140.78

Non-authoritative answer:
Name:	google.com
Address: 2a00:1450:4017:815::200e
```

PR #54, opened 2024-01-12
Added clean-up for network component

berkayoz:
Added clean-up for network component


k8s/lib.sh
```diff
@@ -31,6 +31,29 @@ k8s::common::setup_env() {
...
+  do
+    ip link delete ${link} || true
+  done
+
+  iptables-save | grep -iv cilium | iptables-restore
+  ip6tables-save | grep -iv cilium | ip6tables-restore
+  iptables-legacy-save | grep -iv cilium | iptables-legacy-restore
+  ip6tables-legacy-save | grep -iv cilium | ip6tables-legacy-restore
+
+  default_interface=$(ip route get 1.1.1.1 | sed -n 's/.*dev \([^\ ]*\).*/\1/p')
```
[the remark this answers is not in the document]
berkayoz: That works better, updated thanks.

snap/snapcraft.yaml
```diff
@@ -388,3 +388,11 @@ plugs:
...
     interface: docker-support
     privileged-containers: false
+
+hooks:
+  disconnect-plug-network-control:
+    plugs:
+      - network
+      - network-bind
+      - network-control
+      - firewall-control
```
neoaggelos, line 398: newline at EOF

snap/snapcraft.yaml
```diff
@@ -388,3 +388,11 @@ plugs:
   docker-unprivileged:
     interface: docker-support
     privileged-containers: false
+
+hooks:
+  disconnect-plug-network-control:
```
neoaggelos, line 393: should we do it on `remove` instead?

k8s/lib.sh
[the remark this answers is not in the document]
neoaggelos: note that might still not be sufficient (e.g. for airgap). in the future, we might have to be more careful to know which interface cilium is using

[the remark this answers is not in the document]
neoaggelos: i'd like to have a FIXME here just so that it is not lost.

k8s/lib.sh
```diff
@@ -31,6 +31,29 @@ k8s::common::setup_env() {
   _K8S_ENV_SETUP_ONCE="1"
 }
 
+# Cleanup configuration left by the network component
+#   - Iptables Rules
+#   - Network Interfaces
+#   - Traffic Control(tc) rules
+k8s::remove::network() {
```
bschimke95, line 38: I understand that you do not want to mention Cilium in the signature as CNIs might change. But this removal process will always be somewhat CNI specific so it makes sense to add some context.

Could you add some links/explanations where this comes from (e.g. codepointers to cilium cleanup etc)? Thanks!

k8s/lib.sh
```diff
@@ -31,6 +31,29 @@ k8s::common::setup_env() {
...
+  do
+    ip link delete ${link} || true
+  done
+
+  iptables-save | grep -iv cilium | iptables-restore
+  ip6tables-save | grep -iv cilium | ip6tables-restore
+  iptables-legacy-save | grep -iv cilium | iptables-legacy-restore
+  ip6tables-legacy-save | grep -iv cilium | ip6tables-legacy-restore
+
+  default_interface=$(ip route show default | sed -n 's/.*dev \([^\ ]*\).*/\1/p')
```
bschimke95, line 49: Maybe it makes sense to put this in a util?

bschimke95 (commented):
Great stuff Berkay, thanks for figuring this out! Just minor nits

bschimke95 (approved):
LGTM

PR #55, opened 2024-01-14
Implement `dns` component and Components Refactor

mateoflorido:
## Overview
This Pull Request introduces a mechanism for separating logic for component management, allowing for the configuration of these components. Additionally, the DNS component is implemented.
## Changes
- Introduced custom handlers for the `dns` and `network` components.
- Refactored the disable logic.
- Refactored the `network` component to adhere to the new management logic.
- Updated CoreDNS to version 1.29.0.

## Manual Testing
- Pods
```
NAMESPACE     NAME                               READY   STATUS    RESTARTS   AGE
kube-system   cilium-operator-56bdb99ff6-j5w2q   1/1     Running   0          8m5s
kube-system   cilium-v4q8z                       1/1     Running   0          8m5s
kube-system   ck-dns-coredns-7f648ff498-z7zrj    1/1     Running   0          8m
```
- nslookup
```
root@k8s-test:~# /snap/k8s/current/bin/kubectl --kubeconfig /var/snap/k8s/common/etc/kubernetes/admin.conf run --rm -it --image alpine -- sh
If you don't see a command prompt, try pressing enter.
/ # nslookup canonical.com
Server:		10.152.183.10
Address:	10.152.183.10:53

Non-authoritative answer:
Name:	canonical.com
Address: 2620:2d:4000:1::28
Name:	canonical.com
Address: 2620:2d:4000:1::26
Name:	canonical.com
Address: 2620:2d:4000:1::27

Non-authoritative answer:
Name:	canonical.com
Address: 185.125.190.21
Name:	canonical.com
Address: 185.125.190.29
Name:	canonical.com
Address: 185.125.190.20

/ # nslookup google.com
Server:		10.152.183.10
Address:	10.152.183.10:53

Non-authoritative answer:
Name:	google.com
Address: 2800:3f0:4005:411::200e

Non-authoritative answer:
Name:	google.com
Address: 142.250.218.142
```
## Notes
- Testing will be introduced in a follow-up PR.
- Custom configuration will be introduced once the machinery to provide custom configuration is in place.


src/k8s/pkg/component/hook.go
```diff
@@ -12,6 +12,62 @@ type valuesHook func() (map[string]any, error)
...
+func dnsValues() (map[string]any, error) {
+	nameserverStr, err := utils.GetServiceArgument("kubelet", "--resolv-conf")
+	if err != nil {
+		_, err := utils.LocateValidResolvConf()
+		if err != nil {
+			nameserverStr = "8.8.8.8 8.8.4.4"
+		} else {
+			nameserverStr = "/etc/resolv.conf"
+		}
+	}
```
neoaggelos, line 27: this should be handled with the kubelet patch, so no need to adjust it

src/k8s/pkg/component/hook.go
```diff
@@ -12,6 +12,62 @@ type valuesHook func() (map[string]any, error)
...
+	var dnsIP string
+	// If not using a custom service CIDR, then use the well-known address 10.152.183.10
+	if serviceClusterCIDR == "10.152.183.0/24" {
+		dnsIP = "10.152.183.10"
+	}
+
+	values := map[string]any{
+		"service": map[string]any{
+			"clusterIP": dnsIP,
+		},
```
neoaggelos, line 44: Since we will be retrieving the service IP later on from the apiserver, we can avoid this entirely.

src/k8s/pkg/component/postconfig.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	dnsIP := svc.Spec.ClusterIP
+
+	clusterDomain, err := utils.GetServiceArgument("kubelet", "--cluster-domain")
+	if err != nil || clusterDomain != "cluster.local" {
+		err := utils.UpdateServiceArgs("cluster-domain", "cluster.local", "kubelet")
+		if err != nil {
+			return fmt.Errorf("failed to update cluster-domain argument: %w", err)
+		}
```
neoaggelos, line 32: i wonder if we should be simply automatically setting `--cluster-domain` in the default args of kubelet. that way, we would only have to set `--cluster-dns` in this place.

src/k8s/pkg/utils/file.go
```diff
@@ -76,6 +76,30 @@ func GetServiceArgument(service string, argument string) (string, error) {
...
+
+	regexPattern := regexp.MustCompile(`(?m)^` + regexp.QuoteMeta(argument) + `=.*$`)
+	if regexPattern.Match(fileContent) {
+		fileContent = regexPattern.ReplaceAll(fileContent, []byte(replaceLine))
+	} else {
+		fileContent = append(fileContent, []byte(replaceLine+"\n")...)
+	}
+
+	return os.WriteFile(configFile, fileContent, 0o644)
+}
```
neoaggelos, line 101: please see https://github.com/canonical/k8s-snap/pull/49/files#diff-506efc1ed4c8530427a69550ef2bbc3caba657b9a33dde1851cc3b1f07ecf17aR39 (there was also some discussion around why want to avoid a regex for updating arguments in the same PR).

My preference would be to wait until #49 is merged and adjust the code. We need to move #49 forward.

neoaggelos (commented):
Left some initial comments. Good work overall, but I believe we will have to do some adjustments on this PR before it goes in.

src/k8s/pkg/utils/file.go
mateoflorido: I see, I'll wait until #49 is merged to refactor the PR. Thanks for the heads up :)

mateoflorido:
Using custom nameservers and service ip:
```
root@k8s-test:~# k8s enable dns --upstream-nameservers=1.1.1.1,8.8.8.8 --service-ip=10.152.183.11
DNS component enabled
root@k8s-test:~# /snap/k8s/current/bin/kubectl --kubeconfig /var/snap/k8s/common/etc/kubernetes/admin.conf run --rm -it --image alpine -- sh
If you don't see a command prompt, try pressing enter.
/ # nslookup canonical.com
Server:		10.152.183.11
Address:	10.152.183.11:53

Non-authoritative answer:
Name:	canonical.com
Address: 185.125.190.21
Name:	canonical.com
Address: 185.125.190.29
Name:	canonical.com
Address: 185.125.190.20

Non-authoritative answer:
Name:	canonical.com
Address: 2620:2d:4000:1::27
Name:	canonical.com
Address: 2620:2d:4000:1::26
Name:	canonical.com
Address: 2620:2d:4000:1::28
```

src/k8s/api/v1/component.go
```diff
@@ -13,6 +13,22 @@ type UpdateComponentRequest struct {
 	Status ComponentStatus `json:"status"`
 }
 
+// UpdateDNSComponentRequest is used to update the DNS component state.
+type UpdateDNSComponentRequest struct {
+	Config *DNSComponentConfig `json:"config,omitempty"`
+	Status string              `json:"status"`
```
neoaggelos, line 19: you can use the `ComponentStatus` alias here, like in `UpdateComponentRequest`.

src/k8s/cmd/k8s/k8s_enable_dns.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+		config := api.DNSComponentConfig{
+			ServiceIP:           dnsConfig.ServiceIP,
+			UpstreamNameservers: dnsConfig.UpstreamNameservers,
+			ClusterDomain:       dnsConfig.ClusterDomain,
+		}
+
+		request := api.UpdateDNSComponentRequest{
+			Status: "enabled",
```
neoaggelos, line 37: Use the "enum"-like `api.ComponentEnable` here

src/k8s/pkg/component/component.go
```diff
@@ -16,6 +16,8 @@ import (
 type ComponentManager interface {
 	// Enable enables a k8s component.
 	Enable(name string) error
+	// EnableWithValues enables a k8s component with custom configuration.
+	EnableWithValues(name string, values map[string]any) error
```
neoaggelos, line 20: We fully control callers here, so let's just change `Enable`?

```suggestion
	// Enable enables a k8s component, optionally specifying custom configuration options.
	Enable(name string, values map[string]any) error
```

src/k8s/pkg/component/dns.go
```diff
@@ -0,0 +1,107 @@
...
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	api "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/k8s/pkg/utils"
+)
+
+func EnableDNSComponent(request api.UpdateDNSComponentRequest) error {
```
neoaggelos, line 13: A minor note on this. `EnableDNSComponent` is internal machinery, whereas `api.UpdateDNSComponentRequest` is "public" API.

If we had to change the API in any way (or this implementation), we don't want to have to make this change in multiple places. As such, I would recommend changing the definition to:

```
func EnableDNSComponent(upstreamNameserver, clusterDomain, serviceIP string) error
```

This way, we can change the API (or the EnableDNSComponent) without having to track down changes across multiple files.

src/k8s/pkg/k8s/client/component.go
```diff
@@ -23,6 +23,20 @@ func (c *Client) ListComponents(ctx context.Context) ([]api.Component, error) {
...
+func (c *Client) UpdateDNSComponent(ctx context.Context, request api.UpdateDNSComponentRequest) error {
+	queryCtx, cancel := context.WithTimeout(ctx, time.Second*30)
+	defer cancel()
+
+	var response api.UpdateDNSComponentResponse
+	// TODO: This URL is a temporary measure to prevent collisions with the /k8sd/components/{name} path
+	err := c.mc.Query(queryCtx, "PUT", lxdApi.NewURL().Path("k8sd", "components", "dns:enable"), request, &response)
+	if err != nil {
+		clientURL := c.mc.URL()
+		return fmt.Errorf("failed to query endpoint on %q: %w", clientURL.String(), err)
```
neoaggelos, line 35: Side-note, when wrapping errors we should add context that the function that failed did not have, e.g.

```go
return fmt.Errorf("enable dns component failed: %w", err)
```

src/k8s/pkg/k8sd/api/component.go
```diff
@@ -21,6 +21,12 @@ var k8sdComponents = rest.Endpoint{
 	Get:  rest.EndpointAction{Handler: componentsGet, AllowUntrusted: false},
 }
 
+// TODO: This endpoint is a temporary measure to prevent collisions with the /k8sd/components/{name} path
+var k8sdDNSComponent = rest.Endpoint{
+	Path: "k8sd/components/dns:enable",
```
neoaggelos, line 26: thanks for keeping the todo here and in the client code :+1:

src/k8s/pkg/utils/dns.go
```diff
@@ -0,0 +1,61 @@
...
+		isValid := processResolvConf(path)
+		if isValid {
+			return path, nil
+		}
+	}
+	return "", fmt.Errorf("no suitable resolv.conf file found")
+}
+
+// processResolvConf processes a single resolv.conf file and checks if it contains only non-loopback addresses.
+func processResolvConf(path string) bool {
```
neoaggelos, line 40: Do we still need this here? Kubelet should be able to handle the default resolv.conf already.

neoaggelos (commented):
Excellent work @mateoflorido! Leaving some follow-up comments

src/k8s/pkg/utils/kubectl.go
```diff
@@ -0,0 +1,46 @@
+package utils
```
bschimke95, line 1: seems like we have done a bit of duplicate work here:
https://github.com/canonical/k8s-snap/pull/57/files

I would not reimplement the go-k8s library but rather - if possible - provide more abstraction.
For example, from my PR: Not implementing "k8s.GetNodes" and then iterate on the node objects but rather "k8s.ClusterReady".

bschimke95 (commented):
Great work @mateoflorido 
Left some comments.

src/k8s/cmd/k8s/k8s_disable.go
```diff
@@ -9,36 +10,60 @@ import (
...
-	}
-)
+type disableFunc func(ctx context.Context, client *client.Client) error
+
+var disableActions = map[string]disableFunc{
+	"dns":     disableDns,
+	"network": disableNetwork,
+}
+
+var disableCmd = &cobra.Command{
```
neoaggelos, line 20: This looks great, but I would recommend having separate commands here as well. It is much easier to be able to specify custom arguments in the future (if we have to), same as we have in the enable command.

src/k8s/cmd/k8s/k8s_enable_dns.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+
+	api "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/k8s/pkg/k8s/client"
+	"github.com/spf13/cobra"
+)
+
+var dnsConfig struct {
```
neoaggelos, line 11: Just a nit, same as other commands, `enableDNSCmdConfig`

src/k8s/pkg/component/component.go
```diff
@@ -14,8 +14,9 @@ import (
 
 // ComponentManager defines an interface for managing k8s components.
 type ComponentManager interface {
-	// Enable enables a k8s component.
+	// Enable enables a k8s component, optionally specifying custom configuration options.
 	Enable(name string) error
+	EnableWithValues(name string, values map[string]any) error
```
neoaggelos, line 19: if `Enable(name)` is equivalent to `EnableWithValues(name, nil)`, then let's only have a single one and update all callers:

```go
Enable(name string, values map[string]any) error
```

neoaggelos (commented):
Almost there, nice work @mateoflorido! Leaving a few more comments

src/k8s/pkg/component/dns.go
```diff
@@ -0,0 +1,138 @@
...
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/k8s/pkg/utils"
+)
+
+func EnableDNSComponent(clusterDomain, serviceIP string, upstreamNameservers []string, s snap.Snap) error {
```
neoaggelos, line 13: nit; as a matter of precedence, make `s snap.Snap` the first argument.

src/k8s/pkg/component/component.go
```diff
@@ -83,7 +84,7 @@ func NewManager(snap snap.Snap) (*helmClient, error) {
 }
 
 // Enable enables a specified component.
-func (h *helmClient) Enable(name string) error {
+func (h *helmClient) Enable(name string, values map[string]any) error {
```
neoaggelos, line 87: yes, see `helmClient` does not implement the `ComponentManager` with this. A common approach to ensure we don't break the interface is to add the following at the end of the file, which ensures that a `*helmClient` can be assigned to a `ComponentManager` interface:

```go
var _ ComponentManager = &helmClient{}
```

src/k8s/pkg/utils/kubectl.go
bschimke95, line 1: With that, we have two utils doing the same thing:
https://github.com/canonical/k8s-snap/tree/main/src/k8s/pkg/utils/k8s

Please merge them into one.

bschimke95 (commented):
Great work Mateo, two final comments

src/k8s/pkg/component/dns.go
```diff
@@ -0,0 +1,136 @@
...
+	upstreamNameserver := "/etc/resolv.conf"
+	if clusterDomain == "" {
+		clusterDomain = "cluster.local"
+	}
+
+	if len(upstreamNameservers) > 0 {
+		upstreamNameserver = strings.Join(upstreamNameservers, " ")
+	}
+
+	values := map[string]any{
```
neoaggelos, line 28: Can we also set `service.name=coredns` and `deployment.name=coredns`? Will reduce some of the stutter in the deployed names.

- https://github.com/coredns/helm/blob/master/charts/coredns/values.yaml#L60
- https://github.com/coredns/helm/blob/master/charts/coredns/values.yaml#L376

src/k8s/pkg/component/dns.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	dnsIP, err := k8s.GetServiceClusterIP(ctx, client, "ck-dns-coredns", "kube-system")
+	if err != nil {
+		return fmt.Errorf("failed to get dns service: %w", err)
+	}
+
```
neoaggelos, line 78: These are part of the cluster config. Can you add a TODO to store them with `database.SetClusterConfig`?

I'm okay with not delaying this PR any longer, we will return to it at a later time

tests/e2e/tests/test_dns.py
```diff
@@ -0,0 +1,78 @@
...
+    assert "10.152.183.1 kubernetes.default.svc.foo.local" in result.stdout.decode()
+
+    result = h.exec(
+        instance_id,
+        ["k8s", "kubectl", "exec", "busybox", "--", "nslookup", "canonical.com"],
+        capture_output=True,
+        check=False,
+    )
+
+    assert not ("can't resolve" in result.stdout.decode())
```
neoaggelos, line 76:
```suggestion
    assert "can't resolve" not in result.stdout.decode()
```

neoaggelos (commented):
Just a couple final nits, great work!

bschimke95 (approved):
LGTM (when CI passes)

PR #56, opened 2024-01-15
Add k8s kubectl REST endpoint and CLI command

bschimke95:


src/k8s/pkg/k8sd/api/cluster_config.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/canonical/microcluster/state"
 )
 
-var k8sdClusterConfig = rest.Endpoint{
+var k8sdConfig = rest.Endpoint{
```
bschimke95, line 14: minor drive-by cleanup: The name didn't match the API endpoint

src/k8s/pkg/utils/sys.go
```diff
@@ -3,20 +3,27 @@ package utils
...
 
 // runCommand executes a command with a given context.
 // runCommand returns nil if the command completes successfully and the exit code is 0.
 func RunCommand(ctx context.Context, command ...string) error {
+	return RunCommandWithCustomWriter(ctx, os.Stdout, os.Stderr, command...)
+}
+
+// RunCommandWithCustomWriter executes a command with a given context and allows to set a custom writer
+// RunCommandWithCustomWriter returns nil if the command completes successfully and the exit code is 0.
+func RunCommandWithCustomWriter(ctx context.Context, stdout io.Writer, stderr io.Writer, command ...string) error {
```
bschimke95, line 19: This is somewhat obsolete for this PR but I felt like this can be a useful util for future work. I can remove it if you want.

neoaggelos: i prefer not to clutter utils, let's only add this in the future if needed

src/k8s/cmd/k8s/k8s_kubectl.go
```diff
@@ -0,0 +1,49 @@
...
+		DisableFlagParsing: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if rootCmdOpts.logDebug {
+				logrus.SetLevel(logrus.TraceLevel)
+			}
+
+			// Allow users to provide their own kubeconfig but
+			// fallback to the admin config if nothing is provided.
+			if os.Getenv("KUBECONFIG") == "" {
+				os.Setenv("KUBECONFIG", utils.SnapCommonPath("/etc/kubernetes/admin.conf"))
```
neoaggelos, line 26: `SnapCommonPath` is not needed here, this is running inside the snap mount namespace, therefore `/etc/kubernetes/admin.conf` will work.

src/k8s/cmd/k8s/k8s_kubectl.go
```diff
@@ -0,0 +1,49 @@
...
+					os.Getenv("KUBECONFIG"),
+				},
+				args...,
+			)
+
+			err := utils.RunCommand(cmd.Context(), command...)
+			if err != nil {
+				fmt.Println(err)
+				return fmt.Errorf("failed to execute kubectl command: %w", err)
+			}
```
neoaggelos, line 40: Do we have a reason to mess with the kubectl output? I would prefer just creating a `exec.Command`, hook up stdout and stderr then run and return the status code.

src/k8s/cmd/k8s/k8s_kubectl.go
```diff
@@ -0,0 +1,49 @@
...
+			if os.Getenv("KUBECONFIG") == "" {
+				os.Setenv("KUBECONFIG", utils.SnapCommonPath("/etc/kubernetes/admin.conf"))
+			}
+			command := append(
+				[]string{utils.SnapPath("bin/kubectl"),
+					"--kubeconfig",
+					os.Getenv("KUBECONFIG"),
+				},
+				args...,
+			)
```
neoaggelos, line 34: `kubectl` should already be available on the PATH when running inside the snap context, is this needed?

neoaggelos (commented):
Looks better, some comments

src/k8s/cmd/k8s/k8s_kubectl.go
neoaggelos: It's OK if we don't use a `util` that does not really fit our purpose here, this is a single one-off wrapper command.

neoaggelos: Can you check if we could instead use https://pkg.go.dev/syscall#Exec, which handles everything for us, and ensures we are as close to a proper `kubectl` invocation as possible?

bschimke95: Thanks for the recommendation - this worked!

src/k8s/cmd/k8s/k8s_kubectl.go
```diff
@@ -0,0 +1,48 @@
...
+				return fmt.Errorf("kubectl not found")
+			}
+
+			command := append(
+				[]string{"kubectl"},
+				args...,
+			)
+			// completly replace the executable with kubectl
+			// as we want to be as close as possible to a "real"
+			// kubectl invokation.
```
neoaggelos, line 40:
```suggestion
			// kubectl invocation.
```

neoaggelos (approved):
LGTM! only see a typo that slipped through!

neoaggelos (approved):

PR #57, opened 2024-01-16
Add k8s client and ClusterReady check

bschimke95:
This commit introduces a client to query the k8s cluster. 
It wraps the official go-k8s client library and introduces `ClusterReady` in the k8s Status.


neoaggelos (approved):
LGTM

A challenge to keep in mind with these is that we must make sure to keep the go libraries up to date as we move forward.

PR #58, opened 2024-01-16
Format cluster status CLI output

bschimke95:
Adds flags to specify in which format the CLI output should be displayed.

Current output (the `:` seem to be a bug in the components endpoint that I address in a follow-up PR)
```
ubuntu@init:~/k8s$ sudo k8s status
k8s is running
high-availability: no

components:
  : 
  : 
  : 
  dns: disable
  storage: disable
  network: enable
```

### DIFF
--- a/src/k8s/cmd/k8s/formatter/formatter.go
+++ b/src/k8s/cmd/k8s/formatter/formatter.go
@@ -1,0 +1,54 @@
+package formatter
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"gopkg.in/yaml.v2"
+)
+
+type Formatter interface {
+	// Print formattes the output.
+	Print(any) error
+}
+
+// New creates a new formatter based on passed type
+// Can be "plain", "json", "yaml".
+func New(formatterType string, writer io.Writer) (Formatter, error) {
+	switch formatterType {
+	case "plain":
+		return plainFormatter{writer: writer}, nil
+	case "json":
+		return jsonFormatter{writer: writer}, nil
+	case "yaml":
+		return yamlFormatter{writer: writer}, nil
+	default:
+		return nil, fmt.Errorf("unknown formatter type %s", formatterType)
+	}
+}
+
+type plainFormatter struct {
+	writer io.Writer
+}
+
+func (p plainFormatter) Print(data any) error {
+	_, err := fmt.Fprint(p.writer, data)
+	return err
+}
+
+type jsonFormatter struct {
+	writer io.Writer
+}
+
+func (j jsonFormatter) Print(data any) error {
+	return json.NewEncoder(j.writer).Encode(data)
+}
+
+type yamlFormatter struct {
+	writer io.Writer
+}
+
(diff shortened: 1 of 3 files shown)


neoaggelos (commented):
Looks fine, I would recommend considering a more unified approach towards this, where we have a `type Formatter interface { Write(any) error }`, and 3 implementations (`jsonFormatter`, `yamlFormatter`, `plainFormatter`). The CLI initialization takes care of picking the right one based on the flag.

Then individual commands would not need to care about reimplementing this switch-case.

I'm OK with starting like this for the `k8s status` command, we can re-iterate and clean this up later.

bschimke95:
Yeah, was also thinking about this.
Yesterday, I felt like it is not worth right now to implement an abstraction for this.
Today I thought different and implemented it - thanks for the suggestion.

src/k8s/cmd/k8s/formatter/formatter.go
```diff
@@ -0,0 +1,64 @@
...
+type yamlFormatter struct {
+	writer io.Writer
+}
+
+func (y yamlFormatter) Write(data any) error {
+	b, err := yaml.Marshal(data)
+	if err != nil {
+		return fmt.Errorf("failed to marshal yaml: %w", err)
+	}
+	_, err = fmt.Fprint(y.writer, string(b))
```
neoaggelos, line 62:
```suggestion
     return yaml.NewEncoder(y.writer).Encode(data)
```

src/k8s/cmd/k8s/formatter/formatter.go
```diff
@@ -0,0 +1,64 @@
...
+	writer io.Writer
+}
+
+func (j jsonFormatter) Write(data any) error {
+	b, err := json.MarshalIndent(data, "", "\t")
+	if err != nil {
+		return fmt.Errorf("failed to marshal json: %w", err)
+	}
+	_, err = fmt.Fprint(j.writer, string(b))
+	return err
```
neoaggelos, line 50:
```suggestion
     return json.NewEncoder(j.writer).Encode(data)
```

neoaggelos (commented):
`json` and `yaml` already support `io.Writer`

src/k8s/cmd/k8s/k8s_status.go
```diff
@@ -34,12 +39,13 @@ var (
...
+			f, err := formatter.New(statusCmdOpts.outputFormat, cmd.OutOrStdout())
+			if err != nil {
+				return fmt.Errorf("failed to create formatter: %w", err)
+			}
+			err = f.Write(clusterStatus)
+			if err != nil {
+				return fmt.Errorf("failed to write output: %w", err)
 			}
 
 			return nil
```
neoaggelos, line 51: i would probably make this into a `formatter.Print(any) error`, so that all commands can do

```
return formatter.Print(format, _something_)
```

neoaggelos (approved):
LGTM, just a final comment

bschimke95:
Done, merge once the parent branches are merged to `main`

PR #59, opened 2024-01-17
Support loopback devices on LXD profile

neoaggelos:
### Summary

Update the LXD profile to allow containers to mount loopback devices.

Note that adding these devices on the profile does not clutter the host or the container with extra loop devices. It only ensures that they can be used as needed.

### Test

```bash
# start a container with this profile
lxc launch ubuntu:22.04 -p default -p k8s-e2e test-container
lxc shell test-container

# create a loopback device based on a sparse file on disk
truncate -s 1G a.img
losetup -fP a.img
```

### DIFF
--- a/tests/e2e/lxd-profile.yaml
+++ b/tests/e2e/lxd-profile.yaml
@@ -26,3 +26,333 @@ devices:
     path: /proc/sys/net/netfilter/nf_conntrack_max
     source: /proc/sys/net/netfilter/nf_conntrack_max
     type: disk
+  dev-loop-control:
+    major: "10"
+    minor: "237"
+    path: /dev/loop-control
+    type: unix-char
+  dev-loop0:
+    major: "7"
+    minor: "0"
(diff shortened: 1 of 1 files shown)


tests/e2e/lxd-profile.yaml
```diff
@@ -26,3 +26,333 @@ devices:
     path: /proc/sys/net/netfilter/nf_conntrack_max
     source: /proc/sys/net/netfilter/nf_conntrack_max
     type: disk
+  dev-loop-control:
```
bschimke95, line 29: do we need 64 loopback devices ever?

bschimke95 (approved):
LGTM

ktsakalozos:
DO we know what happens if we have two lxc containers on the same host? Do they see the same devices? How is this case handled?

bschimke95:
What's the status here? This is open for a while already.

PR #61, opened 2024-01-17
Run cilium end-to-end connectivity tests

neoaggelos:
### Summary

Adds the Cilium end-to-end connectivity tests. These require the cilium CLI to be downloaded and run.

There is one test that fails on LXD, see the inline note in the test.

### DIFF
--- a/tests/e2e/tests/test_cilium_e2e.py
+++ b/tests/e2e/tests/test_cilium_e2e.py
@@ -1,0 +1,119 @@
+#
+# Copyright 2023 Canonical, Ltd.
+#
+import logging
+import platform
+from pathlib import Path
+
+import pytest
+from e2e_util import config, harness, util
+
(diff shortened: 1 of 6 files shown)


tests/e2e/tests/test_cilium_e2e.py
```diff
@@ -0,0 +1,119 @@
...
+            "-n",
+            "kube-system",
+            "-c",
+            "cilium-agent",
+            "--",
+            "cilium",
+            "status",
+            "--brief",
+        ],
+        condition=lambda p: p.stdout.decode().strip() == "OK",
```
bschimke95, line 97: nit: we should probably let `h.exec` just return the decoded string by default and add a flag to return the command output if this is really needed.

bschimke95 (approved):
LGTM

PR #62, opened 2024-01-17
Polish k8s status CLI command

bschimke95:
* Adds `--wait-ready` flag that blocks until a node is ready
* Adds `--timeout`
* `k8s status` always waits for API server to become ready (with timeout)
* Fixes a bug in `GetComponents` that added empty component entries to the result.

```
ubuntu@my-node:~/k8s$ sudo k8s status --wait-ready
k8s is ready.
high-availability: no

control-plane nodes:
  my-node: 10.84.220.162

components:
  dns        disabled
  network    enabled
  storage    disabled
```

### DIFF
--- a/src/k8s/cmd/k8s/k8s_status.go
+++ b/src/k8s/cmd/k8s/k8s_status.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/canonical/k8s/cmd/k8s/formatter"
 	"github.com/canonical/k8s/pkg/k8s/client"
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
@@ -23,16 +22,10 @@ var (
 		Short:  "Retrieve the current status of the cluster",
 		Hidden: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if rootCmdOpts.logDebug {
-				logrus.SetLevel(logrus.TraceLevel)
-			}
-
 			c, err := client.NewClient(cmd.Context(), client.ClusterOpts{
-				StorageDir:    clusterCmdOpts.storageDir,
-				RemoteAddress: clusterCmdOpts.remoteAddress,
-				Port:          clusterCmdOpts.port,
-				Verbose:       rootCmdOpts.logVerbose,
-				Debug:         rootCmdOpts.logDebug,
+				StorageDir: clusterCmdOpts.storageDir,
+				Verbose:    rootCmdOpts.logVerbose,
+				Debug:      rootCmdOpts.logDebug,
 			})
 			if err != nil {
 				return fmt.Errorf("failed to create client: %w", err)
--- a/src/k8s/cmd/k8s/k8s_status.go
+++ b/src/k8s/cmd/k8s/k8s_status.go
@@ -1,7 +1,9 @@
 package k8s
 
 import (
+	"context"
 	"fmt"
+	"time"
 
 	"github.com/canonical/k8s/cmd/k8s/formatter"
 	"github.com/canonical/k8s/pkg/k8s/client"
@@ -12,6 +14,8 @@ import (
 var (
 	statusCmdOpts struct {
 		outputFormat string
+		timeout      time.Duration
+		waitReady    bool
 	}
 
 	statusCmd = &cobra.Command{
@@ -34,7 +38,15 @@ var (
 				return fmt.Errorf("failed to create client: %w", err)
 			}
 
-			clusterStatus, err := c.ClusterStatus(cmd.Context())
+			const minTimeout = 3
+			if statusCmdOpts.timeout < minTimeout*time.Second {
+				cmd.PrintErrf("Timeout %v is less than minimum of %ds. Using the minimum %ds instead.\n", statusCmdOpts.timeout, minTimeout, minTimeout)
+				statusCmdOpts.timeout = minTimeout * time.Second
+			}
+
+			timeoutCtx, cancel := context.WithTimeout(cmd.Context(), statusCmdOpts.timeout)
+			defer cancel()
+			clusterStatus, err := c.ClusterStatus(timeoutCtx, statusCmdOpts.waitReady)
 			if err != nil {
 				return fmt.Errorf("failed to get cluster status: %w", err)
 			}
@@ -51,4 +63,6 @@ var (
 func init() {
 	rootCmd.AddCommand(statusCmd)
 	rootCmd.PersistentFlags().StringVar(&statusCmdOpts.outputFormat, "format", "plain", "Specify in which format the output should be printed. One of plain, json or yaml")
+	rootCmd.PersistentFlags().DurationVar(&statusCmdOpts.timeout, "timeout", 90*time.Second, "The max time to wait for the K8s API server to be ready.")
+	rootCmd.PersistentFlags().BoolVar(&statusCmdOpts.waitReady, "wait-ready", false, "If set, the command will block until at least one cluster node is ready.")
 }
--- a/src/k8s/pkg/k8s/client/cluster.go
+++ b/src/k8s/pkg/k8s/client/cluster.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	apiv1 "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/k8s/pkg/utils"
 	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/lxd/shared/api"
 )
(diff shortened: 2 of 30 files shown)


src/k8s/pkg/k8s/client/cluster.go
```diff
@@ -61,12 +61,15 @@ func (c *Client) Init(ctx context.Context) (apiv1.ClusterMember, error) {
 }
 
 // ClusterStatus returns the current status of the cluster.
-func (c *Client) ClusterStatus(ctx context.Context) (apiv1.ClusterStatus, error) {
-	queryCtx, cancel := context.WithTimeout(ctx, time.Second*30)
+func (c *Client) ClusterStatus(ctx context.Context, timeout int, waitReady bool) (apiv1.ClusterStatus, error) {
```
neoaggelos, line 64: `timeout` should just be passed through the caller context.

src/k8s/pkg/k8sd/api/cluster.go
```diff
@@ -22,7 +23,13 @@ var k8sdCluster = rest.Endpoint{
...
 
 func clusterGet(s *state.State, r *http.Request) response.Response {
-	status, err := impl.GetClusterStatus(r.Context(), s)
+	var req apiv1.GetClusterStatusRequest
+	err := json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		return response.SmartError(fmt.Errorf("failed to decode request data: %w", err))
+	}
+
+	status, err := impl.GetClusterStatus(r.Context(), s, req.WaitReady)
```
neoaggelos, line 32: An issue with this is that the client that does `k8s status --wait-ready` will block while waiting on the server, and the server will then block waiting on something else.

I am not very fond of this approach, as it can easily backfire. A different approach could be that the server returns a status field `Ready: true/false` then the client retries the request if not. This has much clearer semantics and does not keep long-running server requests.

neoaggelos (changes requested):

src/k8s/pkg/k8sd/api/cluster.go
bschimke95: I went with this approach as otherwise we would need to reimplement the `wait-ready` logic in the charm and other potential clients that use the REST API.
We don't want to have any business logic in the CLI - relying on the client to handle the `wait-ready` would be against that.

If you really feel we should go with the `Ready: true/false` approach, then this is fine for me and I'll change it. Just wanted to explain why I went down this path.

neoaggelos: I see the reasoning, indeed. However, the logic should just be a retry loop, so i still believe this is worth doing:

```
while !status.is_ready():
    status = client.GetStatus()
```

We should avoid long running RPCs unless we absolutely have to, since handling timeouts can be tricky (the client does not know where the timeout is, is it k8sd or the upstream service?)

bschimke95: Alright, makes sense - I refactored it.

src/k8s/cmd/k8s/k8s_status.go
```diff
@@ -51,4 +57,6 @@ var (
 func init() {
 	rootCmd.AddCommand(statusCmd)
 	rootCmd.PersistentFlags().StringVar(&statusCmdOpts.outputFormat, "format", "plain", "Specify in which format the output should be printed. One of plain, json or yaml")
+	rootCmd.PersistentFlags().IntVar(&statusCmdOpts.timeout, "timeout", 90, "The max time in seconds to wait for the K8s API server to be ready.")
```
neoaggelos, line 60: Use a `DurationVar`, and `time.Duration` instead.

src/k8s/pkg/utils/control.go
```diff
@@ -0,0 +1,36 @@
...
+				return
+			default:
+				ready := checkFunc()
+				if ready {
+					ch <- struct{}{}
+					return
+				}
+			}
+			time.Sleep(time.Second)
+		}
```
neoaggelos, line 25: You can avoid the extra goroutine and double context cancellation checks by using a `<-time.After(time.Second)` case for the retries instead

src/k8s/pkg/k8sd/api/cluster.go
```diff
@@ -22,6 +23,12 @@ var k8sdCluster = rest.Endpoint{
 }
 
 func clusterGet(s *state.State, r *http.Request) response.Response {
+	var req apiv1.GetClusterStatusRequest
+	err := json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		return response.SmartError(fmt.Errorf("failed to decode request data: %w", err))
+	}
+
```
neoaggelos, line 31: do we need any arguments here?

src/k8s/pkg/k8s/client/cluster.go
```diff
@@ -61,17 +62,22 @@ func (c *Client) Init(ctx context.Context) (apiv1.ClusterMember, error) {
 }
 
 // ClusterStatus returns the current status of the cluster.
-func (c *Client) ClusterStatus(ctx context.Context) (apiv1.ClusterStatus, error) {
-	queryCtx, cancel := context.WithTimeout(ctx, time.Second*30)
-	defer cancel()
-
+func (c *Client) ClusterStatus(ctx context.Context, waitReady bool) (apiv1.ClusterStatus, error) {
+	request := apiv1.GetClusterStatusRequest{
+		WaitReady: waitReady,
```
neoaggelos, line 67: do we need this?

neoaggelos (commented):
Thanks, looks better now, did another pass

src/k8s/cmd/k8s/k8s_status.go
```diff
@@ -34,7 +38,14 @@ var (
 				return fmt.Errorf("failed to create client: %w", err)
 			}
 
-			clusterStatus, err := c.ClusterStatus(cmd.Context())
+			if statusCmdOpts.timeout < 3*time.Second {
+				fmt.Printf("Timeout %v is less than minimum of 3s. Using the minimum 3s instead.\n", statusCmdOpts.timeout)
```
neoaggelos, line 42: maybe log this to stderr instead

src/k8s/pkg/k8s/client/cluster.go
```diff
@@ -61,17 +62,18 @@ func (c *Client) Init(ctx context.Context) (apiv1.ClusterMember, error) {
...
+	checkFunc := func() (bool, error) {
+		err := c.mc.Query(ctx, "GET", api.NewURL().Path("k8sd", "cluster"), nil, &response)
+		if err != nil {
+			return false, err
+		}
+		return response.ClusterStatus.Ready, nil
 	}
-	return response.ClusterStatus, nil
+
+	err := utils.WaitUntilReady(ctx, checkFunc)
```
neoaggelos, line 75: it's also fine to inline the checkFunc in here, it will still be easy to read

src/k8s/pkg/k8sd/api/impl/k8s-dqlite.go
```diff
@@ -89,38 +89,24 @@ func createClusterInitFile(voters []string, host string) error {
...
-	case <-ctx.Done():
-		return ctx.Err()
-	case <-time.After(time.Minute):
-		return fmt.Errorf("node (%s) did not finish joining the cluster within time", host)
-	case <-ch:
-		return nil
+		return strings.Contains(string(out), host), nil
 	}
+
+	if err := utils.WaitUntilReady(ctx, checkFunc); err != nil {
```
neoaggelos, line 108: same here, you can inline the func

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -4,10 +4,23 @@ import (
...
+// WaitApiServerReady waits until the kube-apiserver becomes available.
+func WaitApiServerReady(ctx context.Context, client *k8sClient) error {
+	checkFunc := func() (bool, error) {
+		_, err := client.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+		// We want to retry if an error occurs (=API server not ready)
+		// returning the error would abort, thus checking for nil
+		return err == nil, nil
+	}
+
+	return utils.WaitUntilReady(ctx, checkFunc)
```
neoaggelos, line 21: same here, you can inline the check

neoaggelos (commented):
Almost there, a few final nits and we should be good to proceed

neoaggelos (approved):
LGTM!

src/k8s/cmd/k8s/k8s_status.go
```diff
@@ -34,7 +38,14 @@ var (
 				return fmt.Errorf("failed to create client: %w", err)
 			}
 
-			clusterStatus, err := c.ClusterStatus(cmd.Context())
+			if statusCmdOpts.timeout < 3*time.Second {
+				cmd.PrintErrf("Timeout %v is less than minimum of 3s. Using the minimum 3s instead.\n", statusCmdOpts.timeout)
+				statusCmdOpts.timeout = 3 * time.Second
+			}
```
addyess, line 44: could the `3` be pulled out to a constant?

PR #63, opened 2024-01-17
drop extraneous check=True from e2e tests

neoaggelos:
### Summary

`check=True` is set by default on `util.run` at https://github.com/canonical/k8s-snap/blob/main/tests/e2e/tests/e2e_util/util.py#L18

Cleanup the code a bit, in hope that future tests do not clutter by imitating the existing ones.


mateoflorido (approved):
LGTM

PR #69, opened 2024-01-22
Introduce functionality for managing cluster config

bschimke95:
Incorporates a new table named "cluster_configs" designed to store comprehensive configuration settings pertinent to the entire cluster, excluding node-specific configurations.

Simultaneously, eliminates the outdated certificate table, as certificates are now seamlessly integrated into the cluster configuration storage.

This PR only contains the machinery to store/retrieve the configs, the CLI and API functionality will be integrated in a follow-up PR.


src/k8s/pkg/k8sd/database/cluster_configs.go
```diff
@@ -0,0 +1,112 @@
...
+	clusterConfigsStmts = map[string]int{
+		"upsert-config": mustPrepareStatement("cluster-configs", "upsert-config.sql"),
+		"select-all":    mustPrepareStatement("cluster-configs", "select-all.sql"),
+	}
+)
+
+type ClusterConfig struct {
+	// K8sCaCert represents the Kubernetes Certificate Authority certificate.
+	// Empty if we don't use self-signed certificates.
+	K8sCaCert string `mapstructure:"k8s-ca-crt,omitempty"`
```
neoaggelos, line 23: Let's avoid abbreviations on struct field members, so that it's easier to track their use across the codebase.

e.g. instead I would recommend


	K8sCertificateAuthority string `mapstructure:"k8s-ca-crt,omitempty"`

	K8sCertificateAuthorityKey string `mapstructure:"k8s-ca-crt,omitempty"`

 Also, for apiserver configs, prefer `APIServerRBAC` capitalization.

src/k8s/pkg/k8sd/database/cluster_configs.go
```diff
@@ -0,0 +1,112 @@
...
+	configMap := make(map[string]any)
+	if err := mapstructure.Decode(clusterConfig, &configMap); err != nil {
+		return fmt.Errorf("failed to decode cluster config: %w", err)
+	}
+
+	for key, value := range configMap {
+		if err := UpsertClusterConfigEntry(ctx, tx, key, value); err != nil {
+			return fmt.Errorf("upsert query (write %s->%s) failed: %w", key, value, err)
+		}
+	}
```
neoaggelos, line 80: This works, but requires N queries for N updates. This should not be a problem now, but something to keep in mind.

src/k8s/pkg/k8sd/database/cluster_configs.go
```diff
@@ -0,0 +1,112 @@
...
+	return nil
+}
+
+// UpsertConfigs inserts or updates a cluster config instance.
+// Members that are not set will be emptied.
+func UpsertClusterConfigs(ctx context.Context, tx *sql.Tx, clusterConfig ClusterConfig) error {
+	configMap := make(map[string]any)
+	if err := mapstructure.Decode(clusterConfig, &configMap); err != nil {
+		return fmt.Errorf("failed to decode cluster config: %w", err)
+	}
```
neoaggelos, line 74: I would personally prefer avoiding `mapstructure` as a dependency as this is something we could do without much trouble, but this is also fine.

Reason for this: it's currently not clear to a reader what the mapping of cluster config field <-> database row looks like.

src/k8s/pkg/k8sd/database/sql/queries/cluster-configs/upsert-config.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO cluster_configs (key, value)
+VALUES (?, ?)
+ON CONFLICT (key) DO UPDATE SET value = EXCLUDED.value;
```
neoaggelos, line 3: Are we sure that this works properly on sqlite? I would recommend adding some test that verifies that the query works as we expect and that we don't end up having conflicting configs in the table.

neoaggelos (commented):
As a generic comment:

I would rename `UpsertClusterConfigs` to `SetClusterConfig`, since the semantics are clear that this is a "PUT" operation.

I would also rename `UpsertClusterConfigEntry` to `UpdateClusterConfig`.

The idea is that the exposed API names make sense to the callers without having to look up the implementation to understand how they work.

Overall great work! Leaving some more comments

src/k8s/pkg/k8sd/database/sql/queries/cluster-configs/upsert-config.sql
bschimke95: I have a basic unittest that checks if keys can properly be written, read and overwritten. Seems to work, however I'm by far no expert on SQL so if there is any extra test case that I should check, I happily do it.

src/k8s/pkg/k8sd/database/cluster_configs.go
bschimke95: The intention was to have a 1:1 mapping between db key and member.
But I see the problem with it - let's make it more explicit.

src/k8s/pkg/k8sd/database/cluster_configs.go
bschimke95: `mapstructure`` is already part of our (indirect) dependencies, so I just went on using it. 

Once the reader understands that `mapstructure: ...` represents the database key I think it is quite obvious for the reader how the mapping is done, isn't it? How would you make it more obvious? 

Let's keep it for now and if it causes confusion we can still iterate on it in the future.

neoaggelos: Yup, tis OK to go with this

src/k8s/pkg/k8sd/database/cluster_configs.go
bschimke95: True, I added a TODO for it.

PR #73, opened 2024-01-24
Use microcluster hooks for control plane node clustering

neoaggelos:
### Summary

Refactor a few key pieces in the clustering logic for control plane nodes. We switch to using the hooks provided by microcluster instead.

This is pre-liminary work to support worker only nodes.

### Changes

- Rely on the `OnBootstrap` hook to bootstrap a cluster, avoid a second call to `POST /k8sd/cluster`
- Rely on the `PostJoin` hook for joining cluster nodes
- Cluster nodes that join retrieve any cluster configuration from the database (`database.GetClusterConfig()`), they do not reach out to other control plane nodes
- Remove Rest API endpoints that would proxy the request to the local socket. We instead reach out to the socket directly
- Move k8s-dqlite configuration to the `setup` package, where all other service configuration code also lives.
- Added some TODOs for extra refactors that are pending. Left out of this PR to make it easier to review
- Add some structure to the `ClusterConfig` type, add some more tests. This is also work in progress, and is to be completed on a follow up PR. Kept changes here minimal to make it easier to digest the PR.
- Slightly changes the k8s-dqlite clustering logic, adding TODOs for followup work
- Changes the k8sd state dir to `/var/lib/k8sd/state` from `/var/lib/k8sd`, since `/var/lib/k8sd` can be a bind mount and causes the following failure when removing nodes https://github.com/canonical/microcluster/blob/1cbda575bffe4419defcdff82ea539e718846051/internal/rest/resources/cluster.go#L272-L275

```
# why change /var/lib/k8sd -> /var/lib/k8sd/state
root@t1:~# k8s remove-node t2
Error: failed to remove node from cluster: Failed to remove the s directory: open /var/lib: permission denied
```

### Testing

- Test the `k8s bootstrap`, `k8s add-node` `k8s join-cluster`, `k8s remove-node` and verify they work
- Existing integration tests should cover regressions


src/k8s/cmd/k8s/k8s_cluster.go
```diff
@@ -7,14 +7,14 @@ import (
...
-		storageDir string
+		stateDir string
 	}
 )
 
 func init() {
-	rootCmd.PersistentFlags().StringVar(&clusterCmdOpts.storageDir, "storage-dir", path.Join(os.Getenv("SNAP_COMMON"), "/var/lib/k8sd"), "Directory with the dqlite datastore")
+	rootCmd.PersistentFlags().StringVar(&clusterCmdOpts.stateDir, "state-dir", path.Join(os.Getenv("SNAP_COMMON"), "/var/lib/k8sd/state"), "Directory with the dqlite datastore")
 
 	// By default, the storage dir is set to a fixed directory in the snap.
```
bschimke95, line 17: nit: state dir

bschimke95 (approved):
Great work, Angelos!

PR #76, opened 2024-01-24
Added gateway support

berkayoz:
Added gateway API support to the snap through Cilium. The requirements are the same with Ingress but seems to operate normal as well. 

The PR creates a basic helm chart that repackages upstream CRDs since an upstream chart doesn't exist. This has the downside of needing to update versions in both `components.yaml` and `version.sh`.


src/k8s/pkg/k8sd/api/component.go
```diff
@@ -139,3 +144,21 @@ func ingressComponentPut(s *state.State, r *http.Request) response.Response {
...
+
+	err := json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		return response.SmartError(fmt.Errorf("failed to decode request: %w", err))
+	}
+
+	if req.Status == api.ComponentEnable {
+		err = component.EnableGatewayComponent(snap)
+	} else {
+		err = component.DisableGatewayComponent(snap)
```
bschimke95, line 161: You don't check for that error if I see this correctly.

bschimke95 (commented):
Looks good, minor comment

src/k8s/pkg/k8sd/api/component.go
```diff
@@ -139,3 +144,21 @@ func ingressComponentPut(s *state.State, r *http.Request) response.Response {
...
+	}
+	return response.SyncResponse(true, &api.UpdateGatewayComponentResponse{})
```
addyess, line 163: obviously not right -- but maybe something like this?

```suggestion
    if err != nil {
       return response.SmartError(fmt.Errorf("failed to toggle gateway: %w", err))
    }
	return response.SyncResponse(true, &api.UpdateGatewayComponentResponse{})
```

tests/e2e/tests/test_gateway.py
```diff
@@ -0,0 +1,118 @@
...
+from pathlib import Path
+
+import pytest
+from e2e_util import config, harness, util
+from e2e_util.config import MANIFESTS_DIR
+
+LOG = logging.getLogger(__name__)
+
+
+def test_gateway(h: harness.Harness, tmp_path: Path):
```
addyess, line 14: I'm really starting to see much more similarity in each of these E2E tests.  They look more copy/paste than clearly defining what's being tested.  Once all the components are in place, i'd love to take another pass at these tests to make them more unified

berkayoz: The similarity being there is true, however most of them are checking if pods are ready/deployed and making sure everything is settled. For example gateway and ingress basically provides the same thing functionality wise, so they both create some deployment and check if we can route requests. 

Most of these tests are verifying the simplest mode of operations, I believe our tests will grow more meaningful as components get different configuration options. +1 to revisiting our tests to make them more unified/clean.

snap/snapcraft.yaml
```diff
@@ -83,6 +83,12 @@ parts:
     source: build-scripts
     override-build: ./build-component.sh helm
 
+  gateway-api-crds:
+    after: [build-deps, helm]
+    plugin: nil
+    source: build-scripts
+    override-build: ./build-component.sh gateway-api-crds
+
```
neoaggelos, line 91: let's rather not do this at build time, better to just have the script around and have the chart checked in the repository.

tests/e2e/tests/test_gateway.py
neoaggelos: +1, let's revisit the tests _in the future_ to clean things up. I'm all fine with letting the tests be written independently at the moment, while we are still figuring out what they will look like.

neoaggelos (changes requested):
Overall looks OK, but let's keep the charts bundled in the repository.

build-scripts/components/gateway-api-crds/build.sh
```diff
@@ -0,0 +1,22 @@
...
+"$HELM" create gateway-api
+rm -rf gateway-api/templates/*
+rm -rf gateway-api/charts
+cp config/crd/standard/* gateway-api/templates/
+cp config/crd/experimental/gateway.networking.k8s.io_tlsroutes.yaml gateway-api/templates/
+sed -i 's/^\(version: \).*$/\1'"${VERSION:1}"'/' gateway-api/Chart.yaml
+sed -i 's/^\(appVersion: \).*$/\1'"${VERSION:1}"'/' gateway-api/Chart.yaml
+"$HELM" package gateway-api
+
+cp gateway-api-*.tgz "${INSTALL}/k8s/components/charts/"
```
neoaggelos, line 22: let's just make this a `build-scripts/update-gateway-api-chart.sh`, which builds the chart and ensures it's in place

it can just call `helm`, and fail if it is not available on the system.

src/k8s/pkg/component/gateway.go
```diff
@@ -0,0 +1,85 @@
...
+
+	err = manager.Enable("gateway", values)
+	if err != nil {
+		return fmt.Errorf("failed to enable gateway component: %w", err)
+	}
+
+	manager, err = NewManager(s)
+	if err != nil {
+		return fmt.Errorf("failed to get component manager: %w", err)
+	}
```
neoaggelos, line 28: do we need a second manager instance?

neoaggelos (approved):
LGTM, just a final comment

src/k8s/pkg/component/gateway.go
berkayoz: Yes, this is related to Helm capabilities( as an example ` .Capabilities.APIVersions`) being fetched at `Init` time. If we don't recreate the manager instance, the cilium chart doesn't install the `GatewayClass` see [here](https://github.com/cilium/cilium/blob/39a474967deb6e04f09076433b0e5af7646fbabc/install/kubernetes/cilium/templates/cilium-gateway-api-class.yaml#L2)

neoaggelos: Ah, thanks. OK, fine with this short-term, let's not block the PR, we'll return to this in the future.

PR #81, opened 2024-01-25
k8s remove-node: disconnect k8s-dqlite node

bschimke95:
## Summary
Currently, the `k8s remove-node` command only removes the k8sd node from the cluster.
This PR adds logic to remove the node also from the k8s-dqlite cluster. 

## Details
The k8s-dqlite cluster might be left unusable if we remove the only voter left. So, it is validated that this does not happen *before* removing the node from any cluster.

Also, a `--timeout` flag is introduced on `k8s remove-node` to confirm with the CLI spec.

Finally, a `dqlite` wrapper command is introduced for convenience. I tried to use the `go-dqlite` library for those tasks but failed on some details there and moved on (for now.


src/k8s/pkg/utils/dqlite/k8s-dqlite.go
```diff
@@ -0,0 +1,40 @@
...
+	K8sDqliteDefaultPort = 9000
+)
+
+// GetK8sDqliteClusterMembers queries the local k8s-dqlite datastore for its members.
+//
+// TODO:
+// This should be done by using the go-dqlite client implementation.
+// However, when I tried to use it the client connects, but returns an empty cluster member list.
+func GetK8sDqliteClusterMembers(ctx context.Context, snap snap.Snap) ([]NodeInfo, error) {
+	c, err := client.DefaultNodeStore(snap.CommonPath("/var/lib/k8s-dqlite/cluster.yaml"))
```
berkayoz, line 22: Not sure if I'm missing something, isn't this still using the go-dqlite client?

bschimke95: Yes, the comment is a bit misleading.
The client has a `Client` and `NodeStore`. Client is a "real" client to perform actions on the cluster (remove, add etc.) while node store simply parses the cluster informations and has no additional functionality. 
I decided to keep this as otherwise I'd need to capture and parse the output from the dqlite wrapper call which seems error prone and clumpsy.

tests/e2e/tests/e2e_util/util.py
```diff
@@ -225,14 +225,20 @@ def setup_k8s_snap(h: harness.Harness, instance_id: str, snap_path: Path):
...
-            h.exec(instance, ["hostname"], capture_output=True).stdout.decode().strip()
-        )
         result = (
             stubbornly(retries=15, delay_s=5)
             .on(h, instance)
             .until(lambda p: " Ready" in p.stdout.decode())
-            .exec(["k8s", "kubectl", "get", "node", hostname, "--no-headers"])
+            .exec(
+                ["k8s", "kubectl", "get", "node", hostname(h, instance), "--no-headers"]
+            )
```
addyess, line 234: well, that's not really any more readable. 

I like the `util.hostname(...)` method -- that's fine

just keep :

```python
hostname = util.hostname(h, instance)
...
exec(["k8s", "kubectl", "get", "node", hostname, "--no-headers"])
```

tests/e2e/tests/test_clustering.py
```diff
@@ -32,6 +33,7 @@ def add_node(h: harness.Harness, cluster_node: str, joining_node: str) -> str:
         ["k8s", "add-node", joining_node],
         capture_output=True,
     )
+    assert out.returncode == 0
```
addyess, line 36: i think `h.exec` will raise an exception if  the returncode is non-zero.  There shouldn't be a need for this

tests/e2e/tests/test_clustering.py
```diff
@@ -52,9 +54,19 @@ def test_clustering(h: harness.Harness, tmp_path: Path):
...
 
     util.wait_until_k8s_ready(h, instances)
 
+    # TODO: Remove if --wait-ready for `join-cluster` is implemented.
+    time.sleep(20)
+    out = h.exec(
+        cluster_node,
+        ["k8s", "remove-node", util.hostname(h, joining_node)],
+    )
+    assert out.returncode == 0
```
addyess, line 70: here too

.github/workflows/e2e.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     name: Build
     runs-on: ubuntu-20.04
 
-    steps:
+    steps:
```
neoaggelos, line 14: looks like this slipped

src/k8s/cmd/k8s/k8s_remove_node.go
```diff
@@ -27,7 +30,16 @@ var (
 				return fmt.Errorf("failed to create cluster client: %w", err)
 			}
 
-			err = client.RemoveNode(cmd.Context(), name, removeNodeCmdOpts.force)
+			// TODO: Apply this check for all command where a timeout is required, do not repeat in each command.
+			const minTimeout = 3
```
neoaggelos, line 34: `3 * time.Second`

neoaggelos (commented):
Looks good, some comments

tests/e2e/tests/test_clustering.py
```diff
@@ -52,9 +53,18 @@ def test_clustering(h: harness.Harness, tmp_path: Path):
...
 
+    h.exec(cluster_node, ["k8s", "kubectl", "get", "nodes", "-A"])
+
     token = add_node(h, cluster_node, joining_node)
     join_cluster(h, joining_node, token)
 
     util.wait_until_k8s_ready(h, instances)
 
+    # TODO: Remove if --wait-ready for `join-cluster` is implemented.
+    time.sleep(20)
```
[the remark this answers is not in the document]
bschimke95: This waits for the k8sd to move from pending to ready so we cannot use kubectl here. 
I just retry the `remove-node` command now.

neoaggelos (approved):

PR #84, opened 2024-01-26
Initial support for worker nodes

neoaggelos:
### Summary

Implement support for worker nodes.

### Changes

- Add database methods to create and retrieve a token that can be used to join worker nodes.
- Add `POST /1.0/k8sd/worker/info` for worker nodes to retrieve join information (cluster CA, tokens, etc)
- Add e2e test for worker nodes
- Extend `k8s add-node` and `k8s join-cluster` commands to support worker nodes.
- Left a couple of TODOs for things we must return to and resolve.

### Testing

```
# control plane node
$ k8s bootstrap
$ k8s add-node w1 --worker

# worker node
k8s join-cluster "$token"
```


src/k8s/pkg/k8sd/types/worker.go
```diff
@@ -0,0 +1,52 @@
...
+func (t *InternalWorkerNodeToken) Decode(encoded string) error {
+	raw, err := encoding.DecodeString(encoded)
+	if err != nil {
+		return fmt.Errorf("failed to deserialize token: %w", err)
+	}
+	var st serializableWorkerNodeInfo
+	if err := json.Unmarshal(raw, &st); err != nil {
+		return fmt.Errorf("failed to unmarshal token: %w", err)
+	}
+	if st.Magic != internalWorkerNodeTokenSerializeMagicString {
```
bschimke95, line 46: Why are we doing this?

src/k8s/pkg/k8sd/database/sql/queries/cluster-configs/insert-worker-token.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO
+    cluster_configs(key, value)
+VALUES
+    ("worker-token", ?)
```
bschimke95, line 4: So, on the worker-node we store the token of for that node in `k8sd` right? 
Why are we doing this? For the control-planes it makes sense to store it in the DB so that other nodes can access it, but wouldn't a file be sufficient for the worker?
Also, it feels a bit weird to use the `cluster-configs` DB as a version->yaml store in the control-plane and as a key->value store in the worker node.

bschimke95 (commented):
Great work Angelos!
I did a first pass and left some comments.

src/k8s/pkg/k8sd/database/sql/queries/cluster-configs/insert-worker-token.sql
neoaggelos: No, this is stored on the control plane nodes to authenticate requests to `POST /k8sd/worker/info`.

Ι re-used `cluster_configs` since it is already a "generic" key-value place for configs. Kept it out of the `ClusterConfig` struct, in order to have runtime secrets and persistent cluster configs separate. Note that we have separate methods to access this, so it's fine if we are using the same table, our "API" ensures we never break things.

src/k8s/pkg/k8sd/types/worker.go
neoaggelos: we embed a magic string in the token to make sure that the encoding comes from us. if the magic string is missing, then we know that this token maybe be manipulated or invalid, preventing headaches down the line.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	if err := setup.RenderKubeProxyKubeconfig(snap, response.KubeProxyToken, response.CA); err != nil {
+		return fmt.Errorf("failed to render kube-proxy kubeconfig: %w", err)
+	}
+
+	if err := setup.InitAPIServerProxy(snap, response.APIServers); err != nil {
+		return fmt.Errorf("failed to configure k8s-apiserver-proxy: %w", err)
+	}
+
+	// TODO: mark node as worker
```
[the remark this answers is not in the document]
neoaggelos: the `k8sd` database will be there because we have initialized a microcluster app locally. `k8s-dqlite` might also not be in use, (e.g. if using an external etcd).

It's best to declare _intent_ rather than rely on figuring this out from side-effects.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -0,0 +1,213 @@
...
+	if err := json.NewDecoder(httpResponse.Body).Decode(&wrappedResp); err != nil {
+		return fmt.Errorf("failed to parse HTTP response: %w", err)
+	}
+	if httpResponse.StatusCode != 200 {
+		return fmt.Errorf("HTTP request for worker node info failed: %s", wrappedResp.Error)
+	}
+	response := wrappedResp.Metadata
+
+	snap := snap.SnapFromContext(s.Context)
+	if err := setup.InitFolders(snap.DataPath("args")); err != nil {
```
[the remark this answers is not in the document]
neoaggelos: there is already a huge TODO block in https://github.com/canonical/k8s-snap/pull/84/files#diff-ada4410d5c18dc109c5a39ecc63cd082390bb335b42f92667e971b14cbe44c19R3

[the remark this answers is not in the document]
bschimke95: Thanks - missed that.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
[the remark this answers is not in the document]
bschimke95: fair.

src/k8s/pkg/k8sd/types/worker.go
bschimke95: Alright, thanks for the explanation.

src/k8s/pkg/k8sd/database/sql/queries/cluster-configs/insert-worker-token.sql
neoaggelos: currently only one worker token is used, indeed. the token is created here https://github.com/canonical/k8s-snap/blob/b9dcfc2176b882bfebb1bcec53902880cacd4780/src/k8s/pkg/k8sd/database/worker.go#L39-L65, therefore the race condition would occur if:

- two control plane nodes attempt to `add-node --worker` at the same time.
- both database transactions start before the other finishes, and see an empty token
- both set the token to something, then one of the writes is lost.

This is already a quite hard situation to get into, and the single-writer of sqlite should be enough protection for _most_ cases.

Multiple workers can still join, `CheckWorkerToken` will work without issues.

bschimke95: makes sense - thanks!

bschimke95 (approved):
LGTM

PR #86, opened 2024-01-27
Shrinkify many tests using a setup fixture

addyess:
I noticed that many of the e2e tests were doing the complete same tasks again and again.  I've pushed this reuse through to conftest.py into a pytest fixture called "instances" which starts the requested number of instances for a test, bootstraps the first node, then yields the instances as a list to the test procedure. 

While testing, i was using the juju test harness and have applied a number of improvements to that harness

The exec method now behaves more like the lxd and multipass harness in that it uses juju's stderr, stdout, returncode and even provides a way to pass in stdin through the `input` argument. 

Lastly, this creates a new `Instance` object that's returned from harness.new_instance which maps through the common harness methods like send_file, pull_file, exec, and delete_instance so you don't have to keep calling `h.exec(instance_id, ...`  but instead `instance.exec(....)` which is just much easier to read

### DIFF
--- a/tests/e2e/tests/conftest.py
+++ b/tests/e2e/tests/conftest.py
@@ -77,7 +77,8 @@ def instances(
         instances.append(instance)
         util.setup_k8s_snap(instance, snap_path)
 
-    instances[0].exec(["k8s", "bootstrap"])
-    util.setup_network(instances[0])
+    first_node, *_ = instances
+    first_node[0].exec(["k8s", "bootstrap"])
+    util.setup_network(first_node)
 
     yield instances
--- a/tests/e2e/tests/conftest.py
+++ b/tests/e2e/tests/conftest.py
@@ -11,6 +11,19 @@ from e2e_util import config, harness, util
 LOG = logging.getLogger(__name__)
 
 
+def _harness_clean(h: harness.Harness):
+    "Clean up created instances within the test harness."
+
+    if config.SKIP_CLEANUP:
+        LOG.warning(
+            "Skipping harness cleanup. "
+            "It is your job now to clean up cloud resources"
+        )
+    else:
+        LOG.debug("Cleanup")
+        h.cleanup()
+
+
 @pytest.fixture(scope="session")
 def h() -> harness.Harness:
     LOG.debug("Create harness for %s", config.SUBSTRATE)
@@ -29,11 +42,7 @@ def h() -> harness.Harness:
 
     yield h
 
-    if config.SKIP_CLEANUP:
-        return
-
-    LOG.debug("Cleanup")
-    h.cleanup()
+    _harness_clean(h)
 
 
 def pytest_configure(config):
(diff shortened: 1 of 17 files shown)


tests/e2e/tests/conftest.py
```diff
@@ -32,3 +34,50 @@ def h() -> harness.Harness:
...
+    if not node_count_marker:
+        return 1
+    node_count_arg, *_ = node_count_marker.args
+    return int(node_count_arg)
+
+
+@pytest.fixture(scope="function")
+def instances(
+    h: harness.Harness, node_count: int, tmp_path: Path
+) -> List[harness.Instance]:
```
eaudetcobello, line 58: If the function yields should it's return type indicate that it's a generator?

tests/e2e/tests/e2e_util/util.py
```diff
@@ -213,24 +208,24 @@ def setup_network(h: harness.Harness, instance_id: str):
...
+def setup_k8s_snap(i: harness.Instance, snap_path: Path):
     LOG.info("Install snap")
-    h.send_file(instance_id, config.SNAP, snap_path)
-    h.exec(instance_id, ["snap", "install", snap_path, "--dangerous"])
+    i.send_file(config.SNAP, snap_path)
+    i.exec(["snap", "install", snap_path, "--dangerous"])
 
     LOG.info("Initialize Kubernetes")
-    h.exec(instance_id, ["/snap/k8s/current/k8s/connect-interfaces.sh"])
+    i.exec(["/snap/k8s/current/k8s/connect-interfaces.sh"])
```
mateoflorido, line 217: nit: What do you think on `s/i/instance`? All the other methods use instance, and I am not very fond of single-char arguments.

addyess: no complaints from me

tests/e2e/tests/conftest.py
addyess: hmm... yeah i suppose that's more clear 

in this case it's wrapped by a  `pytest.fixture` which will do things to the yielded value.  It ends up looking more like a contextmanager

tests/e2e/tests/conftest.py
```diff
@@ -32,3 +34,50 @@ def h() -> harness.Harness:
...
+    instances: List[util.Instance] = []
+
+    for _ in range(node_count):
+        # Create <node_count> instances and setup the k8s snap in each.
+        instance = h.new_instance()
+        instances.append(instance)
+        util.setup_k8s_snap(instance, snap_path)
+
+    instances[0].exec(["k8s", "bootstrap"])
+    util.setup_network(instances[0])
```
eaudetcobello, line 81: nitpick: node_count() uses unpacking syntax to access a single element. Do you want to do the same thing here ?
i.e. instance, *_ = instances

neoaggelos (commented):
Can we also update `tests/e2e/README.md`? E.g. the `Writing an End to End test` section, to make sure that the doc is up to date.

PR #87, opened 2024-01-29
ensure all k8s binaries are symlinks

neoaggelos:
### Summary

Ensure symlinked binaries are symlinks on the resulting snap. Reduces the size of the unpacked snap (by default, the `cp` command would create multiple identical binaries).


bschimke95 (approved):
LGTM

PR #90, opened 2024-01-30
Cleanup API endpoint definitions and error codes

neoaggelos:
### Summary

Move the endpoint definitions in `endpoints.go`, such that we can easily look up the available APIs. Add missing endpoint names, comments and use a consistent naming scheme for the handlers.

In a similar vein, cleanup some parts of the components to return consistent error codes, depending on the error that occurs.

### DIFF
--- a/src/k8s/pkg/k8sd/api/component.go
+++ b/src/k8s/pkg/k8sd/api/component.go
@@ -11,41 +11,10 @@ import (
 	"github.com/canonical/k8s/pkg/k8sd/api/impl"
 	"github.com/canonical/k8s/pkg/snap"
 	"github.com/canonical/lxd/lxd/response"
-	"github.com/canonical/microcluster/rest"
 	"github.com/canonical/microcluster/state"
 )
 
-var k8sdComponents = rest.Endpoint{
-	Path: "k8sd/components",
-	Get:  rest.EndpointAction{Handler: componentsGet, AllowUntrusted: false},
-}
-
-var k8sdDNSComponent = rest.Endpoint{
-	Path: "k8sd/components/dns",
-	Put:  rest.EndpointAction{Handler: dnsComponentPut, AllowUntrusted: false},
-}
-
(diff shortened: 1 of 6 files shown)


src/k8s/pkg/k8sd/api/component.go
```diff
@@ -11,195 +11,138 @@ import (
...
-}
-
-func componentsGet(s *state.State, r *http.Request) response.Response {
+func getComponents(s *state.State, r *http.Request) response.Response {
 	snap := snap.SnapFromContext(s.Context)
 
 	components, err := impl.GetComponents(snap)
 	if err != nil {
-		return response.SmartError(fmt.Errorf("failed to get components: %w", err))
+		return response.InternalError(fmt.Errorf("failed to get components: %w", err))
```
bschimke95, line 22: nit: SmartError will return better error codes and fallback to `InternalError` as a generic error. 
IMHO we should use `SmartError` instead of `InternalError`.

src/k8s/pkg/k8sd/api/component.go
```diff
@@ -11,195 +11,138 @@ import (
...
 
 	result := api.GetComponentsResponse{
 		Components: components,
 	}
-
 	return response.SyncResponse(true, &result)
 }
 
-func dnsComponentPut(s *state.State, r *http.Request) response.Response {
+func putDNSComponent(s *state.State, r *http.Request) response.Response {
```
bschimke95, line 31: Thanks for cleaning up such inconsistencies!

bschimke95 (approved):
Very nice - I like the overview we get now in `endpoints.go`

src/k8s/pkg/k8sd/api/component.go
neoaggelos: Reasoning: I feel better declaring this explicitly, especially for 4xx vs 5xx errors (so that the client can differentiate between something he did wrong, or whether something is wrong with k8sd). For example, a file not found error would be a "500", not a "404".

PR #95, opened 2024-01-31
Reduce patch for strict confinement

neoaggelos:
### Summary

Handle differences between strict and classic confinement, so that the patch for a strictly confined k8s is kept minimal.

### Changes

- Apply `runc` apparmor patches only when building the strictly confined snap
- Do not use a custom runtime dir for cilium, this will be handled with mounts if needed
- Use canonical dirs for all components (e.g. `/var/lib/k8s-dqlite` everywhere)
- Apply containerd workload profile for strict confinement only
- Ensure `/var/lib/kubelet` is a shared mount on strict only
- Ensure `/var/lib/run/containerd` is a tmpfs


src/k8s/cmd/k8s/k8s_kubectl.go
```diff
@@ -21,6 +21,10 @@ var (
 			if os.Getenv("KUBECONFIG") == "" {
 				os.Setenv("KUBECONFIG", "/etc/kubernetes/admin.conf")
 			}
+			// Set a default editor that comes with the snap so that 'kubectl edit' works
+			if os.Getenv("EDITOR") == "" {
+				os.Setenv("EDITOR", "nano")
```
bschimke95, line 26: Do we need to pack this with the snap in strict confinement?

build-scripts/print-patches-for.py
```diff
@@ -72,22 +73,29 @@ def get_patches_for(component: str, version_string: str) -> list:
...
-            patches_dir.iterdir(), reverse=True
-        )  # sort by reverse to handle exact matches
-        if path.is_dir()
-    ]
-    patch_version = find_suitable_patch_version(candidates, component_version)
-    if patch_version is None:
-        return []
-
-    return [p.resolve().as_posix() for p in (patches_dir / patch_version).iterdir()]
+    patches = []
```
bschimke95, line 76: I probably miss something here but this looks like all patches (`patches` and `strict-patches`) are loaded and applied. Where do we check if the snap is strict and only then apply the strict patches?

bschimke95 (approved):
LGTM, just to questions for my own understanding.

ktsakalozos:
LGTM +1

build-scripts/print-patches-for.py
neoaggelos: For classic builds, we'll change https://github.com/canonical/k8s-snap/pull/95/files#diff-e5250e12ec707f3ef82f07821039437b2e4fc1aa76e04504069abd4baae255d7R8 to only check in the `patches` folder instead.

We could parse the snapcraft.yaml, but that breaks remote builds (since snapcraft.yaml is out of the source dir of the part)

src/k8s/cmd/k8s/k8s_kubectl.go
neoaggelos: We already do in https://github.com/canonical/k8s-snap/blob/main/snap/snapcraft.yaml#L148

PR #98, opened 2024-02-01
Configure less loop devices on the containers by default

neoaggelos:
### Summary

Follow up PR on #59. Be a little more conservative with how many devices are passed through in the LXD profile, since it causes performance regressions on our e2e tests.


bschimke95 (approved):
LGTM

PR #99, opened 2024-02-01
Improve clustering resilience and CLI

bschimke95:
## Summary
By default, microcluster commands are not retriable if something fails in the hooks (other microcluster apps ignore this, a fix is expected to be delivered this cycle). For now, we manually roll back node assignments and configs in the CLI.
Additionally, several checks and flags were added to improve CLI handling and resilience.

Note that this PR only cleans up control-plane nodes. Workers need a bit more work (see TODO in code) that will be tackled in follow-up PRs.

## Changes

* Cleanup node if `k8s bootstrap` or `k8s join-cluster` fails
* Add proper error message if k8s is not run with root
* `k8s join-cluster` checks if the node is already bootstrapped and provides an actionable error message.
* `k8s status` shows nodes and components even if k8s is not ready.
* `k8s bootstrap` check if the node is already bootstrapped.
* Add snap convenience functions.
* Add test to remove a node in `test_clustering.py` again as `--wait-ready` makes this check reliably pass now.

### DIFF
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -81,8 +81,7 @@ func (c ClusterStatus) String() string {
 	if c.Ready {
 		result.WriteString("k8s is ready.")
 	} else {
-		result.WriteString("k8s is not ready.\n")
-		return result.String()
+		result.WriteString("k8s is not ready.")
 	}
 	result.WriteString("\n")
 
--- a/src/k8s/cmd/k8s/k8s_bootstrap.go
+++ b/src/k8s/cmd/k8s/k8s_bootstrap.go
@@ -32,7 +32,7 @@ var (
 			}
 
 			if c.IsBootstrapped(cmd.Context()) {
-				return fmt.Errorf("Node is already bootstrapped.")
+				return fmt.Errorf("k8s cluster already bootstrapped")
 			}
 
 			config := apiv1.BootstrapConfig{}
--- a/src/k8s/cmd/k8s/k8s_bootstrap.go
+++ b/src/k8s/cmd/k8s/k8s_bootstrap.go
@@ -31,6 +31,10 @@ var (
 				return fmt.Errorf("failed to create client: %w", err)
 			}
 
+			if c.IsBootstrapped(cmd.Context()) {
+				return fmt.Errorf("Node is already bootstrapped.")
+			}
+
 			config := apiv1.BootstrapConfig{}
 			if bootstrapCmdOpts.interactive {
 				config = getConfigInteractively(cmd.Context())
--- a/src/k8s/cmd/k8s/k8s_join_cluster.go
+++ b/src/k8s/cmd/k8s/k8s_join_cluster.go
@@ -1,8 +1,10 @@
 package k8s
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/canonical/k8s/pkg/config"
 	"github.com/canonical/k8s/pkg/k8s/client"
@@ -14,6 +16,7 @@ var (
 	joinNodeCmdOpts struct {
 		name    string
 		address string
+		timeout time.Duration
 	}
 
 	joinNodeCmd = &cobra.Command{
@@ -38,7 +41,7 @@ var (
 				)
 			}
 
-			client, err := client.NewClient(cmd.Context(), client.ClusterOpts{
+			c, err := client.NewClient(cmd.Context(), client.ClusterOpts{
 				StateDir: clusterCmdOpts.stateDir,
 				Verbose:  rootCmdOpts.logVerbose,
 				Debug:    rootCmdOpts.logDebug,
@@ -47,12 +50,22 @@ var (
 				return fmt.Errorf("failed to create cluster client: %w", err)
 			}
 
-			err = client.JoinNode(cmd.Context(), joinNodeCmdOpts.name, joinNodeCmdOpts.address, token)
+			if c.IsBootstrapped(cmd.Context()) {
+				return fmt.Errorf("A k8s cluster is already running on this node.")
+			}
+			const minTimeout = 3
+			if joinNodeCmdOpts.timeout < minTimeout*time.Second {
+				cmd.PrintErrf("Timeout %v is less than minimum of %ds. Using the minimum %ds instead.\n", joinNodeCmdOpts.timeout, minTimeout, minTimeout)
+				joinNodeCmdOpts.timeout = minTimeout * time.Second
+			}
+
+			timeoutCtx, cancel := context.WithTimeout(cmd.Context(), joinNodeCmdOpts.timeout)
+			defer cancel()
+
+			err = c.JoinNode(timeoutCtx, joinNodeCmdOpts.name, joinNodeCmdOpts.address, token)
 			if err != nil {
 				return fmt.Errorf("failed to join cluster: %w", err)
 			}
-
-			// TODO: (with flag) wait until the node has joined the cluster (e.g. node appears on Kubernetes)
 
 			fmt.Println("Joined the cluster.")
 			return nil
@@ -63,6 +76,7 @@ var (
 func init() {
 	joinNodeCmd.Flags().StringVar(&joinNodeCmdOpts.name, "name", "", "The name of the joining node. defaults to hostname")
 	joinNodeCmd.Flags().StringVar(&joinNodeCmdOpts.address, "address", "", "The address (IP:Port) on which the nodes REST API should be available")
+	joinNodeCmd.Flags().DurationVar(&joinNodeCmdOpts.timeout, "timeout", 90*time.Second, "The max time to wait for the node to be ready.")
 
 	rootCmd.AddCommand(joinNodeCmd)
 }
--- a/src/k8s/cmd/k8s/k8s_join_cluster.go
+++ b/src/k8s/cmd/k8s/k8s_join_cluster.go
@@ -51,11 +51,11 @@ var (
 			}
 
 			if c.IsBootstrapped(cmd.Context()) {
-				return fmt.Errorf("A k8s cluster is already running on this node.")
+				return fmt.Errorf("k8s cluster already bootstrapped")
 			}
 			const minTimeout = 3 * time.Second
 			if joinNodeCmdOpts.timeout < minTimeout {
-				cmd.PrintErrf("Timeout %v is less than minimum of %v. Using the minimum %v instead.\n", joinNodeCmdOpts.timeout, minTimeout, minTimeout)
+				cmd.PrintErrf("Timeout %v is less than minimum of %v, using the minimum %v instead.\n", joinNodeCmdOpts.timeout, minTimeout, minTimeout)
 				joinNodeCmdOpts.timeout = minTimeout
 			}
 
--- a/src/k8s/cmd/k8s/k8s_join_cluster.go
+++ b/src/k8s/cmd/k8s/k8s_join_cluster.go
@@ -53,10 +53,10 @@ var (
 			if c.IsBootstrapped(cmd.Context()) {
 				return fmt.Errorf("A k8s cluster is already running on this node.")
 			}
-			const minTimeout = 3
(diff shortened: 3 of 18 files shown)


src/k8s/api/v1/types.go
```diff
@@ -82,7 +82,6 @@ func (c ClusterStatus) String() string {
 		result.WriteString("k8s is ready.")
 	} else {
 		result.WriteString("k8s is not ready.\n")
```
neoaggelos, line 84: consistent whitespace
```suggestion
		result.WriteString("k8s is not ready.")
```

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -31,6 +31,13 @@ var (
 				return fmt.Errorf("failed to create client: %w", err)
 			}
 
+			if c.IsBootstrapped(cmd.Context()) {
+				fmt.Println("Node is already bootstrapped.")
+				// This is not an error as we "archived" the
+				// goal to bootstrap the cluster on this node.
```
neoaggelos, line 37: do you mean `achieved`? also, single-line comments are fine here, no need to keep 80 chars

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -63,6 +82,8 @@ var (
 func init() {
 	joinNodeCmd.Flags().StringVar(&joinNodeCmdOpts.name, "name", "", "The name of the joining node. defaults to hostname")
 	joinNodeCmd.Flags().StringVar(&joinNodeCmdOpts.address, "address", "", "The address (IP:Port) on which the nodes REST API should be available")
+	joinNodeCmd.Flags().BoolVar(&joinNodeCmdOpts.force, "force", false, "If set, a currently running cluster will be purged before joining.")
```
neoaggelos, line 85: no, let's not do this. there's lots of things that can go wrong if we allow this (e.g. existing runtime configuration being only one of them).

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -63,6 +82,8 @@ var (
...
+	joinNodeCmd.Flags().BoolVar(&joinNodeCmdOpts.waitReady, "wait-ready", false, "If set, the command blocks until the node is fully functional.")
```
neoaggelos, line 86: let's also add a timeout?

src/k8s/pkg/utils/microcluster.go
```diff
@@ -0,0 +1,31 @@
...
+// dqlite node is properly setup yet.
+func WaitForNodeToBeReady(ctx context.Context, client *client.Client, nodeName string) error {
+	return WaitUntilReady(ctx, func() (bool, error) {
+		clusterMembers, err := client.GetClusterMembers(ctx)
+		if err != nil {
+			return false, fmt.Errorf("failed to get cluster members: %w", err)
+		}
+
+		for _, member := range clusterMembers {
+			fmt.Printf("%s->%s\n", member.Name, member.Role)
```
neoaggelos, line 21: is this debugging or useful info?

src/k8s/pkg/utils/sys.go
```diff
@@ -3,22 +3,38 @@ package utils
...
+
+func RunsWithRootPrivilege(ctx context.Context) (bool, error) {
+	out := strings.Builder{}
+	errOut := strings.Builder{}
+	err := RunCommandWithCustomWriter(ctx, &out, &errOut, "id", "-u")
+	if err != nil {
+		return false, fmt.Errorf("failed to check for privilege: %w, %s", err, errOut.String())
+	}
+	return strings.TrimSpace(out.String()) == "0", nil
+}
```
neoaggelos, line 40: use https://pkg.go.dev/os/user#Current instead

neoaggelos: this means we can also revert the `RunCommand` changes?

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -47,14 +51,29 @@ var (
...
 				return fmt.Errorf("failed to join cluster: %w", err)
 			}
 
-			// TODO: (with flag) wait until the node has joined the cluster (e.g. node appears on Kubernetes)
-
 			fmt.Println("Joined the cluster.")
+
+			if joinNodeCmdOpts.waitReady {
+				fmt.Println("Waiting for the node to become ready.")
+				utils.WaitForNodeToBeReady(cmd.Context(), c.Mc, joinNodeCmdOpts.name)
```
neoaggelos, line 74: this needs exposing the microcluster client on the package, so that we can call the `Mc.GetClusterMembers()` directly. We have already added our `client.ClusterStatus()`, should the role just be included there instead?

neoaggelos (changes requested):
Good overall direction, did a first pass

src/k8s/cmd/k8s/k8s_join_cluster.go
neoaggelos: ah, it is already? https://github.com/canonical/k8s-snap/blob/main/src/k8s/api/v1/types.go#L54

src/k8s/cmd/k8s/k8s_bootstrap.go
neoaggelos: the more i think about it, i would prefer if `k8s bootstrap` fails on a cluster that has already started. `bootstrap` is a one-time action, not a desired state.

neoaggelos: is there any use-case we would like to accomodate by allow this?

bschimke95: I'd say "bootstrapped" is a desired state. 
With `k8s bootstrap` the user wants to achieve something -> bootstrapping a cluster. I don't think that this command should fail if the cluster is already bootstrapped - this is the expected outcome for the user.

Also, we basically offload the check for a bootstrapped cluster to the user. How does the user know if the cluster is bootstrapped - he kicks some APIs and checks for results. This may or may not be sufficient (the API call can fail for other reasons). We as the devs can provide a more robust implementation for this.
See, e.g. the [k8s charm](https://github.com/canonical/k8s-operator/blob/main/charms/k8s/src/charm.py#L89).

bschimke95: resolved offline - not returning an error can get messy (e.g. if a joined node is bootstrapped or if the init config is changed)

src/k8s/pkg/k8s/client/client.go
```diff
@@ -46,6 +46,6 @@ func NewClient(ctx context.Context, opts ClusterOpts) (*Client, error) {
 	return &Client{
 		opts: opts,
 		m:    m,
-		mc:   microClient,
+		Mc:   microClient,
```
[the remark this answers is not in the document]
bschimke95: Nope not anymore

src/k8s/pkg/utils/sys.go
bschimke95: What do you mean? This only checks if the k8s command was executed as root and returns a proper error message if not.

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -63,6 +83,8 @@ var (
 func init() {
 	joinNodeCmd.Flags().StringVar(&joinNodeCmdOpts.name, "name", "", "The name of the joining node. defaults to hostname")
 	joinNodeCmd.Flags().StringVar(&joinNodeCmdOpts.address, "address", "", "The address (IP:Port) on which the nodes REST API should be available")
+	joinNodeCmd.Flags().BoolVar(&joinNodeCmdOpts.waitReady, "wait-ready", false, "If set, the command blocks until the node is fully functional.")
```
ktsakalozos, line 86: This is not good. For the user a Ready node means a k8s node that has registered with the API server as is reporting "Ready". The fact we are using k8sd and dqlite here does not concern the user. The event of the node registering with dqlite means nothing to the user.

We should indeed wait for the node to register with dqlite and if it does not we should report it as a failure.

src/k8s/pkg/utils/sys.go
neoaggelos: I mean we don't need the `RunCommandWithCustomWriter` (not yet)

src/k8s/cmd/k8s/k8s_join_cluster.go
bschimke95: Ok, then let's drop this flag altogether.
The join command will wait for dqlite to come up and we have a `--timeout` flag if the user don't want to wait for too long.
If the user wants to wait for the node to be ready he needs to fallback to `k8s kubectl wait ...` (`k8s status --wait-ready` only waits for ONE node to be ready - not a particular one).

cc @neoaggelos

src/k8s/cmd/k8s/k8s_join_cluster.go
bschimke95: discussed offline - this is indeed a bad idea and I remove it.

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -47,13 +50,23 @@ var (
 				return fmt.Errorf("failed to create cluster client: %w", err)
 			}
 
-			err = client.JoinNode(cmd.Context(), joinNodeCmdOpts.name, joinNodeCmdOpts.address, token)
+			if c.IsBootstrapped(cmd.Context()) {
+				return fmt.Errorf("A k8s cluster is already running on this node.")
+			}
+			const minTimeout = 3
```
neoaggelos, line 56: nit, and use `%v` instead of `%ds` below
```suggestion
			const minTimeout = 3 * time.Second
```

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -104,34 +104,32 @@ func onBootstrapWorkerNode(s *state.State, encodedToken string) error {
...
 		return fmt.Errorf("failed to configure k8s-apiserver-proxy: %w", err)
 	}
 
 	// TODO: mark node as worker
 
-	for _, service := range []string{"containerd", "k8s-apiserver-proxy", "kubelet", "kube-proxy"} {
-		if err := snap.StartService(s.Context, fmt.Sprintf("k8s.%s", service)); err != nil {
-			return fmt.Errorf("failed to start service %s: %w", service, err)
-		}
+	if err := snap.StartServices(s.Context, sn, snap.WorkerServices); err != nil {
```
neoaggelos, line 131: better make this a `snap.StartWorkerServices`, and `snap.StartControlPlaneServices` below (make it hard for the caller to pass a wrong argument)

src/k8s/pkg/snap/snap.go
```diff
@@ -33,6 +33,15 @@ func NewSnap(snapDir, snapDataDir, snapCommonDir string, options ...func(s *snap
 	return s
 }
 
+// NewDefaultSnap returns a snap configured with the default snap environment.
+func NewDefaultSnap() Snap {
+	return NewSnap(
```
neoaggelos, line 38: okay to have this in a single line

neoaggelos (commented):
Almost there

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
neoaggelos: also, I prefer `s` to `sn`, like we do in other places

src/k8s/cmd/k8s/k8s_join_cluster.go
bschimke95: huh, weird - I thought I have already changed that - thanks!

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
bschimke95: This was colliding with `s state` but just renamed `state` now.

neoaggelos (approved):

PR #100, opened 2024-02-01
Unit testing harness for ComponentManager

louiseschmidtgen:
## Description
This PR adds unit test harness(fixtures and mocks) to our `ComponentManager` utility which interacts with the helm client/cli to operate on components. 

#### Interface
In order to be able to override the mocked Helm action configuration, @bschimke95  advised me to use an `HelmConfigInitializer` interface that calls the real or fake `InitializeHelmClientConfig` function. This function either returns a fake action Configuration or the real one. 
The interface is added to the Helm Client and set when the new Component manager instance is created. This required adding the initializer as an argument to the Component manager and update references to the function.

#### Mock Snap
The mock for the snap builds on the snap mock in this repo. I used the snap mock's pathprefix to be able to give the snap a fake `k8s/components/components.yaml` file to build it's configuration from.

#### Mock Action
My Mock for the `helmClient.actionConfig` is heavily inspired by the helm action package and its associated tests and uses parts to create the `actionConfig`, as well as mocks for component releases.

#### Test Set-up
The `createNewManager` function is supposed to make adding new tests easier and is supposed to provide the default HelmClient mock to test the other functions with. I'm open to suggestions on making it easier to use all the mocks and fixtures since there are indeed a lot of them in here.

#### The Bug
During unit testing a little bug crawled out of the List() function:

Instead of setting `allComponents[index] = Component{Name: name}`, I changed it to  `allComponents = append(allComponents, Component{Name: name})` to avoid overwriting components at the same index.

Previously, the `componentsMap` did not change its length so the index didn’t increment so we were just overwriting the component.
```
for name, component := range h.config {
		index := len(componentsMap)
```

[Bug Fix Commit](https://github.com/canonical/k8s-snap/commit/fb5eae50ac7bb415f66300f553eb3172daa1c2fc)

We could consider doing the bug-fix in a separate PR.

### DIFF
--- a/src/k8s/pkg/component/component.go
+++ b/src/k8s/pkg/component/component.go
@@ -177,21 +177,25 @@ func (h *helmClient) List() ([]Component, error) {
 
 	list := action.NewList(actionConfig)
 	releases, err := list.Run()
+
 	if err != nil {
 		return nil, fmt.Errorf("failed to list components: %w", err)
 	}
 
-	allComponents := make([]Component, len(h.config))
+	allComponents := make([]Component, 0)
 	componentsMap := make(map[string]int)
+
+	// Loop through components and populate allComponents and componentsMap
 	for name, component := range h.config {
 		index := len(componentsMap)
-		allComponents[index] = Component{Name: name}
+
+		allComponents = append(allComponents, Component{Name: name})
 		componentsMap[component.ReleaseName] = index
 	}
 
+	// Loop through releases and update statuses in allComponents
 	for _, release := range releases {
-		index, ok := componentsMap[release.Name]
-		if ok {
+		if index, ok := componentsMap[release.Name]; ok {
 			allComponents[index].Status = true
 		}
 	}
--- a/src/k8s/pkg/component/component.go
+++ b/src/k8s/pkg/component/component.go
@@ -78,20 +78,28 @@ func logAdapter(format string, v ...any) {
 }
 
 // NewManager creates a new Component manager instance.
-func NewManager(snap snap.Snap, initializer HelmConfigInitializer) (*helmClient, error) {
-	if initializer == nil {
+func NewManager(snap snap.Snap, initializers ...HelmConfigInitializer) (*helmClient, error) {
+	var initializer HelmConfigInitializer
+
+	if len(initializers) > 0 {
+		initializer = initializers[0]
+	} else {
+		// If no initializer provided, use a default one
 		initializer = &HelmClientIntitializer{}
 	}
 
 	viper.SetConfigName("components")
 	viper.SetConfigType("yaml")
 	viper.AddConfigPath(snap.Path("k8s/components"))
-	err := viper.ReadInConfig()
-	if err != nil {
+
+	if err := viper.ReadInConfig(); err != nil {
 		return nil, err
 	}
+
 	config := make(map[string]componentDefinition)
-	err = viper.Unmarshal(&config)
+	if err := viper.Unmarshal(&config); err != nil {
+		return nil, err
+	}
 
 	return &helmClient{
 		config:      config,
--- a/src/k8s/pkg/component/component.go
+++ b/src/k8s/pkg/component/component.go
@@ -12,6 +12,29 @@ import (
 	"helm.sh/helm/v3/pkg/cli"
 )
 
+type HelmConfigInitializer interface {
+	// Initializes a fresh Helm Configuration
+	InitializeHelmClientConfig() (*action.Configuration, error)
+}
+
+type HelmClientIntitializer struct{}
+
+func (r *HelmClientIntitializer) InitializeHelmClientConfig() (*action.Configuration, error) {
+	settings := cli.New()
+	settings.KubeConfig = "/etc/kubernetes/admin.conf"
+	actionConfig := new(action.Configuration)
+	err := actionConfig.Init(
+		settings.RESTClientGetter(),
+		settings.Namespace(),
+		os.Getenv("HELM_DRIVER"),
+		logAdapter,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize action config: %w", err)
+	}
+	return actionConfig, nil
+}
+
 // ComponentManager defines an interface for managing k8s components.
 type ComponentManager interface {
 	// Enable enables a k8s component, optionally specifying custom configuration options.
@@ -34,8 +57,9 @@ type componentDefinition struct {
 
 // helmClient implements the ComponentManager interface
 type helmClient struct {
-	config map[string]componentDefinition
-	snap   snap.Snap
+	config      map[string]componentDefinition
+	snap        snap.Snap
+	initializer HelmConfigInitializer
 }
 
 // Component defines the name and status of a k8s Component.
@@ -49,7 +73,11 @@ func logAdapter(format string, v ...any) {
 }
 
 // NewManager creates a new Component manager instance.
-func NewManager(snap snap.Snap) (*helmClient, error) {
+func NewManager(snap snap.Snap, initializer HelmConfigInitializer) (*helmClient, error) {
+	if initializer == nil {
+		initializer = &HelmClientIntitializer{}
+	}
+
 	viper.SetConfigName("components")
 	viper.SetConfigType("yaml")
 	viper.AddConfigPath(snap.Path("k8s/components"))
@@ -61,25 +89,10 @@ func NewManager(snap snap.Snap) (*helmClient, error) {
 	err = viper.Unmarshal(&config)
 
 	return &helmClient{
-		config: config,
-		snap:   snap,
+		config:      config,
+		snap:        snap,
+		initializer: initializer,
 	}, nil
-}
-
-func (h *helmClient) initializeHelmClientConfig() (*action.Configuration, error) {
-	settings := cli.New()
-	settings.KubeConfig = "/etc/kubernetes/admin.conf"
-	actionConfig := new(action.Configuration)
-	err := actionConfig.Init(
-		settings.RESTClientGetter(),
-		settings.Namespace(),
-		os.Getenv("HELM_DRIVER"),
-		logAdapter,
-	)
-	if err != nil {
-		return nil, fmt.Errorf("failed to initialize action config: %w", err)
-	}
-	return actionConfig, nil
 }
 
 // Enable enables a specified component.
@@ -89,7 +102,7 @@ func (h *helmClient) Enable(name string, values map[string]any) error {
 		return fmt.Errorf("invalid component %s", name)
 	}
 
-	actionConfig, err := h.initializeHelmClientConfig()
+	actionConfig, err := h.initializer.InitializeHelmClientConfig()
 	if err != nil {
 		return fmt.Errorf("failed to initialize Helm client configuration: %w", err)
 	}
@@ -120,7 +133,7 @@ func (h *helmClient) Enable(name string, values map[string]any) error {
 
 // isComponentEnabled checks if a component is enabled.
 func (h *helmClient) isComponentEnabled(name, namespace string) (bool, error) {
-	actionConfig, err := h.initializeHelmClientConfig()
+	actionConfig, err := h.initializer.InitializeHelmClientConfig()
 	if err != nil {
 		return false, fmt.Errorf("failed to initialize Helm client configuration: %w", err)
 	}
@@ -142,7 +155,7 @@ func (h *helmClient) isComponentEnabled(name, namespace string) (bool, error) {
 
 // List lists the status of each k8s component.
 func (h *helmClient) List() ([]Component, error) {
-	actionConfig, err := h.initializeHelmClientConfig()
+	actionConfig, err := h.initializer.InitializeHelmClientConfig()
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize Helm client configuration: %w", err)
 	}
@@ -173,7 +186,7 @@ func (h *helmClient) List() ([]Component, error) {
 
 // Disable disables a specified component.
 func (h *helmClient) Disable(name string) error {
-	actionConfig, err := h.initializeHelmClientConfig()
+	actionConfig, err := h.initializer.InitializeHelmClientConfig()
 	if err != nil {
 		return fmt.Errorf("failed to initialize Helm client configuration: %w", err)
 	}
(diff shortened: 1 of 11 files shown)


src/k8s/pkg/component/component.go
```diff
@@ -142,28 +170,32 @@ func (h *helmClient) isComponentEnabled(name, namespace string) (bool, error) {
...
 
 	list := action.NewList(actionConfig)
 	releases, err := list.Run()
+
 	if err != nil {
 		return nil, fmt.Errorf("failed to list components: %w", err)
 	}
 
-	allComponents := make([]Component, len(h.config))
+	allComponents := make([]Component, 0)
```
bschimke95, line 185: A bit of Go nerding:
you can also specify the capacity of the backing array and still set a size of 0 for the slice:
```
allComponents := make([]Component, 0, len(h.config)
```
See: https://goplay.tools/snippet/-pCq4-6OwX8

This is totally crucial for the performance of the k8s-snap and not premature optimization! ;)

src/k8s/pkg/component/component_test.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return &action.Configuration{
+		Releases:       storage.Init(driver.NewMemory()),
+		KubeClient:     &kubefake.FailingKubeClient{PrintingKubeClient: kubefake.PrintingKubeClient{Out: io.Discard}},
+		Capabilities:   chartutil.DefaultCapabilities,
+		RegistryClient: registryClient,
+		Log: func(format string, v ...interface{}) {
```
bschimke95, line 38: nit: prefer `any` over interfaces{} according to Go style-guide
Also below

src/k8s/pkg/component/component_test.go
```diff
@@ -0,0 +1,243 @@
...
+  namespace: "default"
+`
+
+func createTemporaryTestDirectory(t *testing.T) string {
+	// Create a temporary test directory to mock the snap
+	// <tempDir>
+	// └── k8s/components
+	// 	├── charts
+	// 	└── component.yaml
+	tempDir, err := os.MkdirTemp("", "test")
```
bschimke95, line 135: `t.TempDir()` is your friend. This will also make sure that tests don't interfere etc.

src/k8s/pkg/component/component_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+func addConfigToTestDir(t *testing.T, path string, data string) {
+	// Create a file and add some configs
+	err := os.WriteFile(path, []byte(data), 0644)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func createNewManager(t *testing.T, components string) (*helmClient, string, *action.Configuration) {
```
bschimke95, line 163: nit: Go prefers `mustxxx` for setup steps that fatally fail, e.g. `mustCreateNewManager` and `mustCreateTestDir`

src/k8s/pkg/component/component_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+	for _, rel := range []*release.Release{relStub1, relStub2, relStub3} {
+		if err := store.Create(rel); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	all, err := store.ListReleases()
+	assert.NoError(t, err)
+	assert.Len(t, all, 3, "sanity test: three items added")
```
neoaggelos, line 109: I understand this is using testify, but let's keep `gomega` for all our tests. Have a look at https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/database/kubernetes_auth_tokens_test.go in order to adjust this

```suggestion
	assert.NoError(t, err)
	assert.Len(t, all, 3, "sanity test: three items added")
```

src/k8s/pkg/component/component_test.go
```diff
@@ -0,0 +1,243 @@
...
+func createTemporaryTestDirectory(t *testing.T) string {
+	// Create a temporary test directory to mock the snap
+	// <tempDir>
+	// └── k8s/components
+	// 	├── charts
+	// 	└── component.yaml
+	tempDir, err := os.MkdirTemp("", "test")
+	if err != nil {
+		t.Fatal(err)
+	}
```
neoaggelos, line 138: you can use `t.TempDir()`

src/k8s/pkg/component/component_test.go
```diff
@@ -0,0 +1,243 @@
...
+	"helm.sh/helm/v3/pkg/chartutil"
+	kubefake "helm.sh/helm/v3/pkg/kube/fake"
+	"helm.sh/helm/v3/pkg/registry"
+	"helm.sh/helm/v3/pkg/release"
+	"helm.sh/helm/v3/pkg/storage"
+	"helm.sh/helm/v3/pkg/storage/driver"
+	"helm.sh/helm/v3/pkg/time"
+)
+
+var verbose = flag.Bool("test.log", false, "enable test logging")
```
neoaggelos, line 23: no need for this, we can just rely on `go test -v` which prints all logs.

src/k8s/pkg/component/component_test.go
```diff
@@ -0,0 +1,243 @@
...
+	// Call the List function with the mock HelmClient
+	components, err := mockHelmClient.List()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.NotNil(t, components)
+	assert.Equal(t, 3, len(components))
+
+	assert.Contains(t, components, Component{Name: "one", Status: true})
```
bschimke95, line 239: Our APIs should be deterministic. Let's sort the output of list alphabetically.

src/k8s/pkg/component/component.go
```diff
@@ -44,52 +54,68 @@ type Component struct {
...
-func NewManager(snap snap.Snap) (*helmClient, error) {
+func NewManager(snap snap.Snap, initializers ...HelmConfigInitializer) (*helmClient, error) {
+	var initializer HelmConfigInitializer
+
+	if len(initializers) > 0 {
+		initializer = initializers[0]
+	} else {
+		// If no initializer provided, use a default one
+		initializer = &HelmClientIntitializer{}
+	}
```
neoaggelos, line 89: Is this to allow `NewManager(snap)` to just work and not having to update callers? I'd much refer to be explicit here, and update callers as needed. Let's also make this `NewHelmManager`, to indicate that this uses helm underneath and not a generic implementation (the arguments already are helm specific)

```suggestion
func NewHelmManager(snap snap.Snap, initializer HelmConfigInitializer) (*helmClient, error) {
	if initializer == nil {
		initializer = &HelmClientIntitializer{}
	}
```

bschimke95 (commented):
Great work @louiseschmidtgen!
I did a first pass with some suggestions.

src/k8s/pkg/component/component.go
```diff
@@ -44,52 +54,68 @@ type Component struct {
...
+
+	if err := viper.ReadInConfig(); err != nil {
 		return nil, err
 	}
+
 	config := make(map[string]componentDefinition)
-	err = viper.Unmarshal(&config)
+	if err := viper.Unmarshal(&config); err != nil {
+		return nil, err
+	}
```
neoaggelos, line 102: probably unrelated to this PR but: I wonder if we should just drop `viper` completely, then define the `map[string]componentDefinition` in the Go code. Any reason to always parse the same static yaml file? Could we do this?

src/k8s/pkg/component/component_test.go
```diff
@@ -0,0 +1,243 @@
...
+	err = os.MkdirAll(k8sComponentsDir, 0777)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	k8sComponentsChartsDir := filepath.Join(k8sComponentsDir, "charts")
+	err = os.MkdirAll(k8sComponentsChartsDir, 0777)
+	if err != nil {
+		t.Fatal(err)
+	}
```
neoaggelos, line 150: just the second `MkdirAll` will create all previous dirs as well

neoaggelos (changes requested):
Great work! Left some improvement suggestions.

src/k8s/pkg/component/component.go
neoaggelos: This also needs updating all callers, and they should do `NewHelmManager(snap, nil)`

src/k8s/pkg/component/component.go
```diff
@@ -44,52 +54,68 @@ type Component struct {
...
+	return actionConfig, nil
+}
+
 func logAdapter(format string, v ...any) {
 	logrus.Debugf(format, v...)
 }
 
 // NewManager creates a new Component manager instance.
-func NewManager(snap snap.Snap) (*helmClient, error) {
+func NewManager(snap snap.Snap, initializers ...HelmConfigInitializer) (*helmClient, error) {
```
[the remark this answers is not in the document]
neoaggelos: This is also an option instead of passing `nil` default values, pick whichever you prefer. A simple `nil` arg is probably simpler until we have more configs to consider.

src/k8s/pkg/component/component.go
louiseschmidtgen: Ah, good to know! Learnt something new again :)

src/k8s/pkg/component/component_test.go
louiseschmidtgen: Thanks for the example!

src/k8s/pkg/component/component_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	mocks "github.com/canonical/k8s/pkg/component/mock"
```
neoaggelos, line 9: Any reason to rename from `mock`? 
```suggestion
	"github.com/canonical/k8s/pkg/component/mock"
```

src/k8s/pkg/component/component_test.go
```diff
@@ -180,7 +172,7 @@ func TestNewHelmClientWithValidConfig(t *testing.T) {
 	g.Expect(mockHelmClient).ToNot(BeNil())
 	g.Expect(mockHelmClient).To(BeAssignableToTypeOf(&helmClient{}))
 
-	g.Expect(mockHelmClient.initializer).To(BeAssignableToTypeOf(&MockHelmClientInitializer{}))
+	g.Expect(mockHelmClient.initializer).To(BeAssignableToTypeOf(&mocks.HelmClientInitializer{}))
 	g.Expect(mockHelmClient.snap).To(BeAssignableToTypeOf(&mock.Snap{}))
 	g.Expect(mockHelmClient.config).To(HaveLen(3))
```
neoaggelos, line 177: Are these useful tests? What do we validate here?

src/k8s/pkg/component/interface.go
```diff
@@ -0,0 +1,21 @@
+package component
+
+import "helm.sh/helm/v3/pkg/action"
+
+// HelmComponentManager defines an interface for managing k8s components.
+type HelmComponentManager interface {
```
neoaggelos, line 6: Let's keep the interface name generic. Only implementations should describe _how_ they implement the interface.
```suggestion
// ComponentManager defines an interface for managing k8s components.
type ComponentManager interface {
```

src/k8s/pkg/component/component.go
neoaggelos: ^ this is not addressed, @louiseschmidtgen

neoaggelos (commented):
Nice work! Getting there!

src/k8s/pkg/component/component.go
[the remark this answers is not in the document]
louiseschmidtgen: I'm choosing the `nil` option for now but I will keep the functional pattern in mind if there is more configuration needed later.

src/k8s/pkg/component/component.go
louiseschmidtgen: You are too quick, I was just updating it :)

src/k8s/pkg/component/component_test.go
louiseschmidtgen: I'm also importing the snap mock so go gets confused since both the snap mock and the initializer mock get imported. Do you have a better idea?

src/k8s/pkg/component/component_test.go
louiseschmidtgen: I'll remove this one.

src/k8s/pkg/component/component_test.go
neoaggelos: I would recommend `snapmock` and `componentmock`

louiseschmidtgen: Good call! Thanks

src/k8s/pkg/component/component.go
louiseschmidtgen: Are there any cases we would be updating the `component.yaml`, like a chart version update?

neoaggelos: Not for an installed snap, `component.yaml` is static and immutable. The change requires a new build, therefore it's the same as having it defined in the code

src/k8s/pkg/component/component.go
```diff
@@ -12,17 +13,8 @@ import (
...
-	Enable(name string, values map[string]any) error
-	// List returns a list of enabled components.
-	List() ([]Component, error)
-	// Disable disables a component from the cluster.
-	Disable(name string) error
-	// Refresh updates a k8s component.
-	Refresh(name string) error
-}
+// HelmClientIntitializer implements the HelmConfigInitializer interface
+type HelmClientIntitializer struct{}
```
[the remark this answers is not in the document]
neoaggelos: for lack of a better name (and seeing the conflict with the interface name), this could be called `defaultHelmConfigProvider struct`

src/k8s/pkg/component/component.go
```diff
@@ -12,17 +13,8 @@ import (
...
-	Enable(name string, values map[string]any) error
-	// List returns a list of enabled components.
-	List() ([]Component, error)
-	// Disable disables a component from the cluster.
-	Disable(name string) error
-	// Refresh updates a k8s component.
-	Refresh(name string) error
-}
+// HelmClientInitializer implements the HelmConfigInitializer interface
+type HelmClientInitializer struct{}
```
neoaggelos, line 17: So, this is now great, but still having a `*HelmClientInitializer` be a `HelmConfigInitializer` might be hard to parse and identify issues in the future. Let's instead call them:

(Note also that `provider` is a more common term for this pattern instead of `initializer`)

```
HelmConfigInitializer -> HelmConfigProvider
HelmClientInitializer -> defaultHelmConfigProvider  (it can be private, right?)
```

louiseschmidtgen: Thank you, I've made the change.

neoaggelos (approved):
LGTM!

bschimke95 (approved):
LGTM, great work!

PR #102, opened 2024-02-02
Use cluster config for k8s-dqlite configuration and certificates

bschimke95:
Remove direct read/write of k8s-dqlite certificates in cluster config. 
Adds k8s-dqlite host and port configuration to cluster config and use it to configure the cluster members k8s-dqlite.


src/k8s/pkg/k8sd/database/clusterconfigs/cluster_configs.go
```diff
@@ -51,6 +52,11 @@ type APIServer struct {
 	DatastoreClientKey  string `yaml:"datastore-client-key,omitempty"`
 }
 
+type K8sDqlite struct {
+	Host string `yaml:"host,omitempty"`
```
neoaggelos, line 56: This is not ClusterConfiguration, as it will be different on each node. Instead, change this to specify the subnet (e.g. `193.168.q.0/24`)

We either set this specify this during bootstrap or autodetect based on the default node address.

Also, make shre to use IP addresses for k8s-dqlite, as we have seen hostnames cause subtle issues in the past, especially around the interface used by dqlite.

src/k8s/pkg/k8s/setup/k8s_dqlite.go
```diff
@@ -52,9 +53,13 @@ func JoinK8sDqliteCluster(ctx context.Context, state *state.State, snap snap.Sna
...
-func LeaveK8sDqliteCluster(ctx context.Context, snap snap.Snap, hostname string) error {
-	// TODO: Get dqlite port from config
-	address := fmt.Sprintf("%s:%d", hostname, dqlite.K8sDqliteDefaultPort)
+func LeaveK8sDqliteCluster(ctx context.Context, snap snap.Snap, state *state.State) error {
+	clusterConfig, err := utils.GetClusterConfig(ctx, state)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster config: %w", err)
+	}
+
+	address := fmt.Sprintf("%s:%d", state.Address().Hostname(), clusterConfig.K8sDqlite.Port)
```
neoaggelos, line 62: here, you would use the subnet configuration to find the node IP that matches.

Its fine if we leave the address as-is for now, and rely on the dqlite default to work well

neoaggelos (changes requested):
Needs some changes related to how we pick the dqlite address

src/k8s/pkg/k8s/setup/k8s_dqlite.go
bschimke95: Alright, let's just leave the address out of the cluster config for now.

neoaggelos (approved):

PR #104, opened 2024-02-04
Convert to classic snap

neoaggelos:
### Summary

Convert `k8s` to be a classic snap. Maintain the strict snap under a separate `autoupdate/strict` branch, which is automatically updated on each push to main.

This PR includes some cleanup and path changes so that both classic and strict work without issues.

### Branch and CI changes

- Drop `tmate` from CI (it should be added ad-hoc in our actions when needed. Otherwise it causes CI to delay failures by 10 minutes.
- Add a separate end to end tests action that applies the strict patch, builds the snap and runs the e2e tests for `ubuntu:20.04`. This ensures that the strict patch can always be applied, and that we can gate PRs to `main` if they break strict.  
- Run CI on pushes to `main` (for classic) and `autoupdate/strict` (for strict) branches.
- On every push to `main`, a CI job is running that applies `build-scripts/patches/strict/*.patch` and force-pushes to `autoupdate/strict`.

As a result:

- We must ensure that the strict patch is as minimal as possible, such that conflicts are rare or easy to resolve
- Any changes to the strict patch must be done on `build-scripts/patches/strict/*.patch` with PRs to the `main` branch.

### Changes

- `network-requirements.sh` and `connect-interfaces.sh` scripts are only needed for strict confinement. In the e2e tests, call them through a new `init.sh` script. For classic, this script exists but is a no-op, such that we don't need any changes in the tests themselves.
- Followup #95 with more path changes, such that configs are consistent. The following rule is applied: We make sure to use "well-known" Kubernetes paths for all "user-facing" things like `/etc/cni/net.d`, `/opt/cni/bin`, `/var/lib/kubelet` etc. In classic, these break the snap confinement but we do so either way, otherwise we create lots of compatibility troubles with upstream manifests. Everything else (e.g. containerd, k8s-dqlite, k8sd, other configs) goes under `$SNAP_COMMON`.
- For the change above, there's quite some code sprinkled around with hard-coded paths. This is to avoid breaking function definitions and thus minimize conflicts with #102. These TODOs will be dealt with in the upcoming cleanup of the `setup` package and cluster bootstrapping work, and are irrelevant for this PR.
- Adds a few `mkdir` in places, needed in classic snap as the directories are not there by default (this was not a problem with strict, since the `layout` part of snapcraft.yaml ensure that directories already existed.
- Fixes `*Client.IsBootstrapped()` to use microcluster's `Status()` instead of `ClusterStatus()`, because for the latter the UX is clanky as we rely on the API call to timeout after 3 seconds to decide.
- e2e tests exit on first failure, saving up time for failed tests and making for quicker iterations. This should be changed when we do daily/weekly CI runs in the future.


bschimke95 (approved):
LGTM

PR #106, opened 2024-02-05
Added loadbalancer support

berkayoz:
Added loadbalancer support via Cilium.
Added missing disable command entries


build-scripts/charts/ck-loadbalancer/values.yaml
neoaggelos, line 1: Could we easily add a `values.schema.json` to validate the values?  https://helm.sh/docs/topics/charts/#the-chart-file-structure

build-scripts/charts/ck-loadbalancer/templates/bgp-policy.yaml
```diff
@@ -0,0 +1,18 @@
...
+ virtualRouters:
+ - localASN: {{ .Values.bgp.localASN }}
+   serviceSelector:
+    matchExpressions:
+    - {key: somekey, operator: NotIn, values: ['never-used-value']}
+  {{- with .Values.bgp.neighbors }}
+   neighbors:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+{{- end}}
```
neoaggelos, line 18: newline at EOF, also elsewhere

k8s/components/charts/ck-loadbalancer-0.1.0.tgz
neoaggelos, line 1: as discussed offline, let's just move the chart sources under `k8s/components/charts/ck-loadbalancer`, so this would not be needed

k8s/components/components.yaml
```diff
@@ -16,3 +16,7 @@ gateway:
   release: "ck-gateway"
   chart: "gateway-api-0.7.1.tgz"
   namespace: "kube-system"
+loadbalancer:
+  release: "ck-loadbalancer"
+  chart: "ck-loadbalancer-0.1.0.tgz"
```
neoaggelos, line 21: then set the directory here

src/k8s/api/v1/component.go
```diff
@@ -48,6 +48,24 @@ type UpdateGatewayComponentRequest struct {
...
+// LoadBalancerComponentConfig holds the configuration values for the LoadBalancer component.
+type LoadBalancerComponentConfig struct {
+	CIDRs          []string `json:"cidrs,omitempty"`
+	L2Enabled      bool     `json:"l2Enabled,omitempty"`
+	L2Interfaces   []string `json:"l2Interfaces,omitempty"`
+	BGPEnabled     bool     `json:"bgpEnabled,omitempty"`
+	BGPLocalASN    int      `json:"bgpLocalAsn,omitempty"`
+	BGPPeerAddress string   `json:"bgpPeerAddress,omitempty"`
+	BGPPeerASN     int      `json:"bgpPeerAsn,omitempty"`
+	BGPPeerPort    int      `json:"bgpPeerPort,omitempty"`
```
neoaggelos, line 66: it's fine to have a shallow config, or we could do `BGP` and `L2` blocks. up to you, this is also OK.

src/k8s/cmd/k8s/k8s_disable_storage.go
```diff
@@ -0,0 +1,36 @@
...
+		}
+
+		request := api.UpdateStorageComponentRequest{
+			Status: api.ComponentDisable,
+		}
+
+		err = client.UpdateStorageComponent(cmd.Context(), request)
+		if err != nil {
+			return fmt.Errorf("failed to disable Storage component: %w", err)
+		}
```
neoaggelos, line 31: nit, when no other return values are used later in the code, do the check inline. Also in other places

```
		if err := client.UpdateStorageComponent(cmd.Context(), request); err != nil {
			return fmt.Errorf("failed to disable Storage component: %w", err)
		}
```

src/k8s/pkg/component/loadbalancer.go
```diff
@@ -0,0 +1,129 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/k8s/pkg/utils/k8s"
+)
+
+func EnableLoadBalancerComponent(s snap.Snap, cidrs []string, l2Enabled bool, l2Interfaces []string, bgpEnabled bool, bgpLocalASN int, bgpPeerAddress string, bgpPeerASN int, bgpPeerPort int) error {
+	manager, err := NewManager(s)
```
neoaggelos, line 13: This is now `NewHelmClient(s, nil)`

tests/e2e/tests/test_loadbalancer.py
```diff
@@ -0,0 +1,129 @@
...
+    instance = instances[0]
+    tester_instance = instances[1]
+
+    lb_cidr = ""
+
+    p = instance.exec("ip -o -4 route show to default".split(" "), capture_output=True)
+    instance_default_ip = p.stdout.decode().split(" ")[8]
+
+    p = tester_instance.exec(
+        "ip -o -4 route show to default".split(" "), capture_output=True
```
neoaggelos, line 27: this is an attempt to figure out a CIDR for the loadbalancer, right? could you move it to a helper function with a comment?

Also, please prefer `["cmd", "paths"]` instead of `"cmd paths".split(" ")`

That way, we can return to it in the future (perhaps even set via config like TEST_LOADBALANCER_CIDR).

neoaggelos (commented):
Really nice work, did a first pass

tests/e2e/tests/test_loadbalancer.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+@pytest.mark.node_count(2)
+def test_loadbalancer(instances: List[harness.Instance]):
+    instance = instances[0]
+    tester_instance = instances[1]
+
+    lb_cidr = ""
+
+    p = instance.exec("ip -o -4 route show to default".split(" "), capture_output=True)
```
neoaggelos, line 23: I know it's tedious, but please use `["ip", "-o", ...]` instead. also in other places

src/k8s/api/v1/component.go
neoaggelos: let's leave as is for v0, we can extend later _if needed_

src/k8s/pkg/component/loadbalancer.go
```diff
@@ -0,0 +1,129 @@
...
+		// https://docs.cilium.io/en/v1.14/network/l2-announcements/#sizing-client-rate-limit
+		// Assuming for 50 LB services
+		"k8sClientRateLimit": map[string]any{
+			"qps":   10,
+			"burst": 20,
+		},
+	}
+
+	err = manager.Refresh("network", networkValues)
+	if err != nil {
```
neoaggelos, line 37: also in other places in this file, when you don't need other returned values

```suggestion
	if err := manager.Refresh("network", networkValues); err != nil {
```

neoaggelos (commented):
Nice work @berkayoz! Just a final set of changes and we're good to go!

src/k8s/pkg/k8sd/api/component.go
```diff
@@ -147,3 +147,40 @@ func putGatewayComponent(s *state.State, r *http.Request) response.Response {
...
+			req.Config.CIDRs,
+			req.Config.L2Enabled,
+			req.Config.L2Interfaces,
+			req.Config.BGPEnabled,
+			req.Config.BGPLocalASN,
+			req.Config.BGPPeerAddress,
+			req.Config.BGPPeerASN,
+			req.Config.BGPPeerPort,
+		)
+		if err != nil {
```
neoaggelos, line 173: also here

src/k8s/pkg/k8sd/api/component.go
```diff
@@ -147,3 +147,40 @@ func putGatewayComponent(s *state.State, r *http.Request) response.Response {
...
+			return response.SmartError(fmt.Errorf("failed to enable loadbalancer: %w", err))
+		}
+	case api.ComponentDisable:
+		err = component.DisableLoadBalancerComponent(snap)
+		if err != nil {
```
neoaggelos, line 178: also here

neoaggelos (approved):
LGTM! Two last changes, then we can merge!

PR #108, opened 2024-02-06
Move `ClusterConfig` to internal types package

neoaggelos:
### Summary

This is pre-requisites for driving the cluster using the `ClusterConfig` type.

Move the `ClusterConfig` type into our `k8sd/types` package. All utilities and converters from/to API types also lives inside `types`, such that the `api` package is standalone and does not depend on internal types.

Also adds unit test coverage for merging cluster configs, and fixes a bug where some values were not updated.

### Changes

This is mainly a separation of concerns.

- `pkg/types`: defines internal types and structures, regardless of how they are stored in the database or represented in the API
- `pkg/api`: defines user-facing types
- `pkg/database`: only concerned with database logic, storing and retrieving data, nothing else


src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -193,14 +192,14 @@ func onBootstrapControlPlane(s *state.State, initConfig map[string]string) error
 	clusterConfig.Certificates.CACert = string(caPair.CertPem)
 	clusterConfig.Certificates.CAKey = string(caPair.KeyPem)
 
-	clusterConfig, err = clusterconfigs.Merge(clusterConfig, utils.ConvertBootstrapToClusterConfig(bootstrapConfig))
+	clusterConfig, err = types.MergeClusterConfig(clusterConfig, types.ClusterConfigFromBootstrapConfig(bootstrapConfig))
```
bschimke95, line 195: This reads a bit weird. 
Does it make sense to have a `types/clusterconfigs` module?

bschimke95 (approved):
LGTM, only a minor comment

PR #109, opened 2024-02-06
Unify endpoints for clustering nodes

bschimke95:
Previously, a client needed to call different API endpoints to generate control plane or worker token and joining those different node types to the k8s cluster.
This commit adds a unified endpoint for creating a token (`POST /k8sd/tokens`) and joining a node (`POST /k8sd/cluster/<node>`).

This change:
 * simplifies the API and makes it less confusing for the caller
 * moves business logic from the client/CLI to `k8sd` (e.g. determining the token type)


src/k8s/api/v1/tokens.go
```diff
@@ -0,0 +1,18 @@
+package v1
+
+// WorkerNodeTokenRequest is used to request a token for joining a node to the cluster.
```
neoaggelos, line 3: `TokenRequest`

neoaggelos (commented):
Looks great. I'd avoid having to parse the URL since we have a request body already.

src/k8s/api/v1/cluster_node.go
```diff
@@ -0,0 +1,10 @@
+package v1
+
+// JoinNodeRequest is used to request to add a node to the cluster.
+type JoinNodeRequest struct {
+	Address string `json:"address"`
+	Token   string `json:"token"`
```
neoaggelos, line 6: Let's add the node name here instead of parsing through the URL?

src/k8s/pkg/k8sd/api/endpoints.go
```diff
@@ -12,12 +12,22 @@ var Endpoints = []rest.Endpoint{
...
+	},
 	{
-		Name: "WorkerToken",
-		Path: "k8sd/worker/tokens",
-		Post: rest.EndpointAction{Handler: postWorkerToken},
+		Name: "JoinNode",
+		Path: "k8sd/cluster/{node}",
+		Post: rest.EndpointAction{Handler: postClusterNode},
+		// Joining a node is a bootstrapping action which needs to be
+		// available before k8sd is initialized.
```
neoaggelos, line 27: single-line is fine

neoaggelos (approved):
LGTM

PR #110, opened 2024-02-08
Allow access to the microcluster app from rest endpoints

neoaggelos:
### Summary

Follow PR for #109

### Changes

- API endpoints are built based on the microcluster app
- We can define handlers with access to the microcluster app
- Helper methods for this
- Cleanup postClusterJoin and postClusterTokens based on this
- Make endpoint names consistent


src/k8s/api/v1/cluster_node.go
```diff
@@ -1,11 +1,11 @@
 package v1
 
-// JoinNodeRequest is used to request to add a node to the cluster.
-type JoinNodeRequest struct {
+// JoinClusterRequest is used to request to add a node to the cluster.
```
bschimke95, line 3: Thanks for the catch!
Btw, we probably need to rename things here soon again:
Looking at the CLI spec, it seems like we go with `k8s add` and `k8s join` now.

PR #115, opened 2024-02-12
Improve error message on failed lxc file push

neoaggelos:
### Summary

Improve visibility for lxc file push errors (unfortunately, some are sent to stdout instead of stderr and were dropped, making it hard to debug the reason for the test failures.)


bschimke95 (approved):
LGTM

PR #117, opened 2024-02-12
Pass context to various components

eaudetcobello:
Searched for Enable\w+Component, Disable\w+Component.


src/k8s/pkg/component/dns.go
```diff
@@ -4,13 +4,12 @@ import (
...
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/canonical/k8s/pkg/snap"
 	"github.com/canonical/k8s/pkg/utils/k8s"
 )
 
-func EnableDNSComponent(s snap.Snap, clusterDomain, serviceIP string, upstreamNameservers []string) error {
+func EnableDNSComponent(s snap.Snap, clusterDomain, serviceIP string, upstreamNameservers []string, ctx context.Context) error {
```
neoaggelos, line 12: it's idiomatic that the context be the first argument on all functions

neoaggelos: also update all other places (and caller sites).

neoaggelos, line 12:
```suggestion
func EnableDNSComponent(ctx context.Context, s snap.Snap, clusterDomain, serviceIP string, upstreamNameservers []string) error {
```

neoaggelos (commented):
Great work, let's make sure that we write idiomatic Go code and we're good to go!

neoaggelos (commented):
Needs rebase.

src/k8s/pkg/k8sd/api/component.go
```diff
@@ -43,7 +43,7 @@ func putDNSComponent(s *state.State, r *http.Request) response.Response {
 
 	switch req.Status {
 	case api.ComponentEnable:
-		dnsIP, clusterDomain, err := component.EnableDNSComponent(snap, req.Config.ClusterDomain, req.Config.ServiceIP, req.Config.UpstreamNameservers)
+		dnsIP, clusterDomain, err := component.EnableDNSComponent(s.Context, snap, req.Config.ClusterDomain, req.Config.ServiceIP, req.Config.UpstreamNameservers)
```
bschimke95, line 46: We should use the request context here and below: `r.Context`
This request context is cancelled if the user terminates the request. The state context is only cancelled if `k8sd` terminates.

bschimke95 (commented):
Great work @eaudetcobello !
Only a minor change and we are good to go.

bschimke95 (approved):
LGTM, great work!

PR #118, opened 2024-02-12
Restrict control plane actions on worker nodes

berkayoz:
Added an AccessHandler that restricts certain endpoints to workers.
Also fixed a typo on the Gateway components enable message.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_helm.go
+++ b/src/k8s/cmd/k8s/k8s_helm.go
@@ -6,6 +6,8 @@ import (
 	"os/exec"
 	"syscall"
 
+	"github.com/canonical/k8s/pkg/snap"
+	snaputil "github.com/canonical/k8s/pkg/snap/util"
 	"github.com/spf13/cobra"
 )
 
@@ -15,6 +17,17 @@ var (
 		Hidden:             true,
 		DisableFlagParsing: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			snap := snap.NewSnap(os.Getenv("SNAP"), os.Getenv("SNAP_COMMON"))
+
+			isWorker, err := snaputil.IsWorker(snap)
+			if err != nil {
+				return fmt.Errorf("failed to check if node is a worker: %w", err)
+			}
+
+			if isWorker {
+				return fmt.Errorf("this action is restricted on workers")
+			}
+
 			// Allow users to provide their own kubeconfig but
 			// fallback to the admin config if nothing is provided.
 			if os.Getenv("KUBECONFIG") == "" {
--- a/src/k8s/pkg/snap/util/node.go
+++ b/src/k8s/pkg/snap/util/node.go
@@ -1,0 +1,38 @@
+package snaputil
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/k8s/pkg/utils"
+)
+
+func IsWorker(snap snap.Snap) (bool, error) {
+	return utils.FileExists(path.Join(snap.LockFilesDir(), "worker"))
+}
+
+func MarkAsWorkerNode(snap snap.Snap, mark bool) error {
+	fname := path.Join(snap.LockFilesDir(), "worker")
+
+	if mark {
+		lock, err := os.Create(fname)
+		defer lock.Close()
+		if err != nil {
+			return fmt.Errorf("failed to mark node as worker: %w", err)
+		}
+		if err := os.Chown(fname, snap.UID(), snap.GID()); err != nil {
+			return fmt.Errorf("failed to chown %s: %w", fname, err)
+		}
+		if err := os.Chmod(fname, 0600); err != nil {
+			return fmt.Errorf("failed to chmod %s: %w", fname, err)
+		}
+	} else {
+		if err := os.Remove(fname); err != nil {
+			return fmt.Errorf("failed to unmark node as worker: %w", err)
+		}
+	}
+
+	return nil
+}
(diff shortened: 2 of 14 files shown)


src/k8s/pkg/k8sd/api/worker_handler.go
```diff
@@ -0,0 +1,26 @@
...
+func RestrictWorkers(s *state.State, r *http.Request) response.Response {
+	snap := snap.SnapFromContext(s.Context)
+
+	isWorker, err := snaputil.IsWorker(snap)
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to check if node is a worker: %w", err))
+	}
+
+	if isWorker {
+		return response.InternalError(fmt.Errorf("this action is restricted on workers"))
```
bschimke95, line 22: We should return a 403 Forbidden here instead. That way, we can give nice error messages in the CLI.

src/k8s/cmd/k8s/k8s_helm.go
```diff
@@ -15,6 +17,17 @@ var (
...
 		RunE: func(cmd *cobra.Command, args []string) error {
+			snap := snap.NewSnap(os.Getenv("SNAP"), os.Getenv("SNAP_COMMON"))
+
+			isWorker, err := snaputil.IsWorker(snap)
+			if err != nil {
+				return fmt.Errorf("failed to check if node is a worker: %w", err)
+			}
+
+			if isWorker {
+				return fmt.Errorf("this action is restricted on workers")
```
bschimke95, line 28: Can we have a sentinel error for this? e.g. `ErrDisallowed` or `ErrControlPlaneOnly` (naming is hard). (A TODO is also fine.)
Background: I'm currently working on a PR to resolve errors to human cli errors.

src/k8s/pkg/snap/util/node.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/k8s/pkg/utils"
+)
+
+func IsWorker(s snap.Snap) (bool, error) {
```
bschimke95, line 11: General question: Do we have something like a guideline when to use a util function and when to put it into the `Snap` interface?
e.g. for this function it would also make sense to have it in the interface. Otherwise you'd need to always create a lock file in the tests at `s.WorkerNodeLockFile`. If we have `Snap.IsWorker` we can just mock the return value.

bschimke95 (commented):
Great work, only two comments.

src/k8s/cmd/k8s/k8s_helm.go
berkayoz: I've put a TODO, I can address this after your PR is merged. Depending on which goes in first, feel free to ping me. Thanks!

src/k8s/pkg/snap/util/node.go
berkayoz: Summoning @neoaggelos for the discussion as we've talked about this. Afaik, the snap interface should just hold paths and configs and be simple while functional bits should go in the `snaputils`.

src/k8s/pkg/snap/util/node.go
```diff
@@ -0,0 +1,38 @@
...
+	if err := os.Chown(fname, snap.UID(), snap.GID()); err != nil {
+		return fmt.Errorf("failed to chown %s: %w", fname, err)
+	}
+	if err := os.Chmod(fname, 0700); err != nil {
+		return fmt.Errorf("failed to chmod %s: %w", fname, err)
+	}
+	return nil
+}
+
+func UnmarkAsWorkerNode(snap snap.Snap) error {
```
neoaggelos, line 32: just a nit, "unmark as worker node" does not read very well to me. could we instead have a single `func MarkWorkerNode(snap snap.Snap, mark bool) error`?

src/k8s/pkg/snap/util/node.go
```diff
@@ -0,0 +1,38 @@
...
+	fname := path.Join(snap.LockFilesDir(), "worker")
+	lock, err := os.Create(fname)
+	defer lock.Close()
+	if err != nil {
+		return fmt.Errorf("failed to mark node as worker: %w", err)
+	}
+	if err := os.Chown(fname, snap.UID(), snap.GID()); err != nil {
+		return fmt.Errorf("failed to chown %s: %w", fname, err)
+	}
+	if err := os.Chmod(fname, 0700); err != nil {
```
neoaggelos, line 26: just 0600 is ok

src/k8s/pkg/snap/util/node.go
```diff
@@ -0,0 +1,38 @@
...
+		return fmt.Errorf("failed to chmod %s: %w", fname, err)
+	}
+	return nil
+}
+
+func UnmarkAsWorkerNode(snap snap.Snap) error {
+	err := os.Remove(path.Join(snap.LockFilesDir(), "worker"))
+	if err != nil {
```
neoaggelos, line 34: if err := ...; err != nil

neoaggelos (commented):
Almost there, two final comments

src/k8s/pkg/snap/util/node.go
bschimke95: Well, this is what I thought but we also got e.g. `StartService` there.

neoaggelos: yy, a ServiceManager of sorts should grow from there, but that's something we can/should do at a later time.

neoaggelos (approved):
LGTM, final comment

bschimke95 (approved):
LGTM

PR #121, opened 2024-02-13
Retrieve Kubernetes REST Client from the snap interface

neoaggelos:
### Summary

Extend `type Snap interface` with a `KubernetesRESTClientGetter()`, and use it for initializing Kubernetes clients.

### Changes

- Also cleanup some `if err := ...; err != nil {}` checks in files.


bschimke95 (approved):
LGTM

PR #122, opened 2024-02-13
Fix bootstrap config conversion

louiseschmidtgen:
### Description
Currently, any bootstrap config specified during k8s bootstrap is ignored, because the marshal/unmarshal implementation is broken.
This PR ensures that the conversion of the bootstrap from a map to yaml and back works.

There is a bug because we need tags to ensure proper marshaling and unmarshaling of the bootstrap config.

Adding the yaml tags solves our problem:
```
type BootstrapConfig struct {
  Components []string `yaml:"components"`
  ClusterCIDR string `yaml:"cluster-cidr"`
}
```

### DIFF
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -10,9 +10,9 @@ import (
 
 type BootstrapConfig struct {
 	// Components are the components that should be enabled on bootstrap.
-	Components []string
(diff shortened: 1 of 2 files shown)


neoaggelos (approved):
LGTM, thanks @louiseschmidtgen! Just a minor comment

PR #123, opened 2024-02-13
Add worker node removal

bschimke95:
Unifies control-plane and worker node removal under one endpoint `/1.0/k8sd/cluster/remove`. Also, nodes are now drained and removed from kubernetes.


src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -35,6 +37,12 @@ var (
 				return fmt.Errorf("k8s cluster already bootstrapped")
 			}
 
+			const minTimeout = 3 * time.Second
```
bschimke95, line 40: Sometimes `k8s bootstrap` times out with 30s on low machines. Added the `--timeout` flag and by that, increased the timeout to 90s.
(I still would advocate for a generic timeout flag for all commands ...)

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -0,0 +1,65 @@
...
+		c, err := k8s.NewClient(snap)
+		if err != nil {
+			return response.SmartError(fmt.Errorf("failed to create k8s client: %w", err))
+		}
+
+		if err := c.GracefullyDeleteNode(s.Context, req.Name); err != nil {
+			return response.SmartError(fmt.Errorf("failed to remove k8s node %q: %w", req.Name, err))
+		}
+
+		if err := utils.DeleteWorkerNodeEntry(r.Context(), s, req.Name); err != nil {
```
bschimke95, line 56: @neoaggelos we should probably introduce a `Databaser` interface for the database queries so that we can mock them.

src/k8s/pkg/k8s/client/cluster_node.go
```diff
@@ -22,20 +21,28 @@ func (c *Client) JoinCluster(ctx context.Context, name string, address string, t
 		Address: address,
 		Token:   token,
 	}
-	var response apiv1.JoinClusterResponse
-	err := c.mc.Query(ctx, "POST", api.NewURL().Path("k8sd", "cluster", "join"), request, &response)
+	err := c.mc.Query(ctx, "POST", api.NewURL().Path("k8sd", "cluster", "join"), request, nil)
 	if err != nil {
-		fmt.Fprintln(os.Stderr, "Failed with error:", err)
-
+		fmt.Println("Cleaning up, error was", err)
```
neoaggelos, line 26: this should still go to stderr i think

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -0,0 +1,65 @@
...
+	snap := snap.SnapFromContext(s.Context)
+
+	req := apiv1.RemoveNodeRequest{}
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return response.BadRequest(fmt.Errorf("failed to parse request: %w", err))
+	}
+
+	isControlPlane, err := utils.IsControlPlaneNode(r.Context(), s, req.Name)
+	if err != nil {
+		return response.SmartError(fmt.Errorf("failed to check if node is control-plane: %w", err))
```
neoaggelos, line 27: Use `response.InternalError` for these, also below

src/k8s/pkg/utils/k8s/node.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+)
+
+// DrainNode drains the specified node by evicting its pods gracefully.
+func (c *Client) DrainNode(ctx context.Context, node string) error {
+	pods, err := c.CoreV1().Pods("").List(ctx, metav1.ListOptions{
+		FieldSelector: "spec.nodeName=" + node,
```
neoaggelos, line 14: use a structured field selector here

src/k8s/pkg/utils/node.go
```diff
@@ -0,0 +1,43 @@
...
+		return false, fmt.Errorf("failed to list worker nodes: %w", err)
+	}
+
+	for _, worker := range workers {
+		if worker == name {
+			return true, nil
+		}
+	}
+	return false, nil
+}
```
neoaggelos, line 43: would this be better done with a query on the database, rather than fetching all workers and iterating at the user side?

src/k8s/pkg/utils/k8s/node.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	for _, pod := range pods.Items {
+		if pod.Namespace == "kube-system" {
+			continue
+		} else {
+			if err := c.EvictPod(ctx, pod.Namespace, pod.Name); err != nil {
+				return fmt.Errorf("failed to evict pod %s from namespace %s: %w", pod.Name, pod.Namespace, err)
+			}
+		}
+	}
```
neoaggelos, line 27: this scares me a bit, is this our responsibility?

neoaggelos (commented):
Did a first pass, really nice work overall

src/k8s/pkg/utils/k8s/node.go
bschimke95: obsolete now

src/k8s/pkg/utils/k8s/node.go
bschimke95: let's not do this - as discussed offline.

neoaggelos (approved):

PR #124, opened 2024-02-13
Enable NodeRescript admission plugin

addyess:
enabling the `NodeRestriction` plugin to put restrictions on the kubelet
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction

Among such restrictions are:

> kubelets will only be allowed to modify their own Node API object, and only modify Pod API objects that are bound to their node. kubelets are not allowed to update or remove taints from their Node API object

> Prevents kubelets from adding/removing/updating labels with a node-restriction.kubernetes.io/ prefix. This label prefix is reserved for administrators to label their Node objects for workload isolation purposes, and kubelets will not be allowed to modify labels with that prefix.

### DIFF
--- a/k8s/args/kube-apiserver
+++ b/k8s/args/kube-apiserver
@@ -23,7 +23,7 @@
 --proxy-client-cert-file=/etc/kubernetes/pki/front-proxy-client.crt
 --proxy-client-key-file=/etc/kubernetes/pki/front-proxy-client.key
 #~Enable the aggregation layer
-#--enable-admission-plugins=EventRateLimit
+--enable-admission-plugins=NodeRestriction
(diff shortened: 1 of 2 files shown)


neoaggelos (changes requested):
This should be set in https://github.com/canonical/k8s-snap/blob/f283cd0a5ad1f3848eaa102ee6fdfcfea646861f/src/k8s/pkg/k8sd/setup/kube_apiserver.go#L32 instead

addyess:
> This should be set in
> 
> https://github.com/canonical/k8s-snap/blob/f283cd0a5ad1f3848eaa102ee6fdfcfea646861f/src/k8s/pkg/k8sd/setup/kube_apiserver.go#L32
> instead

Dones

neoaggelos (approved):

PR #127, opened 2024-02-13
[Doc] Adding and removing nodes

eaudetcobello:


docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,98 @@
+# Adding and Removing Nodes
+
+## What you will need
+- A bootstrapped cluster using Canonical K8s (See (getting-started.md)[Getting Started]
+- The `dns` and `network` components enabled on your cluster.
```
louiseschmidtgen, line 5: add installed multipass and a link to the docs: https://multipass.run/install

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+Make sure you can see `k8s is ready.` in the output of the command.
+
+### 2. Start a Multipass VM
+
+Our second node will live on a Multipass VM. Let's launch it:
+
+```
+multipass launch --name expert-elf
```
louiseschmidtgen, line 22: Explain what this command does something like: This command uses `multipass` to launch a virtual machine, running ubuntu 22.04, with the name `expert-elf`. 
You could add `multipass list` to verify the command was successful.

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+**Note**: This step can take a few minutes. It's normal and expected.
+
+### 3. Generate a worker node token from the control plane node
+
+In the shell connected to your control plane node, run the following command to get a token that'll be used by the other node to join the cluster.
```
louiseschmidtgen, line 29: Let's say something like: From a shell in your bootstrapped (add link here again to bootstrapping your cluster) cluster we create a token that is used to allow other nodes to join the cluster.
Briefly, explain the difference between the worker and the control plane node and say in this example we are adding a worker node.

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,98 @@
...
+
+**If you want to create a worker node, append the `--worker` argument.**
+
+```
+sudo k8s add-node expert-elf [--worker]
```
louiseschmidtgen, line 34: Let's remove the brackets around `--worker` here.

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,98 @@
...
+multipass shell expert-elf
+```
+
+The VM doesn't come with the `k8s` snap, so let's install it.
+
+```
+sudo snap install --edge --classic k8s
+```
+
+From the worker node machine, use the `join-cluster` command and pass the token as the last argument.
```
louiseschmidtgen, line 55: make this: and paste **your** join token as the last argument

louiseschmidtgen (changes requested):
Great Docs! @eaudetcobello 
I enjoy that you use the `expert-elf` name it helps with confusion in which vm to run what command.
I would suggest to run through all the steps once more and check that everything works as expected.

docs/src/tutorial/add-remove-nodes.md
eaudetcobello: The k8s status is to be run on the control plane node, my mistake.

docs/src/tutorial/add-remove-nodes.md
eaudetcobello: Agreed for indicating it'll be an Ubuntu 22.04 VM. For the second part, Multipass already outputs `Launched: expert-elf`

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,98 @@
...
+Open a shell to the worker node:
+
+```
+multipass shell expert-elf
+```
+
+The VM doesn't come with the `k8s` snap, so let's install it.
+
+```
+sudo snap install --edge --classic k8s
```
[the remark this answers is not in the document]
eaudetcobello: Don't need those components to join a cluster.

```
ubuntu@expert-santa:~$ sudo snap install --classic --edge k8s
k8s (edge) v1.29.1 from Konstantinos Tsakalozos (kjackal) installed
ubuntu@expert-santa:~$ sudo k8s status
Error: failed to get cluster status: wait check failed: Daemon not yet initialized
ubuntu@expert-santa:~$ sudo k8s join-cluster eyJ0b2tlbiI6Indvcmtlcjo6MmJkYzUzOWEwYzlkNGU3NzNiNzM5NGFjNmIyZDc3N2YwYWE3YzE2YSIsImpvaW5fYWRkcmVzc2VzIjpbIjE5Mi4xNjguMC4xNTk6NjQwMCJdLCJfIjoibSEhIn0
Joined the cluster.
```

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,116 @@
...
+
+**Note**: It's best to give the new node the same name as the hostname of the 
+worker node (in this case the VM).
+
+A base64 string should be printed to your terminal. Keep it close as you 
+will need it in a few steps.
+
+### 4. Join the cluster from the other node
+
+Open a shell to the worker node:
```
evilnick, line 51:
```suggestion
Open a shell on the worker node:
```
small nit

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+Make sure you can see `k8s is ready.` in the output of the command.
+
+### 2. Start a Multipass VM
+
+Our second node will live on Ubuntu 22.04 on a UMultipass VM. Let's launch it:
+
+```
+multipass launch --name expert-elf
```
evilnick, line 23:
```suggestion
multipass launch 22.04 --name expert-elf
```
As we explicitly claim a 22.04 image I think we should make sure we explicitly get one. In a few months they'll be getting 24.04 instead

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,116 @@
...
+### 4. Join the cluster from the other node
+
+Open a shell to the worker node:
+
+```
+multipass shell expert-elf
+```
+
+The VM doesn't come with the `k8s` snap, so let's install it.
+
```
evilnick, line 58: I think we should be careful to make sure that they launch a node running the same k8s as the existing one, and perhaps that should be explained.

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,116 @@
...
+### 5. Delete the expert-elf VM (Optional)
+
+Two commands are needed to delete the expert-elf VM from your system:
+
+```
+multipass remove expert-elf
+multipass purge
+```
+
+### 6. Remove Canonical Kubernetes (Optional)
```
evilnick, line 91: As with the other How to, we don't need this step.

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,116 @@
...
+This command removes the `k8s` snap and automatically creates a snapshot of all data for future restoration.
+
+If you wish to remove the snap without saving a snapshot of its data, add `--purge` to the command:
+
+```
+sudo snap remove k8s --purge
+```
+This option ensures complete removal of the snap and its associated data.
+
+## Next Steps
```
evilnick, line 108: maybe for this one include a reference to multipass, one to the architecture doc we haven't written yet

evilnick (changes requested):
Nice! I have left some comments

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,116 @@
...
+
+```
+sudo k8s status
+```
+
+Make sure you can see `k8s is ready.` in the output of the command.
+
+### 2. Start a Multipass VM
+
+Our second node will live on Ubuntu 22.04 on a UMultipass VM. Let's launch it:
```
neoaggelos, line 20: typo 
```suggestion
Our second node will live on Ubuntu 22.04 on a Multipass VM. Let's launch it:
```

docs/src/tutorial/add-remove-nodes.md
```diff
@@ -0,0 +1,116 @@
...
+sudo k8s join-cluster eyJu...XX0=
+```
+
+After a couple of seconds, you should see: `Joined the cluster.`
+
+Disconnect from the VM and run the `status` command on your control plane node:
+
+```
+sudo k8s status
+```
```
neoaggelos, line 76: This also reads a bit weird. It makes sense for the current setup (control plane runs on the node, then the worker runs on a multipass VM), but that is nowhere near a practical setup for production clusters.

neoaggelos (commented):
As a generic note, it does not seem proper to me that our initial node resides "somewhere" and that our worker node _needs_ multipass. I would expect this how-to to start with:

"You need 2 (or more) machines that you can access via SSH with a root user. For development, you can consider [multipass](link-to-multipass) to quickly spin up VMs on Windows, Linux or macOS. In this how-to, we'll spin up multipass VMs for the instances"

Also, both the control plane and the worker node should be on the _same_ infra. I would not expect the user to have to jump to another guide to start from an undefined state. I'd rather we spin up two multipass instances (or tell the user to use two machines), install `k8s` and bootstrap as control plane on the first, then install `k8s` on the second and join as a worker.

docs/src/tutorial/add-remove-nodes.md
eaudetcobello: What do you suggest instead ?

evilnick:
@neoaggelos yes, that's a good point. I wonder if we should prefer LXD over multipass?

neoaggelos:
> I wonder if we should prefer LXD over multipass?

both are fine for me. I'd personally prefer LXD as well (VMs or containers), but I'm okay with multipass since it also works for Windows and macOS. At the end of the day, it should not matter, but let's make sure that we do the same everywhere (and not have half the docs be "install multipass, then .." and the other "configure lxd, then ..."

docs/src/tutorial/add-remove-nodes.md
eaudetcobello: So two references ?

evilnick (approved):
:cake:

louiseschmidtgen (approved):

PR #130, opened 2024-02-14
Add machinery to make errors human-readable

bschimke95:
We face four different kinds of errors in the k8s-snap. This commit adds machinery to handle all of them:

On the server-side there are:

1. microcluster errors: errors that directly come from microcluster. They are untyped and we need to rely on the error string to handle them.

2. server-side errors: error that happen in our system and that we can formally define.

The errors are not preserved on the client. Microcluster simply sends the error message and creates a new error on the client. This means we cannot rely on error types by default. But we can define custom errors in `api/v1/errors.go` that support a direct string comparison by implementing the `Error` and `Is` interfaces.

For server-side errors we can simply use the exposed error types and the string comparison will always work. The error is basically just serialized on the server and deserialized on the client.

For the microcluster errors we need to be careful on library upgrades. The error messages are not recognized anymore, if microcluster changes the error messages. We can avoid this by adding strict integration tests that validate the explicit CLI output.

On the client-side there are:

3. generic client-side errors: errors that pop up on the client/CLI, e.g. cannot connect to k8sd and apply for all commands.

4. custom client-side errors: errors that are unique for a specific command, e.g. ErrAlreadyBootstrapped.

This commit generically resolve those error by defining an error wrapper (`cmd/k8s/errors/wrapper.go`) that is injected at the root `k8s` CLI command and tries to resolve any error to a human-readable message. Generic errors translations are defined in the wrapper module, custom error messages can be defined in the command (see e.g. `k8s_bootstrap.go`)

### DIFF
--- a/src/k8s/api/v1/errors.go
+++ b/src/k8s/api/v1/errors.go
@@ -1,0 +1,20 @@
+package v1
+
+import (
+	"github.com/canonical/k8s/pkg/utils/errors"
+)
+
+var Errors = []error{
+	&ErrNotBootstrapped{},
+}
+
+// Server-side errors
+type ErrNotBootstrapped struct{}
+
+func (e ErrNotBootstrapped) Error() string {
+	return "no such file or directory"
+}
+
+func (e ErrNotBootstrapped) Is(err error) bool {
+	return errors.DeeplyUnwrapError(err).Error() == e.Error()
+}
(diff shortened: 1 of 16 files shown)


bschimke95:
Closing in favour of: https://github.com/canonical/k8s-snap/pull/131

PR #131, opened 2024-02-14
Refactor CLI to make it testable.

bschimke95:
We face four different kinds of errors in the k8s-snap. This commit adds machinery to handle all of them:

On the server-side there are:

microcluster errors: errors that directly come from microcluster. They are untyped and we need to rely on the error string to handle them.

server-side errors: error that happen in our system and that we can formally define.

The errors are not preserved on the client. Microcluster simply sends the error message and creates a new error on the client. This means we cannot rely on error types by default. But we can define custom errors in api/v1/errors.go that support a direct string comparison by implementing the Error and Is interfaces.

For server-side errors we can simply use the exposed error types and the string comparison will always work. The error is basically just serialized on the server and deserialized on the client.

For the microcluster errors we need to be careful on library upgrades. The error messages are not recognized anymore, if microcluster changes the error messages. We can avoid this by adding strict integration tests that validate the explicit CLI output.

On the client-side there are:

generic client-side errors: errors that pop up on the client/CLI, e.g. cannot connect to k8sd and apply for all commands.

custom client-side errors: errors that are unique for a specific command, e.g. ErrAlreadyBootstrapped.

This commit generically resolve those error by defining an error wrapper (cmd/k8s/errors/wrapper.go) that is injected at the root k8s CLI command and tries to resolve any error to a human-readable message. Generic errors translations are defined in the wrapper module, custom error messages can be defined in the command (see e.g. k8s_bootstrap.go)

Also: 

Puts each CLI command into a function that expects a k8s client. 
Provides a mock implementation of the client so that unit tests for the CLI can be implemented.


src/k8s/cmd/k8s/main.go
```diff
@@ -1,11 +1,14 @@
...
+	"os"
+)
 
 func Main() {
-	if rootCmd.Execute() != nil {
-		// TODO: We need to define actionable error message in the future
-		// That tell the user what went wrong on a high-level and - if possible - how it can be fixed.
+	rootCmd := NewRootCmd()
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Fprintln(rootCmd.ErrOrStderr(), err)
```
neoaggelos, line 11: could it be simpler if we just handle the error here? i'd rather not introduce an error handler that we have to remember to use all our commands with.

It should be enough to have the command's `RunE()` return the error, then call the logic of `resolveError()` here.

neoaggelos (commented):
I see the value in the proposed changes. Left a comment to perhaps make this a bit simpler. No need for an extra errorhandler that mutates the cmd and then wraps the execution: Let's just build the commands as needed (e.g. set `SilenceUsage: true` on all) and run them. They already return an error, and we can handle that error appropriately in `main.go`

src/k8s/cmd/k8s/main.go
neoaggelos: Also, the client should not be initialized during `init()`, but rather as a step for each command. It's okay to keep some _cursed_ global state for this, and have a ConsistentPreRun on commands that need to initialize the client. Let's consider how we can inject a mock at a later time. If it's one place that initializes the client, it will be easy to do.

neoaggelos: The client creation also blocks the e2e tests atm, and the error there makes sense (avoid initializing _runtime_ things that we might not even need in all commands during `init()`)

src/k8s/cmd/k8s/k8s.go
```diff
@@ -11,9 +12,13 @@ var (
...
 		logDebug   bool
 		logVerbose bool
+		stateDir   string
 	}
+	k8sdClient client.Client
+)
 
-	rootCmd = &cobra.Command{
+func NewRootCmd() *cobra.Command {
+	rootCmd := &cobra.Command{
```
neoaggelos, line 21: do we still need the `New$command()` and `new$COmmand()`? I'm okay with it, or let's remove if you don't think it's useful now.

neoaggelos (approved):
Code structure LGTM, excellent work @bschimke95!

Can't speak about the error messages themselves, they look useful to me but we might want to get some input from more people?

src/k8s/cmd/k8s/k8s.go
bschimke95: Let's keep it - we can use it to have isolated unit-tests later.

PR #132, opened 2024-02-15
Initial cluster config is created from bootstrap config and defaults

louiseschmidtgen:
### Description

Currently, when we run `k8s bootstrap --interactive` and we provide a cluster CIDR, we create the defaultconfig and then change the default podCIDR to the new podCIDR. This results in an error as changing podCIDR is not allowed. 

The PR runs a `SetDefaults` to set the default podCIDR after the two configs are merged, whereby the default cluster config does not set the podCIDR yet.

Additionally, there is a method to verify the CIRD from the bootstrap config is valid.
This PR also implements a test for `SetDefaults`.

### DIFF
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -33,6 +33,11 @@ func (b *BootstrapConfig) ToMap() (map[string]string, error) {
 	}, nil
 }
 
+func (b *BootstrapConfig) IsValidCIDR() bool {
+	_, _, err := net.ParseCIDR(b.ClusterCIDR)
+	return err == nil
+}
+
 // BootstrapConfigFromMap converts a string map to a BootstrapConfig struct.
 func BootstrapConfigFromMap(m map[string]string) (*BootstrapConfig, error) {
 	config := &BootstrapConfig{}
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -33,22 +33,6 @@ func (b *BootstrapConfig) ToMap() (map[string]string, error) {
 	}, nil
 }
 
-func (b *BootstrapConfig) Valid() error {
-	clusterCIDRs := strings.Split(b.ClusterCIDR, ",")
-	if len(clusterCIDRs) != 1 && len(clusterCIDRs) != 2 {
-		return fmt.Errorf("invalid number of cluster CIDRs: %d", len(clusterCIDRs))
-	}
-
-	for _, cidr := range clusterCIDRs {
-		_, _, err := net.ParseCIDR(cidr)
-		if err != nil {
-			return fmt.Errorf("invalid CIDR: %w", err)
-		}
-	}
-
-	return nil
-}
-
 // BootstrapConfigFromMap converts a string map to a BootstrapConfig struct.
 func BootstrapConfigFromMap(m map[string]string) (*BootstrapConfig, error) {
 	config := &BootstrapConfig{}
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -33,25 +33,20 @@ func (b *BootstrapConfig) ToMap() (map[string]string, error) {
 	}, nil
 }
 
-func (b *BootstrapConfig) IsValidCIDR() bool {
+func (b *BootstrapConfig) Valid() error {
 	clusterCIDRs := strings.Split(b.ClusterCIDR, ",")
-	if v := len(clusterCIDRs); v != 1 && v != 2 {
-		return false
+	if len(clusterCIDRs) != 1 && len(clusterCIDRs) != 2 {
+		return fmt.Errorf("invalid number of cluster CIDRs: %d", len(clusterCIDRs))
 	}
 
 	for _, cidr := range clusterCIDRs {
-		_, parsed, err := net.ParseCIDR(cidr)
-		switch {
-		case err != nil:
-			return false
-		case parsed.IP.To4() != nil:
-			return true
-		default:
-			return true
+		_, _, err := net.ParseCIDR(cidr)
+		if err != nil {
+			return fmt.Errorf("invalid CIDR: %w", err)
 		}
 	}
 
-	return false
+	return nil
 }
 
 // BootstrapConfigFromMap converts a string map to a BootstrapConfig struct.
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -34,8 +34,24 @@ func (b *BootstrapConfig) ToMap() (map[string]string, error) {
 }
 
 func (b *BootstrapConfig) IsValidCIDR() bool {
-	_, _, err := net.ParseCIDR(b.ClusterCIDR)
-	return err == nil
+	clusterCIDRs := strings.Split(b.ClusterCIDR, ",")
+	if v := len(clusterCIDRs); v != 1 && v != 2 {
+		return false
+	}
+
+	for _, cidr := range clusterCIDRs {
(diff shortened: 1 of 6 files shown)


src/k8s/pkg/k8sd/types/cluster_config.go
```diff
@@ -49,8 +53,8 @@ type K8sDqlite struct {
 	Port int `yaml:"port,omitempty"`
 }
 
-func DefaultClusterConfig() ClusterConfig {
-	return ClusterConfig{
+func SetClusterConfigDefaults(b *apiv1.BootstrapConfig) (ClusterConfig, error) {
+	config := ClusterConfig{
```
neoaggelos, line 57: This should be a `(*ClusterConfig).SetDefaults() error`, that would look like this:

```go
func (c *ClusterConfig) SetDefaults() {
  if c.Network.PodCIDR == "" {
    c.Network.PodCIDR = "10.1.0.0/16"
  }
}
```

src/k8s/pkg/k8sd/types/cluster_config.go
```diff
@@ -64,14 +68,13 @@ func DefaultClusterConfig() ClusterConfig {
 			Port: 9000,
 		},
 	}
-}
 
-// ClusterConfigFromBootstrapConfig extracts the cluster config parts from the BootstrapConfig
-// and maps them to a ClusterConfig.
-func ClusterConfigFromBootstrapConfig(b *apiv1.BootstrapConfig) ClusterConfig {
```
neoaggelos, line 71: Let's keep this function here, this should give us a ClusterConfig seeded from the BootstrapConfig.

Also, two CIDRs (one IPv4, one IPv6) may be specified, so if we want to validate it, use https://github.com/canonical/k8s-snap/blob/7e04e1b16dbb252052e56c237f3449d85eca7212/src/k8s/pkg/component/network.go#L18-L38 for inspration.

```
func ClusterConfigFromBootstrapConfig(b *apiv1.BootstrapConfig) (ClusterConfig, error) {
  if _, _, err := net.ParseCIDR(b.ClusterCIDR); err != nil {
    return ClusterConfig{}, fmt.Errorf("cluster CIDR is not valid: %w", err)
  }
  return ClusterConfig{ .... }, nil
}
```

neoaggelos (commented):
The work is in the right direction, left some comments

src/k8s/pkg/k8sd/types/cluster_config.go
louiseschmidtgen: Thanks Angelos,
the validating CIDR is updated 👍🏼 
I felt like the validation part might be better placed after we get the bootstrap config interactively. Let me know your thoughts on that.

src/k8s/pkg/k8sd/types/cluster_config.go
louiseschmidtgen: Did you want me to keep the merging part of the default config and bootstrap config and then set this default?

src/k8s/pkg/k8sd/types/cluster_config.go
neoaggelos: yeah, probably best to have a `(c *ClusterConfig) Validate() error`

src/k8s/pkg/k8sd/types/cluster_config.go
neoaggelos: usage should look like this:

```
// NOTE: must include error checking
clusterConfig := ClusterConfigFromBootstrapConfig(bootstrapConfig)
clusterConfig.SetDefaults()
clusterConfig.Validate()
```

src/k8s/api/v1/types.go
```diff
@@ -33,6 +33,27 @@ func (b *BootstrapConfig) ToMap() (map[string]string, error) {
 	}, nil
 }
 
+func (b *BootstrapConfig) IsValidCIDR() bool {
```
bschimke95, line 36: Looks like this can be simplified. No need for the switch.

```
func (b *BootstrapConfig) IsValidCIDR() bool {
	clusterCIDRs := strings.Split(b.ClusterCIDR, ",")

	if len(clusterCIDRs) != 1 && len(clusterCIDRs) != 2 {
		return false
	}

	for _, cidr := range clusterCIDRs {
		_, _, err := net.ParseCIDR(cidr)
		if err != nil {
			return false
		}
	}

	return true
}

File: src/k8s/cmd/k8s/k8s_bootstrap.go [hunk 4]
```diff
@@ -42,6 +42,10 @@ var (
 				config.SetDefaults()
 			}
 
+			if !config.IsValidCIDR() {
```
bschimke95, line 45: Should we just have server-side validation? 
Seems cumbersome to implement those validation steps on all clients (e.g. the charm would also need to do this).

Just check on the server and send back a validation error.
(We can also do this as a follow-up step.)

src/k8s/pkg/k8sd/types/cluster_config.go
```diff
@@ -49,10 +49,15 @@ type K8sDqlite struct {
 	Port int `yaml:"port,omitempty"`
 }
 
+func (c *ClusterConfig) SetDefaults() {
```
bschimke95, line 52: I guess this is a PoC implementation and we eventually move `DefaultClusterConfig` into `SetDefaults`?

bschimke95 (commented):
Nice work @louiseschmidtgen 
Left some comments

src/k8s/pkg/k8sd/types/cluster_config.go
louiseschmidtgen: Should we use if statements for the other defaults in case we do decide to add configurations to `ClusterConfigFromBootstrapConfig` or keep it like this?
```
func (c *ClusterConfig) SetDefaults() {
	if c.Network.PodCIDR == "" {
		c.Network.PodCIDR = "10.1.0.0/16"
	}
	c.Network.ServiceCIDR = "10.152.183.0/24"
	c.APIServer.Datastore = "k8s-dqlite"
	c.APIServer.SecurePort = 6443
	c.APIServer.AuthorizationMode = "Node,RBAC"
	c.K8sDqlite.Port = 9000
}
```

louiseschmidtgen: For more than one podCIDR are we expecting `PodCIDR: "10.1.0.0/16, 2001:0db8::/32"` or  no space:`PodCIDR: "10.1.0.0/16,2001:0db8::/32"`?

src/k8s/pkg/k8sd/types/cluster_config_test.go
```diff
@@ -25,6 +25,64 @@ func TestClusterConfigFromBootstrapConfig(t *testing.T) {
 	g.Expect(types.ClusterConfigFromBootstrapConfig(&bootstrapConfig)).To(Equal(expectedConfig))
 }
 
+func TestValidateCIDR(t *testing.T) {
+	g := NewWithT(t)
+	// Create a new BootstrapConfig with default values
+	validConfig := types.ClusterConfig{
+		Network: types.Network{
+			PodCIDR: "10.1.0.0/16, 2001:0db8::/32",
```
neoaggelos, line 33:
```suggestion
			PodCIDR: "10.1.0.0/16,2001:0db8::/32",
```

src/k8s/pkg/k8sd/types/cluster_config.go
neoaggelos: no spaces for the CIDRs. also, yes, let's set these only if the current values are unset, similar to the PodCIDR

neoaggelos (approved):

PR #138, opened 2024-02-15
do not fail if cni plugins already exist

neoaggelos:
### Summary

If any of the cni plugins already exist, do not fail. Fixes the following error:

```
root@t1:~# k8s bootstrap
Failed with error: failed to configure containerd: failed to symlink cni plugin bandwidth: symlink cni /opt/cni/bin/bandwidth: file exists
```

### DIFF
--- a/src/k8s/pkg/k8sd/setup/containerd.go
+++ b/src/k8s/pkg/k8sd/setup/containerd.go
@@ -59,8 +59,24 @@ func Containerd(snap snap.Snap) error {
 	if err := os.Chown(cniBinary, snap.UID(), snap.GID()); err != nil {
 		return fmt.Errorf("failed to chown cni plugin binary: %w", err)
 	}
+
+	// for each of the CNI plugins, ensure they are a symlink to the "cni" binary
 	for _, plugin := range snap.CNIPlugins() {
-		if err := os.Symlink("cni", path.Join(snap.CNIBinDir(), plugin)); err != nil {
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/k8sd/setup/containerd.go
```diff
@@ -59,8 +59,24 @@ func Containerd(snap snap.Snap) error {
...
+			}
+			if err := os.Remove(pluginInstallPath); err != nil {
+				return fmt.Errorf("failed to remove already existing file %s: %w", pluginInstallPath, err)
+			}
+		}
+
+		// add plugin as a symlink for the "cni" binary
+		if err := os.Symlink("cni", pluginInstallPath); err != nil {
 			return fmt.Errorf("failed to symlink cni plugin %s: %w", plugin, err)
 		}
```
addyess, line 81: I stared at this a bit and thought `os.ReadLink` may get called with a path that isn't a symbolic link -- but i assume in that case it will err and we'd end up removing that file -- that's a good thing.  

i was going to offer a suggestion to check the fileInfo from `os.Stat(...)` for `fileInfo.Mode() ^ os.ModeSymlink` but i think it's an over complication.

addyess (approved):
LGTM

src/k8s/pkg/k8sd/setup/containerd.go
neoaggelos: right, if it's not a symlink then we still fail either way and attempt to remove the file

PR #140, opened 2024-02-16
use queue.h from host libc6-dev

neoaggelos:
### Summary

Do not fetch `sys/queue.h` from a remote URL that we don't control, rather use the file that exists on the host. Tested that the file is there for all archs we target.


ktsakalozos:
lgtm +1

ktsakalozos (approved):
lgtm

PR #141, opened 2024-02-16
Add installing info

evilnick:
Pages added:

 - installing index (stub)
 - Installing from a snap (complete)
 - Installing with multipass (stub)
 - explanation/channels

### DIFF
--- a/docs/src/explanation/channels.md
+++ b/docs/src/explanation/channels.md
@@ -35,7 +35,7 @@ For all snaps, you can find out what channels are available by running the
 `info` command, For example:
 
 ```bash
-sudo snap info k8s
+snap info k8s
 ```
 
 ## Updates and switching channels
--- a/docs/src/explanation/channels.md
+++ b/docs/src/explanation/channels.md
@@ -1,0 +1,79 @@
+# Channels
+
+Canonical Kubernetes uses the concept of `channels` to make sure you always get
+the version of Kubernetes you are expecting, and that future upgrades can be
+handled without minimum, if any, disruption.
+
+## Choosing the right channel
+
+When installing or updating Canonical Kubernetes you can (and should in most
+cases) specify a channel. The channel specified is made up of two components;
+the **track** and the **risk level**. 
+
+The track will match the minor version of upstream Kubernetes. For example,
+specifying the `1.30` track will match upstream releases of the same minor
+version ("1.30.0", "1.30.1", "1.30.x" etc.). Releases of Charmed Kubernetes
+closely follow the upstream releases and usually follow within 24 hours.
+
+The 'risk level' component of the channel is one of the following:
+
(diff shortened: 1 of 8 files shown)


docs/src/explanation/channels.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Which channel is right for me?
+
+Choosing the most appropriate channel for your needs depends on a number of
+factors. We can give some general guidance for the following cases:
+
+- **I want to always be on the latest stable version matching a specific upstream K8s
+release (recommended).** 
+
+Specify the release, for example: `--channel=1.25/stable`.
```
kwmonroe, line 73:
```suggestion
Specify the release, for example: `--channel=1.30/stable`.
```
Stop living in the past

docs/src/howto/install/snap.md
```diff
@@ -0,0 +1,69 @@
...
+## What you'll need
+
+This guide assumes the following:
+
+- You are installing on Ubuntu 22.04 or later, **or** another OS which supports
+  snap packages (see [snapd support])
+- You have root or sudo access to the machine
+- You have an internet connection
+- The target machine has sufficient memory and disk space. To accommodate
+  workloads, we recommend a system with at least 20G of disk space and 4G of
```
kwmonroe, line 15: 4G / 20G is fine for a basic install, but I think if we're gonna say *recommend*, we should bump those up. If something like 8G feels too heavy, perhaps emphasize the **at least** part and wrap it in a `<blink>` tag or something.

FWIW, we're upping the charmed k8s k-c-p recommended specs to 8G / 64G in https://github.com/charmed-kubernetes/bundle/pull/892.

docs/src/howto/install/snap.md
```diff
@@ -0,0 +1,69 @@
...
+  memory.
+
+```{note}
+If you cannot meet these requirements, please see the [Installing] page for alternative options.
```
kwmonroe, line 19:
```suggestion
If you cannot meet these requirements, please see the [Installing][installing] page for alternative options.
```

kwmonroe (changes requested):
Lookin good!  There's a couple legit changes plus a couple suggestions for your consideration.

docs/src/explanation/channels.md
evilnick: It was nicer then.

docs/src/howto/install/snap.md
evilnick: I'm pretty sure this link works - I think it just doesn't render properly in GH

kwmonroe: ack, gh preview didn't so i thought it was the different case. i trust you may have been right, but appreciate the update to the links section anyway :)

kwmonroe (approved):
:cake:

PR #145, opened 2024-02-20
Cleanup container shims, networks and mounts when removing k8s

neoaggelos:
### Summary

Cleanup network, processes and mounts when removing k8s snap

### Changes

- Send a SIGKILL to all container shims and `/pause` processes running on the host. Ideally, we would want to do this through the containerd socket using commands like `ctr task {ls,delete}` and `ctr container {ls,delete}`, however all snap services are already stopped by the time the remove hook is running. Therefore, we simply look for all containerd-shim pids (and, recursively, all their children) and send a SIGKILL to them.
- Remove all `cni-$UUID` network namespaces. This is done after killing the container shims.
- Remove all `kubelet` and `containerd` volume mounts.

### DIFF
--- a/tests/e2e/tests/test_cleanup.py
+++ b/tests/e2e/tests/test_cleanup.py
@@ -26,11 +26,6 @@ def test_node_cleanup(instances: List[harness.Instance]):
         )
     ).exec(["ps", "-fea"])
 
-    LOG.info("Waiting for CNI network namespaces to go away...")
-    util.stubbornly(retries=5, delay_s=5).on(instance).until(
-        lambda p: "cni-" not in p.stdout.decode()
-    ).exec(["ip", "netns", "list"])
-
     LOG.info("Waiting for kubelet and containerd mounts to go away...")
     util.stubbornly(retries=5, delay_s=5).on(instance).until(
         lambda p: all(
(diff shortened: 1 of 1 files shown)


addyess:
Is there any e2e test that should be added to demonstrate that a node has been cleaned with the `k8s::remove::containers` during the snap remove hook?

k8s/wrappers/services/containerd
```diff
@@ -7,7 +7,18 @@ k8s::common::setup_env
 # Ensure XDG_RUNTIME_DIR exists, needed by runc for 'kubectl exec' to work
 [ ! -z "${XDG_RUNTIME_DIR:=}" ] && mkdir -p "${XDG_RUNTIME_DIR}"
 
-# Apply the containerd profile
-k8s::common::is_strict && apparmor_parser -r "$SNAP/k8s/profiles/containerd"
+# Apply apparmor profile for "cri-containerd.apparmor.d"
+if ! apparmor_parser -r "$SNAP/k8s/profiles/containerd"; then
+  echo "
```
bschimke95, line 12: Hm, should we just not start the containerd service instead?
As a user I'd rather have a failing service when bringing up the cluster than subtle failures later. 
I'd also say that this message will likely be not seen by the user. At least I wouldn't checkt the journal log if all services come up just fine.
WDYT?

bschimke95 (approved):
LGTM, just a small comment

k8s/wrappers/services/containerd
neoaggelos: so, we might be on a system where AppArmor is not available or enabled, and on such systems a failure here is expected, which is why I started with this "loud silent" warning on the logs.

bschimke95: I see, that makes sense. Thanks for the explanation!

PR #147, opened 2024-02-21
Added metrics server component

berkayoz:
Added metrics server component


tests/e2e/tests/test_metrics_server.py
```diff
@@ -0,0 +1,72 @@
...
+        ["sed", "-i", "-e", "s#= 10250#= 10251#g", "metrics-server/test/e2e_test.go"]
+    )
+
+    instance.exec(["bash", "-c", "mkdir -p ~/.kube"])
+    instance.exec(["bash", "-c", "k8s config > ~/.kube/config"])
+    p = instance.exec(
+        ["bash", "-c", "cd metrics-server && go test test/e2e_test.go -v -count=1"],
+        capture_output=True,
+    )
+    LOG.info(p.stdout.decode())
```
neoaggelos, line 72: If the tests fail, both stdout and stderr will be included in the pytest logs.

```suggestion
    p = instance.exec(
        ["bash", "-c", "cd metrics-server && go test test/e2e_test.go -v -count=1"],
    )
```

k8s/components/charts/metrics-server-3.12.0.tgz
neoaggelos, line 1: How is this file updated? Can we have a script or documentation?

neoaggelos (commented):
Nice work @berkayoz! Left two comments

src/k8s/pkg/component/metrics_server.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/canonical/k8s/pkg/snap"
+)
+
+func EnableMetricsServerComponent(ctx context.Context, s snap.Snap) error {
+	manager, err := NewHelmClient(s, nil)
+	if err != nil {
+		return fmt.Errorf("failed to get component manager: %w", err)
+	}
+
+	values := map[string]any{
```
neoaggelos, line 16: can we set a releaseName here so that this does not deploy as `ck-metrics-server`?

tests/e2e/tests/test_metrics_server.py
```diff
@@ -0,0 +1,72 @@
...
+def test_metrics_server(instances: List[harness.Instance]):
+    if not config.SNAP:
+        pytest.fail("Set TEST_SNAP to the path where the snap is")
+
+    instance = instances[0]
+    instance.exec(["k8s", "enable", "metrics-server"])
+
+    LOG.info("Waiting for metrics-server pod to show up...")
+    util.stubbornly(retries=15, delay_s=5).on(instance).until(
+        lambda p: "ck-metrics-server" in p.stdout.decode()
```
neoaggelos, line 22: so this can be `metrics-server` here as well. also, since this is JSON already, maybe we could parse the JSON instead?

berkayoz: This is a very basic check to see if a pod shows up, and is done similarly in places. If we want to parse these from JSON I think we can create a card/todo for revisiting these.

k8s/components/charts/updating.md
neoaggelos, line 1: This would find its way to the built snap, let's put it under `build-scripts/`? e.g. just the script to do

```
helm pull --repo https://kubernetes-sigs.github.io/metrics-server/ metrics-server
```

at something like `build-scripts/hack/update-metrics-server-chart.sh`?

neoaggelos: similar to the script we have for the gateway api chart

neoaggelos (commented):
Nice work, just a comment on the script

neoaggelos (approved):

PR #152, opened 2024-02-24
enable rbac by default, configurable on bootstrap

kwmonroe:
Ensure rbac is enabled during bootstrap by default.  Only disable it if user explicitly sets enable = `false`.

### DIFF
--- a/src/k8s/pkg/k8sd/types/cluster_config.go
+++ b/src/k8s/pkg/k8sd/types/cluster_config.go
@@ -95,7 +95,16 @@ func (c *ClusterConfig) SetDefaults() {
 // ClusterConfigFromBootstrapConfig extracts the cluster config parts from the BootstrapConfig
 // and maps them to a ClusterConfig.
 func ClusterConfigFromBootstrapConfig(b *apiv1.BootstrapConfig) ClusterConfig {
+	authzMode := "Node,RBAC"
+	// Only disable rbac if explicitly set to false during bootstrap
+	if !(*b.EnableRBAC) {
+		authzMode = "AlwaysAllow"
+	}
+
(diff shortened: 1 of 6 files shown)


src/k8s/pkg/k8sd/types/cluster_config.go
```diff
@@ -95,7 +95,16 @@ func (c *ClusterConfig) SetDefaults() {
 // ClusterConfigFromBootstrapConfig extracts the cluster config parts from the BootstrapConfig
 // and maps them to a ClusterConfig.
 func ClusterConfigFromBootstrapConfig(b *apiv1.BootstrapConfig) ClusterConfig {
+	authzMode := "Node,RBAC"
+	// Only disable rbac if explicitly set to false during bootstrap
+	if !(*b.EnableRBAC) {
```
neoaggelos, line 100: This will fail if `EnableRBAC` is `nil`. Instead, you can do (first check nil, then dereference):
```suggestion
	if v := b.EnableRBAC; v != nil && !*v {
```

neoaggelos: Would be useful to start with having a unit test where `EnableRBAC` is nil to see the failure, then fix here

neoaggelos (changes requested):
Great work, @kwmonroe! Needs one fix, see comment

src/k8s/pkg/k8sd/types/cluster_config.go
kwmonroe: ah nice, thx @neoaggelos!

ktsakalozos:
I see the comment from @neoaggelos is addressed. Merging this.

PR #155, opened 2024-02-25
Enable components specified during bootstrap by default

kwmonroe:
### Summary

Enable components after a successful `k8s bootstrap`. These will include `dns` and `network` by default, or any valid set of component names that the user provided during bootstrap.

### Changes

- Giant switch

### Rationale

We maintain various lists of components in `cmd/k8s/k8s_enable`, `pkg/snap/snap`, a couple interfaces, and as part of our v1 api endpoint list.  We should consolidate those to a single source-of-truth where names, attrs, and methods can be managed together.

But that's not what this PR is about.  This PR ensures that default components are enabled after a successful bootstrap, e.g.:

```bash
$ sudo k8s bootstrap -v -d --interactive
Which components would you like to enable? () [dns, network]: dns, ham, network, storanges
Please set the Cluster CIDR? [10.1.0.0/16]: 
Bootstrapping the cluster. This may take some seconds...
Enabling bootstrap components.
 dns enabled
 skipping unknown component: ham
 network enabled
 skipping unknown component: storanges
Bootstrapped k8s cluster on "ip-172-31-53-214" (172.31.53.214).

$ sudo k8s status 
k8s is ready.
high-availability: no

control-plane nodes:
  ip-172-31-53-214: 172.31.53.214

components:
  dns        enabled
  gateway    disabled
  ingress    disabled
  loadbalancer disabled
  metrics-server disabled
  network    enabled
  storage    disabled

```

### DIFF
--- a/src/k8s/cmd/k8s/k8s_bootstrap.go
+++ b/src/k8s/cmd/k8s/k8s_bootstrap.go
@@ -5,11 +5,11 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"strings"
 	"time"
 
 	apiv1 "github.com/canonical/k8s/api/v1"
-	v1 "github.com/canonical/k8s/api/v1"
 	"github.com/canonical/k8s/cmd/k8s/errors"
 	"github.com/canonical/k8s/pkg/utils"
 	"github.com/spf13/cobra"
@@ -25,6 +25,7 @@ var (
 		apiv1.ErrUnknown:             "An unknown error occured while bootstrapping the cluster:\n",
 		apiv1.ErrAlreadyBootstrapped: "K8s cluster already bootstrapped.",
 	}
+	bootstrappableComponents = []string{"network", "dns", "gateway", "ingress", "storage", "metrics-server"}
 )
 
 func newBootstrapCmd() *cobra.Command {
@@ -37,7 +38,7 @@ func newBootstrapCmd() *cobra.Command {
 			defer errors.Transform(&err, bootstrapCmdErrorMsgs)
 
 			if k8sdClient.IsBootstrapped(cmd.Context()) {
-				return v1.ErrAlreadyBootstrapped
+				return apiv1.ErrAlreadyBootstrapped
 			}
 
 			const minTimeout = 3 * time.Second
@@ -64,8 +65,7 @@ func newBootstrapCmd() *cobra.Command {
 		},
 	}
 
-	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapCmdOpts.interactive, "interactive", false,
-		"Interactively configure the most important cluster options.")
+	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapCmdOpts.interactive, "interactive", false, "Interactively configure the most important cluster options.")
 	bootstrapCmd.PersistentFlags().DurationVar(&bootstrapCmdOpts.timeout, "timeout", 90*time.Second, "The max time to wait for k8s to bootstrap.")
 
 	return bootstrapCmd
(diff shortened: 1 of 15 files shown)


src/k8s/pkg/component/loadbalancer.go
```diff
@@ -73,13 +73,13 @@ func EnableLoadBalancerComponent(ctx context.Context, s snap.Snap, cidrs []strin
 		return fmt.Errorf("failed to create kubernetes client: %w", err)
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	timeoutCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
bschimke95, line 76: drive-by fix: wrong context

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -270,5 +262,57 @@ func onBootstrapControlPlane(s *state.State, initConfig map[string]string) error
...
+			if err := component.EnableGatewayComponent(s.Context, snap); err != nil {
+				return fmt.Errorf("failed to enable gateway: %w", err)
+			}
+
+		case "metrics-server":
+			if err := component.EnableMetricsServerComponent(s.Context, snap); err != nil {
+				return fmt.Errorf("failed to enable metrics-server: %w", err)
+			}
+
+		case "loadbalancer":
```
bschimke95, line 299: This should not be reached but better safe than sorry.

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -75,47 +75,77 @@ func getConfigInteractively(ctx context.Context) apiv1.BootstrapConfig {
...
+
+		if s == "" {
+			return defaultVal
 		}
-	}
-	s = strings.TrimSpace(s)
 
-	if s == "" {
-		return defaultVal
+		// Check if the input is valid
```
kwmonroe, line 122: :heart:

kwmonroe (commented):
This is great!  Not only are you validating input, but you're also storing component options that i originally missed by just mimicking client enable component commands.

Thank you!  :100: (though i can't approve because it's my pr)

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -270,5 +272,59 @@ func onBootstrapControlPlane(s *state.State, initConfig map[string]string) error
...
+		case "network":
+			if err := component.EnableNetworkComponent(s.Context, snap, cfg.Network.PodCIDR); err != nil {
+				return fmt.Errorf("failed to enable network: %w", err)
+			}
+		case "dns":
+			dnsIP, clusterDomain, err := component.EnableDNSComponent(s.Context, snap, "", "", nil)
+			if err != nil {
+				return fmt.Errorf("failed to enable dns: %w", err)
+			}
+			if err := s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
```
louiseschmidtgen, line 287: Note: We will need to duplicate/move this piece of code somewhere so we can also set the dns bits from the charm. (Not needed for this PR)

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -75,47 +75,77 @@ func getConfigInteractively(ctx context.Context) apiv1.BootstrapConfig {
...
+		if options != nil || len(options) > 0 {
+			valid := true
+			sSlice := strings.Split(strings.ReplaceAll(s, " ", ""), ",")
```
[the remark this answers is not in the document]
bschimke95: Good point. I just directly modified the input above so that we can get rid of additional checks upon return.

bschimke95 (approved):

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
bschimke95: Not sure if I understand.
This hook will be executed by microcluster whenever the cluster is bootstrapped. So, it will also be executed if the charm calls the bootstrap endpoint (which the CLI also just calls).

You can also modify the dns bits by using the relevant component endpoint. https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/api/component.go#L51

PR #156, opened 2024-02-26
Automatically label control-plane and worker nodes

neoaggelos:
### Summary

Automatically label worker node and control plane nodes.

Example output from a cluster with two control-plane and one worker node:

```
root@t1:~# k8s kubectl get node
NAME   STATUS   ROLES           AGE   VERSION
t1     Ready    control-plane   14m   v1.29.2
t2     Ready    worker          13m   v1.29.2
t3     Ready    control-plane   11m   v1.29.2
```

### DIFF
--- a/src/k8s/pkg/k8sd/setup/kubelet.go
+++ b/src/k8s/pkg/k8sd/setup/kubelet.go
@@ -22,11 +22,11 @@ var kubeletTLSCipherSuites = []string{
 }
 
 var kubeletControlPlaneLabels = []string{
-	"node-role.kubernetes.io/control-plane=control-plane",
+	"node-role.kubernetes.io/control-plane=",
 }
 
 var kubeletWorkerLabels = []string{
-	"node-role.kubernetes.io/worker=worker",
+	"node-role.kubernetes.io/worker=",
 }
 
 // KubeletControlPlane configures kubelet on a control plane node.
--- a/src/k8s/pkg/k8sd/setup/kubelet.go
+++ b/src/k8s/pkg/k8sd/setup/kubelet.go
@@ -21,8 +21,26 @@ var kubeletTLSCipherSuites = []string{
 	"TLS_RSA_WITH_AES_256_GCM_SHA384",
 }
 
-// Kubelet configures kubelet on the local node.
-func Kubelet(snap snap.Snap, hostname string, nodeIP net.IP, clusterDNS string, clusterDomain string, cloudProvider string) error {
+var kubeletControlPlaneLabels = []string{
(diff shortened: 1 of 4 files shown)


src/k8s/pkg/k8sd/setup/kubelet.go
```diff
@@ -21,8 +21,26 @@ var kubeletTLSCipherSuites = []string{
...
+	"node-role.kubernetes.io/control-plane=",
+}
+
+var kubeletWorkerLabels = []string{
+	"node-role.kubernetes.io/worker=",
+}
+
+// KubeletControlPlane configures kubelet on a control plane node.
+func KubeletControlPlane(snap snap.Snap, hostname string, nodeIP net.IP, clusterDNS string, clusterDomain string, cloudProvider string) error {
+	return kubelet(snap, hostname, nodeIP, clusterDNS, clusterDomain, cloudProvider, nil, append(kubeletControlPlaneLabels, kubeletWorkerLabels...))
```
bschimke95, line 34: Do we really want to append the worker labels to the control plane?

bschimke95 (approved):
LGTM, only a small comment

src/k8s/pkg/k8sd/setup/kubelet.go
neoaggelos: So, currently our control plane are also workers. Optionally, in the future, we could have nodes that are "control-plane" only. For these, we would have to take care of not including the worker labels indeed.

For now, if a workload requires the presence of a worker node to run, we still want a single-node k8s cluster to support it, so we just include the worker labels as well

bschimke95: Makes sense!

addyess:
oh dear -- i do believe these changes (which are very nice) may interfere with the charm performing a similar task. 
The charms may unintentionally attempt to remove labels it doesn't know the snaps have added. 

I think we need to revisit the charm LabelMaker to help it not remove the unexpected labels.

bschimke95:
As I understand, the LabelMaker was used because we were missing labeling in the snap, right?
Can we not just remove the labeling in the charm now or is there a use-case where the charm need to perform its own labeling?

addyess:
the charm config provides a way for custom labels to be applied -- labels the snap knows nothing about. For instance -- some of these labels are:

```bash
juju-application=k8s-worker
juju-charm=k8s-worker
juju.io/cloud=<underlying cloud infra>
```

I just ran the tests again and see that all the labels still seem to be appiled correctly. So there's no worries on these changes

PR #157, opened 2024-02-26
Add architecture doc

evilnick:
This PR does several things:

 - adds the kroki extension for rendering diagrams
 - configures sphinx to use it
 - add UML diags for the architecture
 - Adds an architecture page

This isn't a complete C4 representation of the project at the moment - the architecture doc isn't complete, but it is a very good start to a transparent doc which explains which bit does what.


docs/src/reference/architecture.md
```diff
@@ -0,0 +1,72 @@
+# Architecture
+
+A system architecture document is the starting point for many interested participants in a project, whether you intend contirbuting or simply want to understand how the software is structured. This documentation lays out the current design of Canonical Kubernetes, following the [C4 model].
```
bschimke95, line 3: typo: contributing

docs/src/reference/architecture.md
```diff
@@ -0,0 +1,72 @@
...
+A system architecture document is the starting point for many interested participants in a project, whether you intend contirbuting or simply want to understand how the software is structured. This documentation lays out the current design of Canonical Kubernetes, following the [C4 model]. 
+
+##  System context 
+
+This overview of Canononical Kubernetes demonstrates the interactions of Kubernetes with users and with other systems.
+
+```{kroki} ../assets/overview.puml
+```
+
+Two actors interact with the kubernetes snap:
```
bschimke95, line 12: `Kubernetes` should probably be capitalized here?

docs/src/reference/architecture.md
```diff
@@ -0,0 +1,72 @@
...
+
+- **K8s admin**: The administrator of the cluster interacts directly with the Kubernetes API server. Out of the box our K8s distribution offers admin access to the cluster. That initial user is able to configure the cluster to match their needs and of course create other users that may or may not have the same privileges. The K8s admin is also able to deploy workloads running in the cluster.
```
bschimke95, line 14: maybe `maintain` instead of `deploy` workloads? An admin can e.g. also patch and remove workloads

docs/src/reference/architecture.md
```diff
@@ -0,0 +1,72 @@
...
+
+- **K8s user**: A user consuming the services hosted in the cluster. Users do not have access to the Kubernetes API server. They need to access the cluster through the options (nodeport, ingress, load-balancer) offered by the administrator who deployed the workload they are interested in.
+
+There are non-human users of the k8s snap. In this case that is the [K8s charm]. The K8s charm needs to drive the Kubernetes cluster and to orchestrate the multi-node clustering operations.
```
bschimke95, line 18: This sounds a bit strange.

Maybe: `There are non-human users of the k8s snap, for example the [k8s-operator charm]`

docs/src/reference/architecture.md
```diff
@@ -0,0 +1,72 @@
...
+
+A set of external systems need to be easily integrated with our k8s distribution. We have identified the following:
```
bschimke95, line 20: Further up, k8s distribution is capitalized. I don't have a strong opinion here but we should be consistent with `k8s` or `K8s`.

docs/src/reference/architecture.md
```diff
@@ -0,0 +1,72 @@
...
+ - **Load-balancer**: Although the k8s snap distribution comes with a load-balancer we expect the end customer environment to have a load-balancer and thus we need to integrate with it.
```
bschimke95, line 21: In the `k8s snap` we use `k8s enable loadbalancer` (no hyphen). Same here: I don't have a strong opinion on this but we should choose only one version.

evilnick:
For some reason, the PR previews for RTD don't work with our repo  :(
the best way to preview is to go to the /docs/tools directory and run

```
make serve
```

then follow the link

docs/src/reference/architecture.md
```diff
@@ -0,0 +1,72 @@
...
+## The k8s snap
+
+Looking more closely at what is conatined within the K8s snap istelf:
+
+```{kroki} ../assets/k8s-container.puml
+```
+
+The k8s snap distribution includes the following:
+
+- **Kubectl**: through which users and other systems interact with  Kubernetes and drive the cluster operations.
```
bschimke95, line 35: Only one whitespace between `with` and `Kubernetes` (I don't know if markdown does care.)

docs/src/reference/architecture.md
```diff
@@ -0,0 +1,72 @@
...
+- local storage
+- observability
+
+The cluster manager is responsible for implementing the formation of the cluster. This includes operations such as joining/removing nodes into the cluster and reporting status.
+
+This functionality is exposed via the following interfaces:
+
+- The **CLI**: The CLI is available to only the root user on the k8s snap and all CLI commands are mapped to respective REST calls.
+
+- The **REST API**: apart from serving the CLI, is also used by the charm that has to programmatically drive the Kubernetes cluster.
```
bschimke95, line 66: We provide a `JSON RPC API` to be specific. 
However, people might be more familiar with the `REST API` term even though it is not 100% correct.

docs/src/reference/architecture.md
```diff
@@ -0,0 +1,72 @@
...
+
+```{kroki} ../assets/k8s-container.puml
+```
+
+The k8s snap distribution includes the following:
+
+- **Kubectl**: through which users and other systems interact with  Kubernetes and drive the cluster operations.
+- **K8s upstream services**: These are Kubernetes binaries built from upstream and shipped in the snap.
+- **Components** are the workloads and features we deem important to be available to our users and therefore are shipped in the snap and are enabled, configured and disabled in a guided way.
+- State is backed up by **dqlite** by default, which keeps that state of the Kubernetes cluster as well as the the state we maintain for the needs of the cluster operations.
```
bschimke95, line 38: We may want to add that the Kubernetes `dqlite` store can be changed while the cluster storage will always be `dqlite`

bschimke95 (commented):
Great work, added a few comments.

docs/src/reference/architecture.md
evilnick: It doesn't care in general, but I do, will fix

docs/src/reference/architecture.md
bschimke95: Discussed this with @ktsakalozos offline: We do a wild mix of REST and JSON RPC and should just not use those terms at all. However, we still need to mention the protocol.

So something like `HTTP API` or `we provide an API over/via HTTP` should be used instead.

docs/src/reference/architecture.md
evilnick: okay, reworded

PR #160, opened 2024-02-26
Ensure k8s services are enabled

eaudetcobello:


tests/e2e/tests/test_clustering.py
```diff
@@ -25,6 +25,9 @@ def add_node(
 def join_cluster(instance, token):
     instance.exec(["k8s", "join-cluster", token])
 
+def reboot(instance):
+    instance.exec(["reboot"])
+
```
eaudetcobello, line 30: to remove

src/k8s/pkg/snap/snap.go
```diff
@@ -18,14 +18,22 @@ import (
 type snap struct {
 	snapDir       string
 	snapCommonDir string
+	snapDataDir   string
```
neoaggelos, line 21: we don't need the `snapDataDir` (at least, yet), feel free to remove

src/k8s/pkg/snap/snap_test.go
```diff
@@ -0,0 +1,63 @@
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			mockRunner.CalledWithCommand = []string{} // Resetting the commands for each test case
+			tc.action(context.Background(), tc.service)
+			if lastCmd := mockRunner.CalledWithCommand[0]; lastCmd != tc.expectedCommand {
+				t.Fatalf("Expected command %q, but %q was called instead for service %s", tc.expectedCommand, lastCmd, tc.service)
+			}
+		})
+	}
+}
```
neoaggelos: Some notes on the test added:

1. This will produce tests with names like:

```
TestServiceStartStop/StartService
TestServiceStartStop/StopService
```

which stutters. It would be easier if you just call it `TestServices`

2. It would probably be easier if you created a separate snap instance inside each sub-test, that way you don't need the reset step or accidentally sharing state

3. Use `gomega` instead, see other places.

4. No real need for a table here. It should be simpler (and shorter) to just write the tests explicitly.

Taking the above into account, consider the following:

```go

func TestServices(t *testing.T) {
	t.Run("Start", func(t *testing.T) {
		g := NewWithT(t)
		mockRunner := &MockRunner{}
		snap := snap.NewSnap("testdir", "testdir", snap.WithCommandRunner(mockRunner.Run))

		err := snap.StartService(context.Background(), "service")
		g.Expect(err).To(BeNil())
		g.Expect(mockRunner.CalledWithCommand).To(ConsistOf("snapctl start --enable k8s.test-service"))

		t.Run("Fail", func(t *testing.T) {
			g := NewWithT(t)
			mockRunner.Err = fmt.Errorf("some error")

			err := snap.StartService(context.Background(), "service")
			g.Expect(err).NotTo(BeNil())
		})
	}

	t.Run("Stop", func(t *testing.T) {
		mockRunner := &MockRunner{}
		snap := snap.NewSnap("testdir", "testdir", snap.WithCommandRunner(mockRunner.Run))

		err := snap.StopService(context.Background(), "service")
		g.Expect(err).To(BeNil())
		g.Expect(mockRunner.CalledWithCommand).To(ConsistOf("snapctl stop --disable k8s.test-service"))

		t.Run("Fail", func(t *testing.T) {
			g := NewWithT(t)
			mockRunner.Err = fmt.Errorf("some error")

			err := snap.StartService(context.Background(), "service")
			g.Expect(err).NotTo(BeNil())
		})
	})
}
```

neoaggelos: use the above as inspiration, let's also have unit tests for what happens when the command has failed

neoaggelos (commented):
Looks better, another pass

src/k8s/pkg/snap/snap.go
```diff
@@ -18,14 +18,20 @@ import (
 type snap struct {
 	snapDir       string
 	snapCommonDir string
+	runCommand    func(ctx context.Context, command ...string) error
 }
 
 // NewSnap creates a new interface with the K8s snap.
-// NewSnap accepts the $SNAP and $SNAP_COMMON directories
-func NewSnap(snapDir, snapCommonDir string) *snap {
+// NewSnap accepts the $SNAP, $SNAP_DATA and $SNAP_COMMON, directories, and a number of options.
```
bschimke95, line 25: This comment slipped through - $SNAP_DATA is not set :)

src/k8s/pkg/snap/snap_test.go
```diff
@@ -0,0 +1,107 @@
...
+			mockRunner.Err = fmt.Errorf("some error")
+
+			err := snap.StartService(context.Background(), "service")
+			g.Expect(err).NotTo(BeNil())
+		})
+	})
+}
+
+// MockRunner is a mock implementation of CommandRunner.
+type MockRunner struct {
```
bschimke95, line 92: There is a `snap/mock` directory. Let's move that mock there.

bschimke95 (commented):
Great work, minor comments

neoaggelos (commented):
Nothing to add, please address comments from @bschimke95

neoaggelos (commented):
Just a final comment

src/k8s/pkg/snap/mock/runner/runner.go
```diff
@@ -0,0 +1,25 @@
+package runner
+
+import (
+	"context"
+	"log"
+	"strings"
+)
+
+// MockRunner is a mock implementation of CommandRunner.
+type MockRunner struct {
```
neoaggelos, line 10: This stutters as `runner.MockRunner`. We don't need a sub-package under `pkg/snap/mock`, you can just move this in `mock/runner.go`, then we'll use it as `mock.Runner`

src/k8s/pkg/snap/snap_test.go
```diff
@@ -0,0 +1,89 @@
+package snap
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	. "github.com/canonical/k8s/pkg/snap/mock/runner"
```
neoaggelos, line 8: Avoid `.` imports, as it makes hard to track down where things are coming from

src/k8s/pkg/snap/snap_test.go
```diff
@@ -0,0 +1,89 @@
+package snap
```
neoaggelos, line 1: Make this a separate `snap_test` package. Is this to test the internal `serviceName()` function?

```suggestion
package snap_test
```

neoaggelos (changes requested):

src/k8s/pkg/snap/snap_test.go
neoaggelos: `gomega` is in fact an exception to this rule, because it makes writing test assertions easier.

src/k8s/pkg/snap/snap_test.go
eaudetcobello: Yes it is to test the internal serviceName function.

src/k8s/pkg/snap/mock/runner/runner.go
eaudetcobello: I tried that but I get a circular dependency because mock/mock.go has the `mock` package and imports `snap`.

src/k8s/pkg/snap/snap_test.go
neoaggelos: Yes, please move that test to a separate `snap_internal_test.go` file, which will have `package snap` in the beginning. This will also resolve the circular dependency issues

neoaggelos (changes requested):
Please rebase

src/k8s/pkg/snap/snap_internal_test.go
```diff
@@ -0,0 +1,25 @@
...
+		{"Just k8s", "k8s", "k8s"},
+		{"Empty string", "", "k8s."},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got := serviceName(tc.input)
+			if got != tc.expected {
+				t.Errorf("serviceName(%q) = %q, want %q", tc.input, got, tc.expected)
+			}
```
neoaggelos, line 22: use gomega here

neoaggelos (commented):
One final change and this is good to go, nice work!

bschimke95 (approved):
LGTM

src/k8s/pkg/snap/mock/runner.go
```diff
@@ -0,0 +1,25 @@
+package mock
+
+import (
+	"context"
+	"log"
+	"strings"
+)
+
+// MockRunner is a mock implementation of CommandRunner.
```
neoaggelos, line 9: this one seems to have slipped

src/k8s/pkg/snap/snap_internal_test.go
```diff
@@ -0,0 +1,28 @@
...
+func TestServiceName(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{"With k8s. prefix", "k8s.test-service", "k8s.test-service"},
+		{"Without prefix", "api", "k8s.api"},
+		{"Just k8s", "k8s", "k8s"},
+		{"Empty string", "", "k8s."},
```
neoaggelos, line 18: These are added on the generated test names, so let's make sure that they are proper
```suggestion
		{"WithPrefix", "k8s.test-service", "k8s.test-service"},
		{"NoPrefix", "api", "k8s.api"},
		{"K8s", "k8s", "k8s"},
		{"EmptyString", "", "k8s."},
```

neoaggelos (commented):
Final comments and we're good to go

neoaggelos (approved):

PR #161, opened 2024-02-27
Warn if /sys mount propagation is not shared

bschimke95:
Cilium fails to deploy if `/sys` is not mount with shared propagation.

This PR adds a check to verify that `/sys` is mounted with the propagation type.

### DIFF
--- a/src/k8s/pkg/component/network.go
+++ b/src/k8s/pkg/component/network.go
@@ -92,6 +92,17 @@ func EnableNetworkComponent(ctx context.Context, s snap.Snap, podCIDR string) er
 			},
 			"hostRoot": cgrMnt,
 		}
+	} else {
+		p, err := utils.GetMountPropagation("/sys")
+		if err != nil {
+			return fmt.Errorf("failed to get mount propagation for %s: %w", p, err)
+		}
+		if p == "private" {
+			if s.OnLXD(ctx) {
+				return fmt.Errorf("/sys is not a shared mount on the LXD container, this might be resolved by updating LXD on the host to version 5.0.2 or newer")
+			}
+			return fmt.Errorf("/sys is not a shared mount")
+		}
 	}
 
 	if err := manager.Enable("network", values); err != nil {
--- a/src/k8s/pkg/snap/interface.go
+++ b/src/k8s/pkg/snap/interface.go
@@ -10,9 +10,11 @@ import (
 
 // Snap abstracts file system paths and interacting with the k8s services.
 type Snap interface {
-	Strict() bool // Strict returns true if the snap is installed with strict confinement.
-	UID() int     // UID is the user ID to set on config files.
-	GID() int     // GID is the group ID to set on config files.
+	Strict() bool               // Strict returns true if the snap is installed with strict confinement.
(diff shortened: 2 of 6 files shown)


src/k8s/pkg/component/network.go
```diff
@@ -92,6 +92,17 @@ func EnableNetworkComponent(ctx context.Context, s snap.Snap, podCIDR string) er
...
 			"hostRoot": cgrMnt,
 		}
+	} else {
+		p, err := utils.GetMountPropagation("/sys")
+		if err != nil {
+			return fmt.Errorf("failed to get mount propagation for %s: %w", p, err)
+		}
+		if p == "private" {
+			if s.OnLXD(ctx) {
+				return fmt.Errorf("Please update the LXD version on the host to at least 5.0.2")
```
neoaggelos, line 102: Let's also explain _why_ it's needed

```suggestion
				return fmt.Errorf("/sys is not a shared mount on the LXD container, this might be resolved by updating LXD on the host to version 5.0.2 or newer")
```

src/k8s/pkg/snap/interface.go
```diff
@@ -10,9 +10,11 @@ import (
 
 // Snap abstracts file system paths and interacting with the k8s services.
 type Snap interface {
-	Strict() bool // Strict returns true if the snap is installed with strict confinement.
-	UID() int     // UID is the user ID to set on config files.
-	GID() int     // GID is the group ID to set on config files.
+	Strict() bool               // Strict returns true if the snap is installed with strict confinement.
+	OnLXD(context.Context) bool // OnLXD returns true if the host runs on LXD.
```
neoaggelos, line 14: might want to revisit later, but I think it's a good compromise for now.

src/k8s/pkg/snap/snap.go
```diff
@@ -79,6 +79,10 @@ func (s *snap) Strict() bool {
 	return meta.Confinement == "strict"
 }
 
+func (s *snap) OnLXD(ctx context.Context) bool {
+	return utils.RunCommand(ctx, "grep", "-qa", "container=lxc", "/proc/1/environ") == nil
```
neoaggelos, line 83: You might want to wait and rebase after https://github.com/canonical/k8s-snap/pull/160/files#diff-e8e4848fe5de62c067838047b54f6c6f0ffacb00fc058e00fcf5168ebd9ee625R21

src/k8s/pkg/utils/file.go
```diff
@@ -138,3 +125,38 @@ func ValueInSlice[T comparable](key T, list []T) bool {
 
 	return false
 }
+
+var ErrUnknownMount = errors.New("mount is unknown")
```
neoaggelos, line 129: do we care to export the error here?

neoaggelos (commented):
Looks good overall, nice work @bschimke95. Left some comments

src/k8s/pkg/utils/file.go
bschimke95: People and tests might want to check it.

neoaggelos (approved):

PR #162, opened 2024-02-27
Docs:  readthedocs tidyup

evilnick:
This is to prepare for RTD publishing and fix up stuff like reporting issues, URLs etc.

- reformat a few pages
- fix contents listing
- add canonical-kubernetes slug to config
- properly config issue reporting
- add URL for ogp
- add matrix instead of mattermost

### DIFF
--- a/docs/src/howto/index.md
+++ b/docs/src/howto/index.md
@@ -10,6 +10,8 @@ Overview <self>
 
 ```{toctree}
 :glob:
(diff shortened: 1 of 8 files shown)


evilnick:
I honestly don't think there is any point someone else wasting time trying to find out what all these things do.so... :cowboy_hat_face:

PR #165, opened 2024-02-28
Use PreRunE for sub-commands

bschimke95:
By default, PersistentPreRun hooks are not chained. If a child command overwrites this, the parent hook is not executed.

This caused the global permissions check to be overwritten in the sub-commands and caused misleading error messages when running commands without root permission.

There is a flag to introduce the chaining behavior but this unnecessarily complicates the commands. Since we have shallow command, using `PreRun` hooks is just fine. 

Btw, we are not the only ones that got confused by this feature: https://github.com/spf13/cobra/issues/252


neoaggelos (approved):

kwmonroe (approved):
nice find @bschimke95!

PR #164, opened 2024-02-28
Add storage how-to documentation

bschimke95:
Provides a step by step guide on how to enable, configure and disable the storage component.

PR #166, opened 2024-02-28
How to default dns docs

louiseschmidtgen:
## Description
The how-to guide for working with the default DNS, based on the new CLI specs.


docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,86 @@
...
+
+Canonical Kubernetes includes a default DNS (Domain Name System) essential for internal communication within your cluster. 
+When enabled, the DNS facilitates service discovery
+by assigning each service a DNS name. 
+
+## What you'll need
+
+This guide assumes the following:
+
+- You are installing on Ubuntu 22.04 or later, **or** another OS which supports
```
evilnick, line 11: we don't need any of these I think apart from root/sudo and the last one

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,86 @@
...
+
+Find out wether you have enabled DNS with the following command:
+
+```bash
+sudo k8s status
+```
+
+The default state for the cluster is `dns enabled`.
+
+## Enabling and disabling DNS
```
evilnick, line 30:
```suggestion
## Enable or disable DNS
```

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,86 @@
...
+by assigning each service a DNS name. 
+
+## What you'll need
+
+This guide assumes the following:
+
+- You are installing on Ubuntu 22.04 or later, **or** another OS which supports
+  snap packages (see [snapd support])
+- You have root or sudo access to the machine
+- You have an internet connection
```
bschimke95, line 14: Should we really start _that_ low? I assume the reader has a technical background.
Also, IIRC how-to sections are for people that already know k8s and what they are doing. So we do not need to start from scratch (correct me if I'm wrong @evilnick ).

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,86 @@
...
+- The target machine has sufficient memory and disk space. To accommodate
+  workloads, we recommend a system with ***at least*** 20G of disk space and 4G of
+  memory.
+- You have Canonical Kubernetes installed and a bootstraped cluster. (See: [getting-started-guide](#TODO))
+
+## Is DNS enabled?
+
+Find out wether you have enabled DNS with the following command:
```
bschimke95, line 22: whether

docs/src/howto/networking/default-dns.md
bschimke95, line 30: I think we should agree on one format for the those how-to documentations.
For the storage, I used:
1) Enabling
2) Configuration
3) Disabling

I don't have a strong opinion on which one is more preferable but we should be consistent. @evilnick could you say which one you like better and then we update my or this PR accordingly

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,86 @@
...
+```
+To continue with the `Configuring DNS` section enable dns again.
+
+## Configuring DNS
+Discover your configuration options by running:
+```bash
+sudo k8s set dns –help
+```
+You should see three options:
+- upstream-dns: the dns server used to forward known entries
```
bschimke95, line 63: I would refer to the reference section for all the configuration options. 
Let's just take one example config and explain how to change that.

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,86 @@
...
+- cluster-domain: the cluster domain name
+- dns-ip: the cluster ip to be assigned to the dns service
+
+Set a new DNS server IP for forwarding known entries:
+```bash
+sudo k8s set dns --upstream-dns=<new-ip>
```
bschimke95, line 69: I think the spec uses:
`sudo k8s set dns.upstream-dns=<new>`

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,86 @@
...
+Assign a new cluster IP to the DNS service.
+```bash
+sudo k8s set dns --dns-ip=<new-cluster-ip>
+```
+Replace `<new-ip>`, `<new-domain-name>`, and `<new-cluster-ip>` with the desired values for your DNS configuration.
+
+
+<!-- LINKS -->
+
+[Component Upgrades]: #TODO
```
bschimke95, line 84: Missing brackets

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,86 @@
+# How to use default DNS
+
+Canonical Kubernetes includes a default DNS (Domain Name System) essential for internal communication within your cluster. 
+When enabled, the DNS facilitates service discovery
+by assigning each service a DNS name.
```
bschimke95, line 5: nit: could be on the same line for readability

bschimke95 (commented):
Great work, added some general questions for the how-to documentation.

Also, I think you have to add the docs to the index.md to pop up (See https://github.com/canonical/k8s-snap/pull/164/files).
Have you checked the rendered output with `cd docs/tools && make serve`?

docs/src/howto/networking/default-dns.md
evilnick: Yeah, I already mentioned this in the comment ^

The Canonical style is to use imperatives rather than progressives, so 'Enable xxx' rather than 'Enabling xxx'
If it helps, imaging sticking 'How to' in front of each heading

bschimke95: Thanks! My question was more towards the structure, though.
e.g. should we explain how to enable/disable the component in one section and then how to configure it in another?
Or should we have one section on how to enable the component, then we configure it in the next section and disable it in the last one.
WDYT?

louiseschmidtgen: I like the suggested structure from Benjamin. Then I do not need to say in the middle, hey re-enable what you just turned off if you would like to continue.

docs/src/howto/networking/default-dns.md
louiseschmidtgen: Are you sure? I see:
```
k8s set <functionality> <key=value> ...
k8s get <functionality.key>
```

louiseschmidtgen: Or is that outdated?

bschimke95: you are right - reading is hard :)

But, aren't we both wrong then? It should be:
`k8s set dns upstream-dns=...` without the flag dashes, right?

louiseschmidtgen: I wanted to try it but it's not there yet :) Let me update it...

docs/src/howto/networking/default-dns.md
louiseschmidtgen: @evilnick What ideal structure would you prefer?

docs/src/howto/networking/default-dns.md
louiseschmidtgen: @evilnick Would you prefer one or all? Also should I add a get (ex: `k8s get dns.upstream-dns`) config option?

docs/src/howto/networking/default-dns.md
louiseschmidtgen: The new warning makes it clear not to disable dns now, unless you really want to see some flames. SO we could keep this structure also. I'm happy either way :)

evilnick: yes, definitely. I was a bit confused because at one point I'm sure someone mentioned we would have to set the config for things ***before*** enabling them

docs/src/howto/networking/default-dns.md
```diff
@@ -1,7 +1,8 @@
 # How to use default DNS
 
 Canonical Kubernetes includes a default DNS (Domain Name System) essential for internal communication within your cluster. 
-When enabled, the DNS facilitates service discovery by assigning each service a DNS name. When disabled, you can integrate a custom dns solution into your cluster.
+When enabled, the DNS facilitates service discovery by assigning each service a DNS name. 
+When disabled, you can integrate a custom dns solution into your cluster.
```
eaudetcobello, line 5: Capital DNS :)

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,82 @@
...
+Canonical Kubernetes includes a default DNS (Domain Name System) essential for internal communication within your cluster. 
+When enabled, the DNS facilitates service discovery by assigning each service a DNS name. 
+When disabled, you can integrate a custom dns solution into your cluster.
+
+
+## What you'll need
+
+This guide assumes the following:
+
+- You have root or sudo access to the machine
```
eaudetcobello, line 12: Period at the end :)

eaudetcobello, line 12:
```suggestion
- You have root or sudo access to the machine.
```

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,82 @@
+# How to use default DNS
+
+Canonical Kubernetes includes a default DNS (Domain Name System) essential for internal communication within your cluster. 
+When enabled, the DNS facilitates service discovery by assigning each service a DNS name. 
+When disabled, you can integrate a custom dns solution into your cluster.
```
eaudetcobello, line 5:
```suggestion
When disabled, you can integrate a custom DNS solution into your cluster.
```

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,82 @@
...
+sudo k8s set dns.dns-ip=<new-cluster-ip>
+```
+Replace `<new-ip>`, `<new-domain-name>`, and `<new-cluster-ip>` with the desired values for your DNS configuration.
+
+## Disable DNS
+
+Canonical Kubernetes also allows you to disable the built-in DNS, 
+if you desire a custom solution:
+
+``` {warning} Do not disable DNS unless you have a replacement configured. Disabling DNS will disrupt internal cluster communication. Ensure a suitable custom DNS solution is in place before disabling. You can re-enable dns at any point, and your cluster will return to normal functionality.```
```
eaudetcobello, line 67:
```suggestion
``` {warning} Do not disable DNS unless you have a replacement configured. Disabling DNS will disrupt internal cluster communication. Ensure a suitable custom DNS solution is in place before disabling. You can re-enable DNS at any point, and your cluster will return to normal functionality.```
```

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,82 @@
+# How to use default DNS
+
+Canonical Kubernetes includes a default DNS (Domain Name System) essential for internal communication within your cluster.
```
eaudetcobello, line 3: Maybe say "because it is so essential for internal communication within..." ?

docs/src/howto/networking/default-dns.md
```diff
@@ -0,0 +1,82 @@
...
+When enabled, the DNS facilitates service discovery by assigning each service a DNS name. 
+When disabled, you can integrate a custom dns solution into your cluster.
+
+
+## What you'll need
+
+This guide assumes the following:
+
+- You have root or sudo access to the machine
+- You have a bootstraped Canonical Kubernetes cluster (see the [getting-started-guide]).
```
eaudetcobello, line 13:
```suggestion
- You have a bootstraped Canonical Kubernetes cluster (see the [Getting Started] guide).
```

docs/src/howto/networking/default-dns.md
evilnick: There currently is no reference section to refer to and I would think it was easier for users to find them all in this doc

evilnick (approved):
:cake:

PR #171, opened 2024-02-29
Generate kubeconfig without localhost server address

kwmonroe:
### Summary

Ensure the output from `k8s config` returns a more useful server address, or one that is optionally specified as a command flag.

### Changes

- added `server` flag to k8s config command
- if flag is present, overwrite the cluster `server:` value in the kubeconfig presented to the user
- if flag is absent, overwrite the cluster `server:` value with the hostname of the node


src/k8s/pkg/k8sd/api/kubeconfig.go
```diff
@@ -1,25 +1,39 @@
...
 
 	apiv1 "github.com/canonical/k8s/api/v1"
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/microcluster/state"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 func getKubeconfig(s *state.State, r *http.Request) response.Response {
-	// TODO: Render a new kubeconfig instead of reading the existing one
-	//       when the config can be altered via request parameters.
```
kwmonroe, line 15: I'm not "rendering a new kubeconfig" here, but i am returning one that has been modified by request params.  I think that covers the spirit of the above comment.

src/k8s/cmd/k8s/k8s_config.go
```diff
@@ -2,21 +2,40 @@ package k8s
...
+			snap := snap.NewSnap(os.Getenv("SNAP"), os.Getenv("SNAP_COMMON"))
+
+			isWorker, err := snaputil.IsWorker(snap)
+			if err != nil {
+				return fmt.Errorf("failed to check if node is a worker: %w", err)
+			}
+
+			if isWorker {
+				return fmt.Errorf("this action is restricted on workers")
+			}
```
neoaggelos, line 36: This is not needed here, it's already handled in https://github.com/canonical/k8s-snap/blob/73c17ef31708078874ba4a45f5c599a705c3458e/src/k8s/pkg/k8sd/api/endpoints.go#L46-L50

src/k8s/pkg/k8sd/api/kubeconfig.go
```diff
@@ -1,25 +1,39 @@
...
+	req := apiv1.GetKubeConfigRequest{}
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return response.BadRequest(fmt.Errorf("failed to parse request: %w", err))
+	}
+	server := req.Server
+	if req.Server == "" {
+		server = fmt.Sprintf("https://%s:6443", s.Address().Hostname())
+	}
+
+	config := clientcmd.GetConfigFromFileOrDie("/etc/kubernetes/admin.conf")
```
neoaggelos, line 24: NOTE: `GetConfigFromFileOrDie` will exit in case the file cannot be read. This is not what we want here, we want to handle possible errors instead.

src/k8s/pkg/k8sd/api/kubeconfig.go
neoaggelos: So, here we read in `/etc/kubernetes/admin.conf` (which might have been edited by a user already, and we do not know that.

Starting from the fact that a proper kubeconfig file looks like this:

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURHekNDQWdPZ0F3SUJBZ0lRQXN4WUpvdW45UU5lQmNseUdVMUdNekFOQmdrcWhraUc5dzBCQVFzRkFEQVkKTVJZd0ZBWURWUVFERXcxcmRXSmxjbTVsZEdWekxXTmhNQjRYRFRJME1ESXlPREUyTWpJeU9Gb1hEVE0wTURJeQpPREUyTWpJeU9Gb3dHREVXTUJRR0ExVUVBeE1OYTNWaVpYSnVaWFJsY3kxallUQ0NBU0l3RFFZSktvWklodmNOCkFRRUJCUUFEZ2dFUEFEQ0NBUW9DZ2dFQkFPbEJFQjRmcHFveHBUbDNnK0JNUHNmSWhtVmk4YUpLcFozdU5vYjAKMkI5TEVvQkw5REVNRG9PdVJCbXRsTmExdEFQaXdTclNKN09TaGluaW5aTmZlQTNmQnAwaHZqQ3BhMWg4MytPRQpYYncyZDNlMFF2Z20wWDIyVEs2THFmZUdBd1BkVXJjQ0RMNVk5UTl3QkoxTE5NdUU4eVNNSnZVS0lyMGdJMXdKCmsvVnFvcWdqMWtKQ0NMTldYeWdyQk5DQmJXQlFtaDgrK0Vsd0NjMW85NnNyM0JaOTh0OHBxTXFBMDRjSmYyYjAKM1pQM24yTk9JYVFYaktXbUhMRWo5elV3eDkrM0JqdEZCeldEcTVVQXl6NDg2a0ZpalJsbjRPU0JhMGk3M0VWaQpJSyszK1M3cU1kSkVzL0lJZGlUcHZDM0JHVnE1UVkvTkdJZnBjdXFoR3dvUUlVc0NBd0VBQWFOaE1GOHdEZ1lEClZSMFBBUUgvQkFRREFnS0VNQjBHQTFVZEpRUVdNQlFHQ0NzR0FRVUZCd01DQmdnckJnRUZCUWNEQVRBUEJnTlYKSFJNQkFmOEVCVEFEQVFIL01CMEdBMVVkRGdRV0JCU2lQeWNwUmxJU2w3QmM0S3E1LzdnazFkcElzVEFOQmdrcQpoa2lHOXcwQkFRc0ZBQU9DQVFFQUk5OEo3U2pZR0VCZ2F4dk01OStaYTY2dnVIY2RUY1o3SXo1ZmdGQTJ5QksrCkpEVU9jZEo2VnB3UmlXK0l4eUQ4OCtuL29kNWJqSGt4ZE1kbm9ZOUNZOXcrRHJWL0FwK09rUWRvNDJJZmNQMGkKWTA3UEFiZS9FVFI3cGFxOFNsUGpxTXV0ZERwanRZTVVMdk5HRkpGbll5SXZjTHA1VDdQazRITmsyOEtXWDhUSAp1ei9QR3BZdXRJSGM4NnV1bVZTV3RnM2craUhzdUgzOFQzaVF0RUs4b2dTMEhCRk0yTy8xbzEwZnMycHlkM3QwCm1aTGdyV3VSS3poTXl5a20vVlFTTHpFSE5CbkZUUW5tSXhxLzRJVUoraVd3SmVtVGxTK0twUFBLc2srTHlNUjEKVGhIWUV1d0NlM2czcWt0M2NnYWJCL1ZpRnlKYU1pdkppSVNsTmdaSmZ3PT0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
    server: https://127.0.0.1:6443
  name: k8s
contexts:
- context:
    cluster: k8s
    user: k8s-user
  name: k8s
current-context: k8s
kind: Config
users:
- name: k8s-user
  user:
    token: token::e652c6433bf2aaf16afa0140928f21737bb5dabb
```

The three pieces of information we need are the following:

- Cluster CA, which we can retrieve through the ClusterConfig (by means of https://github.com/canonical/k8s-snap/blob/460c50f7905765911298c850519f3238443c2ff5/src/k8s/pkg/k8sd/api/component.go#L86-L89)
- Server URL, which we have through the node address, or the parameter. Note that, if parameter is not set, we must hardcode 6443 here, but rather use the SecurePort from the cluster config https://github.com/canonical/k8s-snap/blob/a6e1a8bded087ed8006b3dee3d6470af92e5791a/src/k8s/pkg/k8sd/types/cluster_config.go#L44)
- Admin token, which we can retrieve by means of https://github.com/canonical/k8s-snap/blob/73c17ef31708078874ba4a45f5c599a705c3458e/src/k8s/pkg/k8sd/api/kubernetes_auth_tokens.go#L40-L44

We should not be reading in a file here, but rather:
- retrieve the info above
- build a new `*clientcmdapi.Config` object with this info (the same type that is returned by `clientcmd.GetConfigFromFileOrDie()`
- render it (as we do now) and return to the user.

neoaggelos (changes requested):
Nice initial work, left comments

src/k8s/pkg/k8sd/api/kubeconfig.go
kwmonroe: ack.  big thanks for the pointers to kubeconfig components and walkthrough.  i initially thought it better to read/mod `admin.conf` vs rendering one again, *especially* since it may have been tweaked by the user.

but knowing how many times i've broken kubeconfigs with typos, i see the point in having k8s render something with the things it knows should work.

src/k8s/pkg/k8sd/api/kubeconfig.go
```diff
@@ -1,25 +1,60 @@
...
+	clusterCfg, err := utils.GetClusterConfig(r.Context(), s)
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to retrieve cluster config: %w", err))
+	}
+	ca := clusterCfg.Certificates.CACert
+	server := req.Server
+	if req.Server == "" {
+		server = fmt.Sprintf("%s:%d", s.Address().Hostname(), clusterCfg.APIServer.SecurePort)
+	}
+	token, err := impl.GetOrCreateAuthToken(s.Context, s, "kubernetes-admin", []string{"system:masters"})
```
kwmonroe, line 33: assumes we always want this to return the admin kubeconfig.  any value in adding user/group as request params at the moment?

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
```diff
@@ -38,3 +39,14 @@ func Kubeconfig(path string, token string, url string, caPEM string) error {
 	defer file.Close()
 	return renderKubeconfig(file, token, url, caPEM)
 }
+
+// ClientKubeconfig provides a kubeconfig suitable for client manipulations.
+func ClientKubeconfig(token string, url string, caPEM string) ([]byte, error) {
```
neoaggelos, line 44: The name `ClientKubeconfig` here does not describe the usage, and does not help distinguish from the simple `Kubeconfig` just above. Should we call this something like `KubeconfigString`? and return a `(string, error)`? 
```suggestion
// ClientKubeconfig provides a kubeconfig suitable for client manipulations.
func ClientKubeconfig(token string, url string, caPEM string) ([]byte, error) {
```

neoaggelos (commented):
Looks a lot better, some follow up comments

src/k8s/pkg/k8sd/api/kubeconfig.go
```diff
@@ -1,25 +1,60 @@
...
+	}
+	clientConfig, err := clientcmd.NewClientConfigFromBytes(kubeconfig)
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to create ClientConfig: %w", err))
+	}
+	rawConfig, err := clientConfig.RawConfig()
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to create RawConfig: %w", err))
+	}
+	yamlConfig, err := clientcmd.Write(rawConfig)
```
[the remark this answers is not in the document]
kwmonroe: oh good! i had a `"Why not stop here", string(kubeconfig))` followup for you but failed to leave it here.  anyway, the back-n-forth is only because you said:

> build a new *clientcmdapi.Config object with this info (the same type that is returned by clientcmd.GetConfigFromFileOrDie()

i figured you must have had a reason for wanting `Config` objects in the picture (e.g. some indirect template validation against clientcmdapi).  Anyway, we're on the same page.  I can stop slinging bytes around and whittle this down into like 2 lines.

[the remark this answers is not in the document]
neoaggelos: That would have been something like:

```
config := &clientcmd.Config{}
config.CertificateAuthority = ca
config.AuthInfos = // ..
config.Servers = // ...

// render config object to string
```

It would be greatly preferred to the template we have, but keeping the template for now is also fine.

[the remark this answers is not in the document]
kwmonroe: alrighty, we're cooking with gas now.  i dropped the template logic in favor of `clientcmd.Config` generator.

src/k8s/pkg/k8sd/setup/util_kubeconfig_test.go
```diff
@@ -0,0 +1,39 @@
...
+current-context: k8s
+kind: Config
+preferences: {}
+users:
+- name: k8s-user
+  user:
+    token: token
+`, ca)
+
+	actual, _ := setup.KubeconfigString("token", "server", "ca")
```
neoaggelos, line 36: also treat the error, with `g.Expect(err).To(BeNil())`

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
```diff
@@ -1,40 +1,55 @@
...
+
+	return config
 }
 
 // Kubeconfig writes a kubeconfig file to disk.
 func Kubeconfig(path string, token string, url string, caPEM string) error {
-	file, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	config := createConfig(token, url, caPEM)
+	err := clientcmd.WriteToFile(*config, path)
+	if err != nil {
```
neoaggelos, line 41:
```suggestion
	if err := clientcmd.WriteToFile(*config, path); err != nil {
```

neoaggelos: Also, please keep using `clientcmd.Write()` as we have it already, so that we ensure the right permissions (0600) on the file.

neoaggelos (changes requested):
Very nice, almost there, final comments

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
kwmonroe: > please keep using clientcmd.Write() as we have it already, so that we ensure the right permissions (0600)

@neoaggelos i'd rather let [`clientcmd.WriteToFile` manage the file perms](https://github.com/kubernetes/client-go/blob/v0.29.2/tools/clientcmd/loader.go#L439-L441). today, we're both `0600`, but if a new best practice comes out tomorrow to go `0640`, i think client-go will notice before we do. bonus points that it builds the ppath for us as well.  wdyt?

src/k8s/cmd/k8s/k8s_config.go
```diff
@@ -25,4 +31,6 @@ func newKubeConfigCmd() *cobra.Command {
 			return nil
 		},
 	}
+	configCmd.PersistentFlags().StringVar(&configCmdOpts.server, "server", "", "Specify a custom cluster server address for the kubeconfig")
```
bschimke95, line 34: nit: should be lower-case according to https://github.com/canonical/k8s-snap/pull/188
"flag descriptions should start with lowercase and should not end with a '.'"

src/k8s/cmd/k8s/k8s_config.go
```diff
@@ -7,16 +7,22 @@ import (
...
+var (
+	configCmdOpts struct {
+		server string
+	}
+)
+
 func newKubeConfigCmd() *cobra.Command {
-	return &cobra.Command{
-		Use:     "config",
+	configCmd := &cobra.Command{
```
bschimke95, line 17: nit: let's call it `kubeConfigCmd` for consistency

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
```diff
@@ -1,40 +1,54 @@
...
+	return nil
+}
+
+// KubeconfigString provides a stringified kubeconfig.
+func KubeconfigString(token string, url string, caPEM string) (string, error) {
+	config := createConfig(token, url, caPEM)
+	kubeconfig, err := clientcmd.Write(*config)
 	if err != nil {
-		return fmt.Errorf("failed to open file: %w", err)
+		return "", err
```
bschimke95, line 51: please wrap the error

bschimke95 (commented):
great work, just minor comments

src/k8s/cmd/k8s/k8s_config.go
```diff
@@ -7,16 +7,22 @@ import (
...
+		Use:     "config --server <server>",
```
bschimke95, line 18: you don't need to flags here - cobra will take care of this automatically. `config` is sufficient.

bschimke95 (approved):
LGTM, great work!

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
neoaggelos: sure, that works

neoaggelos (approved):

PR #172, opened 2024-02-29
ComponentManager.Enable unit test

eaudetcobello:


src/k8s/pkg/component/component_test.go
```diff
@@ -61,6 +61,12 @@ func buildChart(opts ...chartOption) *chart.Chart {
 	return c.Chart
 }
 
+func withName(name string) chartOption {
```
louiseschmidtgen, line 64: Why don't we pass the chart name straight into the buildChart function?
You should be able to do something like this:
```
myChart := buildChart(
    func(opts *chartOptions) {
        opts.Chart.Metadata.Name = "custom-chart"
    },
   ```
   The opts for-loop in `buildChart` should be able to set the chart options for you.

File: src/k8s/pkg/component/component_test.go [hunk 2]
```diff
@@ -200,3 +216,68 @@ func TestListComponentsWithReleases(t *testing.T) {
...
+	})
+
+	components, err := mockHelmClient.List()
+	g.Expect(err).ShouldNot(HaveOccurred())
+	for _, component := range components {
+		g.Expect(component.Status).To(BeFalse(), "Expected all components to be initially disabled")
+	}
+}
+
+func TestEnableSingleComponent(t *testing.T) {
```
louiseschmidtgen, line 235: What does `TestEnableSingleComponent` test that doesn't get covered in `TestEnableMultipleComponents`?

src/k8s/pkg/component/component_test.go
```diff
@@ -200,3 +216,68 @@ func TestListComponentsWithReleases(t *testing.T) {
 		{Name: "two", Status: true},
 	}))
 }
+
+func TestComponentsInitialState(t *testing.T) {
+	g := NewWithT(t)
+
+	mockHelmClient, _, _ := mustCreateNewHelmClient(t, map[string]types.Component{
+		"one": {ReleaseName: "whiskas-1", Namespace: "default", ManifestPath: "chunky-tuna-0.1.0.tgz"},
```
louiseschmidtgen, line 224: Miny tweak: Can we either re-format my releases to match your formatting or vice versa? Does `make go.fmt` take care of this?

louiseschmidtgen (approved):
Great work @eaudetcobello !
I like the choice of your renaming in `component.go`.
Added a few smaller suggestions. Overall, looks good to me. 👍🏼

src/k8s/pkg/component/component_test.go
eaudetcobello: In upstream code there is withName, withNamespace, etc.
https://github.com/helm/helm/blob/15d116b8548f59959fe9f2fafdaf7c6de8db3ea0/pkg/action/action_test.go#L135

src/k8s/pkg/component/component_test.go
eaudetcobello: Good point

PR #173, opened 2024-02-29
Docs: Add roadmap

evilnick:
There isn't a lot to say here at the moment.

The table is sourced from a simple CSV file to make it easier to update and/or source from elsewhere

### DIFF
--- a/docs/src/reference/roadmap.md
+++ b/docs/src/reference/roadmap.md
@@ -1,8 +1,11 @@
 # Roadmap
 
-The Canonical Kubernetes team  recognises the value of having a public roadmap, letting everyone know the headline features we are working on and the future direction and priorities of the project.
+The Canonical Kubernetes team  recognises the value of having a public roadmap,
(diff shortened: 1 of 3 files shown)


evilnick:
this is it, basically:

image

docs/src/reference/roadmap.md
```diff
@@ -0,0 +1,15 @@
+# Roadmap
+
+The Canonical Kubernetes team recognises the value of having a public roadmap,
```
eaudetcobello, line 3: Could we say something like "enthusiastically supports the idea of a public roadmap" instead of a neutral acknowledgement?

evilnick: sure

evilnick: I will be counting on you to enthusiastically keep it up to date :)

evilnick:
I got a verbal +1 from Perk for this so am going to go ahead and merge

PR #179, opened 2024-02-29
Add GET k8sd/node endpoint

bschimke95:
Provides an endpoint to query the current status of the local k8s node. This is especially useful for the charm that wants to know the current role of specific node. It is not possible to query other nodes for security purposes.

Additionally,
* changed `ClusterMember` to `NodeStatus` to better reflect the new shared use
* added a hidden CLI command `k8s local-node-status` for testing.
* added basic test coverage by integrating assertions into the `test_clustering.py` test.
* use enum for datastore and cluster roles to not rely on random strings.

### DIFF
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -73,22 +73,22 @@ const (
 type NodeStatus struct {
 	// Name is the name for this cluster member that was when joining the cluster.
 	// This is typically the hostname of the node.
-	Name string `mapstructure:"name,omitempty"`
+	Name string `json:"name,omitempty"`
 	// Address is the IP address of the node.
-	Address string `mapstructure:"address,omitempty"`
+	Address string `json:"address,omitempty"`
 	// ClusterRole is the role that the node has within the k8s cluster.
-	ClusterRole ClusterRole `mapstructure:"cluster-role,omitempty"`
+	ClusterRole ClusterRole `json:"cluster-role,omitempty"`
 	// DatastoreRole is the role that the node has within the datastore cluster.
 	// Only applicable for control-plane nodes, empty for workers.
-	DatastoreRole DatastoreRole `mapstructure:"datastore-role,omitempty"`
+	DatastoreRole DatastoreRole `json:"datastore-role,omitempty"`
 }
 
 // ClusterStatus holds information about the cluster, e.g. its current members
 type ClusterStatus struct {
 	// Ready is true if at least one node in the cluster is in READY state.
-	Ready      bool         `mapstructure:"ready,omitempty"`
-	Members    []NodeStatus `mapstructure:"members,omitempty"`
-	Components []Component  `mapstructure:"components,omitempty"`
+	Ready      bool         `json:"ready,omitempty"`
+	Members    []NodeStatus `json:"members,omitempty"`
+	Components []Component  `json:"components,omitempty"`
 }
 
 // HaClusterFormed returns true if the cluster is in high-availability mode (more than two voter nodes).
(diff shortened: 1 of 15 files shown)


src/k8s/cmd/k8s/k8s_local_node_status.go
```diff
@@ -0,0 +1,29 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to get cluster status: %w", err)
+			}
+
+			fmt.Println(clusterStatus)
+			return nil
+		},
+	}
+	return localNodeStatusCmd
+}
```
mateoflorido, line 29: Are we keeping this as a hidden command? What if we could repurpose it and make it a hidden subcommand of the k8s status? If it does not deliver relevant information, I'm OK to just dropping it.

src/k8s/pkg/k8sd/api/impl/k8sd.go
```diff
@@ -58,21 +60,50 @@ func GetClusterMembers(ctx context.Context, s *state.State) ([]apiv1.ClusterMemb
 		return nil, fmt.Errorf("failed to get cluster members: %w", err)
 	}
 
-	members := make([]apiv1.ClusterMember, len(clusterMembers))
+	members := make([]apiv1.NodeStatus, len(clusterMembers))
 	for i, clusterMember := range clusterMembers {
-		fingerprint, err := shared.CertFingerprintStr(clusterMember.Certificate.String())
```
mateoflorido, line 63: Is `fingerprint` no longer required?

mateoflorido (approved):
Just a couple of questions out of curiosity. LGTM!

src/k8s/api/v1/types.go
```diff
@@ -48,28 +48,54 @@ func BootstrapConfigFromMap(m map[string]string) (*BootstrapConfig, error) {
...
+
+// DatastoreRole as provided by dqlite
+type DatastoreRole string
+
+const (
+	DatastoreRoleVoter   DatastoreRole = "voter"
+	DatastoreRoleStandBy DatastoreRole = "stand-by"
+	DatastoreRoleSpare   DatastoreRole = "spare"
+	DatastoreRolePending DatastoreRole = "PENDING"
+	DatastoreRoleUnknown DatastoreRole = "unknown"
```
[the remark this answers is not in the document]
bschimke95: Super weird, isn't it?
Complains directly to the dqlite team please ;)

src/k8s/cmd/k8s/k8s_local_node_status.go
bschimke95: This is really just a hidden command for testing and debugging purpose. We could drop it but I also don't see a problem with keeping it.

src/k8s/pkg/k8sd/api/impl/k8sd.go
bschimke95: It was never. Just a leftover from early days where I randomly copied Microcluster examples.

src/k8s/api/v1/types.go
```diff
@@ -48,28 +48,54 @@ func BootstrapConfigFromMap(m map[string]string) (*BootstrapConfig, error) {
...
+)
+
+// NodeStatus holds information about a node in the k8s cluster.
+type NodeStatus struct {
+	// Name is the name for this cluster member that was when joining the cluster.
+	// This is typically the hostname of the node.
+	Name string `mapstructure:"name,omitempty"`
```
neoaggelos, line 76: what are the mapstructure tags for? let's just use `json`.

neoaggelos (approved):
LGTM, just one comment

PR #181, opened 2024-02-29
Tweaks in docs

ktsakalozos:
Some updates in our docs:
- use `bash` in all cli commands
- the snap now auto enables dns network and metrics server
- remove "components" from various places including our architecture doc
- include the arch.dsl used to produce all arch graphs
- same formatting in configuration options in howtos
- metrics-server is part of the distribution
- RBAC is on by default so we do not mention it
- local-storage vs storage fix
- --help fix

### DIFF
--- a/docs/src/howto/install/snap.md
+++ b/docs/src/howto/install/snap.md
@@ -39,7 +39,7 @@ sudo snap install k8s --classic --channel=latest/edge
 ```
 
 ```{note}
-In `latest/edge` channel is always under active development. This is where you will find the latest features but you may also experience instability.
+The `latest/edge` channel is always under active development. This is where you will find the latest features but you may also experience instability.
 ```
 
 ## Bootstrap the cluster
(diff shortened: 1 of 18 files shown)


docs/src/howto/install/snap.md
```diff
@@ -39,7 +39,7 @@ sudo snap install k8s --classic --channel=latest/edge
 ```
 
 ```{note}
-In the pre-release phase, `latest/edge` is the only channel available. 
+In `latest/edge` channel is always under active development. This is where you will find the latest features but you may also experience instability.
```
kwmonroe, line 42:
```suggestion
The `latest/edge` channel is always under active development. This is where you will find the latest features but you may also experience instability.
```

kwmonroe (commented):
One tiny tweak, but it's looking good -- thanks @ktsakalozos!

kwmonroe (approved):

PR #182, opened 2024-03-01
Changing to get/set notation for functionalities

berkayoz:
The `k8s set` and `k8s get` is implemented and works as expected. There are things that can be considered nice to haves and can be addressed with other PRs

TODO:
* Proper help messages for `get` and `set`
* Auto-completion support for `get` and `set`
* Verify grouped arguments (bgp) are set together and fail if any one of them is missing
* In-code `component` words to be replaced by `functionality`


src/k8s/cmd/k8s/k8s_disable.go
```diff
@@ -5,34 +5,77 @@ import (
...
+			}
+
+			enabled := false
+
+			config := api.UserFacingClusterConfig{}
+
+			switch args[0] {
+			case "network":
+				config.Network = &api.NetworkConfig{
+					Enabled: &enabled,
```
neoaggelos, line 39: we'll have this a lot, does it make sense to add a util function like

```
func Pointer(v T) *T {
  return &v
}
```

then change all these to `utils.Pointer(false)`?

```suggestion
					Enabled: &enabled,
```

src/k8s/cmd/k8s/k8s_enable.go
```diff
@@ -5,37 +5,80 @@ import (
...
+			}
+
+			enabled := true
+
+			config := api.UserFacingClusterConfig{}
+
+			switch args[0] {
+			case "network":
+				config.Network = &api.NetworkConfig{
+					Enabled: &enabled,
```
neoaggelos, line 43: same here

src/k8s/cmd/k8s/k8s_set.go
```diff
@@ -0,0 +1,275 @@
...
+	val, ok := configs[key]
+	if ok {
+		parsed, err := readAsCSV(val)
+		if err != nil {
+			return fmt.Errorf("invalid string list value: %w", err)
+		}
+		*target = parsed
+	}
+	return nil
+}
```
neoaggelos, line 267: let's use `strings.Fields` instead

src/k8s/cmd/k8s/k8s_set.go
```diff
@@ -0,0 +1,275 @@
...
+				_, ok := settingsMap[functionality]
+
+				if !ok {
+					settingsMap[functionality] = map[string]string{}
+				}
+
+				settingsMap[functionality][key] = value
+			}
+
+			for functionality, configs := range settingsMap {
```
neoaggelos, line 94: Let's use a flat map instead, it should be easier to follow here, and possibly extend in the future.

E.g.

```
for _, arg := range args {
  key, value = splitKeyAndValue(arg)
  switch key {
  case "network.enabled":
    // ....
  default:
    // unknown config
  }
}
```

src/k8s/pkg/component/dns.go
```diff
@@ -132,3 +134,130 @@ func DisableDNSComponent(ctx context.Context, s snap.Snap) error {
...
+		err := DisableDNSComponent(ctx, s)
+		if err != nil {
+			return "", "", fmt.Errorf("failed to disable dns: %w", err)
+		}
+		return "", "", nil
+	} else {
+		// do nothing
+	}
+	return "", "", nil
+}
```
neoaggelos, line 263: I see why we do it like so, but we probably want to revisit this logic in the future

src/k8s/pkg/k8sd/types/cluster_config.go
```diff
@@ -112,4 +173,107 @@ func ClusterConfigFromBootstrapConfig(b *apiv1.BootstrapConfig) ClusterConfig {
...
+			config.LoadBalancer = LoadBalancer{
+				Enabled: &[]bool{true}[0],
+			}
+		}
+	}
+
+	return config
+}
+
+func ClusterConfigFromUserFacing(ufConfig *apiv1.UserFacingClusterConfig) ClusterConfig {
```
neoaggelos, line 198: I would also include some validation here. I see the TODO for later, I think that's also OK as long as we don't forget about it

src/k8s/cmd/k8s/k8s_get.go
```diff
@@ -0,0 +1,67 @@
...
+	"slices"
+
+	api "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/k8s/cmd/k8s/errors"
+	"github.com/canonical/k8s/cmd/k8s/formatter"
+	"github.com/spf13/cobra"
+)
+
+func newGetCmd() *cobra.Command {
+	getCmd := &cobra.Command{
```
neoaggelos, line 16: does this only receive one key?

neoaggelos (commented):
Very nice @berkayoz, leaving some comments

src/k8s/cmd/k8s/k8s_get.go
berkayoz: I believe it will be either `functionality` or `functionality.key`, I'm still working on this part.

src/k8s/cmd/k8s/k8s_disable.go
bschimke95: why not just `new(bool)`?

berkayoz: @bschimke95 I think `new(bool)` would only assign zero value(false)

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -272,57 +271,64 @@ func onBootstrapControlPlane(s *state.State, initConfig map[string]string) error
...
-					},
-				}); err != nil {
-					return fmt.Errorf("failed to update cluster configuration for dns=%s domain=%s: %w", dnsIP, clusterDomain, err)
-				}
-				return nil
-			}); err != nil {
-				return fmt.Errorf("database transaction to update cluster configuration failed: %w", err)
-			}
+	if cfg.Network.Enabled != nil {
+		err := component.ReconcileNetworkComponent(s.Context, snap, &[]bool{false}[0], cfg.Network.Enabled, cfg)
```
bschimke95: just use `new(bool)` here and elsewhere.

src/k8s/pkg/k8sd/types/cluster_config.go
```diff
@@ -112,4 +172,93 @@ func ClusterConfigFromBootstrapConfig(b *apiv1.BootstrapConfig) ClusterConfig {
 			Port: b.K8sDqlitePort,
 		},
 	}
+
+	for _, component := range b.Components {
+		switch component {
+		case "network":
+			config.Network.Enabled = &[]bool{true}[0]
```
bschimke95: ok, here you cannot use `new(bool)` but please just create a variable and reference it. Creating 7 slices to get bool references sounds unnecessary.

bschimke95: or better just create a `utils.Pointer(true)` as Angelos suggested

bschimke95 (commented):
Great work! Left some comments.

src/k8s/cmd/k8s/k8s_set.go
berkayoz: Doesn't `strings.Fields` split on whitespace?

src/k8s/pkg/component/dns.go
berkayoz: I am not exactly happy about this as well, the main reason is getting the assigned Service IP to persist it.

src/k8s/api/v1/cluster_config.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+type UpdateClusterConfigRequest struct {
+	Config UserFacingClusterConfig
+}
+
+type UpdateClusterConfigResponse struct {
+}
+
+type UserFacingClusterConfig struct {
```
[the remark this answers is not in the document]
berkayoz: I think we might want to expose stuff other than just functionality configs, the UserFacing is there to improve readability as we already have `ClusterConfig` although still possible since they are in different packages. `types.ClusterConfig` vs `api.ClusterConfig`

src/k8s/pkg/component/dns.go
```diff
@@ -87,6 +88,7 @@ func EnableDNSComponent(ctx context.Context, s snap.Snap, clusterDomain, service
 		return "", "", fmt.Errorf("failed to get dns service: %w", err)
 	}
 
+	// TODO: this should propagate to all nodes
```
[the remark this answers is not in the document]
berkayoz: This is not going to be a part of this card, as it's a big task and really out of scope for this one. I'll create a card.

src/k8s/cmd/k8s/k8s_set.go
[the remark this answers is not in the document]
berkayoz: I think we can create a card for this an maybe with improving bash-completion and help messages.

src/k8s/pkg/utils/vals/pointer.go
```diff
@@ -0,0 +1,13 @@
...
+	return &v
+}
+
+func Deref[T any](p *T) T {
+	if p == nil {
+		var zero T
+		return zero
+	}
+	return *p
+}
```
neoaggelos, line 13: This is only used for optional `bool` values, make this clear:

```suggestion
func OptionalBool(v *bool, defaultValue bool) T {
	if v != nil {
	   return *v
	}
	return defaultValue
}
```

src/k8s/pkg/k8sd/api/cluster_config.go
```diff
@@ -0,0 +1,214 @@
...
+
+	falseByDefault := false
+	trueByDefault := true
+
+	userFacing := api.UserFacingClusterConfig{
+		Network: &api.NetworkConfig{
+			Enabled: &trueByDefault,
+		},
+		DNS: &api.DNSConfig{
+			Enabled:             &trueByDefault,
```
neoaggelos, line 131: use `utils.Pointer` here, also below

src/k8s/pkg/component/storage.go
```diff
@@ -57,3 +67,36 @@ func DisableStorageComponent(s snap.Snap) error {
...
+		return nil
+	} else if !vals.Deref(requestEnabled) {
+		err := DisableStorageComponent(s)
+		if err != nil {
+			return fmt.Errorf("failed to disable local-storage: %w", err)
+		}
+		return nil
+	} else {
+		// do nothing
+	}
```
neoaggelos, line 100: This is checking `if A { } else if !A { } else`. the final else is unnecessary.

Also in other places

```suggestion
	} else if vals.Deref(requestEnabled) {
		err := UpdateStorageComponent(ctx, s, false, clusterConfig.LocalStorage.LocalPath, clusterConfig.LocalStorage.ReclaimPolicy, setDefault)
		if err != nil {
			return fmt.Errorf("failed to enable local-storage: %w", err)
		}
		return nil
	} else {
		err := DisableStorageComponent(s)
		if err != nil {
			return fmt.Errorf("failed to disable local-storage: %w", err)
		}
		return nil
	}
```

src/k8s/pkg/utils/vals/pointer.go
```diff
@@ -0,0 +1,13 @@
+package vals
```
neoaggelos, line 1: better move this under `utils`: `utils.Pointer` and `utils.OptionalBool` are fine as names

neoaggelos (commented):
Also make sure to update all call-sites

src/k8s/cmd/k8s/k8s_set.go
```diff
@@ -0,0 +1,301 @@
...
+			"bgp-mode",
+			"bgp-local-asn",
+			"bgp-peer-address",
+			"bgp-peer-port",
+			"bgp-peer-asn",
+		},
+		"metrics-server": []string{"enabled"},
+	}
+)
+
```
neoaggelos, line 37: you can drop this

src/k8s/cmd/k8s/k8s_set.go
```diff
@@ -0,0 +1,301 @@
...
+				}
+
+				key := matches[2]
+				if !slices.Contains(settings, key) {
+					return fmt.Errorf("invalid key: %s", arg)
+				}
+			}
+
+			return nil
+		},
```
neoaggelos, line 70: also this

src/k8s/cmd/k8s/k8s_set.go
```diff
@@ -0,0 +1,301 @@
...
+						return fmt.Errorf("invalid boolean value for load-balancer.bgp-peer-asn: %w", err)
+					}
+				case "metrics-server.enabled":
+					if config.MetricsServer == nil {
+						config.MetricsServer = &api.MetricsServerConfig{}
+					}
+					if err := setBoolArgument(value, &config.MetricsServer.Enabled); err != nil {
+						return fmt.Errorf("invalid boolean value for metrics-server.enabled: %w", err)
+					}
+				}
```
neoaggelos, line 240: add a  default case here, to handle unknown keys

src/k8s/cmd/k8s/k8s_set.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+func setStringSliceArgument(val string, target *[]string) error {
+	parsed, err := readAsCSV(val)
+	if err != nil {
+		return fmt.Errorf("invalid string list value: %w", err)
+	}
+	*target = parsed
+	return nil
+}
```
neoaggelos, line 296: use `strings.Field` here, this was not addressed

src/k8s/pkg/component/dns.go
```diff
@@ -116,19 +125,46 @@ func DisableDNSComponent(ctx context.Context, s snap.Snap) error {
...
+		return dnsIP, clusterDomain, nil
+	} else if !vals.Deref(requestEnabled) {
+		err := DisableDNSComponent(ctx, s)
+		if err != nil {
+			return "", "", fmt.Errorf("failed to disable dns: %w", err)
+		}
+		return "", "", nil
+	} else {
+		// do nothing
+	}
```
neoaggelos, line 168: same here and in other places

neoaggelos (commented):
Some more comments

src/k8s/cmd/k8s/k8s_set.go
```diff
@@ -0,0 +1,241 @@
...
+					}
+					if err := setBoolArgument(value, &config.DNS.Enabled); err != nil {
+						return fmt.Errorf("invalid boolean value for dns.enabled: %w", err)
+					}
+				case "dns.upstream-nameservers":
+					if config.DNS == nil {
+						config.DNS = &api.DNSConfig{}
+					}
+					if err := setStringSliceArgument(value, &config.DNS.UpstreamNameservers); err != nil {
+						return fmt.Errorf("invalid boolean value for dns.upstream-nameservers: %w", err)
```
neoaggelos, line 46: string slice on the message, also below

neoaggelos (commented):
Final comment, and I think we are good to go

src/k8s/pkg/k8sd/api/cluster_config.go
```diff
@@ -0,0 +1,212 @@
...
+			Enabled:       vals.Pointer(false),
+			LocalPath:     cfg.LocalStorage.LocalPath,
+			ReclaimPolicy: cfg.LocalStorage.ReclaimPolicy,
+			SetDefault:    vals.Pointer(true),
+		},
+		Gateway: &api.GatewayConfig{
+			Enabled: vals.Pointer(false),
+		},
+		MetricsServer: &api.MetricsServerConfig{
+			Enabled: vals.Pointer(false),
```
bschimke95, line 159: this should be enabled by default, shouldn't it?

berkayoz: This is for initializing an empty struct, the default components to be enabled comes from the `BootstrapConfig` as implemented previously. Check `ClusterConfigFromBootstrapConfig` to see how it's adjusted.

bschimke95 (approved):
LGTM

neoaggelos (approved):

PR #184, opened 2024-03-01
Update k8s status command

bschimke95:
Example output:
```
ubuntu@alice:~/k8s$ sudo k8s status
status: ready
high-availability: no
datastore:
  master-nodes:
    - 10.84.220.237:6400
  standy-nodes: none
  spare-nodes:
    - 10.84.220.25:6400

network:
  enabled: true
dns:
  enabled: true
  cluster-domain: cluster.local
  service-ip: 10.152.183.108
  upstream-nameservers:
  - /etc/resolv.conf
metrics-server:
  enabled: true
```

### DIFF
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -186,7 +186,6 @@ func (c ClusterStatus) String() string {
 
 	b, _ := yaml.Marshal(printedConfig)
 	result.WriteString(string(b))
-	result.WriteString("\n")
 
 	return result.String()
 }
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -2,7 +2,6 @@ package v1
 
 import (
 	"fmt"
-	"net"
 	"strings"
 
 	"github.com/canonical/k8s/pkg/utils/vals"
@@ -87,9 +86,9 @@ type NodeStatus struct {
 // ClusterStatus holds information about the cluster, e.g. its current members
 type ClusterStatus struct {
 	// Ready is true if at least one node in the cluster is in READY state.
-	Ready      bool         `json:"ready,omitempty"`
-	Members    []NodeStatus `json:"members,omitempty"`
-	Components []Component  `json:"components,omitempty"`
+	Ready   bool                    `json:"ready,omitempty"`
+	Members []NodeStatus            `json:"members,omitempty"`
+	Config  UserFacingClusterConfig `json:"config,omitempty"`
 }
 
 // HaClusterFormed returns true if the cluster is in high-availability mode (more than two voter nodes).
@@ -103,13 +102,14 @@ func (c ClusterStatus) HaClusterFormed() bool {
 	return voters > 2
 }
 
+// TODO: Print k8s version. However, multiple nodes can run different version, so we would need to query all nodes.
 func (c ClusterStatus) String() string {
 	result := strings.Builder{}
 
 	if c.Ready {
-		result.WriteString("status: ready.")
+		result.WriteString("status: ready")
 	} else {
-		result.WriteString("status: not ready.")
+		result.WriteString("status: not ready")
 	}
 	result.WriteString("\n")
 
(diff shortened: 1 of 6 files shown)


src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -41,21 +41,23 @@ func newBootstrapCmd() *cobra.Command {
...
 				config.SetDefaults()
 			}
+			const minTimeout = 3 * time.Second
+			if bootstrapCmdOpts.timeout < minTimeout {
+				cmd.PrintErrf("Timeout %v is less than minimum of %v. Using the minimum %v instead.\n", bootstrapCmdOpts.timeout, minTimeout, minTimeout)
+				bootstrapCmdOpts.timeout = minTimeout
+			}
+
+			timeoutCtx, cancel := context.WithTimeout(cmd.Context(), joinNodeCmdOpts.timeout)
+			defer cancel()
```
bschimke95, line 57: drive-by fix: context was not using the timeout

src/k8s/api/v1/types.go
```diff
@@ -119,20 +119,74 @@ func (c ClusterStatus) String() string {
...
+	if len(voters) > 0 {
+		result.WriteString(fmt.Sprintf("  master-nodes:\n"))
+		for _, voter := range voters {
+			result.WriteString(fmt.Sprintf("    - %s\n", voter.Address))
+		}
+	} else {
+		result.WriteString(fmt.Sprintf("  master-nodes: none\n"))
+	}
+	if len(standBys) > 0 {
+		result.WriteString(fmt.Sprintf("  standy-nodes:\n"))
```
eaudetcobello, line 147:
```suggestion
		result.WriteString(fmt.Sprintf("  standby-nodes:\n"))
```

src/k8s/api/v1/types.go
```diff
@@ -119,20 +119,74 @@ func (c ClusterStatus) String() string {
...
+	} else {
+		result.WriteString(fmt.Sprintf("  standy-nodes: none\n"))
+	}
+	if len(spares) > 0 {
+		result.WriteString(fmt.Sprintf("  spare-nodes:\n"))
+		for _, spare := range spares {
+			result.WriteString(fmt.Sprintf("    - %s\n", spare.Address))
+		}
+	} else {
+		result.WriteString(fmt.Sprintf("  spare-nodes: none\n"))
```
eaudetcobello, line 160: Could a function be added to reduce repetitiveness of building the string?

src/k8s/api/v1/types.go
```diff
@@ -119,20 +119,74 @@ func (c ClusterStatus) String() string {
...
+		printedConfig.Ingress = c.Config.Ingress
+	}
+	if c.Config.LoadBalancer.Enabled != nil && *c.Config.LoadBalancer.Enabled {
+		printedConfig.LoadBalancer = c.Config.LoadBalancer
 	}
+	if c.Config.LocalStorage.Enabled != nil && *c.Config.LocalStorage.Enabled {
+		printedConfig.LocalStorage = c.Config.LocalStorage
+	}
+	if c.Config.Gateway.Enabled != nil && *c.Config.Gateway.Enabled {
+		printedConfig.Gateway = c.Config.Gateway
```
eaudetcobello, line 181: How about something like this?

```suggestion
		updatePrintedConfigIfEnabled(c.Config.Network, func() {
             printedConfig.Network = c.Config.Network
        })
        updatePrintedConfigIfEnabled(c.Config.DNS, func() {
             printedConfig.DNS = c.Config.DNS
        })
```

eaudetcobello (commented):
Left some comments :)

src/k8s/api/v1/types.go
bschimke95: hm, the `updatePrintedConfigIfEnabled` would just abstract away the nil check, right?
I don't think this is more readable or shorter than the if statement, is it?

src/k8s/api/v1/types.go
bschimke95: Hm, I think it may be slightly more readable but not much.

Also, I would put such function into the `utils` package but we cannot do this as this would cause cyclic dependencies.

A bit of repetitiveness is the price we pay with the explicit configuration, I guess

src/k8s/api/v1/types.go
eaudetcobello: That is true

src/k8s/api/v1/types.go
```diff
@@ -118,20 +119,74 @@ func (c ClusterStatus) String() string {
...
+		case DatastoreRoleVoter:
+			voters = append(voters, node)
+		case DatastoreRoleStandBy:
+			standBys = append(standBys, node)
+		case DatastoreRoleSpare:
+			spares = append(spares, node)
+		}
+	}
+	if len(voters) > 0 {
+		result.WriteString(fmt.Sprintf("  master-nodes:\n"))
```
ktsakalozos, line 139: Do not use the word "master", please use "voter"

bschimke95:
Good point - removed `omitempty` for configuration values and only left it for the functionalities (that way I can set a functionality to nil and it is not rendered by the k8s status command) 

Output now:
```
ubuntu@bob:~/k8s$ sudo k8s status
status: ready
high-availability: no
datastore:
  voter-nodes:
    - 10.84.220.25:6400
  standy-nodes: none
  spare-nodes: none

dns:
  enabled: true
  cluster-domain: cluster.local
  service-ip: 10.152.183.233
  upstream-nameservers:
  - /etc/resolv.conf
ingress:
  enabled: true
  default-tls-secret: ""
  enable-proxy-protocol: false
metrics-server:
  enabled: true
```

berkayoz (approved):
LGTM

src/k8s/pkg/utils/database.go
```diff
@@ -28,6 +30,99 @@ func GetClusterConfig(ctx context.Context, state *state.State) (types.ClusterCon
 	return clusterConfig, nil
 }
 
+// GetUserFacingClusterConfig returns the public cluster config.
+func GetUserFacingClusterConfig(ctx context.Context, state *state.State) (apiv1.UserFacingClusterConfig, error) {
```
neoaggelos, line 34: An important note for this. This is not `utils`: `utils` are supposed to be stand-alone helpers.

I see why this is here, same as the others in the same file. But we have some them in `api/impl` as well. The problem is that this does not allow importing the `pkg/utils` package from `pkg/k8sd/database` code, which is not ideal.

It's OK to deal with this at a later time.

neoaggelos (approved):

louiseschmidtgen (approved):
LGTM

PR #186, opened 2024-03-01
Refactor ParseArgumentLine to support Kubernetes label values

eaudetcobello:
With this PR, the following key-value in Kubelet argument file is now supported by the ParseArgumentLine function:

`--node-labels=node-role.kubernetes.io/control-plane=,node-role.kubernetes.io/worker=`

This is needed for KU-447 specifically for unit testing setup.kubelet.

### DIFF
--- a/src/k8s/pkg/utils/file.go
+++ b/src/k8s/pkg/utils/file.go
@@ -18,6 +18,8 @@ import (
 func ParseArgumentLine(line string) (key string, value string) {
 	line = strings.TrimSpace(line) // Trim leading and trailing white spaces
 
+	// parse "--argument value", "--argument=value", "--argument=value=,othervalue=" variants
+
 	// Find the index of the first occurrence of "=" and space
 	indexEqual := strings.Index(line, "=")
 	indexSpace := strings.Index(line, " ")
--- a/src/k8s/pkg/utils/file.go
+++ b/src/k8s/pkg/utils/file.go
@@ -16,20 +16,32 @@ import (
 // ParseArgumentLine parses a command-line argument from a single line.
 // The returned key includes any dash prefixes.
 func ParseArgumentLine(line string) (key string, value string) {
-	line = strings.TrimSpace(line)
+	line = strings.TrimSpace(line) // Trim leading and trailing white spaces
 
-	// parse "--argument value" and "--argument=value" variants
-	if parts := strings.Split(line, "="); len(parts) >= 2 {
-		key = parts[0]
-		value = parts[1]
-	} else if parts := strings.Split(line, " "); len(parts) >= 2 {
-		key = parts[0]
-		value = strings.Join(parts[1:], " ")
-	} else {
-		key = line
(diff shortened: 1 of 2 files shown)


src/k8s/pkg/utils/file.go
```diff
@@ -16,20 +16,32 @@ import (
 // ParseArgumentLine parses a command-line argument from a single line.
 // The returned key includes any dash prefixes.
 func ParseArgumentLine(line string) (key string, value string) {
-	line = strings.TrimSpace(line)
+	line = strings.TrimSpace(line) // Trim leading and trailing white spaces
 
-	// parse "--argument value" and "--argument=value" variants
```
addyess, line 21: @eaudetcobello i found this comment line SUPER important.  I was trying to figure out why we were looking for " " or "="

qq:  is `--argument         value` legal?

src/k8s/pkg/utils/file.go
```diff
@@ -16,20 +16,32 @@ import (
...
+	indexSpace := strings.Index(line, " ")
+
+	// Determine the split index based on the first occurrence of "=" or space
+	splitIndex := -1
+	if indexEqual == -1 && indexSpace != -1 { // Space but no "="
+		splitIndex = indexSpace
+	} else if indexSpace == -1 && indexEqual != -1 { // "=" but no space
+		splitIndex = indexEqual
+	} else if indexSpace != -1 && indexEqual != -1 { // Both "=" and space
+		splitIndex = min(indexEqual, indexSpace)
```
addyess, line 32: so this means if you encounter:

```--argument=value value value````

then you'd have "--argument", "value value value"?  Is this reasonable

Comment (eaudetcobello) [reply src/k8s/pkg/utils/file.go:32]: Yes, it's in the test cases.

File: src/k8s/pkg/utils/file.go (hunk 1 above)
Comment (eaudetcobello) [reply src/k8s/pkg/utils/file.go:21]: I would think so. There's two test cases for that:

```
		{line: "--key    =", key: "--key", value: "="},
		{line: "--key    = a value=", key: "--key", value: "= a value="},
```

File: src/k8s/pkg/utils/file.go (hunk 2 above)
Comment (eaudetcobello) [reply src/k8s/pkg/utils/file.go]: Simplified the logic quite a bit in the newest commit.

File: src/k8s/pkg/utils/file.go [hunk 3]
```diff
@@ -16,20 +16,28 @@ import (
...
 
-	return
+	if splitIndex == -1 {
+		// If no space or equal sign is found, return the line as key
+		return line, ""
+	}
+
+	// Split the line into key and value based on the split index
+	key = line[:splitIndex]
+	value = strings.TrimSpace(line[splitIndex+1:]) // Remove any leading space in value
```
bschimke95, line 38: super small nit:
decide on one: either comment above or next to the line (I'd prefer above)

bschimke95 (approved):
LGTM, great simplification!

PR #190, opened 2024-03-03
Add docs for reference commands

kwmonroe:
### Summary

Generate reference docs for `k8s $cmd` and add them to our `./docs/src/_parts`.

### Changes

- un-hide `k8s config` and `k8s status` commands
- new `gen_cmd_docs.go` that uses cobra to generate markdown based on our `./src/k8s/cmd/k8s` commands
- new `go.doc` make target that runs as part of the code quality gh workflow
  - if cmd docs are affected but not included in future PRs, this action will fail telling authors to run `make go.doc`
- include initial `./docs/src/_parts` content


evilnick:
Cool. I have a trick for these docs, let me see if I can make it work

evilnick:
cool, that works! Not crazy about some of the formatting, but I can probably strip some of it out.
Do these get regenerated with the build?

src/k8s/cmd/k8s/k8s.go
```diff
@@ -71,5 +71,6 @@ func NewRootCmd() *cobra.Command {
 	rootCmd.AddCommand(newHelmCmd())
 	rootCmd.AddCommand(newKubectlCmd())
 
+	rootCmd.DisableAutoGenTag = true
```
kwmonroe, line 74: without this, we get this at the bottom of our docs:

`generated by cobra`

which isn't bad, per se.. just not necessary in our reference docs.

src/k8s/Makefile
```diff
@@ -6,6 +6,10 @@ go.fmt:
 	go mod tidy
 	go fmt ./...
 
+go.docs: go.doc
+go.doc:
+	$(DQLITE_BUILD_SCRIPTS_DIR)/static-go-doc.sh gen_cmd_docs.go
+
```
kwmonroe, line 12: `go run gen_cmd_docs.go` needs `dqlite.h`

src/k8s/hack/gen_cmd_docs.go
```diff
@@ -0,0 +1,17 @@
+package main
```
kwmonroe, line 1: this impl of cobra needs to be `go run`, which needs a `main`, which would conflict with our `./src/k8s/cmd/main.go` -- hence this is in our `hack` util dir.

evilnick:
This gets a :cake: from me for how the docs turn out - not sure if anyone wants to review the code changes

src/k8s/hack/gen_cmd_docs.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func main() {
+	k8sCmd := k8s.NewRootCmd()
+
+	err := doc.GenMarkdownTree(k8sCmd, "../../../docs/src/_parts/commands")
+	if err != nil {
+		log.Fatal(err)
+	}
+}
```
neoaggelos, line 17: could we make this a hidden `k8s documentation` command? That way you can avoid the `-C ${DIR}` things below

kwmonroe: i can make this a `k8s $cmd` and generate docs at runtime, but i'm not sure how to call it from a Makefile / gh action that can determine if a PR-under-test is missing docs.

kwmonroe: aha!  call it like you would any other command, so long as it's built/linked in the go.doc target.  thanks @neoaggelos!

kwmonroe (commented):
LGTM with some clutch plays from the team.

PR #192, opened 2024-03-04
Request enabled should be treated as true if not set in request

berkayoz:
Request enabled should be treated as true if not set in request, otherwise the functionality won't be refreshed with new settings.


bschimke95 (approved):
LGTM

PR #197, opened 2024-03-04
Multipass install

evilnick:
Adds the install on Multipass page

### DIFF
--- a/docs/src/howto/install/multipass.md
+++ b/docs/src/howto/install/multipass.md
@@ -1,1 +1,42 @@
 # Install with Multipass (Ubuntu/Mac/Windows)
+
+**Multipass** (https://multipass.run/) is a simple way to run Ubuntu in a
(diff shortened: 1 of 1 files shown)


docs/src/howto/install/multipass.md
```diff
@@ -1 +1,121 @@
...
+
+```no-highlight
+multipass launch 22.04 --name k8s-node -m 4G -d 20G -c 2
+```
+
+This command specifies:
+
+ - **22.04**: The Ubuntu image used as the basis for the instance 
+ - **--name**: The name by which you will refer to the instance
+ - **-m**: The memory to allocate
```
neoaggelos, line 65: can we use the long option names instead?

evilnick: yes, i did think that myself. it makes the example command very long, but you are undoubtedly correct

PR #201, opened 2024-03-04
Add unit tests for k8sd/setup/kubelet

eaudetcobello:


src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,94 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// testFixture prepares and returns a test environment setup.
+func testFixture(t *testing.T) *mock.Snap {
+	g := NewWithT(t)
+	dir := t.TempDir()
+
+	// Ensure the mock CNI binary can be written.
+	require.NoError(t, os.WriteFile(path.Join(dir, "mockcni"), []byte("echo hi"), 0600))
```
neoaggelos, line 22: This mixes `gomega` and require. Please use `g.Expect(err).To(BeNil())`

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,94 @@
...
+			{"--eviction-hard", "'memory.available<100Mi,nodefs.available<1Gi,imagefs.available<1Gi'"},
+			{"--fail-swap-on", "false"},
+			{"--hostname-override", "dev"},
+			{"--kubeconfig", path.Join(kubernetesConfigDir, "kubelet.conf")},
+			{"--node-labels", expectedLabels},
+			{"--read-only-port", "0"},
+			{"--register-with-taints", ""},
+			{"--root-dir", kubeletRootDir},
+			{"--serialize-image-pulls", "false"},
+			{"--tls-cipher-suites", strings.Join(setup.GetKubeletTLSCipherSuites(), ",")},
```
neoaggelos, line 84: Note that this is not really testing anything useful. Either hardcode the list (which must be updated when we change the list of TLS cipher suites), ensure it is not empty, ensure it has a few ones we care about, or ignore completely.

I would recommend going with a hardcoded list here.

```suggestion
			{"--tls-cipher-suites", strings.Join(setup.GetKubeletTLSCipherSuites(), ",")},
```

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,94 @@
...
-	// TODO: setup a mock snap with paths, then verify setup.Kubelet().
-	// See TestContainerd() for inspiration.
+	g := NewWithT(t)
+
+	s := testFixture(t)
+
+	kubernetesPKIDir := s.KubernetesPKIDir()
+	containerdSocketDir := s.ContainerdSocketDir()
+	kubernetesConfigDir := s.KubernetesConfigDir()
+	kubeletRootDir := s.KubeletRootDir()
```
neoaggelos, line 59: this works, but maybe has unnecessary setup work. Consider creating a mock snap that only sets the 4 particular dirs we care about here.

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,94 @@
...
+
+	g.Expect(setup.KubeletControlPlane(s, "dev", nil, "10.152.1.1", "test-cluster.local", "")).To(BeNil())
+
+	t.Run("Args", func(t *testing.T) {
+		expectedLabels := strings.Join(append(setup.GetKubeletControlPlaneLabels(), setup.GetKubeletWorkerLabels()...), ",")
```
neoaggelos, line 64: same here, prefer hardcoding the labels instead. if the implementation changes, the tests will need to be adapted as well. if the implementation drops them by accident, then the tests will serve as a backup

neoaggelos (commented):
Going in the right direction, some comments

src/k8s/pkg/k8sd/setup/kubelet.go
```diff
@@ -52,7 +52,7 @@ func kubelet(snap snap.Snap, hostname string, nodeIP net.IP, clusterDNS string,
 		"--fail-swap-on":                 "false",
 		"--hostname-override":            hostname,
 		"--kubeconfig":                   path.Join(snap.KubernetesConfigDir(), "kubelet.conf"),
-		"--node-labels":                  strings.Join(labels, ","),
+		"--node-labels":                  strings.Join(append(kubeletControlPlaneLabels, kubeletWorkerLabels...), ","),
 		"--read-only-port":               "0",
```
neoaggelos, line 56: this does not look right

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,149 @@
...
+			{"--node-labels", expectedLabels},
+			{"--read-only-port", "0"},
+			{"--register-with-taints", ""},
+			{"--root-dir", s.Mock.KubeletRootDir},
+			{"--serialize-image-pulls", "false"},
+			{"--tls-cipher-suites", strings.Join(kubeletTLSCipherSuites, ",")},
+			{"--cloud-provider", "provider"},
+			{"--cluster-dns", "10.152.1.1"},
+			{"--cluster-domain", "test-cluster.local"},
+			{"--node-ip", net.IP("192.168.0.1").String()},
```
neoaggelos, line 89: prefer static over auto-generation
```suggestion
			{"--node-ip", "192.168.0.1"},
```

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,149 @@
...
+			{"--eviction-hard", "'memory.available<100Mi,nodefs.available<1Gi,imagefs.available<1Gi'"},
+			{"--fail-swap-on", "false"},
+			{"--hostname-override", "dev"},
+			{"--kubeconfig", path.Join(s.Mock.KubernetesConfigDir, "kubelet.conf")},
+			{"--node-labels", expectedLabels},
+			{"--read-only-port", "0"},
+			{"--register-with-taints", ""},
+			{"--root-dir", s.Mock.KubeletRootDir},
+			{"--serialize-image-pulls", "false"},
+			{"--tls-cipher-suites", strings.Join(kubeletTLSCipherSuites, ",")},
```
neoaggelos, line 85: prefer static here

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,149 @@
...
+
+		// Call the kubelet control plane setup function
+		g.Expect(setup.KubeletControlPlane(s, "dev", net.IP("192.168.0.1"), "10.152.1.1", "test-cluster.local", "provider")).To(BeNil())
+
+		// Ensure the kubelet arguments file has the expected arguments and values
+		tests := []struct {
+			key         string
+			expectedVal string
+		}{
+			{"--anonymous-auth", "false"},
```
neoaggelos, line 70: use named fields
```suggestion
			{key: "--anonymous-auth", expectedVal: "false"},
```

neoaggelos (commented):
Stylistic comments, but please deal with them. Further, I am missing unit tests for worker node kubelets

src/k8s/pkg/k8sd/setup/kubelet.go
neoaggelos: the tests for worker nodes should have caught this

eaudetcobello: Wow good catch

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,238 @@
...
+		g.Expect(err).To(BeNil())
+		g.Expect(len(args)).To(Equal(len(tests)))
+	})
+
+	t.Run("Setup control plane with all optional arguments missing", func(t *testing.T) {
+		g := NewWithT(t)
+
+		// Create a mock snap
+		s, dir := testFixture(t)
+		defer os.RemoveAll(dir)
```
bschimke95, line 99: The test framework will take care of cleaning this up.
From the `t.TempDir` documenation:
```
// TempDir returns a temporary directory for the test to use.
// The directory is automatically removed by Cleanup when the test and
// all its subtests complete.
```

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,238 @@
...
+		args, err := utils.ParseArgumentFile(path.Join(s.Mock.ServiceArgumentsDir, "kubelet"))
+		g.Expect(err).To(BeNil())
+		g.Expect(len(args)).To(Equal(len(tests)))
+	})
+
+	t.Run("Setup worker with all possible arguments", func(t *testing.T) {
+		g := NewWithT(t)
+
+		// Create a mock snap
+		s, dir := testFixture(t)
```
bschimke95, line 143: Let's make this explicit and follow Go conventions:

`mustSetupSnapAndDirectory(t)`

bschimke95 (commented):
Looks good - just some nits

src/k8s/pkg/k8sd/setup/kubelet_test.go
eaudetcobello: Good to know !

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,236 @@
...
+
+		// Ensure the kubelet arguments file has exactly the expected number of arguments
+		args, err := utils.ParseArgumentFile(path.Join(s.Mock.ServiceArgumentsDir, "kubelet"))
+		g.Expect(err).To(BeNil())
+		g.Expect(len(args)).To(Equal(len(tests)))
+	})
+
+	t.Run("Setup with missing service arguments file should fail", func(t *testing.T) {
+
+	})
```
neoaggelos, line 235: This is not testing anything?

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,236 @@
...
+			{key: "--serialize-image-pulls", expectedVal: "false"},
+			{key: "--tls-cipher-suites", expectedVal: kubeletTLSCipherSuites},
+			{key: "--cloud-provider", expectedVal: "provider"},
+			{key: "--cluster-dns", expectedVal: "10.152.1.1"},
+			{key: "--cluster-domain", expectedVal: "test-cluster.local"},
+			{key: "--node-ip", expectedVal: "192.168.0.1"},
+		}
+		for _, tc := range tests {
+			t.Run(tc.key, func(t *testing.T) {
+				val, err := snaputil.GetServiceArgument(s, "kubelet", tc.key)
```
neoaggelos, line 175: Sub-tests must get their own `gomega` instance
```suggestion
			t.Run(tc.key, func(t *testing.T) {
				g := NewWithT(t)
				val, err := snaputil.GetServiceArgument(s, "kubelet", tc.key)
```

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,236 @@
...
+			{key: "--kubeconfig", expectedVal: path.Join(s.Mock.KubernetesConfigDir, "kubelet.conf")},
+			{key: "--node-labels", expectedVal: expectedWorkerLabels},
+			{key: "--read-only-port", expectedVal: "0"},
+			{key: "--register-with-taints", expectedVal: ""},
+			{key: "--root-dir", expectedVal: s.Mock.KubeletRootDir},
+			{key: "--serialize-image-pulls", expectedVal: "false"},
+			{key: "--tls-cipher-suites", expectedVal: kubeletTLSCipherSuites},
+		}
+		for _, tc := range tests {
+			t.Run(tc.key, func(t *testing.T) {
```
neoaggelos, line 220: same here

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,236 @@
...
+		g.Expect(err).To(BeNil())
+		g.Expect(len(args)).To(Equal(len(tests)))
+	})
+
+	t.Run("Setup control plane with all optional arguments missing", func(t *testing.T) {
+		g := NewWithT(t)
+
+		// Create a mock snap
+		s, dir := mustSetupSnapAndDirectories(t, mustReturnMockForKubelet)
+		defer os.RemoveAll(dir)
```
neoaggelos, line 97: `t.TempDir()` already cleans up, the defer here is not needed. Also, no need to return the `dir` either.

src/k8s/pkg/k8sd/setup/kubelet.go
```diff
@@ -74,5 +74,6 @@ func kubelet(snap snap.Snap, hostname string, nodeIP net.IP, clusterDNS string,
 	if _, err := snaputil.UpdateServiceArguments(snap, "kubelet", args, nil); err != nil {
 		return fmt.Errorf("failed to render arguments file: %w", err)
 	}
+
```
neoaggelos, line 77: revert

```suggestion
```

neoaggelos (changes requested):
Needs some more work

bschimke95 (commented):
Two more things and we are good to go

src/k8s/pkg/k8sd/setup/kubelet_test.go
```diff
@@ -1,8 +1,245 @@
...
+		g := NewWithT(t)
+		s := mustSetupSnapAndDirectories(t, setKubeletMock)
+
+		s.Mock.ServiceArgumentsDir = "nonexistent"
+
+		g.Expect(setup.KubeletControlPlane(s, "", nil, "", "", "")).ToNot(Succeed())
+		g.Expect(setup.KubeletWorker(s, "", nil, "", "", "")).ToNot(Succeed())
+
+		g.Expect(setup.KubeletControlPlane(s, "dev", net.ParseIP("192.168.0.1"), "10.152.1.1", "test-cluster.local", "provider")).ToNot(Succeed())
+		g.Expect(setup.KubeletWorker(s, "dev", net.ParseIP("192.168.0.1"), "10.152.1.1", "test-cluster.local", "provider")).ToNot(Succeed())
```
neoaggelos, line 243: What are we testing here? Test one invocation, or make separate tests

eaudetcobello: Testing that invoking KubeletControlPlane/KubeletWorker with different arguments always results in an error if the required directory does not exist.

neoaggelos: Let's make them separate tests then. Also, testing with empty values is not very useful (the implementation could be failing with an empty hostname, so your test does not test what you think it does).

neoaggelos (approved):

PR #202, opened 2024-03-04
Get or create join tocken on get-join-token

louiseschmidtgen:
## Description
Functionality to return the existing join-tocken when the `get-join-token` command is re-run.

### DIFF
--- a/src/k8s/pkg/k8sd/api/cluster_tokens.go
+++ b/src/k8s/pkg/k8sd/api/cluster_tokens.go
@@ -28,13 +28,33 @@ func postClusterJoinTokens(m *microcluster.MicroCluster, s *state.State, r *http
 	if req.Worker {
 		token, err = createWorkerToken(s)
 	} else {
-		token, err = m.NewJoinToken(req.Name)
+		token, err = getOrCreateJoinTocken(m, req.Name)
 	}
 	if err != nil {
 		return response.InternalError(fmt.Errorf("failed to create token: %w", err))
 	}
 
 	return response.SyncResponse(true, &apiv1.TokensResponse{EncodedToken: token})
+}
+
+func getOrCreateJoinTocken(m *microcluster.MicroCluster, tokenName string) (string, error) {
+	// grab token if it exists and return it
+	records, err := m.ListJoinTokens()
+	if err != nil {
+		return "", err
+	}
+	for _, record := range records {
+		if record.Name == tokenName {
+			return record.Token, nil
+		}
+	}
+
+	// if token does not exist, create a new one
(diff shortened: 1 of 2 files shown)


src/k8s/pkg/k8sd/api/cluster_tokens.go
```diff
@@ -40,6 +40,26 @@ func postClusterJoinTokens(m *microcluster.MicroCluster, s *state.State, r *http
 	return response.SyncResponse(true, &apiv1.TokensResponse{EncodedToken: token})
 }
 
+func getOrCreateJoinTocken(m *microcluster.MicroCluster, tokenName string) (string, error) {
+	// grab token if it exists and return it
+	records, err := m.ListJoinTokens()
+	if err != nil {
+		return "", err
```
neoaggelos, line 47: if we can't list the existing tokens, might as well try to create a new one instead of returning the error.

neoaggelos: for example, you can log the error here and do `m.NewJoinToken` instead. If I am creating a token, I don't expect to get a "failed to list tokens" error instead.

src/k8s/pkg/k8sd/api/cluster_tokens.go
```diff
@@ -40,6 +40,26 @@ func postClusterJoinTokens(m *microcluster.MicroCluster, s *state.State, r *http
...
+	for _, record := range records {
+		if record.Name == tokenName {
+			return record.Token, nil
+		}
+	}
+
+	// if token does not exist, create a new one
+	token, err := m.NewJoinToken(tokenName)
+	if err != nil {
+		return "", err
```
neoaggelos, line 58: wrap the error and add relevant context, e.g.
```suggestion
		return "", fmt.Errorf("failed to generate a new microcluster join token: %w", err)
```

src/k8s/pkg/k8sd/api/cluster_tokens.go
```diff
@@ -31,7 +31,7 @@ func postClusterJoinTokens(m *microcluster.MicroCluster, s *state.State, r *http
 	if req.Worker {
 		token, err = createWorkerToken(s)
 	} else {
-		token, err = m.NewJoinToken(hostname)
+		token, err = getOrCreateJoinTocken(m, hostname)
```
neoaggelos, line 34:
```suggestion
		token, err = getOrCreateJoinToken(m, hostname)
```

neoaggelos (commented):
Nice work, a couple of comments, mostly on style

PR #206, opened 2024-03-05
Unit test for ComponentManager.Disable

eaudetcobello:
And a small refactor in TestComponentInitialState and TestEnableMultipleComponents so they use isComponentEnabled instead of List which has already been tested, this increases coverage.


src/k8s/pkg/component/component_test.go
```diff
@@ -266,10 +274,76 @@ func TestEnableMultipleComponents(t *testing.T) {
...
-	components, err := mockHelmClient.List()
-	g.Expect(err).ShouldNot(HaveOccurred())
-	g.Expect(components).To(ConsistOf(
-		Component{Name: "one", Status: true},
-		Component{Name: "two", Status: true},
-	), "Expected all components to be enabled")
+	// Reenable should not error
+	for name := range mockHelmClient.components {
+		err := mockHelmClient.Enable(name, map[string]interface{}{})
+		g.Expect(err).ShouldNot(HaveOccurred())
```
bschimke95, line 280: Looked up what is best practice here as I used to do:
`g.Expect(err).To(BeNil)`

Seems like this is the recommended way (btw, prefer `any` over `interface{}`):
```
g.Expect(mockHelmClient.Enable(name, map[string]any{})).Should(Succeed())
```
https://github.com/onsi/gomega/blob/master/matchers.go#L98

src/k8s/pkg/component/component_test.go
```diff
@@ -266,10 +274,76 @@ func TestEnableMultipleComponents(t *testing.T) {
...
 	}
 
-	components, err := mockHelmClient.List()
-	g.Expect(err).ShouldNot(HaveOccurred())
-	g.Expect(components).To(ConsistOf(
-		Component{Name: "one", Status: true},
-		Component{Name: "two", Status: true},
-	), "Expected all components to be enabled")
+	// Reenable should not error
+	for name := range mockHelmClient.components {
```
bschimke95, line 278: nit: you could put those steps into sub-tests:

```
t.Run("Reenable", func(...) {...})
```
That way, it is easier to see what went wrong.

bschimke95 (commented):
Looks great - just two small nits

src/k8s/pkg/component/component_test.go
```diff
@@ -266,10 +274,78 @@ func TestEnableMultipleComponents(t *testing.T) {
...
+		{"whiskas-2", "default"},
+	}
+	for _, tc := range releases {
+		t.Run(tc.release, func(t *testing.T) {
+			g.Expect(mockHelmClient.isComponentEnabled(tc.release, tc.namespace)).To(BeTrue(), "Expected all components to enabled")
+		})
+	}
+
+	// Component does not exist at all
+	err := mockHelmClient.Enable("non-existent", map[string]interface{}{})
```
bschimke95, line 298: Please use the subtests also for the other tests here and below

src/k8s/pkg/component/component_test.go
```diff
@@ -266,10 +271,83 @@ func TestEnableMultipleComponents(t *testing.T) {
...
+	}
+
+	// Component does not exist at all
+	t.Run("EnableNonExistent", func(t *testing.T) {
+		g.Expect(mockHelmClient.Enable("non-existent", map[string]any{})).ShouldNot(Succeed())
+	})
+}
+
+func TestDisableComponent(t *testing.T) {
+	g := NewWithT(t)
```
louiseschmidtgen, line 302: Can we turn this into a setup piece that we run before each test? It would be nice to have a fresh test set-up for each test, this way the tests are independent of each other (not crucial here but a good practice).

src/k8s/pkg/component/component_test.go
```diff
@@ -266,10 +271,83 @@ func TestEnableMultipleComponents(t *testing.T) {
...
+
+	mockHelmClient, tempDir, _ := mustCreateNewHelmClient(t, map[string]types.Component{
```
louiseschmidtgen, line 304: This piece of code exists in multiple places. We can try and avoid code duplication by writing a helper function.

louiseschmidtgen (commented):
Nice work @eaudetcobello! I've got some small suggestions, mainly about making the test set-up a bit nicer.

src/k8s/pkg/component/component_test.go
eaudetcobello: Thanks. Took the opportunity to make other improvements as well.

src/k8s/pkg/component/component_test.go
```diff
@@ -102,6 +103,26 @@ func mustMakeMeSomeReleases(store *storage.Storage, t *testing.T) (all []*releas
 	return all
 }
 
+func mustMakeMeSomeComponents() map[string]types.Component {
```
bschimke95, line 106: a minor comment: the `must` convention in Go indicates that you perform some setup steps that could fail and then panic.
This function should can just be a variable declaration. You don't perform any setup steps.

bschimke95 (approved):
LGTM

PR #215, opened 2024-03-07
Add unit tests for k8sd/setup/k8s_apiserver_proxy

eaudetcobello:
Needs https://github.com/canonical/k8s-snap/pull/201


src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy_test.go
```diff
@@ -1,8 +1,70 @@
...
+
+func TestK8sApiServerProxy(t *testing.T) {
+	t.Run("Args", func(t *testing.T) {
+		g := NewWithT(t)
+
+		s := mustSetupSnapAndDirectories(t, setK8sApiServerMock)
+
+		g.Expect(setup.K8sAPIServerProxy(s, []string{}))
+
+		os.Create(path.Join(s.Mock.ServiceExtraConfigDir, "k8s-apiserver-proxy.json"))
```
neoaggelos, line 30: why is this?

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy_test.go
```diff
@@ -1,8 +1,70 @@
...
+	}
+}
+
+func TestK8sApiServerProxy(t *testing.T) {
+	t.Run("Args", func(t *testing.T) {
+		g := NewWithT(t)
+
+		s := mustSetupSnapAndDirectories(t, setK8sApiServerMock)
+
+		g.Expect(setup.K8sAPIServerProxy(s, []string{}))
```
neoaggelos, line 28: Use `nil` instead of `[]string{}`

neoaggelos (commented):
We should also test that the `k8s-apiserver-proxy.json` file has the right list of apiservers

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy_test.go
```diff
@@ -1,8 +1,91 @@
...
```
neoaggelos, line 30: This is not right. `g.Expect()` does not validate anything. You need to to add `.To(...)`

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy_test.go
neoaggelos: this is not addressed

neoaggelos (commented):
Needs a few final fixes

PR #218, opened 2024-03-08
Register control-plane Nodes with the same taints

addyess:
Provides some of the infrastructure to register control plane nodes with a set of taints.


src/k8s/api/v1/types.go
```diff
@@ -23,6 +25,7 @@ type BootstrapConfig struct {
 // SetDefaults sets the fields to default values.
 func (b *BootstrapConfig) SetDefaults() {
 	b.Components = []string{"dns", "metrics-server", "network"}
+	b.Taints = []string{}
```
neoaggelos, line 28: It's okay to remove this line, `b.Taints` will just be nil.

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -126,6 +126,8 @@ func getConfigInteractively(stdin io.Reader, stdout io.Writer, stderr io.Writer)
 	config.ServiceCIDR = askQuestion(stdin, stdout, stderr, "Please set the Service CIDR:", nil, config.ServiceCIDR, nil)
 	rbac := askBool(stdin, stdout, stderr, "Enable Role Based Access Control (RBAC)?", []string{"yes", "no"}, "yes")
 	*config.EnableRBAC = rbac
+	taints := askQuestion(stdin, stdout, stderr, "Any taints on the control nodes?", nil, "", nil)
+	config.Taints = strings.Split(taints, ",")
```
neoaggelos, line 130: I think it's okay to not set this during interactive bootstrap. If someone knows that they want particular taints to be set, then they should configure them in the config file instead.

src/k8s/api/v1/types.go
```diff
@@ -11,6 +11,8 @@ import (
 type BootstrapConfig struct {
 	// Components are the components that should be enabled on bootstrap.
 	Components []string `yaml:"components"`
+	// Taints on the control-plane during bootstrap
+	Taints []string `yaml:"taints"`
```
neoaggelos, line 15: Should they be called `ControlPlaneTaints`?

src/k8s/pkg/k8sd/types/cluster_config.go
```diff
@@ -43,9 +43,10 @@ type Certificates struct {
 }
 
 type Kubelet struct {
-	CloudProvider string `yaml:"cloud-provider,omitempty"`
-	ClusterDNS    string `yaml:"cluster-dns,omitempty"`
-	ClusterDomain string `yaml:"cluster-domain,omitempty"`
+	CloudProvider      string   `yaml:"cloud-provider,omitempty"`
+	ClusterDNS         string   `yaml:"cluster-dns,omitempty"`
+	ClusterDomain      string   `yaml:"cluster-domain,omitempty"`
+	RegisterWithTaints []string `yaml:"register-with-taints,omitempty"`
```
neoaggelos, line 49: Should the name state that these are only for control-plane nodes? e.g. `ControlPlaneTaints`?

neoaggelos (commented):
Nice work, a few minor comments.

src/k8s/api/v1/types.go
```diff
@@ -28,6 +30,7 @@ type BootstrapConfig struct {
 // SetDefaults sets the fields to default values.
 func (b *BootstrapConfig) SetDefaults() {
 	b.Components = []string{"dns", "metrics-server", "network", "gateway"}
+	b.ControlPlaneTaints = nil
```
bschimke95, line 33: not required. this is the default value already

src/k8s/cmd/k8s/k8s_bootstrap_test.go
```diff
@@ -48,12 +50,13 @@ cluster-cidr: "10.244.0.0/16"
...
 bananas: 5`,
 		expectedConfig: apiv1.BootstrapConfig{
-			Components:    []string{"dns", "metrics-server", "network", "gateway"},
-			ClusterCIDR:   "10.244.0.0/16",
-			ServiceCIDR:   "10.152.183.0/24",
-			EnableRBAC:    vals.Pointer(true),
-			K8sDqlitePort: 9000,
-			Datastore:     "k8s-dqlite",
+			Components:         []string{"dns", "metrics-server", "network", "gateway"},
+			ControlPlaneTaints: nil,
```
bschimke95, line 54: same here

bschimke95 (commented):
looks good, only small nits.
Please resolve the conflicts and then we should be good to go

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -188,7 +188,8 @@ func getConfigInteractively(stdin io.Reader, stdout io.Writer, stderr io.Writer)
 	config.ServiceCIDR = vals.Pointer(serviceCIDR)
 
 	// TODO: any other configs we care about in the interactive bootstrap?
-
+	taints := askQuestion(stdin, stdout, stderr, "Any taints on the control nodes?", nil, "", nil)
+	config.ControlPlaneTaints = strings.FieldsFunc(taints, func(c rune) bool { return c == ',' })
 	return config
```
neoaggelos, line 193: does it make sense to ask such a question when bootstrapping interactively?

src/k8s/pkg/k8sd/types/cluster_config_kubelet.go
```diff
@@ -46,6 +55,10 @@ func KubeletFromConfigMap(m map[string]string) (Kubelet, error) {
 	if v, ok := m["cluster-domain"]; ok {
 		c.ClusterDomain = &v
 	}
+	if v, ok := m["control-plane-taints"]; ok {
+		s := strings.FieldsFunc(v, func(c rune) bool { return c == ',' })
+		c.ControlPlaneTaints = &s
+	}
```
neoaggelos, line 61: Do we ever need to update this flag in running kubelets? Probably not, and this would also find its way to worker nodes as well.

I'd recommend not exposing this in the configmap at all.

src/k8s/pkg/k8sd/types/cluster_config_kubelet_test.go
```diff
@@ -79,6 +99,7 @@ func TestKubelet(t *testing.T) {
 
 				k, err := types.KubeletFromConfigMap(tc.configmap)
 				g.Expect(err).To(BeNil())
+				fmt.Printf("k=%v\n", k)
```
neoaggelos, line 102: this is for debugging right?

src/k8s/cmd/k8s/k8s_bootstrap.go
addyess: if not, i'll just drop it

bschimke95 (approved):
LGTM

PR #221, opened 2024-03-11
Machinery to configure containerd registry auths and mirrors

neoaggelos:
### Summary

Add machinery to configure containerd registries and mirrors


bschimke95 (approved):
LGTM, only two small nits

berkayoz (approved):
LGTM

PR #224, opened 2024-03-11
Fix should check returned error before deferring lock.Close()

eaudetcobello:


src/k8s/pkg/snap/util/node.go
```diff
@@ -18,10 +18,10 @@ func MarkAsWorkerNode(snap snap.Snap, mark bool) error {
 
 	if mark {
 		lock, err := os.Create(fname)
-		defer lock.Close()
 		if err != nil {
 			return fmt.Errorf("failed to mark node as worker: %w", err)
 		}
+		defer lock.Close()
```
neoaggelos, line 24: As a note, we don't _have_ to defer this, you can just call the lock.Close() directly.

neoaggelos (approved):

src/k8s/pkg/snap/util/node.go
eaudetcobello: Done

PR #231, opened 2024-03-13
tidy usage of snap in k8sd client code

neoaggelos:
### Summary

Minor PR to clean things up about how we use the snap interface from the k8sd client.


src/k8s/pkg/k8s/client/client.go
```diff
@@ -37,7 +35,7 @@ type k8sdClient struct {
 func NewClient(ctx context.Context, opts ClusterOpts) (*k8sdClient, error) {
 	// TODO: pass snap through opts instead, do not create here.
```
bschimke95, line 36: seems outdated. The snap is optional now.

bschimke95 (approved):
LGTM, only a minor nit

PR #235, opened 2024-03-13
Use unique worker tokens

bschimke95:
Create a unique token for each worker. Previously, all workers shared the same token.

This increases security and allows workers to authenticate on endpoints on a per-node basis.

### DIFF
--- a/src/k8s/pkg/k8sd/api/worker_access_handler.go
+++ b/src/k8s/pkg/k8sd/api/worker_access_handler.go
@@ -1,11 +1,15 @@
 package api
 
 import (
+	"context"
+	"database/sql"
 	"fmt"
 	"net/http"
 
+	"github.com/canonical/k8s/pkg/k8sd/database"
 	"github.com/canonical/k8s/pkg/snap"
 	snaputil "github.com/canonical/k8s/pkg/snap/util"
+	"github.com/canonical/k8s/pkg/utils"
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/microcluster/state"
 )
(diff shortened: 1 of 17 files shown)


src/k8s/pkg/k8sd/types/worker.go
```diff
@@ -8,44 +8,32 @@ import (
...
-const internalWorkerNodeTokenSerializeMagicString = "m!!"
-
 var encoding = base64.RawStdEncoding
 
-type serializableWorkerNodeInfo struct {
-	InternalWorkerNodeToken
-	Magic string `json:"_"`
-}
-
 // Encode a worker node token to a base64-encoded string.
```
neoaggelos, line 20: Let's keep this in. If `microcluster` ever changes their tokens to have both `secret` and `join_addresses`, then our check in https://github.com/canonical/k8s-snap/pull/235/files#diff-05ae23d2944a3f9c66ecf2daeccede5edd805e16c27117a47fd45cdd643a8ab4R35 will fail.

src/k8s/pkg/k8sd/database/worker.go
```diff
@@ -24,27 +24,28 @@ var (
...
-func CheckWorkerNodeToken(ctx context.Context, tx *sql.Tx, token string) (bool, error) {
+// CheckWorkerNodeToken returns true if the specified token can be used to join the specified node on the cluster.
+func CheckWorkerNodeToken(ctx context.Context, tx *sql.Tx, nodeName string, token string) (bool, error) {
 	selectTxStmt, err := cluster.Stmt(tx, workerStmts["select-token"])
 	if err != nil {
 		return false, fmt.Errorf("failed to prepare select statement: %w", err)
 	}
 	var realToken string
-	if selectTxStmt.QueryRowContext(ctx).Scan(&realToken) == nil {
+	if selectTxStmt.QueryRowContext(ctx, nodeName).Scan(&realToken) == nil {
```
neoaggelos, line 34: `fmt.Sprintf("worker-token::%s", nodeName)` and revert the query above.

neoaggelos: also in other places

src/k8s/pkg/k8sd/api/worker_access_handler.go
```diff
@@ -24,3 +28,38 @@ func RestrictWorkers(s *state.State, r *http.Request) response.Response {
 
 	return response.EmptySyncResponse
 }
+
+// TokenAuthentication access handler checks if the worker is allowed to access this endpoint with the provided token.
+// TokenAuthentication expects the "worker-name" and "worker-secret" fields to be set in the request header.
+func TokenAuthentication(s *state.State, r *http.Request) response.Response {
```
neoaggelos, line 34: Let's give it a name that describes the usage without having to lookup the implementation. Also, make this a closure that accepts the header names and returns the handler

```suggestion
func ValidateWorkerInfoAccessHandler(nodeHeaderName string, tokenHeaderName string) func(s *state.State, r *http.Request) response.Response {
```

then used as `AccessHandler: ValidateWorkerInfoAccessHandler("worker-name", "worker-token")`

neoaggelos (commented):
Nice work, did a first pass

neoaggelos (approved):
LGTM

PR #237, opened 2024-03-13
Support single-node deployments

neoaggelos:
### Summary

Small improvements to make the single-node cluster for development UX better.

### Changes

- Configurable `hostname` and `address` in the `k8s bootstrap` command.
- Always include machine IP addresses as SANs to the generated self-signed certificates.
- Do not set the kubelet `--node-ip` if a loopback address is used.

### Testing

You can bootstrap a cluster using the loopback address and get a functional cluster

```bash
k8s bootstrap --address 127.0.0.1:6400
```

### DIFF
--- a/src/k8s/pkg/k8s/client/cluster.go
+++ b/src/k8s/pkg/k8s/client/cluster.go
@@ -7,8 +7,6 @@ import (
 	"time"
 
 	apiv1 "github.com/canonical/k8s/api/v1"
-	"github.com/canonical/k8s/pkg/config"
-	"github.com/canonical/k8s/pkg/utils"
 	"github.com/canonical/k8s/pkg/utils/control"
 	"github.com/canonical/k8s/pkg/utils/k8s"
 	"github.com/canonical/lxd/lxd/util"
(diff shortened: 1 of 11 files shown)


src/k8s/pkg/k8s/client/cluster.go
```diff
@@ -22,21 +20,7 @@ func (c *k8sdClient) IsBootstrapped(ctx context.Context) bool {
...
-	// TODO: this should be done on the server side, but we cannot currently hijack the microcluster bootstrap endpoint.
-	hostname, err := utils.CleanHostname(rawHostname)
-	if err != nil {
-		return apiv1.NodeStatus{}, fmt.Errorf("invalid hostname %q: %w", rawHostname, err)
-	}
-
-	// Get system addrPort.
-	addrPort := util.CanonicalNetworkAddress(util.NetworkInterfaceAddress(), config.DefaultPort)
-
+func (c *k8sdClient) Bootstrap(ctx context.Context, hostname string, address string, bootstrapConfig apiv1.BootstrapConfig) (apiv1.NodeStatus, error) {
```
bschimke95, line 23: Maybe just add those new parameters to the `BootstrapConfig`?

bschimke95 (approved):
LGTM

src/k8s/pkg/k8s/client/cluster.go
neoaggelos: hmm, the `BootstrapConfig` is passed verbatim to the bootstrap hook, whereas `name` and `address` here are explicitly needed by microcluster.

It would look someting like:

```
NewCluster(BootstrapConfig.Name, BootstrapConfig.Address, BootstrapConfig)
```

which is not very clean

bschimke95: fair - let's keep it.

PR #240, opened 2024-03-13
k8s status fixes

eaudetcobello:
Moved status checks to server side, refactored getting nodes and checking cluster status to reduce code duplication

### DIFF
--- a/src/k8s/cmd/k8s/k8s_status.go
+++ b/src/k8s/cmd/k8s/k8s_status.go
@@ -27,16 +27,6 @@ func newStatusCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
 			}
 
-			// TODO(neoaggelos): this must be done on the server side
-			// fail fast if we're not explicitly waiting and we can't get kube-apiserver endpoints
-			if !opts.waitReady {
-				if !client.IsKubernetesAPIServerReady(cmd.Context()) {
-					cmd.PrintErrln("Error: There are no active kube-apiserver endpoints, cluster status is unavailable")
-					env.Exit(1)
-					return
-				}
-			}
-
 			status, err := client.ClusterStatus(cmd.Context(), opts.waitReady)
 			if err != nil {
 				cmd.PrintErrf("Error: Failed to retrieve the cluster status.\n\nThe error was: %v\n", err)
--- a/src/k8s/pkg/k8s/client/cluster.go
+++ b/src/k8s/pkg/k8s/client/cluster.go
@@ -8,7 +8,6 @@ import (
 
 	apiv1 "github.com/canonical/k8s/api/v1"
 	"github.com/canonical/k8s/pkg/utils/control"
-	"github.com/canonical/k8s/pkg/utils/k8s"
 	"github.com/canonical/lxd/shared/api"
 )
 
@@ -65,17 +64,4 @@ func (c *k8sdClient) KubeConfig(ctx context.Context, request apiv1.GetKubeConfig
 		return "", fmt.Errorf("failed to GET /k8sd/kubeconfig: %w", err)
 	}
 	return response.KubeConfig, nil
-}
-
-// IsKubernetesAPIServerReady checks if kube-apiserver is reachable.
-func (c *k8sdClient) IsKubernetesAPIServerReady(ctx context.Context) bool {
-	kc, err := k8s.NewClient(c.snap)
-	if err != nil {
-		return false
-	}
-	_, err = kc.GetKubeAPIServerEndpoints(ctx)
-	if err != nil {
-		return false
-	}
-	return err == nil
 }
--- a/src/k8s/pkg/k8s/client/cluster.go
+++ b/src/k8s/pkg/k8s/client/cluster.go
@@ -50,8 +50,6 @@ func (c *k8sdClient) Bootstrap(ctx context.Context, hostname string, address str
 func (c *k8sdClient) ClusterStatus(ctx context.Context, waitReady bool) (apiv1.ClusterStatus, error) {
 	var response apiv1.GetClusterStatusResponse
 
-	// TODO(neoaggelos): this must be done on the server side
-	// fail fast if we're not explicitly waiting and we can't get kube-apiserver endpoints
 	if !waitReady && !c.IsKubernetesAPIServerReady(ctx) {
 		return apiv1.ClusterStatus{}, fmt.Errorf("there are no active kube-apiserver endpoints, cluster status is unavailable")
 	}
--- a/src/k8s/pkg/k8s/client/cluster.go
+++ b/src/k8s/pkg/k8s/client/cluster.go
@@ -49,7 +49,6 @@ func (c *k8sdClient) Bootstrap(ctx context.Context, hostname string, address str
 // ClusterStatus returns the current status of the cluster.
 func (c *k8sdClient) ClusterStatus(ctx context.Context, waitReady bool) (apiv1.ClusterStatus, error) {
 	var response apiv1.GetClusterStatusResponse
-
 	err := control.WaitUntilReady(ctx, func() (bool, error) {
 		if err := c.mc.Query(ctx, "GET", api.NewURL().Path("k8sd", "cluster"), nil, &response); err != nil {
 			return false, fmt.Errorf("failed to GET /k8sd/cluster: %w", err)
--- a/src/k8s/pkg/k8s/client/cluster.go
+++ b/src/k8s/pkg/k8s/client/cluster.go
@@ -49,6 +49,13 @@ func (c *k8sdClient) Bootstrap(ctx context.Context, hostname string, address str
 // ClusterStatus returns the current status of the cluster.
 func (c *k8sdClient) ClusterStatus(ctx context.Context, waitReady bool) (apiv1.ClusterStatus, error) {
 	var response apiv1.GetClusterStatusResponse
+
+	// TODO(neoaggelos): this must be done on the server side
+	// fail fast if we're not explicitly waiting and we can't get kube-apiserver endpoints
+	if !waitReady && !c.IsKubernetesAPIServerReady(ctx) {
+		return apiv1.ClusterStatus{}, fmt.Errorf("there are no active kube-apiserver endpoints, cluster status is unavailable")
+	}
+
 	err := control.WaitUntilReady(ctx, func() (bool, error) {
 		if err := c.mc.Query(ctx, "GET", api.NewURL().Path("k8sd", "cluster"), nil, &response); err != nil {
 			return false, fmt.Errorf("failed to GET /k8sd/cluster: %w", err)
--- a/src/k8s/pkg/k8s/client/cluster.go
+++ b/src/k8s/pkg/k8s/client/cluster.go
@@ -50,10 +50,6 @@ func (c *k8sdClient) Bootstrap(ctx context.Context, hostname string, address str
 func (c *k8sdClient) ClusterStatus(ctx context.Context, waitReady bool) (apiv1.ClusterStatus, error) {
 	var response apiv1.GetClusterStatusResponse
 
-	if !waitReady && !c.IsKubernetesAPIServerReady(ctx) {
-		return apiv1.ClusterStatus{}, fmt.Errorf("there are no active kube-apiserver endpoints, cluster status is unavailable")
-	}
-
 	err := control.WaitUntilReady(ctx, func() (bool, error) {
 		if err := c.mc.Query(ctx, "GET", api.NewURL().Path("k8sd", "cluster"), nil, &response); err != nil {
 			return false, fmt.Errorf("failed to GET /k8sd/cluster: %w", err)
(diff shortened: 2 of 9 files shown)


src/k8s/pkg/k8s/client/cluster.go
```diff
@@ -66,6 +66,11 @@ func (c *k8sdClient) Bootstrap(ctx context.Context, bootstrapConfig apiv1.Bootst
 // ClusterStatus returns the current status of the cluster.
 func (c *k8sdClient) ClusterStatus(ctx context.Context, waitReady bool) (apiv1.ClusterStatus, error) {
 	var response apiv1.GetClusterStatusResponse
+
+	if !waitReady && !c.IsKubernetesAPIServerReady(ctx) {
```
bschimke95, line 70: Please move this to the GET /cluster endpoint. You moved the check into the client implementation. So, it will still be executed on the client-side.

eaudetcobello: Thanks!

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -318,7 +318,8 @@ func onBootstrapControlPlane(s *state.State, initConfig map[string]string) error
 		return fmt.Errorf("failed to create k8s client: %w", err)
 	}
 
-	if err := client.WaitApiServerReady(s.Context); err != nil {
+	err = client.WaitApiServerReady(s.Context)
+	if err != nil {
```
neoaggelos, line 322: check errors inline
```suggestion
	if err := client.WaitApiServerReady(s.Context); err != nil {
```

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -181,8 +181,9 @@ func onPostJoin(s *state.State, initConfig map[string]string) error {
 		return fmt.Errorf("failed to create kubernetes client: %w", err)
 	}
 
-	if err := client.WaitApiServerReady(s.Context); err != nil {
-		return fmt.Errorf("kube-apiserver did not become ready: %w", err)
+	err = client.WaitApiServerReady(s.Context)
+	if err != nil {
```
neoaggelos, line 185: same here

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -9,24 +9,30 @@ import (
...
 )
 
-// WaitApiServerReady waits until the kube-apiserver becomes available.
+func (c *Client) listNodes(ctx context.Context) (*v1.NodeList, error) {
+	nodes, err := c.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list nodes: %v", err)
+	}
+	return nodes, nil
+}
```
neoaggelos, line 18: Let's remove this, as it is only wrapping one line of useful code: `c.CoreV1().Nodes().List(ctx, metav1.ListOptions{})`, without any other transformation or error handling.

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -181,8 +181,9 @@ func onPostJoin(s *state.State, initConfig map[string]string) error {
...
+		return fmt.Errorf("k8s api server did not become ready in time: %w", err)
```
neoaggelos, line 186: why `k8s api server` here? `kube-apiserver` is the standard service name

```suggestion
		return fmt.Errorf("kube-apiserver did not become ready: %w", err)
```

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -9,24 +9,30 @@ import (
...
+		_, err := c.listNodes(ctx)
 		return err == nil, nil
 	})
 }
 
-// ClusterReady checks the status of all nodes in the Kubernetes cluster.
-// If at least one node is in READY state it will return true.
-func (c *Client) ClusterReady(ctx context.Context) (bool, error) {
-	nodes, err := c.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+func (c *Client) IsClusterReady(ctx context.Context) (bool, error) {
```
neoaggelos, line 27: `IsClusterReady` can mean a lot of things, let's use a name that lets callers know what this is doing without having to loop up the implementation:

```suggestion
func (c *Client) HasAnyReadyNodes(ctx context.Context) (bool, error) {
```

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -9,24 +9,30 @@ import (
...
+	nodes, err := c.listNodes(ctx)
+
 	if err != nil {
-		return false, fmt.Errorf("failed to list nodes: %v", err)
+		return false, err
+	}
+
+	if len(nodes.Items) == 0 {
+		return false, fmt.Errorf("cluster has no nodes")
 	}
```
neoaggelos, line 36: This is problematic. Why do we consider it an "error" if we got back a response but no kubelets have registered yet? Isn't this handled by the `return false, nil` at the end of the function?

neoaggelos (changes requested):
Getting there! Please also rebase with latest `main`

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -9,24 +9,30 @@ import (
...
+	nodes, err := c.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list nodes: %v", err)
+	}
+	return nodes, nil
+}
+
 func (c *Client) WaitApiServerReady(ctx context.Context) error {
 	return control.WaitUntilReady(ctx, func() (bool, error) {
-		// TODO: use the /readyz endpoint instead
```
neoaggelos, line 15: please keep this TODO around

src/k8s/pkg/k8sd/api/impl/k8sd.go
```diff
@@ -17,28 +17,24 @@ import (
...
 	if err != nil {
-		return apiv1.ClusterStatus{}, fmt.Errorf("failed to get cluster members: %w", err)
+		return apiv1.ClusterStatus{}, fmt.Errorf("failed to create k8s client: %w", err)
 	}
 
-	config, err := utils.GetUserFacingClusterConfig(ctx, s)
+	ready, err := client.IsClusterReady(ctx)
 	if err != nil {
-		return apiv1.ClusterStatus{}, fmt.Errorf("failed to get user-facing cluster config: %w", err)
+		return apiv1.ClusterStatus{}, fmt.Errorf("failed to check if cluster is ready: %w", err)
```
neoaggelos, line 37:
```suggestion
		return apiv1.ClusterStatus{}, fmt.Errorf("failed to check if cluster has Ready nodes: %w", err)
```

neoaggelos (changes requested):

src/k8s/pkg/utils/k8s/status.go
eaudetcobello: Yep, you're right.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -319,8 +319,8 @@ func onBootstrapControlPlane(s *state.State, initConfig map[string]string) error
 		return fmt.Errorf("failed to create k8s client: %w", err)
 	}
 
-	if err := client.WaitApiServerReady(s.Context); err != nil {
-		return fmt.Errorf("k8s api server did not become ready in time: %w", err)
+	if client.WaitApiServerReady(s.Context); err != nil {
+		return fmt.Errorf("kube api-server did not become ready in time: %w", err)
```
neoaggelos, line 323: Otherwise you are returning the previous `err`
```suggestion
	if err := client.WaitApiServerReady(s.Context); err != nil {
		return fmt.Errorf("kube api-server did not become ready in time: %w", err)
```

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -2,31 +2,27 @@ package k8s
...
 	"context"
-	"fmt"
 
 	"github.com/canonical/k8s/pkg/utils/control"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// WaitApiServerReady waits until the kube-apiserver becomes available.
 func (c *Client) WaitApiServerReady(ctx context.Context) error {
```
neoaggelos, line 11: Keep the go doc comment

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -2,31 +2,27 @@ package k8s
...
 		// returning the error would abort, thus checking for nil
+		_, err := c.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
 		return err == nil, nil
 	})
 }
 
-// ClusterReady checks the status of all nodes in the Kubernetes cluster.
-// If at least one node is in READY state it will return true.
-func (c *Client) ClusterReady(ctx context.Context) (bool, error) {
+func (c *Client) HasReadyNodes(ctx context.Context) (bool, error) {
```
neoaggelos, line 21: same here

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -40,5 +36,5 @@ func (c *Client) ClusterReady(ctx context.Context) (bool, error) {
 		}
 	}
 
-	return false, nil
+	return false, nil // Cluster has no ready nodes
```
neoaggelos, line 39: Avoid inline comments like this, which would mean that someone has to lookup the implementation. Instead make sure they are in the go doc.

neoaggelos (changes requested):
Getting there

src/k8s/pkg/utils/k8s/status.go
eaudetcobello: Don't know why I lost all those comments.. sorry 'bout that

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -320,7 +320,7 @@ func onBootstrapControlPlane(s *state.State, initConfig map[string]string) error
 	}
 
 	if err := client.WaitApiServerReady(s.Context); err != nil {
-		return fmt.Errorf("k8s api server did not become ready in time: %w", err)
+		return fmt.Errorf("kube api-server did not become ready in time: %w", err)
```
neoaggelos, line 323: the service name is `kube-apiserver`
```suggestion
		return fmt.Errorf("kube-apiserver did not become ready in time: %w", err)
```

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -13,20 +12,20 @@ import (
...
-// If at least one node is in READY state it will return true.
-func (c *Client) ClusterReady(ctx context.Context) (bool, error) {
+// HasReadyNodes checks the status of all nodes in the Kubernetes cluster.
+// Returns true if there is atleast one ready node, false otherwise.
+func (c *Client) HasReadyNodes(ctx context.Context) (bool, error) {
 	nodes, err := c.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+
 	if err != nil {
-		return false, fmt.Errorf("failed to list nodes: %v", err)
+		return false, err
```
neoaggelos, line 28: revert, we need to wrap the error here

src/k8s/pkg/utils/k8s/status.go
```diff
@@ -13,20 +12,20 @@ import (
...
+		_, err := c.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
 		return err == nil, nil
 	})
 }
 
-// ClusterReady checks the status of all nodes in the Kubernetes cluster.
-// If at least one node is in READY state it will return true.
-func (c *Client) ClusterReady(ctx context.Context) (bool, error) {
+// HasReadyNodes checks the status of all nodes in the Kubernetes cluster.
+// Returns true if there is atleast one ready node, false otherwise.
```
neoaggelos, line 23:
```suggestion
// HasReadyNodes returns true if there is at least one Ready node in the cluster, false otherwise.
```

src/k8s/cmd/k8s/k8s_status.go
```diff
@@ -27,16 +27,6 @@ func newStatusCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
 			}
 
-			// TODO(neoaggelos): this must be done on the server side
-			// fail fast if we're not explicitly waiting and we can't get kube-apiserver endpoints
-			if !opts.waitReady {
-				if !client.IsKubernetesAPIServerReady(cmd.Context()) {
```
neoaggelos, line 33: Please also remove the implementation of `client.IsKubernetesAPIServerReady()` as it not used anywhere anymore (remove from client implementation and the interface)

neoaggelos (commented):
Final changes and this is good to go!

neoaggelos (approved):

neoaggelos:
Make sure to clean the squash commit message (the PR description with proper wording should suffice)

PR #253, opened 2024-03-15
refactor and cleanup pkg/k8s/client

neoaggelos:
## Summary

Refactor and cleanup some of the pkg/k8s/client

## Changes

- Update the client interface to use the API request structs where possible
- Rename some structs to match the API requests
- Update error messages to be consistent

### DIFF
--- a/src/k8s/pkg/k8s/client/client.go
+++ b/src/k8s/pkg/k8s/client/client.go
@@ -5,64 +5,36 @@ import (
 	"fmt"
 
 	"github.com/canonical/k8s/pkg/snap"
-	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/microcluster/client"
 	"github.com/canonical/microcluster/microcluster"
 )
 
-// ClusterOpts contains options for cluster queries.
-type ClusterOpts struct {
-	// Verbose enables info level logging.
-	Verbose bool
-	// Debug enables trace level logging.
-	Debug bool
-	// Snap is the snap instance.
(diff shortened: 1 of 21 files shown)


src/k8s/pkg/k8s/client/client.go
```diff
@@ -5,64 +5,36 @@ import (
...
 	return &k8sdClient{
-		snap: opts.Snap,
+		snap: snap,
 		m:    m,
-		mc:   microClient,
+		mc:   mc,
 	}, nil
 }
-
-func (c *k8sdClient) Query(ctx context.Context, method string, path *api.URL, in any, out any) error {
```
bschimke95, line 63: nit: you could keep the wrapper (maybe make it private) and assemble to proper error message here instead of doing that manually in each call. This may become outdated.

```
func (c *k8sdClient) Query(ctx context.Context, method string, path *api.URL, in any, out any) error {
	if err := c.mc.Query(ctx, method, path, in, out); err != nil {
		return fmt.Errorf("failed to %s %s: %w, method, path, err)
	}
	return nil
}
```

bschimke95 (approved):
LGTM, great changes. I like how they simplify e.g. the client mock.

src/k8s/pkg/k8s/client/client.go
neoaggelos: i'd rather we keep this on a client side. functions should not wrap the error with info the caller already has, and we might want to add more info in the future

PR #254, opened 2024-03-15
Configure External Datastore

mateoflorido:
## Overview
We are missing some checks for external datastores. Therefore, while attempting to join other control plane units, these fail to join the cluster. This pull request aims to fix that issue


bschimke95 (approved):
LGTM, good catch

neoaggelos (approved):

PR #256, opened 2024-03-17
Use optional fields for ClusterConfig

neoaggelos:
### Summary

Update all cluster configurations to use `*<type>` fields, such that they can be set to empty values. The changes are extensive but are mostly to introduce structure to the code, and ensure that all checks happen in the right place

### Notes

- This PR introduces huge breaking changes in the cluster configuration. These were left out during the initial development to get a v0 working implementation, but are required going forward. After merging this PR, any running clusters that migrate to this will be effectively bricked
- There is no point in working towards a migration path _for now_, as there has not been any stable releases yet. For the same reason, it is best to do the changes _before_ there is a stable release.

### Changes

- `apiv1.UserFacingClusterConfig` does not have nil structs, all individual fields can be `nil` instead.
- Add `Getter()` functions for all fields, to avoid having to check for pointers all around the code. To facilitate this, all zero values and nil values of fields should behave identically.
- `types.ClusterConfig` rework, same as above. All fields can be `nil`, and all structs are never nil.
- Move convert functions from API <-> internal types in the `types` package
- Move merge config validations into the `types` package
- Add extensive unit tests for all cluster config field updates, and also test for specific scenarios that match our validations.
- `database.SetClusterConfig()` now also returns the merged cluster config, simplifying code in the caller sites.
- Use `json` for storing the cluster configuration.
- Ensure `types.ClusterConfig.SetDefaults()` sets all fields such that they are not nil.
- Update all call sites (there are a lot of these, since the internal types are used throughout the API and hooks). Most are about using the designated Getters instead of the struct fields.

### Previous issues

- Validating the merged cluster configuration happened in two steps, in `types.MergeClusterConfig` and then in `api.validateConfig`. This is now handled properly in `types.MergeClusterConfig` instead. Extensive tests have been added to validate the expected behaviours

- Default values for the cluster config were split between `(*types.ClusterConfig).SetDefaults()` and `utils.GetUserFacingClusterConfig`, making it hard to track around the code if a particular value is nil, or if the user will be presented with slightly misleading configs.

 - Converting to and from userfacing config has been moved to the `types` package, it does not belong in `utils`, especially not coupled with retrieving the cluster config from the database

- It was impossible to "unset" configurations after they had been configured (e.g. reset the `load-balancer.cidrs`) field. We now use optional fields for everything (`*type`). We now consider `nil` values instead of `zero` values as "don't update". Further, we ensure that for each field the zero and the nil value behave in the same way, such that we can have straightforward `Get$Field()` and simplify the code.

This now works:

```
root@t1:~# k8s get load-balancer.cidrs 
[]
root@t1:~# k8s set load-balancer.cidrs=10.0.100.0/24
Configuration updated.
root@t1:~# k8s get load-balancer.cidrs 
[10.0.100.0/24]
root@t1:~# k8s set load-balancer.cidrs=
Configuration updated.
root@t1:~# k8s get load-balancer.cidrs 
[]
```

- Fixes a bug where setting dns.enabled=false and a dns.service-ip did not work, as it was overwritten to an empty config.


src/k8s/pkg/k8sd/database/cluster_config.go
```diff
@@ -19,28 +19,28 @@ var (
 
 // SetClusterConfig updates the cluster configuration with any non-empty values that are set.
 // SetClusterConfig will attempt to merge the existing and new configs, and return an error if any protected fields have changed.
```
bschimke95, line 21: nit: update comment to include that the merged config is returned now.

src/k8s/pkg/k8sd/database/cluster_config.go
```diff
@@ -59,7 +59,7 @@ func GetClusterConfig(ctx context.Context, tx *sql.Tx) (types.ClusterConfig, err
 	}
 
 	var clusterConfig types.ClusterConfig
-	if err := yaml.Unmarshal([]byte(s), &clusterConfig); err != nil {
+	if err := json.Unmarshal([]byte(s), &clusterConfig); err != nil {
```
bschimke95, line 62: out of curiosity: why the change to json?

src/k8s/pkg/k8sd/types/cluster_config_defaults.go
```diff
@@ -0,0 +1,100 @@
...
+		c.LoadBalancer.Enabled = vals.Pointer(false)
+	}
+	if c.LoadBalancer.CIDRs == nil {
+		c.LoadBalancer.CIDRs = vals.Pointer([]string{})
+	}
+	if c.LoadBalancer.L2Mode == nil {
+		c.LoadBalancer.L2Mode = vals.Pointer(false)
+	}
+	if c.LoadBalancer.L2Interfaces == nil {
+		c.LoadBalancer.L2Interfaces = vals.Pointer([]string{})
```
bschimke95, line 65: having a pointer to a slice seems a bit redundant but this is probably for consistency?

src/k8s/pkg/k8sd/types/cluster_config_validate.go
```diff
@@ -0,0 +1,30 @@
+package types
+
+import (
+	"fmt"
+	"net"
+	"strings"
+)
+
+func (c *ClusterConfig) Validate() error {
```
bschimke95, line 9: nit: can you add a short comment what is validated here?

bschimke95 (approved):
Great work as always @neoaggelos 
Just a few nits.

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -129,26 +129,26 @@ func onPostJoin(s *state.State, initConfig map[string]string) error {
 		}
 		cluster := make([]string, len(members))
 		for _, member := range members {
-			cluster = append(cluster, fmt.Sprintf("%s:%d", member.Address.Addr(), cfg.K8sDqlite.Port))
+			cluster = append(cluster, fmt.Sprintf("%s:%d", member.Address.Addr(), cfg.Datastore.GetK8sDqlitePort()))
```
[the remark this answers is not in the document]
neoaggelos: decided against it, as `GetDatastorePort()` has a generic name, and the semantics for that value when not using k8s-dqlite would be blurry (and confusing). Instead, `GetK8sDqlitePort()` clearly means that this is for the k8s-dqlite datastore instead

neoaggelos:
Waiting for #279 to rebase

src/k8s/pkg/k8sd/controllers/node_configuration_test.go
```diff
@@ -109,21 +130,31 @@ func TestConfigPropagation(t *testing.T) {
...
+
+			// TODO: this is to ensure that the controller has handled the event. This should ideally
+			// be replaced with something like a "<-sentCh" instead
 			time.Sleep(100 * time.Millisecond)
 
-			for ekey, evalue := range tc.expectedUpdates {
+			for ekey, evalue := range tc.expectArgs {
 				val, err := snaputil.GetServiceArgument(s, "kubelet", ekey)
-				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(err).To(Succeed())
```
bschimke95, line 149: I think succeed is only used for single return values. 
`Expect(err).To(Succeed())` reads wrong (although it might work)

https://github.com/onsi/gomega/blob/437995126acee58e6237f963913f54cb6386dd0a/matchers.go#L103

bschimke95 (approved):
LGTM

PR #257, opened 2024-03-18
POST k8sd/cluster endpoint to bootstrap a cluster

louiseschmidtgen:
## Description 
This PR adds the  `POST k8sd/cluster` endpoint to the k8s-snap.

The endpoints calls `postClusterBootstrap` which then calls the `cluster/control` endpoint in microcluster in order to create a new cluster.

### DIFF
--- a/src/k8s/api/v1/cluster.go
+++ b/src/k8s/api/v1/cluster.go
@@ -8,7 +8,7 @@ type GetClusterStatusResponse struct {
 	ClusterStatus ClusterStatus `json:"status"`
 }
 
-// PostClusterBootstrapRequest is used to bootstrap the cluster.
+// PostClusterBootstrapRequest is used to bootstrap the cluster by using the endpoint "POST 1.0/k8sd/cluster".
 type PostClusterBootstrapRequest struct {
 	Name    string          `json:"name"`
 	Address string          `json:"address"`
--- a/src/k8s/api/v1/cluster.go
+++ b/src/k8s/api/v1/cluster.go
@@ -10,10 +10,9 @@ type GetClusterStatusResponse struct {
 
 // PostClusterBootstrapRequest is used to bootstrap the cluster.
 type PostClusterBootstrapRequest struct {
-	Bootstrap bool            `json:"bootstrap"`
-	Name      string          `json:"name"`
-	Address   string          `json:"address"`
-	Config    BootstrapConfig `json:"config"`
+	Name    string          `json:"name"`
+	Address string          `json:"address"`
+	Config  BootstrapConfig `json:"config"`
 }
 
 // GetKubeConfigRequest is used to ask for the admin kubeconfig
--- a/src/k8s/api/v1/cluster.go
+++ b/src/k8s/api/v1/cluster.go
@@ -8,6 +8,14 @@ type GetClusterStatusResponse struct {
 	ClusterStatus ClusterStatus `json:"status"`
 }
 
+// PostClusterBootstrapRequest is used to bootstrap the cluster.
+type PostClusterBootstrapRequest struct {
+	Bootstrap bool            `json:"bootstrap"`
+	Name      string          `json:"name"`
+	Address   string          `json:"address"`
+	Config    BootstrapConfig `json:"config"`
+}
+
 // GetKubeConfigRequest is used to ask for the admin kubeconfig
 type GetKubeConfigRequest struct {
 	Server string `json:"server"`
--- a/src/k8s/pkg/k8s/client/mock/mock.go
+++ b/src/k8s/pkg/k8s/client/mock/mock.go
@@ -10,10 +10,8 @@ import (
 // Client is a mock implementation for k8s Client.
 type Client struct {
 	BootstrapCalledWith struct {
-		Ctx             context.Context
-		Name            string
-		Address         string
-		BootstrapConfig apiv1.BootstrapConfig
+		Ctx     context.Context
+		request apiv1.PostClusterBootstrapRequest
 	}
 	BootstrapClusterMember apiv1.NodeStatus
 	BootstrapErr           error
@@ -53,11 +51,9 @@ type Client struct {
 	UpdateClusterConfigErr        error
 }
 
-func (c *Client) Bootstrap(ctx context.Context, name string, address string, bootstrapConfig apiv1.BootstrapConfig) (apiv1.NodeStatus, error) {
+func (c *Client) Bootstrap(ctx context.Context, request apiv1.PostClusterBootstrapRequest) (apiv1.NodeStatus, error) {
 	c.BootstrapCalledWith.Ctx = ctx
-	c.BootstrapCalledWith.Name = name
-	c.BootstrapCalledWith.Address = address
-	c.BootstrapCalledWith.BootstrapConfig = bootstrapConfig
+	c.BootstrapCalledWith.request = request
 	return c.BootstrapClusterMember, c.BootstrapErr
 }
 
--- a/src/k8s/pkg/k8s/client/mock/mock.go
+++ b/src/k8s/pkg/k8s/client/mock/mock.go
@@ -11,7 +11,7 @@ import (
 type Client struct {
 	BootstrapCalledWith struct {
 		Ctx     context.Context
-		request apiv1.PostClusterBootstrapRequest
+		Request apiv1.PostClusterBootstrapRequest
 	}
 	BootstrapClusterMember apiv1.NodeStatus
 	BootstrapErr           error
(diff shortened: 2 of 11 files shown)


src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	//Convert Bootstrap config to map
+	config, err := req.BootstrapConfig.ToMap()
+	if err != nil {
+		return response.BadRequest(fmt.Errorf("failed to convert bootstrap config to map: %w", err))
+	}
+
+	// Clean hostname
+	// TODO: Move this to the server side
```
bschimke95, line 32: this is the server side. Copy paste error? :)

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+
+func postClusterBootstrap(m *microcluster.MicroCluster, s *state.State, r *http.Request) response.Response {
+	req := apiv1.ClusterBootstrapRequest{}
+	req.BootstrapConfig.SetDefaults()
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return response.BadRequest(fmt.Errorf("failed to parse request: %w", err))
+	}
+
+	//Convert Bootstrap config to map
+	config, err := req.BootstrapConfig.ToMap()
```
bschimke95, line 26: You might want to wait for https://github.com/canonical/k8s-snap/pull/256 to land. I assume there might be some conflicts.

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	// Set timeout
+	timeout := 30 * time.Second
+	if deadline, set := s.Context.Deadline(); set {
+		timeout = time.Until(deadline)
+	}
+
+	if err := m.Ready(int(timeout / time.Second)); err != nil {
+		return response.BadRequest(fmt.Errorf("cluster did not come up in time: %w", err))
```
bschimke95, line 45: A `InternalError` is more appropriate. `BadRequests` are the users fault.

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	// TODO Check if already Bootstrapped
```
bschimke95, line 48: There should already be a util function to check for this.

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Bootstrap the cluster
+	// TODO where do we grab the address from?
+	address := util.CanonicalNetworkAddress(util.NetworkInterfaceAddress(), req.BootstrapConfig.K8sDqlitePort)
+	if err := m.NewCluster(hostname, address, config, timeout); err != nil {
+		// TODO(neoaggelos): only return error that bootstrap failed
+		fmt.Fprintln(os.Stderr, "Failed with error:", err)
```
bschimke95, line 55: We already return the error, no need to log it here. Feel free to still log it but then use the `log` module.

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+
+		// c.CleanupNode(ctx, hostname)
```
bschimke95, line 57: I guess it makes sense to also move this on the server-side. I was never a big fan of let the client handle the cleanup.

bschimke95 (commented):
Great work @louiseschmidtgen, did a first pass

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
louiseschmidtgen: There is a new card to move node cleanup server side.

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Set timeout
+	timeout := 30 * time.Second
+	if deadline, set := s.Context.Deadline(); set {
+		timeout = time.Until(deadline)
+	}
+
+	if err := m.Ready(int(timeout / time.Second)); err != nil {
+		return response.InternalError(fmt.Errorf("cluster did not come up in time: %w", err))
+	}
```
neoaggelos, line 44: this `.Ready()` talks to the microcluster socket to verify it's running. here, we are already on the server side, so this is not useful.

neoaggelos (commented):
Looks good so far, there is still work to do on the client side to use this endpoint instead

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// Check if the cluster is already bootstrapped
+	_, err = m.Status()
+	if err == nil {
+		return response.BadRequest(fmt.Errorf("cluster is already bootstrapped"))
+	}
+
+	// Set timeout
+	timeout := 30 * time.Second
+	if deadline, set := s.Context.Deadline(); set {
```
bschimke95, line 44: nit: we have this now in a bunch of places. Consider introducing a util `timeoutFromCtx(ctx, default)`

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,56 @@
...
+		timeout = time.Until(deadline)
+	}
+
+	// Bootstrap the cluster
+	address := util.CanonicalNetworkAddress(util.NetworkInterfaceAddress(), req.BootstrapConfig.K8sDqlitePort)
```
bschimke95, line 49: this should not be the `k8sDqlitePort`

bschimke95 (commented):
server side looks good. 
Please also update the client-side.

src/k8s/api/v1/cluster.go
```diff
@@ -8,6 +8,14 @@ type GetClusterStatusResponse struct {
 	ClusterStatus ClusterStatus `json:"status"`
 }
 
+// PostClusterBootstrapRequest is used to bootstrap the cluster.
+type PostClusterBootstrapRequest struct {
+	Bootstrap bool            `json:"bootstrap"`
```
neoaggelos, line 13: this is not used anywhere, is it?

louiseschmidtgen: Should be in use now :)

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -0,0 +1,56 @@
...
+	"net/http"
+
+	apiv1 "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/k8s/pkg/utils"
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/microcluster"
+	"github.com/canonical/microcluster/state"
+)
+
+func postClusterBootstrap(m *microcluster.MicroCluster, s *state.State, r *http.Request) response.Response {
```
neoaggelos, line 15: I don't see the `.Bootstrap` field of the request used anywhere, is there a need for it?

louiseschmidtgen: I put it in because the API spec suggests we would like it? I don't see a purpose for it either... I'll remove it.

src/k8s/api/v1/cluster.go
```diff
@@ -8,6 +8,13 @@ type GetClusterStatusResponse struct {
 	ClusterStatus ClusterStatus `json:"status"`
 }
 
+// PostClusterBootstrapRequest is used to bootstrap the cluster.
+type PostClusterBootstrapRequest struct {
```
bschimke95, line 12: Our naming has become a bit weird here.
We initially had "<method><endpoint-name>" for the API side function. 
So, it would be consistent to name this `PostClusterRequest` but that is even more weird. 
Well, the API will likely change several times anyway so we got some time to think about it...

src/k8s/pkg/k8s/client/mock/mock.go
```diff
@@ -10,10 +10,8 @@ import (
 // Client is a mock implementation for k8s Client.
 type Client struct {
 	BootstrapCalledWith struct {
-		Ctx             context.Context
-		Name            string
-		Address         string
-		BootstrapConfig apiv1.BootstrapConfig
+		Ctx     context.Context
+		request apiv1.PostClusterBootstrapRequest
```
bschimke95, line 14: Make this public

bschimke95 (commented):
the CI passes now, some final comments and we are good to merge.

src/k8s/api/v1/cluster.go
louiseschmidtgen: I agree, my thought here was since we are also doing `GetClusterStatusRequest` instead of `GetClusterRequest` it might be a bit better in conveying what it is used for.

bschimke95 (approved):
LGTM, great work

PR #258, opened 2024-03-20
Propagate node configuration through kubernetes api

berkayoz:
Added `NodeConfigurationController` that handles watching and reconciling arguments on services. 
Restructured components(functionalities) to reduce code duplication and help with updating node config(`k8sd-config` configmap)

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -21,8 +21,10 @@ import (
 	"github.com/canonical/k8s/pkg/snap"
 	snaputil "github.com/canonical/k8s/pkg/snap/util"
 	"github.com/canonical/k8s/pkg/utils"
+	"github.com/canonical/k8s/pkg/utils/k8s"
 	"github.com/canonical/k8s/pkg/utils/vals"
 	"github.com/canonical/microcluster/state"
+	"github.com/mitchellh/mapstructure"
 )
 
 // onBootstrap is called after we bootstrap the first cluster node.
@@ -304,6 +306,23 @@ func onBootstrapControlPlane(s *state.State, initConfig map[string]string) error
 			return nil
 		}); err != nil {
 			return fmt.Errorf("database transaction to update cluster configuration failed: %w", err)
+		}
+
+		var data map[string]string
+		if err := mapstructure.Decode(types.NodeConfig{
+			ClusterDNS:    dnsIP,
+			ClusterDomain: cfg.Kubelet.ClusterDomain,
+		}, &data); err != nil {
+			return fmt.Errorf("failed to encode node config: %w", err)
+		}
+
+		client, err := k8s.NewClient(snap.KubernetesRESTClientGetter(""))
+		if err != nil {
+			return fmt.Errorf("failed to create kubernetes client: %w", err)
+		}
+
+		if _, err := client.UpdateConfigMap(s.Context, "kube-system", "k8sd-config", data); err != nil {
+			return fmt.Errorf("failed to update node configs: %w", err)
 		}
 	}
 
--- a/src/k8s/pkg/k8sd/app/hooks_start.go
+++ b/src/k8s/pkg/k8sd/app/hooks_start.go
@@ -9,8 +9,8 @@ import (
 func onStart(s *state.State) error {
 	snap := snap.SnapFromContext(s.Context)
 
-	configController := controllers.NewNodeConfigurationController()
-	go configController.Run(s.Context, snap)
+	configController := controllers.NewNodeConfigurationController(snap)
+	go configController.Run(s.Context)
 
 	return nil
 }
--- a/src/k8s/pkg/k8sd/app/hooks_start.go
+++ b/src/k8s/pkg/k8sd/app/hooks_start.go
@@ -1,15 +1,20 @@
 package app
 
 import (
+	"context"
+
 	"github.com/canonical/k8s/pkg/k8sd/controllers"
 	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/k8s/pkg/utils/k8s"
 	"github.com/canonical/microcluster/state"
 )
 
 func onStart(s *state.State) error {
 	snap := snap.SnapFromContext(s.Context)
 
-	configController := controllers.NewNodeConfigurationController(snap)
+	configController := controllers.NewNodeConfigurationController(snap, func(ctx context.Context) *k8s.Client {
+		return k8s.RetryNewClient(ctx, snap.KubernetesNodeRESTClientGetter("kube-system"))
+	})
 	go configController.Run(s.Context)
 
 	return nil
--- a/src/k8s/pkg/k8sd/app/hooks_start.go
+++ b/src/k8s/pkg/k8sd/app/hooks_start.go
@@ -1,0 +1,16 @@
+package app
+
+import (
+	"github.com/canonical/k8s/pkg/k8sd/controllers"
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/microcluster/state"
+)
+
+func onStart(s *state.State) error {
+	snap := snap.SnapFromContext(s.Context)
+
+	configController := controllers.NewNodeConfigurationController()
+	go configController.Run(s.Context, snap)
+
+	return nil
+}
--- a/src/k8s/pkg/k8sd/controllers/node_configuration.go
+++ b/src/k8s/pkg/k8sd/controllers/node_configuration.go
@@ -1,0 +1,79 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/canonical/k8s/pkg/k8sd/types"
+	"github.com/canonical/k8s/pkg/snap"
+	snaputil "github.com/canonical/k8s/pkg/snap/util"
+	"github.com/canonical/k8s/pkg/utils/k8s"
+	"github.com/mitchellh/mapstructure"
+	v1 "k8s.io/api/core/v1"
+)
+
+type NodeConfigurationController struct {
+}
+
+func NewNodeConfigurationController() *NodeConfigurationController {
+	return &NodeConfigurationController{}
+}
+
+func (c *NodeConfigurationController) Run(ctx context.Context, snap snap.Snap) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-time.After(3 * time.Second):
+		default:
+		}
+
+		client, err := k8s.NewClient(snap.KubernetesNodeRESTClientGetter("kube-system"))
+		if err != nil {
+			// This fails when the node is not bootstrapped or joined
+			log.Println(fmt.Errorf("failed to create kubernetes node client: %w", err))
+			continue
+		}
+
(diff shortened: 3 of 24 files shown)


src/k8s/pkg/k8sd/types/node_config.go
```diff
@@ -0,0 +1,7 @@
+package types
+
+type NodeConfig struct {
+	CloudProvider string `mapstructure:"cloud-provider,omitempty"`
```
bschimke95, line 4: no need to pull in mapstructure - just use `json`

src/k8s/pkg/utils/k8s/node_config.go
bschimke95, line 1: please add some test coverage here.

src/k8s/pkg/utils/k8s/node_config.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/canonical/k8s/pkg/k8sd/types"
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/mitchellh/mapstructure"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	applyv1 "k8s.io/client-go/applyconfigurations/core/v1"
+)
+
+func (c *Client) WatchNodeConfig(ctx context.Context, snap snap.Snap, reconcile func(ctx context.Context, snap snap.Snap, config types.NodeConfig) error, cancel func()) {
```
bschimke95, line 16: Can we just make this a `WatchResource(namespace, name)`?

src/k8s/pkg/utils/k8s/node_config.go
```diff
@@ -0,0 +1,57 @@
...
+
+				if err := reconcile(ctx, snap, config); err != nil {
+					log.Println(fmt.Errorf("failed to reconcile node config: %w", err))
+				}
+			}
+		}
+	}
+}
+
+func (c *Client) UpdateNodeConfig(ctx context.Context, config types.NodeConfig) error {
```
bschimke95, line 45: same here

bschimke95 (commented):
great work @berkayoz - did a first pass on the propagation part

src/k8s/pkg/k8sd/types/node_config.go
berkayoz: We can technically use `json` but wouldn't that mean doing a `Marshal` and `Unmarshal` just to convert between `map[string]string` and `NodeConfig`. Seems a bit inefficient/hacky imo.

src/k8s/pkg/utils/k8s/node_config.go
berkayoz: Lets leave the unit tests and an e2e test for the propagation to a separate card, we have big and conflicting PRs on the way. I'll pull the tests in for the next pulse, which will be easier to rebase and work with.

src/k8s/pkg/utils/k8s/node_config.go
berkayoz: We can do that with generics and pushing the data handling(by passing the whole resource) down the line, I kind of followed the spec. I don't have a strong preference, would like to hear from @neoaggelos on this as well.

src/k8s/pkg/utils/k8s/client.go
```diff
@@ -23,3 +23,15 @@ func NewClient(snap snap.Snap) (*Client, error) {
...
+	config, err := snap.KubernetesNodeRESTClientGetter("kube-system").ToRESTConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build Kubernetes REST config: %w", err)
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Kubernetes clientset: %w", err)
+	}
+	return &Client{clientset}, nil
+}
```
neoaggelos, line 37: Rather, make this a single `NewClient()` that accepts the restclientgetter and the namespace as arguments. The kubernetes client package does not _care_ about what kind of client it will create.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -325,62 +325,105 @@ func onBootstrapControlPlane(s *state.State, initConfig map[string]string) error
 	}
 
 	if cfg.Network.Enabled != nil {
-		err := component.ReconcileNetworkComponent(s.Context, snap, vals.Pointer(false), cfg.Network.Enabled, cfg)
-		if err != nil {
+		network := &component.Network{
+			PodCIDR: cfg.Network.PodCIDR,
+		}
```
neoaggelos, line 330: not a huge fan of this, as there are no clear semantics as to which fields are inputs and which are outputs. I'd recommend reverting these changes

src/k8s/pkg/k8sd/controllers/node_configuration.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if config.ClusterDomain != "" {
+		var err error
+		kubeletChanged, err = snaputil.UpdateServiceArguments(snap, "kubelet", map[string]string{
+			"--cluster-domain": config.ClusterDomain,
+		}, nil)
+		if err != nil {
+			return fmt.Errorf("failed to update kubelet arguments: %w", err)
+		}
+	}
```
neoaggelos, line 74: create an `updateMap map[string]string` and `deleteList []string`, then call `snaputil.UpdateServiceArguments()` once

src/k8s/pkg/k8sd/controllers/node_configuration.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if kubeletChanged {
```
neoaggelos, line 76: two notes here: i'd call this `mustRestartKubelet`, to be clear on the intent of how this will be used. secondly, note that this value might not be correct as the code is structured (if both clusterDNS and clusterDomain are set, only the clusterDomain is considered for restarting kubelet, which is not right.

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/microcluster/state"
+)
+
+func onStart(s *state.State) error {
+	snap := snap.SnapFromContext(s.Context)
+
+	configController := controllers.NewNodeConfigurationController()
+	go configController.Run(s.Context, snap)
+
```
neoaggelos, line 14: i'd love us to have an option like `--disable-node-configuration-controller` in k8sd. Just a note for the future, don't do this just yet

neoaggelos (changes requested):
Nice work so far, but needs some improvements

src/k8s/pkg/utils/k8s/node_config.go
neoaggelos: The kubernetes client should use kubernetes terms. So, let's make this a `WatchConfigMap(ctx, namespace, name, reconcile func(o *corev1.ConfigMap) error)`

Note that the reconcile function only gets the configmap object, everything else would come as a closure from the caller site

src/k8s/pkg/utils/k8s/client.go
berkayoz: I've updated it, not exactly sure if this was what you envisioned. Let me know if this needs more changing, thank you!

src/k8s/pkg/k8sd/controllers/node_configuration.go
```diff
@@ -0,0 +1,82 @@
...
+	return &NodeConfigurationController{}
+}
+
+func (c *NodeConfigurationController) Run(ctx context.Context, snap snap.Snap) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
```
neoaggelos, line 30: change the `time.Sleep()` at the end of the loop with a `<-time.After(3 * time.Second)` here. This interval should be something larger like 10 seconds (or more?). That way you don't need multiple `time.Sleep()`s in the loop, and you don't have to wait for sleep to finish to see that the context has been canceled.

src/k8s/pkg/k8sd/controllers/node_configuration.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			log.Println(fmt.Errorf("failed to create kubernetes node client: %w", err))
+			// This fails when the node is not bootstrapped or joined, acts like a retry timer
+			time.Sleep(3 * time.Second)
+			continue
+		}
+
+		reconcile := c.reconcile(ctx, snap)
+		client.WatchConfigMap(ctx, "kube-system", "k8sd-config", reconcile)
+
```
neoaggelos, line 42: then use a closure to capture the needed state here:

```suggestion
		client.WatchConfigMap(ctx, "kube-system", "k8sd-config", func(configMap *v1.ConfigMap) error { return c.reconcile(ctx, snap, configMap) })

```

src/k8s/pkg/k8sd/controllers/node_configuration.go
```diff
@@ -0,0 +1,82 @@
...
+func (c *NodeConfigurationController) reconcile(ctx context.Context, snap snap.Snap) func(*v1.ConfigMap) error {
+	return func(configMap *v1.ConfigMap) error {
+		var nodeConfig types.NodeConfig
+		if err := mapstructure.Decode(configMap.Data, &nodeConfig); err != nil {
+			return fmt.Errorf("failed to decode node config: %w", err)
+		}
+
+		var kubeletUpdateMap map[string]string = map[string]string{}
+		var kubeletDeleteList []string = []string{}
```
neoaggelos, line 57: more idiomatic

```suggestion
		kubeletUpdateMap := make(map[string]string)
		var kubeletDeleteList []string
```

src/k8s/pkg/k8sd/controllers/node_configuration.go
berkayoz: I'm not exactly sure if the interval should be something larger. Mainly 2 concerns:
* The initial start of the controller might be a bit delayed, which increases the time for nodes to settle.
* If for some reason the `Watch` fails, it will take 10 seconds to recover

I think a small interval is okay, the only worry I would have is for the failure on getting a kubernetes client when the cluster is not bootstrapped/joined. Especially in the scenario where someone installed the `k8s` snap and never interacted with it. We would print an error message into the logs every 3 seconds, that can eventually exhaust disk space.

neoaggelos (approved):
LGTM, how is this tested? Is there value in having an e2e test for this? Join a control plane and a worker, update a config like the dns.service-ip, and check that all nodes eventually use that?

src/k8s/pkg/k8sd/controllers/node_configuration.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+type NodeConfigurationController struct {
+}
+
+func NewNodeConfigurationController() *NodeConfigurationController {
+	return &NodeConfigurationController{}
+}
+
+func (c *NodeConfigurationController) Run(ctx context.Context, snap snap.Snap) error {
```
neoaggelos, line 24: Make the snap and the k8s client a part of the controller itself, and consider adding a unit test (using a mock snap and client)

neoaggelos: Otherwise we create clients in a loop, which works but will eventually lead to fd leaks

neoaggelos: If we need to recreate the client (e.g. because config files are not yet ready, consider making this a function that creates the client

bschimke95 (approved):
LGTM

PR #265, opened 2024-03-22
Load-balancer how to

evilnick:
I originally wrote a much larger doc but I have stripped all that content out and will save for an explanation page (we should also cover using external load-balancers first before adding that.
So, this is fairly straightforward.

### DIFF
--- a/docs/src/howto/networking/default-loadbalancer.md
+++ b/docs/src/howto/networking/default-loadbalancer.md
@@ -1,0 +1,81 @@
+# How to use the default load-balancer
+
+Canonical Kubernetes includes a default load-balancer. As this is not an
+essential service for all deployments, it is not enabled by default. This guide
+explains how to configure and enable the loadbalancer.
+
+## What you'll need
+
+This guide assumes the following:
+
+- You have root or sudo access to the machine.
+- You have a bootstraped Canonical Kubernetes cluster (see the [Getting
+  Started][getting-started-guide] guide).
+
+## Check the status and configuration
+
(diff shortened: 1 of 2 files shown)


docs/src/howto/networking/default-loadbalancer.md
```diff
@@ -0,0 +1,81 @@
+# How to use the default load-balancer
+
+Canonical Kubernetes includes a default load-balancer. As this is not an
+essential service for all deployments, it is not enabled by default. This guide
+explains how to configure and enable the loadbalancer.
```
eaudetcobello, line 5: load-balancer or loadbalancer?

docs/src/howto/networking/default-loadbalancer.md
```diff
@@ -0,0 +1,81 @@
...
+sudo k8s get load-balancer
+```
+This should output a list of values like this:
+
+
+- `cidrs` - a list containing [cidr] definitions of the pool of IP addresses to use
+- `l2-mode` - whether L2 mode (failover) is turned on
+- `l2-interfaces`  - optional list of interfaces to announce services over( defaults to all)
+- `bgp-mode` - whether BGP mode is active.
+- `bgp-local-asn` - the local Autonomous System Number(ASN)
```
eaudetcobello, line 38:
```suggestion
- `bgp-local-asn` - the local Autonomous System Number (ASN)
```

eaudetcobello (commented):
Cool! Just a couple typos

kwmonroe (approved):
couple drive-by tweaks.  otherwise lgtm.

PR #273, opened 2024-03-25
Add kubeconfig tests and bootstrapping validation

bschimke95:
Just noticed that we do not have any test coverage for `k8s config`. 
Also, added a pre-flight `IsBootstrapped` check for.


tests/e2e/tests/test_kubeconfig.py
neoaggelos, line 1: not a huge fan of a separate e2e test just for this. could we perhaps have this as part of the smoke test? or just a unit test really?

berkayoz (approved):
LGTM

PR #274, opened 2024-03-25
Rename e2e to integration

bschimke95:
The E2E keyword gets confusing when talking about actual upstream E2E/conformance tests and the integration tests in the snap. This PR renames the integration tests to avoid confusion.


addyess (approved):
Super duper.  Looks amazing

PR #279, opened 2024-03-28
Added unit and integration test for node config propagation

berkayoz:
Added unit and integration test for node config propagation, follow up on #258

### DIFF
--- a/src/k8s/pkg/component/gateway.go
+++ b/src/k8s/pkg/component/gateway.go
@@ -47,22 +47,23 @@ func UpdateGatewayComponent(ctx context.Context, s snap.Snap, isRefresh bool) er
 	// There is a race condition where the cilium resources can change
 	// while we try to restart them, which fails with:
 	// the object has been modified; please apply your changes to the latest version and try again
-	if err := control.RetryFor(3, func() error {
+	attempts := 3
+	if err := control.RetryFor(attempts, func() error {
 		if err := client.RestartDeployment(ctx, "cilium-operator", "kube-system"); err != nil {
 			return fmt.Errorf("failed to restart cilium-operator deployment: %w", err)
 		}
 		return nil
 	}); err != nil {
-		return err
+		return fmt.Errorf("failed to restart cilium-operator deployment after %d attempts: %w", attempts, err)
 	}
 
-	if err := control.RetryFor(3, func() error {
+	if err := control.RetryFor(attempts, func() error {
 		if err := client.RestartDaemonset(ctx, "cilium", "kube-system"); err != nil {
 			return fmt.Errorf("failed to restart cilium daemonset: %w", err)
 		}
 		return nil
 	}); err != nil {
-		return err
+		return fmt.Errorf("failed to restart cilium daemonset after %d attempts: %w", attempts, err)
 	}
 
 	return nil
--- a/src/k8s/pkg/component/gateway.go
+++ b/src/k8s/pkg/component/gateway.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/canonical/k8s/pkg/k8sd/types"
 	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/k8s/pkg/utils/control"
 	"github.com/canonical/k8s/pkg/utils/k8s"
 	"github.com/canonical/k8s/pkg/utils/vals"
 )
@@ -43,12 +44,25 @@ func UpdateGatewayComponent(ctx context.Context, s snap.Snap, isRefresh bool) er
 		return fmt.Errorf("failed to create kubernetes client: %w", err)
 	}
 
-	if err := client.RestartDeployment(ctx, "cilium-operator", "kube-system"); err != nil {
-		return fmt.Errorf("failed to restart cilium-operator deployment: %w", err)
+	// There is a race condition where the cilium resources can change
+	// while we try to restart them, which fails with:
+	// the object has been modified; please apply your changes to the latest version and try again
+	if err := control.RetryFor(3, func() error {
+		if err := client.RestartDeployment(ctx, "cilium-operator", "kube-system"); err != nil {
+			return fmt.Errorf("failed to restart cilium-operator deployment: %w", err)
+		}
+		return nil
+	}); err != nil {
+		return err
 	}
 
-	if err := client.RestartDaemonset(ctx, "cilium", "kube-system"); err != nil {
-		return fmt.Errorf("failed to restart cilium daemonset: %w", err)
+	if err := control.RetryFor(3, func() error {
+		if err := client.RestartDaemonset(ctx, "cilium", "kube-system"); err != nil {
+			return fmt.Errorf("failed to restart cilium daemonset: %w", err)
+		}
+		return nil
+	}); err != nil {
+		return err
 	}
 
 	return nil
--- a/src/k8s/pkg/k8sd/api/cluster_bootstrap.go
+++ b/src/k8s/pkg/k8sd/api/cluster_bootstrap.go
@@ -40,7 +40,7 @@ func postClusterBootstrap(m *microcluster.MicroCluster, s *state.State, r *http.
 	}
 
 	// Set timeout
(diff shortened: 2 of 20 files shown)


src/k8s/pkg/k8sd/controllers/node_configuration_test.go
```diff
@@ -0,0 +1,130 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	g := NewWithT(t)
+
+	dir := t.TempDir()
+
+	s := &mock.Snap{
+		Mock: mock.Mock{
+			KubernetesPKIDir:    path.Join(dir, "pki"),
```
bschimke95, line 33: We keep setting up these standard dirs and IDs. Maybe create a `mock.NewDefaultMock(snapDir string)` method?

src/k8s/pkg/k8sd/controllers/node_configuration_test.go
```diff
@@ -0,0 +1,130 @@
...
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+func TestConfigPropagation(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
bschimke95, line 24: You don't need to cancel this context (here and in other places).

src/k8s/pkg/utils/k8s/configmap_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	clientset := fake.NewSimpleClientset(existingObjs...)
+	client := &Client{Interface: clientset}
+
+	updateData := map[string]string{
+		"existing-key": "change-value",
+		"new-key":      "new-value",
+	}
+	cm, err := client.UpdateConfigMap(ctx, "kube-system", "test-config", updateData)
+	g.Expect(err).To(BeNil())
```
bschimke95, line 107: nit: `Not(HaveOccured())`

bschimke95 (commented):
looks great, just some small nits

src/k8s/pkg/k8sd/controllers/node_configuration_test.go
berkayoz: Thanks for the review! I've checked tests in general and what kind of mocks required by a unit test varied a lot so just went with the simple approach. I suggest creating a separate card to improve/standardize mocks for all of the tests in this regard.

bschimke95: fair

bschimke95 (approved):
LGTM

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -40,7 +40,7 @@ func postClusterBootstrap(m *microcluster.MicroCluster, s *state.State, r *http.
 	}
 
 	// Set timeout
-	timeout := utils.TimeoutFromCtx(s.Context, 30*time.Second)
+	timeout := utils.TimeoutFromCtx(s.Context, 60*time.Second)
```
bschimke95, line 43: That is not the correct context. Slipped through in the review. cc @louiseschmidtgen 
It should be `r.Context()`. This context includes a timeout, so the default value is not used. Could you please also check that the other occurrences of `TimeoutFromCtx` use the correct context? Thanks!

src/k8s/pkg/utils/k8s/configmap_test.go
```diff
@@ -0,0 +1,107 @@
...
+	})
+
+	defer watcher.Stop()
+
+	for i, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+
+			watcher.Add(tc.configmap)
+			time.Sleep(100 * time.Millisecond)
```
neoaggelos, line 72: Is this to make sure that the watch function has executed? If so, don't do a sleep, but rather use a channel.

```
doneCh := make(chan struct{})

// in the goroutine
go watch...(...) { 
   doneCh <-struct{}{}
}()

// then outside the loop, replace the time.Sleep with

select {
case <-doneCh:
case <-time.After(time.Second):
  t.Fatal("Timed out waiting for watch to complete")
}
```

tests/integration/tests/test_config_propagation.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import List
+
+import pytest
+from test_util import harness, util
+
+LOG = logging.getLogger(__name__)
+
+
+@pytest.mark.node_count(3)
+def test_config_propagation(instances: List[harness.Instance]):
```
neoaggelos, line 14: note that this test will also work without our controller (since nodes get the right dns config when they join).

make sure to change the config then validate the new config is propagated (e.g. edit the cluster-domain)

neoaggelos (changes requested):

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
berkayoz: @bschimke95 I've changed the context the `r.Context()` but the default value still seems to be used as I got test failures, but  bumping the default timeout to 45 seconds gets me a passing CI. I don't have more background on this, would appreciate if you could take a look into it as well.

src/k8s/pkg/component/gateway.go
```diff
@@ -43,12 +44,25 @@ func UpdateGatewayComponent(ctx context.Context, s snap.Snap, isRefresh bool) er
...
+	// There is a race condition where the cilium resources can change
+	// while we try to restart them, which fails with:
+	// the object has been modified; please apply your changes to the latest version and try again
+	if err := control.RetryFor(3, func() error {
+		if err := client.RestartDeployment(ctx, "cilium-operator", "kube-system"); err != nil {
+			return fmt.Errorf("failed to restart cilium-operator deployment: %w", err)
+		}
+		return nil
+	}); err != nil {
+		return err
```
neoaggelos, line 56: wrap the error here

src/k8s/pkg/component/gateway.go
```diff
@@ -43,12 +44,25 @@ func UpdateGatewayComponent(ctx context.Context, s snap.Snap, isRefresh bool) er
...
 
-	if err := client.RestartDaemonset(ctx, "cilium", "kube-system"); err != nil {
-		return fmt.Errorf("failed to restart cilium daemonset: %w", err)
+	if err := control.RetryFor(3, func() error {
+		if err := client.RestartDaemonset(ctx, "cilium", "kube-system"); err != nil {
+			return fmt.Errorf("failed to restart cilium daemonset: %w", err)
+		}
+		return nil
+	}); err != nil {
+		return err
```
neoaggelos, line 65: also here, something like `failed to do foo after N attempts: %w`

neoaggelos (approved):
LGTM, very nice work @berkayoz! Left some final comments

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -13,7 +14,20 @@ func onStart(s *state.State) error {
 	snap := snap.SnapFromContext(s.Context)
 
 	configController := controllers.NewNodeConfigurationController(snap, func(ctx context.Context) *k8s.Client {
-		return k8s.RetryNewClient(ctx, snap.KubernetesNodeRESTClientGetter("kube-system"))
+		for {
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(3 * time.Second):
+			default:
```
neoaggelos, line 22: related, since we are doing fixes on this PR. the default case here will ignore the `time.After` above. remove it

src/k8s/pkg/k8sd/controllers/node_configuration.go
neoaggelos, line 1: Also remove the `default:` case from line 36 in the controller's loop here

PR #280, opened 2024-03-28
Fix joining under high load

### LINKED ISSUE #271: Failed Cluster Join Attempt Blocks Worker from Joining
We have encountered issues where the leader node, particularly when its CPU is highly utilized, fails to respond to a joining worker after a bootstrap request. Therefore, the joining worker believes it has not completed the bootstrap process.

Attempting to join again results in the leader node indicating that a worker with the same identifier already exists in the cluster, thereby blocking the joining process for the affected worker.

Logging from the environment indicates that the worker attempted to bootstrap with a valid token. However, the server did not respond within an acceptable time frame:
```
        body={"type":"error","status":"","status_code":0,"operation":"","error_code":500,"error":"failed to join k8sd cluster as worker: HTTP request for worker node info failed: failed to retrieve list of known kube-apiserver endpoints: failed to get endpoints for kubernetes service: Get \"https://127.0.0.1:6443/api/v1/namespaces/default/endpoints/kubernetes\": dial tcp 127.0.0.1:6443: connect: connection refused","metadata":null}
```
Subsequent attempts to contact the join endpoint fail because the server thinks that this worker has already joined the cluster:
```
charms.contextual_status.ReconcilerError: Found expected exception: Error status 500
        method=POST
        endpoint=/1.0/k8sd/cluster/join
        reason=Internal Server Error
        body={"type":"error","status":"","status_code":0,"operation":"","error_code":500,"error":"failed to join k8sd cluster as worker: Failed to initialize local remote entry: A remote with name \"juju-3cb80e-2\" already exists","metadata":null}
```

bschimke95:
Makes the k8s-snap joining process more robust and fixes #271 
See integration test runs here: https://github.com/canonical/k8s-operator/pull/54

This PR:
 * Ensures that the `kubernetes API` server is ready
 * Let the worker config request wait for the right context and not a fixed timeout
 * Adds a retry to the k8s endpoint request.

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -50,10 +50,13 @@ func onBootstrapWorkerNode(s *state.State, encodedToken string) error {
 	}
 
 	// TODO(neoaggelos): figure out how to use the microcluster client instead
-
+	timeout := 30 * time.Second
+	if deadline, set := s.Context.Deadline(); set {
+		timeout = time.Until(deadline)
+	}
 	// create an HTTP client that ignores https
 	httpClient := &http.Client{
-		Timeout: 10 * time.Second,
+		Timeout: timeout,
 		Transport: &http.Transport{
 			TLSClientConfig: &tls.Config{
 				InsecureSkipVerify: true,
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -10,7 +10,6 @@ import (
 	"net"
 	"net/http"
 	"path"
-	"time"
 
 	apiv1 "github.com/canonical/k8s/api/v1"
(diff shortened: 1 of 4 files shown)


bschimke95:
The integration tests pass now for >5 times in a row. https://github.com/canonical/k8s-operator/actions/runs/8432630507?pr=54

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -50,10 +50,13 @@ func onBootstrapWorkerNode(s *state.State, encodedToken string) error {
 	}
 
 	// TODO(neoaggelos): figure out how to use the microcluster client instead
-
+	timeout := 30 * time.Second
+	if deadline, set := s.Context.Deadline(); set {
```
berkayoz, line 54: Is `s.Context.Deadline()` ever set on this call? If not we will again use a fixed timeout. Should we adjust this to handle a cancelled context like we've done yesterday?

berkayoz (approved):
Looks good overall, small question.

PR #284, opened 2024-04-02
Node address port fixes

neoaggelos:
### Summary

Fixes around the node microcluster address and port

### Changes

- Remove `ListenPort` config from the microcluster app configuration, and deprecate the `--port` k8sd argument. This port is the default HTTPS port for the microcluster app, before the node is bootstrapped. After bootstrapping, this is unused, and the node address specified is used instead.
- Automatically add the default port (6400), if only the node address is specified during bootstrap (e.g. via `k8s bootstrap --address 1.1.1.1`)

### DIFF
--- a/k8s/args/k8sd
+++ b/k8s/args/k8sd
@@ -1,2 +1,1 @@
---port=6400
 --state-dir=${SNAP_COMMON}/var/lib/k8sd/state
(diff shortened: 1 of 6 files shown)


berkayoz (approved):
LGTM

PR #285, opened 2024-04-02
add datastore to k8s status

louiseschmidtgen:
This PR grabs the datastore type and the datastore-url from the cluster config to display them in the 'k8s status' command.

For external we now have (relevant snippet of the output):
```
status: ready
high-availability: no
datastore:
  datastore: external
  datastore-url: I-am-a-postgres-url
```
For internal we get:
```
status: ready
high-availability: yes
datastore:
  datastore: k8s-dqlite
  voter-nodes:
    - 192.168.0.1
    - 192.168.0.2
    - 192.168.0.3
  standby-nodes: none
  spare-nodes: none
```

### DIFF
--- a/src/k8s/api/v1/cluster_config.go
+++ b/src/k8s/api/v1/cluster_config.go
@@ -27,6 +27,7 @@ type UserFacingClusterConfig struct {
 	LocalStorage  *LocalStorageConfig  `json:"local-storage,omitempty" yaml:"local-storage,omitempty"`
 	Gateway       *GatewayConfig       `json:"gateway,omitempty" yaml:"gateway,omitempty"`
 	MetricsServer *MetricsServerConfig `json:"metrics-server,omitempty" yaml:"metrics-server,omitempty"`
+	APIServer     *APIServerConfig     `json:"apiserver,omitempty" yaml:"apiserver,omitempty"`
 }
 
 type DNSConfig struct {
@@ -65,6 +66,11 @@ type NetworkConfig struct {
 	Enabled *bool `json:"enabled,omitempty" yaml:"enabled"`
 }
 
+type APIServerConfig struct {
+	Datastore    string `json:"datastore,omitempty" yaml:"datastore"`
+	DatastoreURL string `json:"datastore-url,omitempty" yaml:"datastore-url"`
+}
+
 type GatewayConfig struct {
 	Enabled *bool `json:"enabled,omitempty" yaml:"enabled"`
 }
@@ -82,6 +88,14 @@ func (c UserFacingClusterConfig) String() string {
 }
 
 func (c NetworkConfig) String() string {
+	b, err := yaml.Marshal(c)
+	if err != nil {
+		return fmt.Sprintf("%#v\n", c)
+	}
+	return string(b)
+}
+
+func (c APIServerConfig) String() string {
 	b, err := yaml.Marshal(c)
 	if err != nil {
 		return fmt.Sprintf("%#v\n", c)
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -112,26 +112,22 @@ func (c ClusterStatus) HaClusterFormed() bool {
 	return voters > 2
 }
 
-// TODO: Print k8s version. However, multiple nodes can run different version, so we would need to query all nodes.
-func (c ClusterStatus) String() string {
+func (c ClusterStatus) datastoreToString() string {
 	result := strings.Builder{}
 
-	if c.Ready {
-		result.WriteString("status: ready")
-	} else {
-		result.WriteString("status: not ready")
+	// Datastore
+	if c.Config.APIServer != nil && c.Config.APIServer.Datastore != "" {
+		result.WriteString(fmt.Sprintf("  datastore: %s\n", c.Config.APIServer.Datastore))
+		// Datastore URL for external only
+		if c.Config.APIServer.Datastore == "external" {
+			if c.Config.APIServer.DatastoreURL != "" {
+				result.WriteString(fmt.Sprintf("  datastore-url: %s\n", c.Config.APIServer.DatastoreURL))
+			}
+			return result.String()
+		}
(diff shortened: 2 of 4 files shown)


src/k8s/api/v1/cluster_config.go
```diff
@@ -65,6 +66,11 @@ type NetworkConfig struct {
 	Enabled *bool `json:"enabled,omitempty" yaml:"enabled"`
 }
 
+type APIServerConfig struct {
+	Datastore    string `json:"datastore,omitempty" yaml:"datastore"`
+	DatastoreURL string `json:"datastore-url,omitempty" yaml:"datastore-url"`
+}
+
```
neoaggelos, line 73: Note that the `UserFacingClusterConfig` struct (message, in RPC terms) is used in multiple places, like for example `k8s get/set`. We do not want this as it causes confusion:

- we can now pass such a struct to `k8s set`, but it will be silently ignored. 
- `k8s get` will give me back an APIServerConfig that will always be empty.

Please instead only add the field to `type ClusterStatus struct` instead.

neoaggelos (changes requested):
Going in the right direction, however the API needs changes.

src/k8s/api/v1/types.go
```diff
@@ -170,6 +180,35 @@ func (c ClusterStatus) String() string {
...
+	if c.Ready {
+		result.WriteString("status: ready")
+	} else {
+		result.WriteString("status: not ready")
+	}
+	result.WriteString("\n")
+
+	// High availability
+	result.WriteString("high-availability: ")
+	if c.HaClusterFormed() {
```
bschimke95, line 200: nit: can you make `HaClusterFormed` also private - we don't need it elsewhere and I probably made it public by mistake.

src/k8s/pkg/utils/database.go
```diff
@@ -30,6 +30,20 @@ func GetClusterConfig(ctx context.Context, state *state.State) (types.ClusterCon
 	return clusterConfig, nil
 }
 
+// GetUserFacingClusterConfig returns the public cluster config.
```
bschimke95, line 33: Please update

bschimke95 (commented):
good work, did a first pass and added some comments.

src/k8s/api/v1/types_test.go
```diff
@@ -117,17 +117,47 @@ func TestString(t *testing.T) {
...
+					Network: &NetworkConfig{Enabled: vals.Pointer(true)},
+					DNS:     &DNSConfig{Enabled: vals.Pointer(true)},
+				},
+				APIServer: &APIServerConfig{Datastore: "external", DatastoreURL: "I-am-a-postgres-url"},
+			},
+			expectedOutput: `status: ready
+high-availability: no
+datastore:
+  datastore: external
+  datastore-url: I-am-a-postgres-url
```
neoaggelos, line 159: same here, this is a bit redundant

neoaggelos (commented):
Looks better, but left some more comments, especially around naming and code structure

src/k8s/api/v1/types.go
```diff
@@ -93,16 +93,30 @@ type NodeStatus struct {
...
+	ExternalURL string `json:"datastore-url,omitempty"`
+}
+
+func (c Datastore) String() string {
+	b, err := yaml.Marshal(c)
+	if err != nil {
+		return fmt.Sprintf("%#v\n", c)
+	}
+	return string(b)
+}
```
neoaggelos, line 107: is this used somewhere? or just leftover from previous iteration?

src/k8s/api/v1/types.go
```diff
@@ -93,16 +93,30 @@ type NodeStatus struct {
...
 // ClusterStatus holds information about the cluster, e.g. its current members
 type ClusterStatus struct {
 	// Ready is true if at least one node in the cluster is in READY state.
-	Ready   bool                    `json:"ready,omitempty"`
-	Members []NodeStatus            `json:"members,omitempty"`
-	Config  UserFacingClusterConfig `json:"config,omitempty"`
+	Ready     bool                    `json:"ready,omitempty"`
+	Members   []NodeStatus            `json:"members,omitempty"`
+	Config    UserFacingClusterConfig `json:"config,omitempty"`
+	Datastore *Datastore              `json:"datastore,omitempty"`
```
neoaggelos, line 115: no need for a `*Datastore`, just `Datastore` should be ok

src/k8s/api/v1/types.go
```diff
@@ -112,26 +126,22 @@ func (c ClusterStatus) HaClusterFormed() bool {
...
-	}
-	result.WriteString("\n")
-
-	result.WriteString("high-availability: ")
-	if c.HaClusterFormed() {
-		result.WriteString("yes")
-	} else {
-		result.WriteString("no")
+	// Datastore
+	if c.Datastore != nil && c.Datastore.Type != "" {
```
neoaggelos, line 133: avoiding `*Datastore` will also simplify this

src/k8s/pkg/k8sd/api/cluster.go
```diff
@@ -25,6 +25,15 @@ func (e *Endpoints) getClusterStatus(s *state.State, r *http.Request) response.R
 		return response.InternalError(fmt.Errorf("failed to get user-facing cluster config: %w", err))
 	}
 
+	clusterConfig, err := utils.GetClusterConfig(s.Context, s)
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to get cluster config: %w", err))
+	}
+	datastoreConfig := apiv1.Datastore{
+		Type:        *clusterConfig.Datastore.Type,
+		ExternalURL: *clusterConfig.Datastore.ExternalURL,
```
neoaggelos, line 34:
```suggestion
		Type:        clusterConfig.Datastore.GetType(),
		ExternalURL: clusterConfig.Datastore.GetExternalURL(),
```

neoaggelos (approved):

PR #286, opened 2024-04-03
Add cilium-dbg cleanup to snap removal hook

berkayoz:
Add cilium-dbg cleanup to snap removal hook for proper cleanup of node


k8s/lib.sh
```diff
@@ -69,6 +69,9 @@ k8s::remove::network() {
   do
     tc filter del dev $default_interface ${d} || true
   done
+
+  rm -rf /var/run/cilium/cilium.pid
+  "${SNAP}/usr/bin/cilium-dbg" cleanup --all-state --force
```
bschimke95, line 74: What happens if this fails? Will the remove-hook continue or will it abort the uninstall process?

berkayoz: Good catch, thanks!

neoaggelos (commented):
Looks great, also needs a bump in the cilium rock tag (when that is ready)

neoaggelos (approved):

PR #287, opened 2024-04-03
Nightly Latest/Edge K8s Snap Workflow

louiseschmidtgen:
## Description
This Workflow is triggered by a daily cronjob. The integration test is run against the latest edge k8s snap.

See the results of the nightly run here: https://github.com/canonical/k8s-snap/actions/runs/8604245353/job/23577911227

- Tests pass against Latest/Edge Snap on Ubuntu 22.04, 20.04.
- Integration test fails against Beta Snap (expected: running an older version).
- Fails on the arm64 runner (expected for now).

Should we disable any tests in the matrix while they're broken?


.github/workflows/nightly-test.yaml
```diff
@@ -0,0 +1,61 @@
...
+  schedule:
+    - cron: '0 0 * * *'  # Runs every midnight
+
+jobs:
+  test-integration:
+    name: Integration Test ${{ matrix.os }} ${{ matrix.runner == 'Ubuntu_ARM64_4C_16G_01' && 'arm64' || 'amd64' }}
+    strategy:
+      matrix:
+        os: ["ubuntu:20.04", "ubuntu:22.04"]
+        runner: ["ubuntu-20.04", "Ubuntu_ARM64_4C_16G_01"]
```
addyess, line 17: ok, so this wasn't immediately clear to me until i spent some time thinking about it

how about this

```suggestion
        os: ["ubuntu:20.04", "ubuntu:22.04"]
        arch: ["amd64", "arm64"]
```


and then when you use apply the runs on, use your github expression:

```yaml
runs-on: ${{ matrix.arch == "arm64" && "Ubuntu_ARM64_4C_16G_01" || "ubuntu-20.04" }}
```
```

File: .github/workflows/nightly-test.yaml [hunk 2]
```diff
@@ -0,0 +1,61 @@
+name: Nightly Latest/Edge Tests
+
+on:
+  # TODO: remove, this is just testing
+  push: 
+    branches: 
+      - KU-91/nightly-runs
```
addyess, line 7: Can we run this on release branches too??

```suggestion
    branches: 
      - KU-91/nightly-runs
      - main
      - 'release-[0-9]+.[0-9]+'
```

.github/workflows/nightly-test.yaml
```diff
@@ -0,0 +1,61 @@
...
+        run: |
+          pip3 install tox==4.13
+      - name: Install lxd
+        run: |
+          sudo snap refresh lxd --channel 5.19/stable
+          sudo lxd init --auto
+          sudo usermod --append --groups lxd $USER
+          sg lxd -c 'lxc version'
+      - name: Create build directory
+        run: mkdir -p build
```
addyess, line 50: in order to know which snap to fetch for this branch you can use this:

```suggestion
        run: mkdir -p build
      - name: Determine Release branch
        id: snap-release
        env:
          BRANCH: ${{ github.base_ref || github.ref }}
        run: |
          BRANCH=${BRANCH#refs/heads/}  # strip off refs/heads/ if it exists
          if [[ "${BRANCH}" == "main" ]]; then
            echo "release=latest/edge" >> "$GITHUB_OUTPUT"
          elif [[ "${BRANCH}" =~ "^release-[0-9]+\.[0-9]+$" ]]; then
            echo "release=${BRANCH:-8}/edge" >> "$GITHUB_OUTPUT"
          else
            echo "Failed to determine matching strict branch for ${BRANCH}"
            echo "release=edge" >> $GITHUB_OUTPUT
          fi     
```

.github/workflows/nightly-test.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Install latest/edge k8s snap
+        run: |
+          cd build
+          snap download k8s --channel=edge --basename k8s
```
addyess, line 54: i'd LOVE if this channel was based on which branch this was running on

```suggestion
          snap download k8s --channel=${release} --basename k8s
```

addyess (changes requested):
It'd be neat to see this run on multiple branches.  Don't forget to remove the stray branch before merging.  I know it's just there for testing the workflow

.github/workflows/nightly-test.yaml
louiseschmidtgen: We sure can, I don't think we want to run this every time we push to the main branch though? We'd be running the integration test twice?

.github/workflows/nightly-test.yaml
louiseschmidtgen: I was thinking it is looking a bit scary in the name. I like this much better!

.github/workflows/nightly-test.yaml
louiseschmidtgen: Neat!

.github/workflows/nightly-test.yaml
addyess: OH RIGHT not on push here... i was hoping to run the cron on every branch that matches main or release-1xx

.github/workflows/nightly-test.yaml
louiseschmidtgen: Maybe compromise on matrix for now? `releases: ["latest/edge", "1.30-classic/beta"]`?

.github/workflows/nightly-test.yaml
```diff
@@ -0,0 +1,47 @@
...
+          pip3 install tox==4.13
+      - name: Install lxd
+        run: |
+          sudo snap refresh lxd --channel 5.19/stable
+          sudo lxd init --auto
+          sudo usermod --append --groups lxd $USER
+          sg lxd -c 'lxc version'
+      - name: Create build directory
+        run: mkdir -p build
+      - name: Install latest/edge k8s snap
```
addyess, line 37: oopsie
```suggestion
      - name: Install $${ matrix.releases }} k8s snap
```

.github/workflows/nightly-test.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - cron: '0 0 * * *'  # Runs every midnight
+
+jobs:
+  test-integration:
+    name: Integration Test ${{ matrix.os }} ${{ matrix.arch }} ${{ matrix.releases }}
+    strategy:
+      matrix:
+        os: ["ubuntu:20.04", "ubuntu:22.04"]
+        arch: ["amd64", "arm64"]
+        releases: ["latest/edge", "1.30-classic/beta"]
```
addyess, line 14: lets get this going with one branch first .. reduce this to a list of one "latest/edge"

addyess (changes requested):
Tiny nits

addyess (approved):
LGTM

PR #289, opened 2024-04-03
transfer leadership before removing the leader node

bschimke95:
In a two node dqlite cluster, one is leader/voter and one is a spare node.
If the leader node should be removed, the leadership needs to be transferred to the remaining node first. 
The `go-dqlite` client provides a Transfer function that transfers the leadership to another node. However, this requires the other node to either be a voter or standby, not spare.
To work around this limitation, this PR:

1. assign the spare node the standby role
1. transfer leadership to the new standby node
1. create a new client to the new leader
1. remove the old leader node from the cluster

This approach is a bit problematic for several reasons. If a spare node is promoted to standby it needs to copy the whole database. For large dbs this might take significant amount of time. 

So, this implementation serves as a workaround. A proper fix - have the non-leader node as standby by default - will be delivered by in https://github.com/canonical/go-dqlite/issues/293

### DIFF
--- a/src/k8s/pkg/client/dqlite/remove.go
+++ b/src/k8s/pkg/client/dqlite/remove.go
@@ -3,6 +3,9 @@ package dqlite
 import (
 	"context"
 	"fmt"
+	"time"
+
+	"github.com/canonical/k8s/pkg/utils/control"
 )
 
 func (c *Client) RemoveNodeByAddress(ctx context.Context, address string) error {
@@ -18,6 +21,7 @@ func (c *Client) RemoveNodeByAddress(ctx context.Context, address string) error 
 	var (
 		memberExists, clusterHasOtherVoters bool
 		memberToRemove                      NodeInfo
+		failOverMemberID                    *uint64
 	)
 	for _, member := range members {
 		switch {
@@ -27,6 +31,12 @@ func (c *Client) RemoveNodeByAddress(ctx context.Context, address string) error 
 
 		case member.Address != address && member.Role == Voter:
 			clusterHasOtherVoters = true
+
+		case member.Address != address && member.Role == Spare:
+			// Mark as fail-over node. This is only used in a two node setup,
+			// where the leader node is removed. The fail-over node will be
+			// promoted to leader before the existing leader is removed.
+			failOverMemberID = &member.ID
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


src/k8s/pkg/client/dqlite/remove.go
```diff
@@ -27,21 +31,49 @@ func (c *Client) RemoveNodeByAddress(ctx context.Context, address string) error
...
+		if err != nil {
+			return fmt.Errorf("failed to create dqlite client: %w", err)
+		}
 	}
 
-	return nil
+	// Remove the node from the cluster. Retry as the leadership transfer might still be in progress.
+	// For a large database this might take some time.
+	return control.RetryFor(10, func() error {
+		if err := client.Remove(ctx, memberToRemove.ID); err != nil {
```
addyess, line 73: does this `client.Remove` fail after a specific amount of time?  I'm trying to count the seconds this loop would run

* 10 retries
* undetermined number of seconds for the `client.Remove`
* 5 seconds of sleep back off

could be as QUICK as 0 seconds on success, to heat death of the universe if `client.Remove` never times out

was the intent instead to retry 10 times with 5 seconds between calls with some maximum timeout of 60s?

addyess (commented):
Mostly questions that i think may need an answer.  Otherwise i don't have enough information to offer a review

src/k8s/pkg/client/dqlite/remove.go
bschimke95: Good point, the timeout of `client.Remove` is not something we have control over. 
However, we provide a context that, if cancelled, will cause the `client.Remove` to stop.

However, the `RetryFor` function did not check if the context was already cancelled before retrying which causes the function the hang as it retries again.

I added a fix for that and a unittests to verify the behavior.

src/k8s/pkg/client/dqlite/remove.go
```diff
@@ -27,21 +31,48 @@ func (c *Client) RemoveNodeByAddress(ctx context.Context, address string) error
...
+			return fmt.Errorf("failed to create dqlite client: %w", err)
+		}
 	}
 
-	return nil
+	// Remove the node from the cluster. Retry as the leadership transfer might still be in progress.
+	// For a large database this might take some time.
+	return control.RetryFor(ctx, 10, func() error {
+		if err := client.Remove(ctx, memberToRemove.ID); err != nil {
+			time.Sleep(5 * time.Second)
```
neoaggelos, line 73: this can be dangerous. is this backoff? if so, it should be specified properly in `control.RetryFor`, and `control.RetryFor` should use `<-time.After(timeout)` instead. For some more background consider the difference between:

```go
select {
case <-ctx.Done():
  // do stuff
case <-time.After(5 * time.Second):
}
```

and 

```go
time.Sleep(5 * time.Second)
<-ctx.Done()

// do stuff
```

in the second case, the `time.Sleep` takes control away from our code, and a context cancellation will block until the sleep completes.

neoaggelos (changes requested):
Code looks good, but left a comment regarding `control.RetryFor`

src/k8s/pkg/client/dqlite/remove_test.go
```diff
@@ -48,17 +48,23 @@ func TestRemoveNodeByAddress(t *testing.T) {
...
-			g.Expect(client.RemoveNodeByAddress(ctx, memberToRemove)).ToNot(BeNil())
+			// Removing the last voter should succeed and leadership should be transfered.
+			g.Expect(client.RemoveNodeByAddress(ctx, memberToRemove.Address)).To(Succeed())
 
 			members, err = client.ListMembers(ctx)
 			g.Expect(err).To(BeNil())
-			g.Expect(members).To(HaveLen(2))
+			g.Expect(members).To(HaveLen(1))
+			g.Expect(members[0].Role == dqlite.Voter)
+			g.Expect(members[0].Address).ToNot(Equal(memberToRemove.Address))
```
louiseschmidtgen, line 67: Do we have a test for edge cases like `remove the last node in the cluster`?

louiseschmidtgen (commented):
Great Work! This will be great to have for all those 2 node clusters.

src/k8s/pkg/client/dqlite/remove_test.go
bschimke95: Added a test that this fails.

src/k8s/pkg/client/dqlite/remove.go
bschimke95: Good point, moved the delay to the `RetryFor` function.

src/k8s/pkg/utils/control/retry.go
```diff
@@ -1,10 +1,22 @@
 package control
 
-func RetryFor(retryCount int, retryFunc func() error) error {
+import (
+	"context"
+	"time"
+)
+
+// RetryFor will retry a given function for the given amount of times.
+// RetryFor will wait for delayBetweenRetry between retries.
```
neoaggelos, line 9:
```suggestion
// RetryFor will wait for backoff between retries.
```

src/k8s/pkg/utils/control/retry_test.go
```diff
@@ -0,0 +1,64 @@
...
+	})
+
+	t.Run("Retry fails with context cancellation", func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+		defer cancel()
+
+		retryCount := 3
+
+		err := RetryFor(ctx, retryCount, time.Second, func() error {
+			time.Sleep(200 * time.Millisecond)
```
neoaggelos, line 41: i see why this is here, but in general we should avoid having `time.Sleep()` or unit tests that are based on time.

neoaggelos (approved):

src/k8s/pkg/utils/control/retry_test.go
bschimke95: ack

PR #290, opened 2024-04-03
Extend sbom archive with rock repos

neoaggelos:
### Summary

Extend the SBOM archive with links and commit hashes of the rocks we use


bschimke95 (approved):
LGTM

PR #291, opened 2024-04-04
Wait for microcluster to be running before starting node config controller

neoaggelos:
### Summary

Do not start the node config controller before the microcluster node is running (i.e. the database is open, see https://github.com/canonical/microcluster/blob/5043bc984c83e9ecfdba1de731ec05678b166f13/internal/rest/rest.go#L188

### Changes

- Add a `sync.WaitGroup` in `*app.App`, which is marked as complete once the database is open.
- Move the node config controller creation to `*app.App`, and accept the `waitReady` func for the wait group
- Move the node config controller create k8s client inside the package, to simplify how dependencies are passed.

### DIFF
--- a/src/k8s/pkg/k8sd/controllers/node_configuration.go
+++ b/src/k8s/pkg/k8sd/controllers/node_configuration.go
@@ -14,19 +14,43 @@ import (
 )
 
 type NodeConfigurationController struct {
-	snap            snap.Snap
-	createK8sClient func(ctx context.Context) *k8s.Client
+	snap         snap.Snap
+	waitReady    func()
+	newK8sClient func() (*k8s.Client, error)
 }
 
-func NewNodeConfigurationController(snap snap.Snap, createK8sClient func(ctx context.Context) *k8s.Client) *NodeConfigurationController {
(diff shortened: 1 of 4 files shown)


src/k8s/pkg/k8sd/controllers/node_configuration.go
```diff
@@ -14,24 +14,48 @@ import (
...
+		}
 	}
 }
 
 func (c *NodeConfigurationController) Run(ctx context.Context) {
-	client := c.createK8sClient(ctx)
+	// wait for microcluster node to be ready
+	c.waitReady()
+
+	client, err := c.retryNewK8sClient(ctx)
```
bschimke95, line 49: why does the k8s client might fail if the cluster is ready?

bschimke95 (approved):
LGTM, only a small question

src/k8s/pkg/k8sd/controllers/node_configuration.go
neoaggelos: database is open, but could be that we hit a race condition where kubeconfig files are not yet in place

PR #293, opened 2024-04-04
pki: handle both RSA PRIVATE KEY and PRIVATE KEY

neoaggelos:
### Summary

Handle both RSA PRIVATE KEY (PKCS#1) and PRIVATE KEY (PKCS#8) blocks for rsa keys. We still require the presence of an RSA private key.

@mateoflorido


bschimke95 (approved):
LGTM, maybe add a test case?

PR #296, opened 2024-04-04
Add (Datastore).ToKubeAPIServerArguments() helper

neoaggelos:
### Summary

Move the generation of kube-apiserver arguments from the datastore configuration into `pkg/k8sd/types`, and add focused unit tests

### Changes

- Add `(Datastore).ToKubeAPIServerArguments()`
- Use it in `setup.KubeAPIServer()`
- Update callers and tests
- Add unit tests for ToKubeAPIServerArguments()

### DIFF
--- a/src/k8s/pkg/k8sd/types/cluster_config_datastore.go
+++ b/src/k8s/pkg/k8sd/types/cluster_config_datastore.go
@@ -1,5 +1,10 @@
 package types
 
+import (
+	"fmt"
+	"path"
+)
+
 type Datastore struct {
 	Type *string `json:"type,omitempty"`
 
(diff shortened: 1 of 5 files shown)


src/k8s/pkg/k8sd/types/cluster_config_datastore.go
```diff
@@ -24,3 +29,45 @@ func (c Datastore) GetExternalClientKey() string  { return getField(c.ExternalCl
 func (c Datastore) Empty() bool {
 	return c.Type == nil && c.K8sDqlitePort == nil && c.K8sDqliteCert == nil && c.K8sDqliteKey == nil && c.ExternalURL == nil && c.ExternalCACert == nil && c.ExternalClientCert == nil && c.ExternalClientKey == nil
 }
+
+// DatastoreSnapProvider is to avoid circular dependency for snap.Snap in Datastore.ToKubeAPIServerArguments()
```
bschimke95, line 33: Typo DatastorePathsProvider

bschimke95 (approved):
LGTM, great cleanup. Only a minor typo

src/k8s/pkg/k8sd/types/cluster_config_datastore.go
```diff
@@ -24,3 +29,45 @@ func (c Datastore) GetExternalClientKey() string  { return getField(c.ExternalCl
...
+		updateArgs = make(map[string]string)
+		deleteArgs []string
+	)
+
+	switch c.GetType() {
+	case "k8s-dqlite":
+		updateArgs["--etcd-servers"] = fmt.Sprintf("unix://%s", path.Join(p.K8sDqliteStateDir(), "k8s-dqlite.sock"))
+		deleteArgs = []string{"--etcd-cafile", "--etcd-certfile", "--etcd-keyfile"}
+	case "external":
+		updateArgs["--etcd-servers"] = c.GetExternalURL()
```
addyess, line 52: we should keep in mind that this is really `ExternalURLs` -- `etcd-servers` is a comma separated list of  servers.  I sort of wish the data structure was a `[]string` rather than `string`.

> List of etcd servers to connect with (scheme://ip:port), comma separated.

The endpoints can be https or http of course, so the cert fields are only necessary for `https`

src/k8s/pkg/k8sd/types/cluster_config_datastore.go
```diff
@@ -24,3 +29,45 @@ func (c Datastore) GetExternalClientKey() string  { return getField(c.ExternalCl
...
+// according the datastore configuration.
+func (c Datastore) ToKubeAPIServerArguments(p DatastorePathsProvider) (map[string]string, []string) {
+	var (
+		updateArgs = make(map[string]string)
+		deleteArgs []string
+	)
+
+	switch c.GetType() {
+	case "k8s-dqlite":
+		updateArgs["--etcd-servers"] = fmt.Sprintf("unix://%s", path.Join(p.K8sDqliteStateDir(), "k8s-dqlite.sock"))
```
addyess, line 49: why can't the default `ExternalURLs` in in the Datastore type just be 
` fmt.Sprintf("unix://%s", path.Join(p.K8sDqliteStateDir(), "k8s-dqlite.sock"))`?

also, could you not also run TLS over the socket?  Why does `k8s-dqlite` never require certificates if it COULD in the future? 

I guess i wonder if this function could get rid of the switch entirely and just render the args based on the Datastore class's 4 attributes however they may be set?

src/k8s/pkg/k8sd/types/cluster_config_datastore.go
```diff
@@ -24,3 +29,45 @@ func (c Datastore) GetExternalClientKey() string  { return getField(c.ExternalCl
...
+		for _, loop := range []struct {
+			arg  string
+			cert string
+			path string
+		}{
+			{cert: c.GetExternalCACert(), arg: "--etcd-cafile", path: "ca.crt"},
+			{cert: c.GetExternalClientCert(), arg: "--etcd-certfile", path: "client.crt"},
+			{cert: c.GetExternalClientKey(), arg: "--etcd-keyfile", path: "client.key"},
+		} {
+			if loop.cert != "" {
```
addyess, line 64: any need to check if the file exists first?   I see the comment above saying  that we're just setting the paths here.  Do we expect them to be written?  If the certs are updated, does something come along and restart the service (i know that's a different PR)

addyess (commented):
Questions mostly.  the code looks fine.

src/k8s/pkg/k8sd/types/cluster_config_datastore.go
neoaggelos: good point, this is something for a separate PR (such that all `setup.Ensure*` functions return whether any of the files were changed

src/k8s/pkg/k8sd/types/cluster_config_datastore.go
neoaggelos: The API can be `[]string`, regardless of the internal type.

src/k8s/pkg/k8sd/types/cluster_config_datastore.go
neoaggelos: let's follow up on this separate from this PR

PR #299, opened 2024-04-04
Support External Certificates

mateoflorido:
## Overview
This pull request introduces a mechanism to specify external certificates during the bootstrap and joining processes in k8s-snap.

### Further Information
- Testing will be implemented in a follow-up PR. Manual testing validation:
```
root@kcp-1:~# k8s kubectl get no -A
NAME    STATUS   ROLES                  AGE     VERSION
kcp-1   Ready    control-plane,worker   7m29s   v1.29.3
kcp-2   Ready    control-plane,worker   3m9s    v1.29.3
kw-1    Ready    worker                 48s     v1.29.3
```


src/k8s/pkg/k8sd/types/cluster_config_certificates.go
```diff
@@ -22,5 +24,11 @@ func (c Certificates) GetAPIServerKubeletClientKey() string {
 	return getField(c.APIServerKubeletClientKey)
 }
 func (c Certificates) Empty() bool {
```
eaudetcobello, line 26:
```suggestion
// Check if every field of the certificate is Nil, returns false otherwise
func (c Certificates) Empty() bool {
```

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -16,6 +18,11 @@ import (
 func (a *App) onPostJoin(s *state.State, initConfig map[string]string) error {
 	snap := a.Snap()
 
+	joinClusterConfig, err := apiv1.JoinClusterConfigFromMap(initConfig)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal join config: %w", err)
+	}
```
eaudetcobello, line 24:
```suggestion
    if joinClusterConfig, err := apiv1.JoinClusterConfigFromMap(initConfig); err != nil {
	    return fmt.Errorf("failed to unmarshal join config: %w", err)
    }
```

src/k8s/pkg/k8sd/pki/worker.go
```diff
@@ -49,3 +49,7 @@ func (c *WorkerNodePKI) CompleteCertificates() error {
 	}
 	return nil
 }
+
+func (c *WorkerNodePKI) IsKubeletPresent() bool {
```
eaudetcobello, line 53: I get that this operates on a WorkerNodePKI object but could we name it something like "AreKubeletCertsSet" so the call would look like "certificates.AreKubeletCertsSet()"? I think it'd be more immediately clear what the function is checking for.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -220,6 +231,23 @@ func (a *App) onBootstrapControlPlane(s *state.State, initConfig map[string]stri
...
+	certificates.ServiceAccountKey = bootstrapConfig.ServiceAccountKey
+
+	certificates.APIServerKubeletClientCert = bootstrapConfig.APIServerKubeletClientCert
+	certificates.APIServerKubeletClientKey = bootstrapConfig.APIServerKubeletClientKey
+
+	certificates.APIServerCert = bootstrapConfig.APIServerCert
+	certificates.APIServerKey = bootstrapConfig.APIServerKey
+	certificates.KubeletCert = bootstrapConfig.KubeletCert
+	certificates.KubeletKey = bootstrapConfig.KubeletKey
+
```
eaudetcobello, line 250: Side note: this function is getting a bit long. I think it'd be nice to create a card to break it up a bit.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -28,20 +28,25 @@ import (
...
 	token := &types.InternalWorkerNodeToken{}
 	if err := token.Decode(encodedToken); err != nil {
 		return fmt.Errorf("failed to parse worker token: %w", err)
 	}
 
+	joinClusterConfig, err := apiv1.JoinClusterConfigFromMap(initConfig)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal join config: %w", err)
+	}
+
```
eaudetcobello, line 49:
```suggestion
	if joinClusterConfig, err := apiv1.JoinClusterConfigFromMap(initConfig); err != nil {
		return fmt.Errorf("failed to unmarshal join config: %w", err)
	}
```

src/k8s/pkg/k8sd/api/cluster_join.go
```diff
@@ -18,6 +18,11 @@ func (e *Endpoints) postClusterJoin(s *state.State, r *http.Request) response.Re
 		return response.BadRequest(fmt.Errorf("failed to parse request: %w", err))
 	}
 
+	config, err := req.Config.ToMap()
+	if err != nil {
+		return response.BadRequest(fmt.Errorf("failed to convert join config to map: %w", err))
+	}
```
eaudetcobello, line 24:
```suggestion
	if config, err := req.Config.ToMap(); err != nil {
		return response.BadRequest(fmt.Errorf("failed to convert join config to map: %w", err))
	}
```

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -76,5 +88,22 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
 	return cmd
 }
+
+func getJoinClusterConfigFromYaml(filePath string) (apiv1.JoinClusterConfig, error) {
+	config := apiv1.JoinClusterConfig{}
+
+	yamlContent, err := os.ReadFile(filePath)
+	if err != nil {
+		return config, fmt.Errorf("failed to read YAML config file: %w", err)
+	}
```
eaudetcobello, line 101:
```suggestion
	if yamlContent, err := os.ReadFile(filePath); err != nil {
		return config, fmt.Errorf("failed to read YAML config file: %w", err)
	}
```

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -76,5 +88,22 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
+
+	err = yaml.Unmarshal(yamlContent, &config)
+	if err != nil {
+		return config, fmt.Errorf("failed to parse YAML config file: %w", err)
+	}
```
eaudetcobello, line 106:
```suggestion
	if err = yaml.Unmarshal(yamlContent, &config); err != nil {
		return config, fmt.Errorf("failed to parse YAML config file: %w", err)
	}
```

src/k8s/cmd/k8s/k8s_join_cluster.go
eaudetcobello, line 101:
```suggestion
	if yamlContent, err := os.ReadFile(filePath); err != nil {
		return config, fmt.Errorf("failed to read YAML config file: %w", err)
	}
```

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -62,8 +64,18 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
 			}
 
+			joinClusterConfig := apiv1.JoinClusterConfig{}
+			if opts.configFile != "" {
+				joinClusterConfig, err = getJoinClusterConfigFromYaml(opts.configFile)
+				if err != nil {
+					cmd.PrintErrf("Error: Failed to read join configuration from %q.\n\nThe error was: %v\n", opts.configFile, err)
+					env.Exit(1)
+					return
+				}
```
eaudetcobello, line 74:
```suggestion
				if joinClusterConfig, err = getJoinClusterConfigFromYaml(opts.configFile); err != nil {
					cmd.PrintErrf("Error: Failed to read join configuration from %q.\n\nThe error was: %v\n", opts.configFile, err)
					env.Exit(1)
					return
				}
```

src/k8s/api/v1/types.go
```diff
@@ -58,6 +71,32 @@ func BootstrapConfigFromMap(m map[string]string) (*BootstrapConfig, error) {
...
+		"joinClusterConfig": string(config),
+	}, nil
+}
+
+func JoinClusterConfigFromMap(m map[string]string) (*JoinClusterConfig, error) {
+	config := &JoinClusterConfig{}
+	err := yaml.Unmarshal([]byte(m["joinClusterConfig"]), config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal join config: %w", err)
+	}
```
eaudetcobello, line 96:
```suggestion
	if err := yaml.Unmarshal([]byte(m["joinClusterConfig"]), config); err != nil {
		return nil, fmt.Errorf("failed to unmarshal join config: %w", err)
	}
```

src/k8s/api/v1/types.go
```diff
@@ -58,6 +71,32 @@ func BootstrapConfigFromMap(m map[string]string) (*BootstrapConfig, error) {
...
+	APIServerKey  string `yaml:"apiserver-key,omitempty"`
+	KubeletCert   string `yaml:"kubelet-crt,omitempty"`
+	KubeletKey    string `yaml:"kubelet-key,omitempty"`
+}
+
+func (j *JoinClusterConfig) ToMap() (map[string]string, error) {
+	config, err := yaml.Marshal(j)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal config map: %w", err)
+	}
```
eaudetcobello, line 85:
```suggestion
	if config, err := yaml.Marshal(j); err != nil {
		return nil, fmt.Errorf("failed to marshal config map: %w", err)
	}
```

eaudetcobello (changes requested):
Nice work, couple of stylistic suggestions :)

src/k8s/pkg/k8sd/app/hooks_join.go
mateoflorido: @eaudetcobello I think this is not valid in all the statements. This will scope the `joinClusterConfig` within the if statement.

mateoflorido: https://go.dev/play/p/OG3j0obkKCr

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
mateoflorido: Indeed, we should refactor the function.

src/k8s/pkg/k8sd/app/hooks_join.go
eaudetcobello: Oh cool! Didn't know that. Thanks for the link!

eaudetcobello (approved):

src/k8s/api/v1/types.go
```diff
@@ -24,6 +24,19 @@ type BootstrapConfig struct {
...
+	FrontProxyCACert           string `yaml:"front-proxy-ca-crt"`
+	FrontProxyCAKey            string `yaml:"front-proxy-ca-key"`
+	APIServerKubeletClientCert string `yaml:"apiserver-kubelet-client-crt"`
+	APIServerKubeletClientKey  string `yaml:"apiserver-kubelet-client-key"`
+	ServiceAccountKey          string `yaml:"service-account-key"`
+
+	APIServerCert string `yaml:"apiserver-crt,omitempty"`
+	APIServerKey  string `yaml:"apiserver-key,omitempty"`
+	KubeletCert   string `yaml:"kubelet-crt,omitempty"`
+	KubeletKey    string `yaml:"kubelet-key,omitempty"`
```
addyess, line 39: so here we go --- this config blob continues to grow unchecked.  

it looks to me that you've got "Keys" and "Certs".  Maybe you can make an object called `CertKey`  struct?  

Then include numerous of those here. 

```go
CA:               CertKey
FrontProxy:       CertKey
ServiceAccount:   CertKey
APIServer:        CertKey
APIServerKubelet: CertKey
Kubelet:          CertKey
```

Another question?  what's the different between `APIServerKubelet` and `Kubelet`?  Is the latter for the workers?

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -220,6 +231,23 @@ func (a *App) onBootstrapControlPlane(s *state.State, initConfig map[string]stri
...
+	certificates.ServiceAccountKey = bootstrapConfig.ServiceAccountKey
+
+	certificates.APIServerKubeletClientCert = bootstrapConfig.APIServerKubeletClientCert
+	certificates.APIServerKubeletClientKey = bootstrapConfig.APIServerKubeletClientKey
+
+	certificates.APIServerCert = bootstrapConfig.APIServerCert
+	certificates.APIServerKey = bootstrapConfig.APIServerKey
+	certificates.KubeletCert = bootstrapConfig.KubeletCert
+	certificates.KubeletKey = bootstrapConfig.KubeletKey
```
addyess, line 249: i wish i knew a good reason why all this copying is going on.  Moving things from one datatype to another....sheesh

addyess (commented):
More questions than comments

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -76,5 +88,22 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 	}
 	cmd.Flags().StringVar(&opts.name, "name", "", "node name, defaults to hostname")
 	cmd.Flags().StringVar(&opts.address, "address", "", "microcluster address, defaults to the node IP address")
+	cmd.PersistentFlags().StringVar(&opts.configFile, "config", "", "path to the YAML file containing your custom cluster join configuration")
```
neoaggelos, line 91: let's call this `--file` to be consistent with bootstrap

src/k8s/api/v1/types.go
neoaggelos: `APIServerKubelet` is the client certificate that kube-apiserver uses to authenticate requests to the `kubelet`, e.g. when doing things like `kubectl logs`. `APIServer` is the server certificate that serves TLS on kube-apiserver

Having a `CertKey` is also useful, but then there are things like `ServiceAccountKey` which is just a private key. To avoid dealing with this, we have been using flat structs instead. Why flat? No implied structure is easier to iterate with than some pre-existing (and possibly misleading) notion of structure.

src/k8s/api/v1/types.go
```diff
@@ -58,6 +71,32 @@ func BootstrapConfigFromMap(m map[string]string) (*BootstrapConfig, error) {
...
 }
 
+type JoinClusterConfig struct {
+	APIServerCert string `yaml:"apiserver-crt,omitempty"`
+	APIServerKey  string `yaml:"apiserver-key,omitempty"`
+	KubeletCert   string `yaml:"kubelet-crt,omitempty"`
+	KubeletKey    string `yaml:"kubelet-key,omitempty"`
+}
+
+func (j *JoinClusterConfig) ToMap() (map[string]string, error) {
```
neoaggelos, line 81: Use `ToMicrocluster()`, so that it better conveys what we do, see also https://github.com/canonical/k8s-snap/blob/37b00e5d0ce172c9dae06078f573be102ebdcfe7/src/k8s/api/v1/bootstrap_config.go#L31-L33

src/k8s/api/v1/types.go
```diff
@@ -58,6 +71,32 @@ func BootstrapConfigFromMap(m map[string]string) (*BootstrapConfig, error) {
...
+	config, err := yaml.Marshal(j)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal config map: %w", err)
+	}
+	return map[string]string{
+		"joinClusterConfig": string(config),
+	}, nil
+}
+
+func JoinClusterConfigFromMap(m map[string]string) (*JoinClusterConfig, error) {
```
neoaggelos, line 91: Same here, `JoinClusterConfigFromMicrocluster`

src/k8s/api/v1/types.go
```diff
@@ -58,6 +71,32 @@ func BootstrapConfigFromMap(m map[string]string) (*BootstrapConfig, error) {
 	return config, nil
 }
 
+type JoinClusterConfig struct {
```
neoaggelos, line 74: I would recommend going further and splitting this to two separate `ControlPlaneNodeJoinConfig` and `WorkerNodeJoinConfig` structs. Silently ignoring a `APIServerCert` (or even misleading consumers to provide it) for worker nodes doesn't seem appropriate

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -76,5 +88,22 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
+		return config, fmt.Errorf("failed to read YAML config file: %w", err)
+	}
+
+	err = yaml.Unmarshal(yamlContent, &config)
+	if err != nil {
+		return config, fmt.Errorf("failed to parse YAML config file: %w", err)
+	}
+
+	return config, nil
+}
```
neoaggelos, line 109: adjust accordingly with https://github.com/canonical/k8s-snap/blob/ca19306f76be9d166eaba9b606eca543b0cec1ce/src/k8s/cmd/k8s/k8s_bootstrap.go#L144-L157

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -28,20 +28,25 @@ import (
...
 	if workerToken, ok := initConfig["workerToken"]; ok {
-		return a.onBootstrapWorkerNode(s, workerToken)
+		return a.onBootstrapWorkerNode(s, workerToken, initConfig)
 	}
 
 	return a.onBootstrapControlPlane(s, initConfig)
 }
 
-func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string) error {
+func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string, initConfig map[string]string) error {
```
neoaggelos, line 37: this starts to look like high coupling. Let's only deal with `initConfig` on the `onBootstrap`, then pass individual types to `onBootstrapWorkerNode` and `onBootstrapControlPlane`.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -102,6 +107,12 @@ func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string) error {
 		KubeletCert: response.KubeletCert,
 		KubeletKey:  response.KubeletKey,
 	}
+
+	if !certificates.AreKubeletCertsSet() {
+		certificates.KubeletCert = joinClusterConfig.KubeletCert
+		certificates.KubeletKey = joinClusterConfig.KubeletKey
+	}
```
neoaggelos, line 114: NOTE: if the user did provide the certificates, we should not ignore them: 

```
if v := joinClusterConfig.KubeletCert; v != "" {
  certificates.KubeletCert = v
}
if v := joinClusterConfig.KubeletKey v != "" {
  certificates.KubeletKey = v
}
```

src/k8s/pkg/k8sd/types/cluster_config_certificates.go
```diff
@@ -21,6 +23,14 @@ func (c Certificates) GetAPIServerKubeletClientCert() string {
...
+// Check if every field of the Certificate struct is Nil, returns false otherwise.
 func (c Certificates) Empty() bool {
-	return c.CACert == nil && c.CAKey == nil && c.FrontProxyCACert == nil && c.FrontProxyCAKey == nil && c.ServiceAccountKey == nil && c.APIServerKubeletClientCert == nil && c.APIServerKubeletClientKey == nil
+	val := reflect.ValueOf(c)
+	for i := 0; i < val.NumField(); i++ {
+		if val.Field(i).Kind() == reflect.Pointer && !val.Field(i).IsNil() {
+			return false
+		}
+	}
+	return true
```
neoaggelos, line 35: not a huge fan of this. i'd rather we instead have a unit test that uses reflect to set individual fields (and would fail in case `.Empty()` does not return true.

If you feel like this grows a lot, then consider rewriting as:

```
switch {
case c.CACert != nil:
  return false
case c.CAKey != nil:
  return false
// ...
}
return true
}
```

neoaggelos (commented):
Very nice work, left some comments

neoaggelos (changes requested):
Great work! Marking as changes requested to make sure that points are addressed

src/k8s/pkg/k8sd/types/cluster_config_certificates.go
louiseschmidtgen: +1 That will help with readability as well.

src/k8s/api/v1/types.go
neoaggelos: also, no need for pointer, you can just return a `JoinClusterConfig` instead

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
neoaggelos: let's keep it as-is. flat is easier to iterate with, we can add more structure to the code once we are comfortable that the interfaces are well tested.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -223,6 +235,23 @@ func (a *App) onBootstrapControlPlane(s *state.State, initConfig map[string]stri
 		AllowSelfSignedCA:         true,
 		IncludeMachineAddressSANs: true,
 	})
+
+	certificates.CACert = bootstrapConfig.GetCACert()
+	certificates.CAKey = bootstrapConfig.GetCAKey()
+
```
neoaggelos, line 241: we can avoid these newlines and below

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -76,5 +87,15 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 	}
 	cmd.Flags().StringVar(&opts.name, "name", "", "node name, defaults to hostname")
 	cmd.Flags().StringVar(&opts.address, "address", "", "microcluster address, defaults to the node IP address")
+	cmd.PersistentFlags().StringVar(&opts.configFile, "file", "", "path to the YAML file containing your custom cluster join configuration")
```
neoaggelos, line 90: PersistentFlags are meant for flags that will be re-used by subcommands. Let's just do flags instead, to avoid possible future confusion.

```suggestion
	cmd.Flags().StringVar(&opts.configFile, "file", "", "path to the YAML file containing your custom cluster join configuration")
```

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -62,8 +63,18 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
 			}
 
+			var joinClusterConfig string
+			if opts.configFile != "" {
+				joinClusterConfig, err = getJoinConfigContents(opts.configFile)
```
neoaggelos, line 68: no need for an extra util function, we don't do much else here than just:

```suggestion
				b, err := os.ReadFile(opts.configFile)
				joinClusterConfig = string(b)
```

src/k8s/pkg/k8sd/types/cluster_config_certificates.go
```diff
@@ -21,6 +21,8 @@ func (c Certificates) GetAPIServerKubeletClientCert() string {
 func (c Certificates) GetAPIServerKubeletClientKey() string {
 	return getField(c.APIServerKubeletClientKey)
 }
+
+// Check if every field of the Certificate struct is Nil, returns false otherwise.
```
neoaggelos, line 25: go docs start with the function name
 
```suggestion
// Empty returns true if all Certificate fields are unset.
```

neoaggelos (approved):
Excellent work @mateoflorido! Just a few final comments!

neoaggelos (approved):

PR #301, opened 2024-04-05
Ku 550/tiobe score

louiseschmidtgen:
## Description
This PR fixes most of the tiobe dashboard security violations.

Violations are mostly resolved by adding error checks and more logging.

Reasoning: 
`Ignoring error conditions may allow an attacker to induce unexpected behavior unnoticed.`
`Detecting a security incident may take longer without sufficient logging. Delaying detection increases potential damage and mitigation effort.`

Some checks/logs feel unnecessary but we can discuss in this PR.

### DIFF
--- a/src/k8s/api/v1/types.go
+++ b/src/k8s/api/v1/types.go
@@ -181,7 +181,10 @@ func (c ClusterStatus) String() string {
 
 	// Status
 	if c.Ready {
-		result.WriteString("status: ready")
+		_, err := result.WriteString("status: ready")
+		if err != nil {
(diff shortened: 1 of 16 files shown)


addyess (changes requested):
I believe we have an understanding that this just moves the ball, but doesn't actually improve the score all that much.  Not only that -- but it further complicates the code. 

it would be nicer to build a map, and marshall the map to yaml.  In places where the output says "none" rather than "null", or "no" rather than "false" -- we need a custom type for corrected yaml marshaling

writing "yaml-like" data out to a string buffer was kind of bad form to begin with

PR #302, opened 2024-04-05
Bootstrap Configuration

neoaggelos:
### Summary

Make sure we can set all fields with `BootstrapConfig` when starting the cluster. Re-use the UserFacingClusterConfig, so that all fields from `k8s set` are available on bootstrap.

### Changes

- Update `apiv1.BootstrapConfig`
- Add unit tests for `apiv1.BootstrapConfig` <-> `types.ClusterConfig`
- Only set defaults on the server side when bootstrapping the cluster
- Use `yaml.UnmarshalStrict` for parsing the config file, such that unknown keys are not silently ignored.
- Test coverage


src/k8s/api/v1/bootstrap_config.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (b *BootstrapConfig) GetDatastoreType() string { return getField(b.DatastoreType) }
+
+// ToMicrocluster converts a BootstrapConfig to a map[string]string for use in microcluster.
+func (b *BootstrapConfig) ToMicrocluster() (map[string]string, error) {
+	config, err := json.Marshal(b)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal bootstrap config: %w", err)
+	}
```
eaudetcobello, line 37:
```suggestion
	if config, err := json.Marshal(b); err != nil {
		return nil, fmt.Errorf("failed to marshal bootstrap config: %w", err)
	}
```

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -143,21 +158,41 @@ func getConfigFromYaml(filePath string) (apiv1.BootstrapConfig, error) {
...
 	components := askQuestion(
 		stdin, stdout, stderr,
 		"Which components would you like to enable?",
 		componentList,
-		strings.Join(config.Components, ", "),
+		"network, dns, gateway, metrics-server",
 		nil,
 	)
-	config.Components = strings.Split(components, ",")
+	for _, component := range strings.FieldsFunc(components, func(r rune) bool { return unicode.IsSpace(r) || r == ',' }) {
```
eaudetcobello, line 169: Can you add a comment for what this is doing ?

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -143,21 +158,41 @@ func getConfigFromYaml(filePath string) (apiv1.BootstrapConfig, error) {
...
+			config.ClusterConfig.Gateway.Enabled = vals.Pointer(true)
+		case "local-storage":
+			config.ClusterConfig.LocalStorage.Enabled = vals.Pointer(true)
+		case "metrics-server":
+			config.ClusterConfig.MetricsServer.Enabled = vals.Pointer(true)
+		}
+	}
+
+	podCIDR := askQuestion(stdin, stdout, stderr, "Please set the Pod CIDR:", nil, "10.1.0.0/16", nil)
+	serviceCIDR := askQuestion(stdin, stdout, stderr, "Please set the Service CIDR:", nil, "10.152.183.", nil)
```
eaudetcobello, line 189: Is this the right default value? 10.152.183. ?

eaudetcobello (approved):
Nice job

src/k8s/api/v1/bootstrap_config.go
```diff
@@ -0,0 +1,51 @@
...
+	ServiceCIDR         *string `json:"service-cidr,omitempty" yaml:"service-cidr,omitempty"`
+	DisableRBAC         *bool   `json:"disable-rbac,omitempty" yaml:"disable-rbac,omitempty"`
+	SecurePort          *int    `json:"secure-port,omitempty" yaml:"secure-port,omitempty"`
+	CloudProvider       *string `json:"cloud-provider,omitempty" yaml:"cloud-provider,omitempty"`
+	K8sDqlitePort       *int    `json:"k8s-dqlite-port,omitempty" yaml:"k8s-dqlite-port,omitempty"`
+	DatastoreType       *string `json:"datastore-type,omitempty" yaml:"datastore-type,omitempty"`
+	DatastoreURL        *string `json:"datastore-url,omitempty" yaml:"datastore-url,omitempty"`
+	DatastoreCACert     *string `json:"datastore-ca-crt,omitempty" yaml:"datastore-ca-crt,omitempty"`
+	DatastoreClientCert *string `json:"datastore-client-crt,omitempty" yaml:"datastore-client-crt,omitempty"`
+	DatastoreClientKey  *string `json:"datastore-client-key,omitempty" yaml:"datastore-client-key,omitempty"`
```
addyess, line 24: `DatastoreType` seems superfluous.  Help me know why its there?  

if we're using a socket for k8sdqlite, isn't that socket file just a DatastoreURL?

secondarily -- `DatastoreURL` should really be a `[]string` because i understand that the args are a comma joined set of URLs. right?

src/k8s/api/v1/bootstrap_config.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// BootstrapConfigFromMicrocluster parses a microcluster map[string]string and retrieves the BootstrapConfig.
+func BootstrapConfigFromMicrocluster(m map[string]string) (BootstrapConfig, error) {
+	config := BootstrapConfig{}
+	if err := json.Unmarshal([]byte(m["bootstrapConfig"]), &config); err != nil {
+		return BootstrapConfig{}, fmt.Errorf("failed to unmarshal bootstrap config: %w", err)
+	}
+	return config, nil
+}
```
addyess, line 51: # ❤️ 

oh microcluster, you're so special

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -125,16 +143,13 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
 
 func getConfigFromYaml(filePath string) (apiv1.BootstrapConfig, error) {
-	config := apiv1.BootstrapConfig{}
-	config.SetDefaults()
-
-	yamlContent, err := os.ReadFile(filePath)
+	b, err := os.ReadFile(filePath)
 	if err != nil {
-		return config, fmt.Errorf("failed to read YAML config file: %w", err)
+		return apiv1.BootstrapConfig{}, fmt.Errorf("failed to read file: %w", err)
```
addyess, line 148: this is a different default config from lines 98->114 above.  why are we returning a valid BootstrapConfig at all and not just `nil`?

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -125,16 +143,13 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
 	}
 
-	err = yaml.Unmarshal(yamlContent, &config)
-	if err != nil {
+	var config apiv1.BootstrapConfig
+	if err := yaml.UnmarshalStrict(b, &config); err != nil {
 		return config, fmt.Errorf("failed to parse YAML config file: %w", err)
```
addyess, line 153: same why does this return a valid default bootstrap config and not `nil`

src/k8s/cmd/k8s/k8s_bootstrap.go
addyess: looks to be splitting the `components` user input on comma or spaces

src/k8s/pkg/k8sd/types/cluster_config_convert.go
```diff
@@ -1,63 +1,51 @@
...
-			ExternalClientCert: vals.Pointer(b.DatastoreClientCert),
-			ExternalClientKey:  vals.Pointer(b.DatastoreClientKey),
+			ExternalURL:        b.DatastoreURL,
+			ExternalCACert:     b.DatastoreCACert,
+			ExternalClientCert: b.DatastoreClientCert,
+			ExternalClientKey:  b.DatastoreClientKey,
 		}
+	default:
+		return ClusterConfig{}, fmt.Errorf("unknown datastore type specified in bootstrap config %q", b.GetDatastoreType())
 	}
```
addyess, line 39: I keep getting hung up here.  why'd we invent our own switch for datastore type baked into the config.  Feels mostly like a cli thing.  Either there is an external datastore or there isn't.  If there's no user configured `DatastoreURL` -- then we should default to the k8s-dqlite socket.

addyess (commented):
More comments than approval/disapproval

src/k8s/api/v1/bootstrap_config.go
mateoflorido: We could use external datastore. If that's the case, we bootstrap the node with `datastore-type: external`

addyess: @mateoflorido no i get that completely.  I just think that if you override the datastore url, you're very obviously using an "external" datastore.   from the kube-apiserver, it just wants to know the following:

```
etcd-servers
etcd-keyfile
etcd-certfile
etcd-cafile
```

it don't have a special name for the type.  This bootstrap config should represent only those things.  The "type" is a made up thing internally to this app but is unnecessary. 


the default should be `etcd-servers` = `DatastoreURL` = `unix:///var/snap..../k8sdqlite.socket`
and if someone wants it to be something else like a URL -- cool so be it.  Why is it necessary to drag this extra stringly typed field?

src/k8s/cmd/k8s/k8s_bootstrap.go
neoaggelos: Note that the return signature is `(apiv1.BootstrapConfig, error)`, not `(*apiv1.BootstrapConfig, error)`. In the latter case, we would return `nil` instead, as you said.

neoaggelos: but good point. the returned `config` would still be empty, but updated the code anyway :+1:

src/k8s/pkg/k8sd/types/cluster_config_convert.go
```diff
@@ -1,63 +1,55 @@
...
+	// Datastore
+	switch b.GetDatastoreType() {
 	case "", "k8s-dqlite":
 		config.Datastore = Datastore{
 			Type:          vals.Pointer("k8s-dqlite"),
-			K8sDqlitePort: vals.Pointer(b.K8sDqlitePort),
+			K8sDqlitePort: b.K8sDqlitePort,
 		}
 	case "external":
+		if len(b.DatastoreServers) == 0 {
```
bschimke95, line 31: Should we introduce a `BootstrapConfig.validate` method that checks that (and other validation, if any)?

bschimke95 (approved):
LGTM, only a minor comment - feel free to ignore.

src/k8s/cmd/k8s/k8s_bootstrap.go
neoaggelos: see above

src/k8s/api/v1/bootstrap_config.go
louiseschmidtgen: @addyess The `DatastoreType` is external for etcd for example. We've got some logic based on the value of this type when we grab the cluster config from the bootstrap config. We're also using it to print our k8s status. In theory we could check if the DatastoreURL is set and that would mean we are dealing with an external datastore but that might make those switch statements quite ugly.

PR #303, opened 2024-04-05
Use gomega NewWithT instead of deprecated NewGomegaWithT

neoaggelos:
Minor PR

```
NewGomegaWithT is deprecated in favor of gomega.NewWithT, which does not stutter.
```


bschimke95 (approved):
LGTM

PR #306, opened 2024-04-08
Update snap base to core24

berkayoz:
*The behaviour of automated elf patching changed in core22, the default is to [not patch auto](https://snapcraft.io/docs/linters-classic#heading--issues-auto)
*Snapcraft has changed in core24 to use the craft-application framework to align with other *craft tools, so some things [have changed](https://forum.snapcraft.io/t/micro-howto-migrate-to-core24/39393)
*Currently only can build with build-base: devel and grade: devel

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -6,23 +6,21 @@ description: |-
   Enjoy the simplified, almost zero, operations and enhanced security posture
   on any infrastructure
 license: GPL-3.0
-grade: stable
+grade: devel
 confinement: classic
-base: core20
+base: core24
+build-base: devel
 environment:
   REAL_PATH: $PATH
   REAL_LD_LIBRARY_PATH: $LD_LIBRARY_PATH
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -45,11 +43,13 @@ parts:
       - pkg-config
       - rsync
       - tcl
+      - sudo
```
neoaggelos, line 46: why do we need sudo ?

neoaggelos (commented):
Looks mostly fine. What's the trade-off of not including `[enable-patchelf]` as a build attribute? I would prefer keeping it out of the parts when it's not required.

Also, please make sure to rebase the strict patch on top of the introduced snapcraft.yaml changes

snap/snapcraft.yaml
berkayoz: The new build bases (that comes with the craft-application update) does not have sudo installed by default and our Makefile(via build scripts) call it

eaudetcobello:
Closing for now, we can reopen when needed.

PR #308, opened 2024-04-08
Ensure*PKI function indicates if [1, +INF[ files were changed

eaudetcobello:


src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
```diff
@@ -70,14 +70,22 @@ func (c *ControlPlaneConfigurationController) reconcile(ctx context.Context, con
...
+
+		if err != nil {
 			return fmt.Errorf("failed to reconcile external datastore certificates: %w", err)
 		}
 
+		if anyUpdated {
+			if err := c.snap.RestartService(ctx, "kube-apiserver"); err != nil {
+				return fmt.Errorf("failed to restart kube-apiserver to apply configuration: %w", err)
+			}
+		}
```
neoaggelos, line 87: note: this might result in multiple kube-apiserver restarts. e.g. if both certificates and arguments have changed. this is not useful, because it might result in failures (e.g. restarting the server without the proper arguments) or extended service downtime.

instead, the service should only be restarted after all configs are in place, see the code suggestions

src/k8s/pkg/k8sd/setup/certificates.go
```diff
@@ -11,54 +11,73 @@ import (
...
+func ensureFile(fname string, contents string, uid, gid int, mode fs.FileMode) (bool, error) {
 	if contents == "" {
-		if err := os.Remove(fname); err != nil && !os.IsNotExist(err) {
-			return fmt.Errorf("failed to delete: %w", err)
+		if err := os.Remove(fname); err != nil {
+			return false, fmt.Errorf("failed to delete: %w", err)
 		}
-		return nil
+
+		return true, nil
```
neoaggelos, line 22: this does not look right to me. `ensureFile(path, "", ...)` will now fail if the file does not exist. but: we do not want to return an error if the file does not exist, it is exactly what we asked `ensureFile` to do.

consider something like this instead. see how we handle the file does not exist case:

```go
if err := os.Remove(fname); err != nil {
  if !os.IsNotExist(err) {
    return false, fmt.Errorf("failed to delete: %w", err)
  }
  return false, nil
}

return true, nil
```

src/k8s/pkg/k8sd/setup/certificates.go
```diff
@@ -11,54 +11,73 @@ import (
...
-		return fmt.Errorf("failed to chown: %w", err)
+		return false, fmt.Errorf("failed to chown: %w", err)
 	}
 	if err := os.Chmod(fname, mode); err != nil {
-		return fmt.Errorf("failed to chmod: %w", err)
+		return false, fmt.Errorf("failed to chmod: %w", err)
 	}
 
-	return nil
+	return true, nil
```
neoaggelos, line 35: this does not look right to me. it will return `true` even if the file contents have not changed.

src/k8s/pkg/k8sd/setup/certificates.go
```diff
@@ -11,54 +11,73 @@ import (
...
+	var anyUpdated bool
+	for fname, content := range files {
+		updated, err := ensureFile(fname, content, uid, gid, mode)
+
+		if err != nil {
+			return false, fmt.Errorf("failed to configure %s: %w", path.Base(fname), err)
+		}
+
+		if updated {
+			anyUpdated = true
```
neoaggelos, line 50: same thing, but a lot more concise:

```suggestion
	var changed bool
	for fname, content := range files {
		if v, err := ensureFile(fname, content, uid, gid, mode); err != nil {
			return false, fmt.Errorf("failed to configure %s: %w", path.Base(fname), err)
		} else if v {
			changed = v
```

src/k8s/pkg/k8sd/setup/certificates_test.go
```diff
@@ -126,9 +121,79 @@ func TestEnsureWorkerPKI(t *testing.T) {
...
+	}
+	certificates := &pki.ExternalDatastorePKI{
+		DatastoreCACert:     "ca_cert",
+		DatastoreClientKey:  "client_key",
+		DatastoreClientCert: "client_cert",
+	}
+
+	if _, err := setup.EnsureExtDatastorePKI(mock, certificates); err != nil {
+		t.Fatalf("EnsureExtDatastorePKI returned unexpected error: %v", err)
+	}
```
neoaggelos, line 147: extend these tests, to test that the `changed` value is what we expect. this would have already caught some of the comments above.

neoaggelos: Consider creating a `certificates_internal_test.go` with a `Test_ensureFiles()` to add extensive coverage

src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
eaudetcobello: Nice, good observation

src/k8s/pkg/k8sd/setup/certificates.go
eaudetcobello: It will return true if the file has been deleted by os.Remove(), is that not right?

src/k8s/pkg/k8sd/setup/certificates.go
eaudetcobello: What does v stand for?

src/k8s/pkg/k8sd/setup/certificates_internal_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+		createdFile, _ := os.ReadFile(fname)
+		g.Expect(string(createdFile) == "new content").To(BeTrue())
+
+		// Change permissions and ensureFile should return that the file was not updated.
+		updated, err = ensureFile(fname, "new content", os.Getuid(), os.Getgid(), 0666)
+		g.Expect(err).To(BeNil())
+		g.Expect(updated).To(BeFalse())
+
+		// TODO: test with a different guid/uid than 1000
```
louiseschmidtgen, line 73: Could you please add a note here that explains why we want to test with a different guid/uid than 1000.

src/k8s/pkg/k8sd/setup/certificates.go
```diff
@@ -11,54 +11,90 @@ import (
...
-	if err := os.WriteFile(fname, []byte(contents), mode); err != nil {
-		return fmt.Errorf("failed to write: %w", err)
+	var contentChanged bool
+
+	if contents != string(origContent) {
+		if err := os.WriteFile(fname, []byte(contents), mode); err != nil {
+			return false, fmt.Errorf("failed to write: %w", err)
+		} else {
+			contentChanged = true
+		}
```
neoaggelos, line 46: no need for an else block here
```suggestion
		}
		contentChanged = true
```

neoaggelos (approved):

PR #309, opened 2024-04-08
Functionality -> Feature

eaudetcobello:
Tried to separate code/docs changes but our git diff CI did not like that... "changes not in repository".

Note: not every instance of the word "functionality" is renamed in docs because "feature" would not be appropriate in those instances. Eg: "If [...] your cluster will return to normal functionality."


louiseschmidtgen (approved):
Great work, I don't see anything wrong with this, it's good to go!

PR #319, opened 2024-04-09
Add fingerprint check when worker joins cluster

louiseschmidtgen:
## Description
This PR addresses Mitm vulnerabilities for joining worker nodes by validating the server. 

A joining node validates the server (cluster member) by checking the servers certificate against the fingerprint of the join token. The joining node adds the certificate to its trusted certificates and the TLS connection is established, the joining node may now join the cluster.

### DIFF
--- a/src/k8s/pkg/k8sd/api/cluster_join.go
+++ b/src/k8s/pkg/k8sd/api/cluster_join.go
@@ -26,6 +26,19 @@ func (e *Endpoints) postClusterJoin(s *state.State, r *http.Request) response.Re
 	internalToken := types.InternalWorkerNodeToken{}
 	// Check if token is worker token
 	if internalToken.Decode(req.Token) == nil {
+		// Check Server Auth
+		// Get remote certificate from the cluster member
+		cert, err := utils.GetRemoteCertificate(req.Address)
+		if err != nil {
+			return response.InternalError(fmt.Errorf("failed to get certificate of cluster member: %w", err))
+		}
+
+		// verify that the fingerprint of the certificate matches the fingerprint of the token
+		fingerprint := utils.CertFingerprint(cert)
+		if fingerprint != internalToken.Fingerprint {
+			return response.BadRequest(fmt.Errorf("server authentication failed: join token fingerprint does not match that of the cluster member"))
+		}
+
 		// valid worker node token - let's join the cluster
 		// The validation of the token is done when fetching the cluster information.
 		if err := e.provider.MicroCluster().NewCluster(hostname, req.Address, map[string]string{"workerToken": req.Token}, 0); err != nil {
--- a/src/k8s/pkg/k8sd/api/cluster_join.go
+++ b/src/k8s/pkg/k8sd/api/cluster_join.go
@@ -39,6 +39,8 @@ func (e *Endpoints) postClusterJoin(s *state.State, r *http.Request) response.Re
 			return response.BadRequest(fmt.Errorf("server authentication failed: join token fingerprint does not match that of the cluster member"))
 		}
 
+		// TODO: add certificate to trusted certificates
+
 		// valid worker node token - let's join the cluster
 		// The validation of the token is done when fetching the cluster information.
 		if err := e.provider.MicroCluster().NewCluster(hostname, req.Address, map[string]string{"workerToken": req.Token}, 0); err != nil {
--- a/src/k8s/pkg/k8sd/api/cluster_join.go
+++ b/src/k8s/pkg/k8sd/api/cluster_join.go
@@ -26,21 +26,6 @@ func (e *Endpoints) postClusterJoin(s *state.State, r *http.Request) response.Re
 	internalToken := types.InternalWorkerNodeToken{}
 	// Check if token is worker token
 	if internalToken.Decode(req.Token) == nil {
-		// Check Server Auth
-		// Get remote certificate from the cluster member
-		cert, err := utils.GetRemoteCertificate(req.Address)
-		if err != nil {
-			return response.InternalError(fmt.Errorf("failed to get certificate of cluster member: %w", err))
-		}
-
-		// verify that the fingerprint of the certificate matches the fingerprint of the token
-		fingerprint := utils.CertFingerprint(cert)
-		if fingerprint != internalToken.Fingerprint {
-			return response.BadRequest(fmt.Errorf("server authentication failed: join token fingerprint does not match that of the cluster member"))
-		}
-
-		// TODO: add certificate to trusted certificates
-
 		// valid worker node token - let's join the cluster
 		// The validation of the token is done when fetching the cluster information.
 		if err := e.provider.MicroCluster().NewCluster(hostname, req.Address, map[string]string{"workerToken": req.Token}, 0); err != nil {
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"path"
 
 	apiv1 "github.com/canonical/k8s/api/v1"
@@ -50,9 +51,12 @@ func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string) error {
 	}
 	// TODO(neoaggelos): figure out how to use the microcluster client instead
 
-	url := "https://" + token.JoinAddresses[0]
+	url, err := url.Parse("https://" + token.JoinAddresses[0])
+	if err != nil {
+		return fmt.Errorf("failed to parse control plane address: %w", err)
+	}
 	// Get remote certificate from the cluster member
-	cert, err := utils.GetRemoteCertificate(url)
+	cert, err := utils.GetRemoteCertificate(url.String())
 	if err != nil {
 		return fmt.Errorf("failed to get certificate of cluster member: %w", err)
 	}
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -51,7 +51,7 @@ func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string) error {
 	// TODO(neoaggelos): figure out how to use the microcluster client instead
 
 	// Get remote certificate from the cluster member
-	cert, err := utils.GetRemoteCertificate(nodeIP.String())
+	cert, err := utils.GetRemoteCertificate(token.JoinAddresses[0])
 	if err != nil {
 		return fmt.Errorf("failed to get certificate of cluster member: %w", err)
 	}
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -63,7 +63,7 @@ func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string) error {
 		return fmt.Errorf("server authentication failed: join token fingerprint does not match that of the cluster member")
 	}
 
-	httpClient, err := utils.CreateHTTPClientWithCert(cert, key)
+	httpClient, err := utils.CreateHTTPClientWithCert(cert)
 	if err != nil {
 		return fmt.Errorf("failed to create HTTP client with certificate: %w", err)
 	}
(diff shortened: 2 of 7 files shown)


src/k8s/pkg/k8sd/api/cluster_join.go
```diff
@@ -26,6 +26,21 @@ func (e *Endpoints) postClusterJoin(s *state.State, r *http.Request) response.Re
...
+		// Check Server Auth
+		// Get remote certificate from the cluster member
+		cert, err := utils.GetRemoteCertificate(req.Address)
+		if err != nil {
+			return response.InternalError(fmt.Errorf("failed to get certificate of cluster member: %w", err))
+		}
+
+		// verify that the fingerprint of the certificate matches the fingerprint of the token
+		fingerprint := utils.CertFingerprint(cert)
+		if fingerprint != internalToken.Fingerprint {
```
neoaggelos, line 38: This should not happen here, but in https://github.com/canonical/k8s-snap/blob/[main](https://github.com/canonical/k8s-snap/blob/b7ae955ee3551d72119d9a46128e09cc5870ebdf/src/k8s/pkg/k8sd/app/hooks_bootstrap.go#L65)/src/k8s/pkg/k8sd/app/hooks_bootstrap.go#L65

so that we can also then create a client with the retrieved certificate

src/k8s/pkg/k8sd/types/worker.go
```diff
@@ -14,6 +14,8 @@ type InternalWorkerNodeToken struct {
 	Secret string `json:"secret"`
 	// JoinAddresses is a list of control-plane addresses that exist in the cluster.
 	JoinAddresses []string `json:"join_addresses"`
+	// Fingerprint is used for server Auth
+	Fingerprint string `json:"fingerprint"`
```
neoaggelos, line 18: NOTE: we also need to set the fingerprint when generating the worker node join token

neoaggelos (commented):
Going in the right direction, see comments

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -49,15 +48,23 @@ func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string) error {
...
-	// create an HTTP client that ignores https
-	httpClient := &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
-			},
-		},
+
+	// Get remote certificate from the cluster member
+	cert, err := utils.GetRemoteCertificate(nodeIP.String())
```
neoaggelos, line 54: `nodeIP.String()` is not right. First of all, we need a `$host:$port` to connect to, not just an address. Secondly, this must be the address of the remote microcluster node, from `token.JoinAddresses[0]`

neoaggelos (changes requested):

src/k8s/pkg/utils/certificate.go
```diff
@@ -26,3 +31,70 @@ func SplitIPAndDNSSANs(extraSANs []string) ([]net.IP, []string) {
...
+	}
+
+	return &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+	}, nil
+}
+
+func GetRemoteCertificate(url string) (*x509.Certificate, error) {
```
neoaggelos, line 70: This makes more sense as `GetRemoteCertificate(hostPort string)`, and then do a GET on `https://$hostPort/` internally. 

```suggestion
func GetRemoteCertificate(url string) (*x509.Certificate, error) {
```

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -49,15 +49,27 @@ func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string) error {
...
-				InsecureSkipVerify: true,
-			},
-		},
+
+	url, err := url.Parse("https://" + token.JoinAddresses[0])
+	if err != nil {
+		return fmt.Errorf("failed to parse control plane address: %w", err)
+	}
+	// Get remote certificate from the cluster member
+	cert, err := utils.GetRemoteCertificate(url.String())
```
neoaggelos, line 59: that way you can just pass `token.JoinAddresses[0]` here

src/k8s/pkg/utils/certificate.go
```diff
@@ -26,3 +31,70 @@ func SplitIPAndDNSSANs(extraSANs []string) ([]net.IP, []string) {
 
 	return ipSANs, dnsSANs
 }
+
+func TLSClientConfig(remoteCert *x509.Certificate) (*tls.Config, error) {
+	if remoteCert == nil {
+		return nil, fmt.Errorf("Invalid remote public key")
```
neoaggelos, line 37: errors are always lower-case. The handler of the error will make this a sentence at the end.

```suggestion
		return nil, fmt.Errorf("invalid remote public key")
```

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -49,15 +48,23 @@ func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string) error {
...
+	// Get remote certificate from the cluster member
+	cert, err := utils.GetRemoteCertificate(token.JoinAddresses[0])
+	if err != nil {
+		return fmt.Errorf("failed to get certificate of cluster member: %w", err)
+	}
+
+	// verify that the fingerprint of the certificate matches the fingerprint of the token
+	fingerprint := utils.CertFingerprint(cert)
+	if fingerprint != token.Fingerprint {
+		return fmt.Errorf("server authentication failed: join token fingerprint does not match that of the cluster member")
```
neoaggelos, line 62: let's make sure this error message helps people understand what is happening
```suggestion
		return fmt.Errorf("fingerprint from token (%q) does not match fingerprint of node %q (%q) (from node %v)", token.Fingerprint, fingerprint, token.JoinAddresses[0])
```

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -49,15 +48,23 @@ func (a *App) onBootstrapWorkerNode(s *state.State, encodedToken string) error {
...
+	}
+
+	httpClient, err := utils.CreateHTTPClientWithCert(cert)
+	if err != nil {
+		return fmt.Errorf("failed to create HTTP client with certificate: %w", err)
```
neoaggelos, line 67:
```suggestion
		return fmt.Errorf("failed to create HTTPS client: %w", err)
```

src/k8s/pkg/k8sd/types/worker.go
```diff
@@ -14,6 +14,8 @@ type InternalWorkerNodeToken struct {
 	Secret string `json:"secret"`
 	// JoinAddresses is a list of control-plane addresses that exist in the cluster.
 	JoinAddresses []string `json:"join_addresses"`
+	// Fingerprint is used for server Auth
```
neoaggelos, line 17: this is not used for server auth

src/k8s/pkg/utils/certificate.go
```diff
@@ -26,3 +32,75 @@ func SplitIPAndDNSSANs(extraSANs []string) ([]net.IP, []string) {
 
 	return ipSANs, dnsSANs
 }
+
+func TLSClientConfig(remoteCert *x509.Certificate) (*tls.Config, error) {
```
neoaggelos, line 36: please add a `// TLSClientConfig returns ...`, as `utils.TLSClientConfig` is a very generic name.

Also, I would recommend naming this something like: `TLSClientConfigWithTrustedCertificate()`

src/k8s/pkg/utils/certificate.go
```diff
@@ -26,3 +32,75 @@ func SplitIPAndDNSSANs(extraSANs []string) ([]net.IP, []string) {
...
+	}
+
+	config := &tls.Config{}
+
+	// Add the public key to the CA pool to make it trusted.
+	config.RootCAs = x509.NewCertPool()
+	remoteCert.IsCA = true
+	remoteCert.KeyUsage = x509.KeyUsageCertSign
+	config.RootCAs.AddCert(remoteCert)
+
```
neoaggelos, line 48: how much of this is required? let's not allow more than the absolute minimum necessary things. Also, `x509.NewCertPool()` ignores the root CAs from the host system, which _might_ be a problem if we use this elsewhere.

src/k8s/pkg/utils/certificate.go
```diff
@@ -26,3 +32,75 @@ func SplitIPAndDNSSANs(extraSANs []string) ([]net.IP, []string) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+	}, nil
+}
```
neoaggelos, line 69: If this is simply wrapping the TLSClientConfig, let's inline it where it's used.

src/k8s/pkg/utils/certificate.go
```diff
@@ -26,3 +32,75 @@ func SplitIPAndDNSSANs(extraSANs []string) ([]net.IP, []string) {
...
+		},
+	}
+
+	url, err := url.Parse("https://" + address)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse address: %w", err)
+	}
+
+	// Connect
+	req, err := http.NewRequest("GET", url.String(), nil)
```
neoaggelos, line 86: Is this to validate the address? That could work _by accident_, but the error message would still be weird as it will be descibing an error with a URL, and the user does not know what to do.

Instead, use `net.SplitHostPort` to verify the `address`, and then you can do `fmt.Sprintf("https://%s", address)` on the HTTP request

neoaggelos (approved):
LGTM, great work @louiseschmidtgen!

src/k8s/pkg/utils/certificate.go
```diff
@@ -26,3 +31,58 @@ func SplitIPAndDNSSANs(extraSANs []string) ([]net.IP, []string) {
 
 	return ipSANs, dnsSANs
 }
+
+func TLSClientConfigWithTrustedCertificate(remoteCert *x509.Certificate) (*tls.Config, error) {
```
bschimke95, line 35: nit: Please add comments/explanation for public functions and some basic unittests. We want tiobe to be happy :)

src/k8s/pkg/utils/certificate.go
```diff
@@ -26,3 +31,58 @@ func SplitIPAndDNSSANs(extraSANs []string) ([]net.IP, []string) {
...
+	}
+
+	config := &tls.Config{}
+
+	remoteCert.IsCA = true
+	config.RootCAs.AddCert(remoteCert)
+
+	// Always use public key DNS name rather than server cert, so that it matches.
+	if len(remoteCert.DNSNames) > 0 {
+		config.ServerName = remoteCert.DNSNames[0]
```
bschimke95, line 47: So basically, we use the `DNSName` of the CA as the server name of the clients TLS config. What happens if there is no `DNSName` set? Will it break if there is no server name set for the config?

Currently, we have our own CA.
What happens if the CA comes from an external vendor (e.g. Vault)? Will this still work?

src/k8s/pkg/utils/certificate.go
```diff
@@ -26,3 +31,58 @@ func SplitIPAndDNSSANs(extraSANs []string) ([]net.IP, []string) {
...
+func GetRemoteCertificate(address string) (*x509.Certificate, error) {
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		},
+	}
+
+	_, _, err := net.SplitHostPort(address)
```
bschimke95, line 62: super small nit: It makes sense to do the input validation at the beginning of the function. Otherwise you create the `httpClient` for no reason.

bschimke95 (commented):
great work, added some minor comments and one for me to understand things better.

src/k8s/pkg/utils/certificate.go
louiseschmidtgen: @neoaggelos What do you think?

src/k8s/pkg/utils/certificate_test.go
```diff
@@ -24,3 +25,19 @@ func TestSplitIPAndDNSSANs(t *testing.T) {
 
 	g.Expect(gotDNSs).To(ConsistOf("cluster.local", "kubernetes.svc.local"))
 }
+
+func TestTLSClientConfigWithTrustedCertificate(t *testing.T) {
+	g := NewWithT(t)
+
+	// Mock certificate and certificate pool for testing
+	remoteCert := &x509.Certificate{
+		DNSNames: []string{"bubblegum.com"},
```
bschimke95, line 34: I like the names your come up with in your tests :D

bschimke95 (commented):
Please also apply the comments/tests to the other two functions you introduced.

src/k8s/pkg/utils/certificate.go
bschimke95: Resolved offline: tldr: here CA = remote certificate, the kubernetes CA is not involved at all.

bschimke95 (approved):
LGTM

PR #322, opened 2024-04-09
fix `k8s enable network` not working

neoaggelos:
### Summary

Fix `k8s enable network` and `k8s disable network` not updating the network feature. The missing check caused https://github.com/canonical/k8s-snap/blob/e6227daf6e17a426eb8088dfc3e5af1ed4f33236/src/k8s/pkg/k8sd/api/cluster_config.go#L49 to fail


bschimke95 (approved):
LGTM

PR #323, opened 2024-04-09
Change Local Storage `set-default` to `default`

mateoflorido:
## Overview
Replace the Local Storage `set-default` config occurrences to `default`.

### DIFF
--- a/src/k8s/api/v1/cluster_config.go
+++ b/src/k8s/api/v1/cluster_config.go
@@ -77,13 +77,13 @@ type LocalStorageConfig struct {
 	Enabled       *bool   `json:"enabled,omitempty" yaml:"enabled,omitempty"`
 	LocalPath     *string `json:"local-path,omitempty" yaml:"local-path,omitempty"`
 	ReclaimPolicy *string `json:"reclaim-policy,omitempty" yaml:"reclaim-policy,omitempty"`
-	SetDefault    *bool   `json:"set-default,omitempty" yaml:"set-default,omitempty"`
+	Default       *bool   `json:"default,omitempty" yaml:"default,omitempty"`
 }
 
 func (c LocalStorageConfig) GetEnabled() bool         { return getField(c.Enabled) }
 func (c LocalStorageConfig) GetLocalPath() string     { return getField(c.LocalPath) }
 func (c LocalStorageConfig) GetReclaimPolicy() string { return getField(c.ReclaimPolicy) }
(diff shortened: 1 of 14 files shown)


ktsakalozos:
I think there should be a docs update too https://github.com/canonical/k8s-snap/blob/main/docs/src/howto/storage.md

louiseschmidtgen (commented):
Thanks for updating the naming, this is cleaning things up nicely! Glad we won't do `GetSetDefault`anymore :)

src/k8s/api/v1/cluster_config.go
```diff
@@ -77,13 +77,13 @@ type LocalStorageConfig struct {
 	Enabled       *bool   `json:"enabled,omitempty" yaml:"enabled,omitempty"`
 	LocalPath     *string `json:"local-path,omitempty" yaml:"local-path,omitempty"`
 	ReclaimPolicy *string `json:"reclaim-policy,omitempty" yaml:"reclaim-policy,omitempty"`
-	SetDefault    *bool   `json:"set-default,omitempty" yaml:"set-default,omitempty"`
+	Default       *bool   `json:"default,omitempty" yaml:"default,omitempty"`
 }
```
neoaggelos, line 81: Just to note: this is a breaking change in the API. We do not have stable releases yet, so we can do it.

src/k8s/pkg/k8sd/types/cluster_config_features.go
```diff
@@ -35,7 +35,7 @@ type LocalStorage struct {
 	Enabled       *bool   `json:"enabled,omitempty"`
 	LocalPath     *string `json:"local-path,omitempty"`
 	ReclaimPolicy *string `json:"reclaim-policy,omitempty"`
-	SetDefault    *bool   `json:"set-default,omitempty"`
+	Default       *bool   `json:"default,omitempty"`
 }
```
neoaggelos, line 39: Of note, this `json:"set-default"` -> `json:"default"` is also a breaking change, which would make the value `nil` in existing clusters.

In the future, we could keep the `json:"set-default"` around, and just rename the struct field.

But again, seeing as there are no stable releases yet, we're good to go with the change.

neoaggelos (approved):
LGTM, nice work @mateoflorido. Adding a couple of notes regarding the breaking changes contained in the PR, this is something we will not be able to do once we get a stable release out the door

bschimke95 (approved):
LGTM, great work

PR #331, opened 2024-04-12
Allow specifying extra SANs when joining control plane nodes

neoaggelos:
### Summary

Allow specifying extra SANs when joining control plane nodes.


eaudetcobello (approved):
LGTM. This is similar to work that was done previously for specifying extra sans on bootstrap.

PR #332, opened 2024-04-12
fix GET /k8sd/cluster panic when not initialized

neoaggelos:
### Summary

Prevent panics in `GET /k8sd/cluster` if node is not yet initialized. This is a regression caused as a side-effect of https://github.com/canonical/k8s-snap/pull/257/files#diff-86fdc39de0c5dec676bb358c40f6de2447b0f4c4570315906d9b1c0e6e2b50eaR11-R19

Also perform some more cleanup in the code (do not retrieve the cluster config twice).

### DIFF
--- a/src/k8s/pkg/k8sd/api/cluster.go
+++ b/src/k8s/pkg/k8sd/api/cluster.go
@@ -13,27 +13,21 @@ import (
 )
 
 func (e *Endpoints) getClusterStatus(s *state.State, r *http.Request) response.Response {
-	snap := e.provider.Snap()
+	// fail if node is not initialized yet
+	if !s.Database.IsOpen() {
+		return response.Unavailable(fmt.Errorf("daemon not yet initialized"))
+	}
 
 	members, err := impl.GetClusterMembers(s.Context, s)
 	if err != nil {
 		return response.InternalError(fmt.Errorf("failed to get cluster members: %w", err))
 	}
-
 	config, err := utils.GetClusterConfig(s.Context, s)
-	if err != nil {
-		return response.InternalError(fmt.Errorf("failed to get user-facing cluster config: %w", err))
-	}
-
-	clusterConfig, err := utils.GetClusterConfig(s.Context, s)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/k8sd/api/cluster.go
```diff
@@ -13,27 +13,21 @@ import (
...
 	if err != nil {
 		return response.InternalError(fmt.Errorf("failed to get cluster members: %w", err))
 	}
-
 	config, err := utils.GetClusterConfig(s.Context, s)
-	if err != nil {
-		return response.InternalError(fmt.Errorf("failed to get user-facing cluster config: %w", err))
-	}
-
-	clusterConfig, err := utils.GetClusterConfig(s.Context, s)
```
addyess, line 28: was this just duplicate code?

src/k8s/pkg/k8sd/api/cluster.go
```diff
@@ -13,27 +13,21 @@ import (
 )
 
 func (e *Endpoints) getClusterStatus(s *state.State, r *http.Request) response.Response {
-	snap := e.provider.Snap()
+	// fail if node is not initialized yet
+	if !s.Database.IsOpen() {
+		return response.Unavailable(fmt.Errorf("daemon not yet initialized"))
```
addyess, line 18: seems like `Unavailable` is still in the 5xx space (503) so this shouldn't change anything on charm

addyess (approved):
Nice thank you

src/k8s/pkg/k8sd/api/cluster.go
neoaggelos: yes

src/k8s/pkg/k8sd/api/cluster.go
neoaggelos: This is trying to imitate https://github.com/canonical/microcluster/blob/5043bc984c83e9ecfdba1de731ec05678b166f13/internal/rest/rest.go#L187-L196 as close as possible

PR #333, opened 2024-04-12
Fixes on node and control plane config controllers

neoaggelos:
### Summary

- Do not panic if watch does not receive a configmap, return an error instead
- Fixup error message in control plane config controller


addyess (approved):
useful context

> context: watch configmap retrieves things from a channel, which are a generic runtime.Object type. We dereference it as a *v1.ConfigMap, but in Go this will panic if the underlying type is different

PR #334, opened 2024-04-13
Make the external datastore configurable via API at runtime

bschimke95:
# Summary
This PR exposes the datastore configuration to the API which makes it possible to adapt the configuration of the external datastore (e.g. server addresses/certificates). It is not allowed to change the datastore type (everything that is not type "external" will fail). 

In the same effort, the external server config is changed to `[]string` instead of `string` to match the internal datatype.


src/k8s/pkg/k8sd/types/cluster_config_datastore.go
```diff
@@ -12,22 +13,22 @@ type Datastore struct {
...
 	K8sDqliteKey  *string `json:"k8s-dqlite-key,omitempty"`
 
-	ExternalURL        *string `json:"external-url,omitempty"`
-	ExternalCACert     *string `json:"external-ca-crt,omitempty"`
-	ExternalClientCert *string `json:"external-client-crt,omitempty"`
-	ExternalClientKey  *string `json:"external-client-key,omitempty"`
+	ExternalServers    *[]string `json:"external-servers,omitempty"`
+	ExternalCACert     *string   `json:"external-ca-crt,omitempty"`
+	ExternalClientCert *string   `json:"external-client-crt,omitempty"`
+	ExternalClientKey  *string   `json:"external-client-key,omitempty"`
```
addyess, line 19: Chance we can just drop the `external` here so they are 

```
servers
ca-crt
client-crt
client-key
```

addyess (changes requested):
to be honest, i'm befuddled why there are so many different struct definitions with the same 4 fields over and over but with differently named keys. 

Can the type be set once? and just reused everywhere with the same key names

if not -- maybe just get the key names the same everywhere

src/k8s/api/v1/cluster_config.go
```diff
@@ -103,6 +107,21 @@ type MetricsServerConfig struct {
...
 func (c MetricsServerConfig) GetEnabled() bool { return getField(c.Enabled) }
 
+type UserFacingDatastoreConfig struct {
+	// Type of the datastore. Needs to be "external".
+	Type       *string   `json:"type,omitempty" yaml:"type,omitempty"`
+	Servers    *[]string `json:"servers,omitempty" yaml:"servers,omitempty"`
+	CACert     *string   `json:"ca-crt,omitempty" yaml:"ca-crt,omitempty"`
+	ClientCert *string   `json:"client-crt,omitempty" yaml:"client-crt,omitempty"`
+	ClientKey  *string   `json:"client-key,omitempty" yaml:"client-key,omitempty"`
+}
```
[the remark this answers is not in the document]
bschimke95: Hm, I see your point. However, in the `BootstrapConfig` we can also set the `k8s-dqlite` port which is obviously a datastore-related config. However, we don't want to set this port during runtime (do we?). 
Hence, I think the flat key approach makes sense for the bootstrap config. 

But I'm also open for discussions here.

src/k8s/api/v1/types.go
```diff
@@ -43,8 +43,8 @@ type NodeStatus struct {
 }
 
 type Datastore struct {
-	Type        string `json:"type,omitempty"`
-	ExternalURL string `json:"external-url,omitempty" yaml:"external-url,omitempty"`
+	Type            string   `json:"type,omitempty"`
+	ExternalServers []string `json:"external-servers,omitempty" yaml:"external-servers,omitempty"`
```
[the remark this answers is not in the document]
neoaggelos: > There is IMHO also no point in perfectly matching the internal representation of configuration with the external-facing one. They (might) eventually diverge in the future if we introduce new features and new API versions

+1 to this. The `pkg/k8sd/types.ClusterConfig` can diverge at will and is not user-facing, so we can handle it as we see fit for the code. The only type users will be aware of is `api/v1.$Foo`.

Note that this is still API

src/k8s/api/v1/cluster_config.go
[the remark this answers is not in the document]
neoaggelos: One approach we can take for now is re-use the `UserFacingDatastoreConfig` in the `UpdateClusterConfigRequest`, which will simply reject fields that are not allowed to changed (e.g. the k8s-dqlite port).

This also gives us a path forward _if_ we ever want to implement this in the future (we probably do not)

src/k8s/pkg/k8sd/types/cluster_config_validate.go
```diff
@@ -20,6 +21,24 @@ func validateCIDRs(cidrString string) error {
 	return nil
 }
 
+// isValidAddress checks if a given address is valid. Allows with and without schema.
+func isValidAddress(address string) bool {
```
louiseschmidtgen, line 25: This looks like a function that could be used in many places. Have you considered putting this in a common place like utils?

tests/integration/templates/etcd/etcd.service
```diff
@@ -22,12 +22,12 @@
     --initial-cluster-state $CLUSTER_STATE \
     --client-cert-auth \
     --trusted-ca-file /tmp/ca-cert.pem \
-    --cert-file /tmp/etcd-cert.pem \
-    --key-file /tmp/etcd-key.pem \
+    --cert-file /tmp/client-cert.pem \
```
louiseschmidtgen, line 25: Would it be possible to name these files to show they are used for the datastore? Like `datastore-client-cert.pem`?

src/k8s/api/v1/types.go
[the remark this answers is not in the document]
addyess: i guess i was asking about the `json` and `yaml` keys being named `external-*`  i had assumed these types would resolve to something that is visible to the users.  So long as they are all consistently named, it's good with me.

src/k8s/api/v1/cluster_config.go
```diff
@@ -13,7 +13,11 @@ type GetClusterConfigResponse struct {
 }
 
 type UpdateClusterConfigRequest struct {
-	Config UserFacingClusterConfig
+	Config UserFacingClusterConfig `json:"config,omitempty" yaml:"config,omitempty"`
```
louiseschmidtgen, line 16: With fields getting added here, would it make any sense to update the name of `UserFacingClusterConfig` to something like `UserFacingClusterFeatureConfig`?

src/k8s/api/v1/types.go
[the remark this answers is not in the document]
bschimke95: Alright, I read of the comments, that we can leave it like this.

src/k8s/api/v1/cluster_config.go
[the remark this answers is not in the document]
bschimke95: I'm not a big fan of this. The API would suggest something we don't support. _IF_ we want to support it later, it would probably be a breaking change, hence `v2` anyway, right?

tests/integration/templates/etcd/etcd.service
bschimke95: Hm, I mean it is the `etcd.service` file which is loaded inside the `EtcdCluster` object. So, if we want to do this it would be `etcd-client-cert.pem`, but I think that is a bit unnecessary tbh.

src/k8s/pkg/k8sd/types/cluster_config_validate.go
bschimke95: Let's keep it here for now, it is kind of specific for the validator (allows schemaless etc.).

src/k8s/api/v1/cluster_config.go
```diff
@@ -13,7 +13,11 @@ type GetClusterConfigResponse struct {
...
+	// DatastoreConfig contains updates of the external datastore configuration.
+	// It is not part of the UserFacingClusterConfig as it is direcly marshalled by the CLI.
+	// The datastore config should not be displayed as part of the configuration in `k8s get`
+	DatastoreConfig UserFacingDatastoreConfig `json:"datastore-config,omitempty" yaml:"datastore-config,omitempty"`
```
neoaggelos, line 20: I would just call this `Datastore` and `json:"datastore"`. I would remove the comment.

src/k8s/pkg/k8sd/api/cluster_config.go
```diff
@@ -27,12 +27,19 @@ func (e *Endpoints) putClusterConfig(s *state.State, r *http.Request) response.R
...
+		return response.BadRequest(fmt.Errorf("failed to updated datastore config: type must be 'external' but is %s", req.DatastoreConfig.GetType()))
+	}
+
 	oldConfig, err := utils.GetClusterConfig(r.Context(), s)
 	if err != nil {
 		return response.InternalError(fmt.Errorf("failed to retrieve cluster configuration: %w", err))
 	}
 
 	requestedConfig := types.ClusterConfigFromUserFacing(req.Config)
+	requestedConfig.Datastore = types.DatastoreConfigFromUserFacing(req.DatastoreConfig)
```
neoaggelos, line 42: This always updates the `datastore` struct in the config, regardless if it's set in the request, right? But these will just be nil so we are OK

src/k8s/pkg/k8sd/api/cluster_config.go
```diff
@@ -27,12 +27,19 @@ func (e *Endpoints) putClusterConfig(s *state.State, r *http.Request) response.R
 		return response.BadRequest(fmt.Errorf("failed to decode request: %w", err))
 	}
 
+	// Changing the datastore configuration is opt-in. We expect the caller to explicitly set the "external" type.
+	// The nil check is required to ensure we only fail if the DatastoreConfig is expected to change.
+	if req.DatastoreConfig.Type != nil && req.DatastoreConfig.GetType() != "external" {
+		return response.BadRequest(fmt.Errorf("failed to updated datastore config: type must be 'external' but is %s", req.DatastoreConfig.GetType()))
```
neoaggelos, line 33: let's use `%q` for quotes, as it also handles escapes for us

```suggestion
		return response.BadRequest(fmt.Errorf("failed to updated datastore config: type must be %q but is %q", "external", req.DatastoreConfig.GetType()))
```

src/k8s/pkg/k8sd/api/cluster_config.go
```diff
@@ -27,12 +27,19 @@ func (e *Endpoints) putClusterConfig(s *state.State, r *http.Request) response.R
 		return response.BadRequest(fmt.Errorf("failed to decode request: %w", err))
 	}
 
+	// Changing the datastore configuration is opt-in. We expect the caller to explicitly set the "external" type.
+	// The nil check is required to ensure we only fail if the DatastoreConfig is expected to change.
+	if req.DatastoreConfig.Type != nil && req.DatastoreConfig.GetType() != "external" {
```
neoaggelos, line 32: i would make this part of `types.DatastoreConfigFromUserFacing` and return a `(Datastore, error)`

neoaggelos (commented):
Nice work, just final comments

src/k8s/pkg/k8sd/api/cluster_config.go
```diff
@@ -33,6 +33,11 @@ func (e *Endpoints) putClusterConfig(s *state.State, r *http.Request) response.R
 	}
 
 	requestedConfig := types.ClusterConfigFromUserFacing(req.Config)
+	requestedConfig.Datastore, err = types.DatastoreConfigFromUserFacing(req.Datastore)
+	if err != nil {
```
neoaggelos, line 37:
```suggestion
	if requestedConfig.Datastore, err = types.DatastoreConfigFromUserFacing(req.Datastore); err != nil {
```

neoaggelos (approved):

tests/integration/tests/test_etcd.py
```diff
@@ -42,3 +44,60 @@ def test_etcd(instances: List[harness.Instance], etcd_cluster: EtcdCluster):
         ["systemctl", "is-active", "--quiet", "snap.k8s.k8s-dqlite"], check=False
     )
     assert p.returncode != 0, "k8s-dqlite service is still active"
+
+    LOG.info("Add new etcd nodes")
+    etcd_cluster.add_nodes(2)
+
+    # Update  server-urls in cluster
```
neoaggelos, line 51:
```suggestion
    # Update server-urls in cluster
```

tests/integration/tests/test_etcd.py
```diff
@@ -42,3 +44,60 @@ def test_etcd(instances: List[harness.Instance], etcd_cluster: EtcdCluster):
...
+            "--data",
+            json.dumps(body),
+            "--unix-socket",
+            "/var/snap/k8s/common/var/lib/k8sd/state/control.socket",
+            "http://localhost/1.0/k8sd/cluster/config",
+        ],
+        capture_output=True,
+    )
+    response = json.loads(resp.stdout.decode())
+    assert response["error_code"] == 400, "changing the datastore type should fail"
```
neoaggelos, line 103: i would expect this should be a unit test instead

tests/integration/tests/test_etcd.py
```diff
@@ -42,3 +44,60 @@ def test_etcd(instances: List[harness.Instance], etcd_cluster: EtcdCluster):
...
+            "http://localhost/1.0/k8sd/cluster/config",
+        ]
+    )
+
+    # Give some time to reconcile and restart services.
+    time.sleep(2)
+
+    # check that we can still connect to the kubernetes cluster
+    k8s_instance.exec(["k8s", "kubectl", "get", "pods", "-A"])
+
```
neoaggelos, line 80: we should instead retry this until it succeeds and use `util.stubbornly`

tests/integration/tests/test_etcd.py
bschimke95: true, I have added a unittest for the converter.
But I also keep this check. It is quick - the cluster is up already -  and we don't have a good way to test the API.

src/k8s/api/v1/types.go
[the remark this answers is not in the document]
addyess: So, back to this specific line -- since it's in `api/v1/types` -- it's not an internal structure?

* can we have this called "servers" in the json and yaml marshalling?  
* is there any need for yaml annotation on the type

src/k8s/api/v1/types.go
```diff
@@ -77,8 +77,9 @@ func (c ClusterStatus) datastoreToString() string {
 		result.WriteString(fmt.Sprintf("  type: %s\n", c.Datastore.Type))
 		// Datastore URL for external only
 		if c.Datastore.Type == "external" {
-			if c.Datastore.ExternalURL != "" {
-				result.WriteString(fmt.Sprintf("  url: %s\n", c.Datastore.ExternalURL))
+			result.WriteString(fmt.Sprintln("  servers:"))
+			for _, serverURL := range c.Datastore.ExternalServers {
+				result.WriteString(fmt.Sprintf("    - %s\n", serverURL))
```
addyess, line 82: Here -- in the "plain-text rendering" -- it looks like `servers`.  that's great.  Would love the same from the API

src/k8s/pkg/k8sd/types/cluster_config_validate.go
```diff
@@ -20,6 +21,24 @@ func validateCIDRs(cidrString string) error {
...
+	if u, err := url.Parse(address); err == nil && u.Host != "" {
+		return true
+	}
+
+	host, port, err := net.SplitHostPort(address)
+	if err == nil {
+		if host == "localhost" || net.ParseIP(host) != nil {
```
addyess, line 32: so?  no fqdn for etcd servers are allowed?  

example)
`172.22.1.1.ec2.internal`  would be disallowed?

tests/integration/tests/test_etcd.py
neoaggelos: sounds good

neoaggelos (approved):

src/k8s/api/v1/types_test.go
```diff
@@ -72,7 +72,7 @@ func TestString(t *testing.T) {
 					Network: NetworkConfig{Enabled: vals.Pointer(true)},
 					DNS:     DNSConfig{Enabled: vals.Pointer(true)},
 				},
-				Datastore: Datastore{Type: "k8s-dqlite", ExternalURL: ""},
+				Datastore: Datastore{Type: "k8s-dqlite", ExternalServers: nil},
```
neoaggelos, line 75: not needed

src/k8s/api/v1/cluster_config.go
```diff
@@ -167,3 +183,11 @@ func (c MetricsServerConfig) String() string {
...
 	return string(b)
 }
+
+func (c UserFacingDatastoreConfig) String() string {
+	b, err := yaml.Marshal(c)
+	if err != nil {
+		return fmt.Sprintf("%#v\n", c)
+	}
+	return string(b)
+}
```
neoaggelos, line 193: is this used anywhere?

src/k8s/pkg/k8sd/types/cluster_config_validate.go
neoaggelos, line 32: ah, good point. any valid hostname should be allowed. if `url.Parse` succeeded above, then we should be good, extra checks here are probably not needed.

let's add unit tests that verify all these

src/k8s/api/v1/cluster_config.go
bschimke95: removed

src/k8s/api/v1/types.go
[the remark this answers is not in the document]
bschimke95: true - updated

src/k8s/api/v1/types_test.go
bschimke95: removed

src/k8s/pkg/k8sd/types/cluster_config_validate.go
bschimke95: loosened the check to allow fqdn and extended unittests

bschimke95: done

addyess (approved):
Dude, you're awesome.  nice work

PR #335, opened 2024-04-15
Use TLS when connecting to apiserver

louiseschmidtgen:
## Description
For security reasons we want to connect to the apiserver using tls. We can use microcluster CA.

### DIFF
--- a/src/k8s/pkg/k8sd/setup/embed/apiserver/auth-token-webhook.conf
+++ b/src/k8s/pkg/k8sd/setup/embed/apiserver/auth-token-webhook.conf
@@ -3,9 +3,10 @@ kind: Config
 clusters:
   - name: k8s-token-auth-service
     cluster:
-      certificate-authority: /var/snap/k8s/common/var/lib/k8sd/state/cluster.crt
+      certificate-authority: "{{ .CA }}"
+      /var/snap/k8s/common/var/lib/k8sd/state
       tls-server-name: 127.0.0.1
-      server: "https://10.0.0.79:6400/1.0/kubernetes/auth/webhook"
+      server: "{{ .URL }}"
 current-context: webhook
 contexts:
 - context:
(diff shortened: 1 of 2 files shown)


src/k8s/pkg/k8sd/setup/kube_apiserver.go
```diff
@@ -52,8 +54,11 @@ func KubeAPIServer(snap snap.Snap, serviceCIDR string, authWebhookURL string, en
 	if err != nil {
 		return fmt.Errorf("failed to open auth-token-webhook.conf: %w", err)
 	}
+
+	microclusterCAPath := filepath.Join(snap.K8sdStateDir(), "cluster.crt")
```
neoaggelos, line 58: let's use `path.Join` like in other places in this file. also, we are not using this anywhere else, so let's inline this below in the struct.

src/k8s/pkg/k8sd/setup/embed/apiserver/auth-token-webhook.conf
```diff
@@ -3,7 +3,9 @@ kind: Config
 clusters:
   - name: k8s-token-auth-service
     cluster:
-      insecure-skip-tls-verify: true
+      certificate-authority: "{{ .CA }}"
+      /var/snap/k8s/common/var/lib/k8sd/state
```
neoaggelos, line 7: this line does not look right

src/k8s/pkg/k8sd/setup/kube_apiserver.go
```diff
@@ -13,6 +14,7 @@ import (
 
 type apiserverAuthTokenWebhookTemplateConfig struct {
 	URL string
+	CA  string
```
neoaggelos, line 17: Let's call this `CAPath` to make it clear that we expect a file path instead of the certificate itself.

bschimke95 (approved):
LGTM

PR #338, opened 2024-04-16
add pprof debug endpoints (disabled by default)

neoaggelos:
To enable, `--pprof-address` needs to be specified in k8sd arguments.


bschimke95 (approved):
LGTM

Issue #321, opened 2024-04-09
cannot bootstrap snap in charm


beliaev-maksim:
Hello,
charm cannot bootstrap snap on my machine

see attached logs
[juju.log](https://github.com/canonical/k8s-snap/files/14920523/juju.log)
[journallog-k8sd.txt](https://github.com/canonical/k8s-snap/files/14920525/journallog-k8sd.txt)
[journallog.txt](https://github.com/canonical/k8s-snap/files/14920526/journallog.txt)

bschimke95:
Hey Maksim,

This basically happens because of a internal timeout of microcluster that we cannot work around yet. 
It is addressed in https://github.com/canonical/microcluster/pull/105.

On our side, we also work towards reducing the overall time that the commands need to finish which eventually also "fixes" this issue. A first effort is done in #339 with a follow-up PR coming soon to make those commands even faster by moving the last pieces into an async approach.

beliaev-maksim:
@bschimke95 if we can fix it, that is ultimate solution

but if we are stuck for now, should we reconsider our docs for minimum hardware reqs?

bschimke95:
It will be fixed this pulse.

bschimke95:
@beliaev-maksim with #353 this should be fixed. Can you verify this on your local machine please?

beliaev-maksim:
@bschimke95 confirm

bschimke95:
nice

Issue #277, opened 2024-03-27
Scaling down K8s charm cluster causes k8s to get stuck waiting for k8sd


amc94:
I initially deployed 3 units of k8s then used juju to remove two, this led the last remaining k8s unit to be stuck with status waiting and the message 'waiting on k8sd'

```
unit-k8s-0: 13:16:15 ERROR unit.k8s/0.juju-log Caught ReconcilerError
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/charms/contextual_status.py", line 101, in on_error
    yield
  File "/usr/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/var/lib/juju/agents/unit-k8s-0/charm/./src/charm.py", line 205, in _check_k8sd_ready
    self.api_manager.check_k8sd_ready()
  File "/var/lib/juju/agents/unit-k8s-0/charm/lib/charms/k8s/v0/k8sd_api_manager.py", line 659, in check_k8sd_ready
    self._send_request(endpoint, "GET", EmptyResponse)
  File "/var/lib/juju/agents/unit-k8s-0/charm/lib/charms/k8s/v0/k8sd_api_manager.py", line 559, in _send_request
    raise InvalidResponseError(
charms.k8s.v0.k8sd_api_manager.InvalidResponseError: Error status 503
	method=GET
	endpoint=/cluster/1.0/ready
	reason=Service Unavailable
	body={"type":"error","status":"","status_code":0,"operation":"","error_code":503,"error":"Daemon is not ready yet","metadata":null}



The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/charms/reconciler.py", line 35, in reconcile
    self.reconcile_function(event)
  File "/var/lib/juju/agents/unit-k8s-0/charm/./src/charm.py", line 433, in _reconcile
    self._check_k8sd_ready()
  File "/usr/lib/python3.10/contextlib.py", line 78, in inner
    with self._recreate_cm():
  File "/usr/lib/python3.10/contextlib.py", line 153, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/charms/contextual_status.py", line 106, in on_error
    raise ReconcilerError(msg) from e
charms.contextual_status.ReconcilerError: Found expected exception: Error status 503
	method=GET
[...]

PR #339, opened 2024-04-16
Decouple update k8sd-config configmap from k8s set API

bschimke95:
This PR adds a controller that updates the `k8sd-config` config map of the cluster asynchronously.
It is triggered by the sending `struct{}` to the controllers `UpdateCh` (e.g. by the bootstrap hook or `k8s set`). 

This is a first effort of making commands return fast instead of e.g. waiting for service restarts and apply changes async. Follow-up work will move the component updates to a controller  as well.

Relates and works towards fixing:
#321 and #277

### DIFF
--- a/src/k8s/pkg/k8sd/api/cluster_config.go
+++ b/src/k8s/pkg/k8sd/api/cluster_config.go
@@ -52,13 +52,6 @@ func (e *Endpoints) putClusterConfig(s *state.State, r *http.Request) response.R
 		return response.InternalError(fmt.Errorf("database transaction to update cluster configuration failed: %w", err))
 	}
 
-	// Trigger an update of the configuration.
-	// Do not wait if the channel is full. The reconcilation loop will apply the most recent changes
-	select {
-	case e.provider.UpdateNodeConfigurationControllerCh() <- struct{}{}:
-	default:
-	}
-
 	if !requestedConfig.Network.Empty() {
 		if err := component.ReconcileNetworkComponent(r.Context(), snap, oldConfig.Network.Enabled, requestedConfig.Network.Enabled, mergedConfig); err != nil {
 			return response.InternalError(fmt.Errorf("failed to reconcile network: %w", err))
@@ -118,6 +111,8 @@ func (e *Endpoints) putClusterConfig(s *state.State, r *http.Request) response.R
 			return response.InternalError(fmt.Errorf("failed to reconcile load-balancer: %w", err))
 		}
 	}
+
+	e.provider.NotifyUpdateConfigMap()
 
 	return response.SyncResponse(true, &api.UpdateClusterConfigResponse{})
 }
--- a/src/k8s/pkg/k8sd/api/cluster_config.go
+++ b/src/k8s/pkg/k8sd/api/cluster_config.go
@@ -13,7 +13,6 @@ import (
 	"github.com/canonical/k8s/pkg/k8sd/database"
 	"github.com/canonical/k8s/pkg/k8sd/types"
 	"github.com/canonical/k8s/pkg/utils"
-	"github.com/canonical/k8s/pkg/utils/k8s"
 	"github.com/canonical/k8s/pkg/utils/vals"
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/microcluster/state"
@@ -53,6 +52,13 @@ func (e *Endpoints) putClusterConfig(s *state.State, r *http.Request) response.R
 		return response.InternalError(fmt.Errorf("database transaction to update cluster configuration failed: %w", err))
 	}
 
+	// Trigger an update of the configuration.
+	// Do not wait if the channel is full. The reconcilation loop will apply the most recent changes
+	select {
+	case e.provider.UpdateNodeConfigurationControllerCh() <- struct{}{}:
+	default:
+	}
+
 	if !requestedConfig.Network.Empty() {
 		if err := component.ReconcileNetworkComponent(r.Context(), snap, oldConfig.Network.Enabled, requestedConfig.Network.Enabled, mergedConfig); err != nil {
 			return response.InternalError(fmt.Errorf("failed to reconcile network: %w", err))
@@ -81,20 +87,6 @@ func (e *Endpoints) putClusterConfig(s *state.State, r *http.Request) response.R
 				return response.InternalError(fmt.Errorf("database transaction to update cluster configuration failed: %w", err))
 			}
 		}
-	}
-
-	cmData, err := mergedConfig.Kubelet.ToConfigMap()
-	if err != nil {
-		return response.InternalError(fmt.Errorf("failed to format kubelet configmap data: %w", err))
-	}
-
-	client, err := k8s.NewClient(snap.KubernetesRESTClientGetter(""))
-	if err != nil {
-		return response.InternalError(fmt.Errorf("failed to create kubernetes client: %w", err))
-	}
-
-	if _, err := client.UpdateConfigMap(r.Context(), "kube-system", "k8sd-config", cmData); err != nil {
-		return response.InternalError(fmt.Errorf("failed to update node config: %w", err))
 	}
 
 	if !requestedConfig.LocalStorage.Empty() {
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -347,13 +347,6 @@ func (a *App) onBootstrapControlPlane(s *state.State, bootstrapConfig apiv1.Boot
 		}
 	}
 
-	// Trigger an update of the configuration.
-	// Do not wait if the channel is full. The reconcilation loop will apply the most recent changes
-	select {
-	case a.updateNodeConfigController.TriggerCh <- struct{}{}:
-	default:
-	}
-
 	if cfg.LocalStorage.GetEnabled() {
 		if err := component.ReconcileLocalStorageComponent(s.Context, snap, vals.Pointer(false), vals.Pointer(true), cfg); err != nil {
 			return fmt.Errorf("failed to reconcile local-storage: %w", err)
@@ -383,6 +376,8 @@ func (a *App) onBootstrapControlPlane(s *state.State, bootstrapConfig apiv1.Boot
 			return fmt.Errorf("failed to reconcile metrics-server: %w", err)
 		}
 	}
+
+	a.NotifyUpdateConfigMap()
 
 	return nil
 }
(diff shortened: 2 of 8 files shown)


src/k8s/pkg/k8sd/api/cluster_config.go
```diff
@@ -53,6 +52,13 @@ func (e *Endpoints) putClusterConfig(s *state.State, r *http.Request) response.R
 		return response.InternalError(fmt.Errorf("database transaction to update cluster configuration failed: %w", err))
 	}
 
+	// Trigger an update of the configuration.
+	// Do not wait if the channel is full. The reconcilation loop will apply the most recent changes
+	select {
+	case e.provider.UpdateNodeConfigurationControllerCh() <- struct{}{}:
+	default:
+	}
+
```
neoaggelos, line 61: we should notify the controller at the end, there's more configuration updates later in this function which might be lost

src/k8s/pkg/k8sd/controllers/update_node_configuration.go
```diff
@@ -0,0 +1,110 @@
...
+		case <-time.After(3 * time.Second):
+		}
+	}
+}
+
+// Run starts the controller.
+// Run accepts a context to manage the lifecycle of the controller.
+// Run accepts a function that retrieves the current cluster configuration.
+// Run will loop everytime a cluster config patch is sent to the UpdateCh.
+func (c *UpdateNodeConfigurationController) Run(ctx context.Context, getClusterConfig func(context.Context) (types.ClusterConfig, error)) {
```
berkayoz, line 57: Aren't we supposed to call this in `hooks_start` ?

src/k8s/pkg/k8sd/controllers/update_node_configuration.go
```diff
@@ -0,0 +1,110 @@
...
+
+		if err := c.reconcile(ctx, client, config); err != nil {
+			log.Println(fmt.Errorf("failed to reconcile cluster configuration: %w", err))
+		}
+
+		select {
+		case c.ReconciledCh <- struct{}{}:
+		case <-time.After(10 * time.Second):
+			log.Println("failed to reconcile cluster config in time - will fetch latest config and retry.")
+			continue
```
neoaggelos, line 94: the `default:` below means that this will never occur

src/k8s/pkg/k8sd/controllers/update_node_configuration_test.go
```diff
@@ -0,0 +1,105 @@
...
+				return &k8s.Client{Interface: clientset}, nil
+			})
+			go ctrl.Run(ctx, configProvider.getConfig)
+
+			select {
+			case ctrl.UpdateCh <- struct{}{}:
+			case <-time.After(channelSendTimeout):
+				g.Fail("Timed out while attempting to trigger controller reconcile loop")
+			}
+			<-ctrl.ReconciledCh
```
neoaggelos, line 96: check with a timeout here as well. if there is a bug in the controller, we don't want the tests to block here

src/k8s/pkg/k8sd/controllers/update_node_configuration.go
```diff
@@ -0,0 +1,110 @@
...
+	return &UpdateNodeConfigurationController{
+		snap:         snap,
+		waitReady:    waitReady,
+		newK8sClient: newK8sClient,
+		UpdateCh:     make(chan struct{}, 1),
+		ReconciledCh: make(chan struct{}, 1),
+	}
+}
+
+func (c *UpdateNodeConfigurationController) retryNewK8sClient(ctx context.Context) (*k8s.Client, error) {
```
berkayoz, line 38: nit: This starts to feel a bit duplicated, I also see the point in having it as  part of the controller.

src/k8s/pkg/k8sd/controllers/update_node_configuration_test.go
```diff
@@ -0,0 +1,105 @@
...
+				Data: tc.initialConfig,
+			}
+			clientset := fake.NewSimpleClientset(configMap)
+
+			if !tc.expectedFailure {
+				clientset.PrependReactor("patch", "configmaps", func(action k8stesting.Action) (handled bool, ret runtime.Object, err error) {
+					configMap.Data = tc.expectedConfig
+					return true, nil, nil
+				})
+			}
```
neoaggelos, line 84: is it simpler to just .Get() the configmap after the reconcile loop is done?

neoaggelos (changes requested):
Really nice work, a couple of comments

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -348,18 +347,11 @@ func (a *App) onBootstrapControlPlane(s *state.State, bootstrapConfig apiv1.Boot
...
-	client, err := k8s.NewClient(snap.KubernetesRESTClientGetter(""))
-	if err != nil {
-		return fmt.Errorf("failed to create kubernetes client: %w", err)
-	}
-
-	if _, err := client.UpdateConfigMap(s.Context, "kube-system", "k8sd-config", cmData); err != nil {
-		return fmt.Errorf("failed to update node configs: %w", err)
+	// Trigger an update of the configuration.
+	// Do not wait if the channel is full. The reconcilation loop will apply the most recent changes
+	select {
```
berkayoz, line 352: This seems alright for now although if we ever depend on that config to be present after this section to do more  bootstrap stuff we might introduce a race condition or some weird errors here. Nothing to worry about, just some food for thought.

berkayoz (commented):
Very nice work, looks good overall. Some small comments.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
bschimke95: There should be no problem. There are two scenarios:
a) the config needs to be set on bootstrap and cannot be changed later.
-> then the config should be applied before this part already and not get the config from the config map anyway.
b) the config can be changed on runtime. Then the reconciliation loop will apply the changes eventually and restart the relevant services. The services might not work in the time between config update and restart but this is fine.

src/k8s/pkg/k8sd/controllers/update_node_configuration.go
bschimke95: yes, thanks

src/k8s/pkg/k8sd/controllers/update_node_configuration.go
bschimke95: ack, let's keep it here for now. If things start to accumulate we can still refactor this.

src/k8s/pkg/k8sd/controllers/update_node_configuration_test.go
bschimke95: true

src/k8s/pkg/k8sd/controllers/update_node_configuration_test.go
bschimke95: fixed

src/k8s/pkg/k8sd/api/cluster_config.go
bschimke95: changed to `NotifyUpdateConfigMap` good idea!

src/k8s/pkg/k8sd/app/provider.go
```diff
@@ -14,5 +14,9 @@ func (a *App) Snap() snap.Snap {
 	return a.snap
 }
 
+func (a *App) NotifyUpdateConfigMap() {
+	a.updateNodeConfigController.TriggerCh <- struct{}{}
```
neoaggelos, line 18: add in the `select { case }` logic here

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -348,18 +347,11 @@ func (a *App) onBootstrapControlPlane(s *state.State, bootstrapConfig apiv1.Boot
...
+	case a.updateNodeConfigController.TriggerCh <- struct{}{}:
+	default:
 	}
```
neoaggelos, line 355: let's use `a.NotifyUpdateConfigMap()` here as well

src/k8s/pkg/k8sd/controllers/update_node_configuration_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+			select {
+			case ctrl.TriggerCh <- struct{}{}:
+			case <-time.After(channelSendTimeout):
+				g.Fail("Timed out while attempting to trigger controller reconcile loop")
+			}
+
+			select {
+			case <-ctrl.ReconciledCh:
+			case <-time.After(channelSendTimeout):
```
neoaggelos, line 92: in this case, `g.Fail()` saying what we timed out waiting for the reconcile to complete

neoaggelos (changes requested):

neoaggelos (approved):

neoaggelos (changes requested):
Just to make sure conflicts are addressed

src/k8s/pkg/k8sd/controllers/update_node_configuration.go
```diff
@@ -0,0 +1,107 @@
...
+// Run accepts a function that retrieves the current cluster configuration.
+// Run will loop everytime the TriggerCh is triggered.
+func (c *UpdateNodeConfigurationController) Run(ctx context.Context, getClusterConfig func(context.Context) (types.ClusterConfig, error)) {
+	c.waitReady()
+
+	client, err := c.retryNewK8sClient(ctx)
+	if err != nil {
+		log.Println(fmt.Errorf("failed to create a Kubernetes client: %w", err))
+	}
+
```
neoaggelos, line 64: move this in the for loop, see https://github.com/canonical/k8s-snap/pull/346/files

src/k8s/pkg/k8sd/types/cluster_config_datastore_convert.go
```diff
@@ -0,0 +1,25 @@
+package types
+
+import apiv1 "github.com/canonical/k8s/api/v1"
+
+// DatastoreConfigFromUserFacing converts UserFacingDatastoreConfig from public API into a Datastore config.
+func DatastoreConfigFromUserFacing(u apiv1.UserFacingDatastoreConfig) Datastore {
```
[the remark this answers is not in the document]
bschimke95: We want to decouple external and internal types. That way, we can safely diverge the public API from the internal types. 
e.g. in APIv2 we might want to support a new datastore config parameter. By decoupling those things we can extend the internal type with the new config and add it to APIv2 while still maintaing APIv1 without this config.

neoaggelos (approved):

PR #340, opened 2024-04-16
[WIP] fix busy loop in node config controller

neoaggelos:
### Summary

### DIFF
--- a/src/k8s/pkg/utils/k8s/configmap.go
+++ b/src/k8s/pkg/utils/k8s/configmap.go
@@ -20,16 +20,17 @@ func (c *Client) WatchConfigMap(ctx context.Context, namespace string, name stri
 		select {
 		case <-ctx.Done():
 			return nil
-		case evt := <-w.ResultChan():
-			if evt.Object != nil {
-				configMap, ok := evt.Object.(*v1.ConfigMap)
-				if !ok {
-					return fmt.Errorf("expected a ConfigMap but received %#v", evt.Object)
-				}
+		case evt, ok := <-w.ResultChan():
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/utils/k8s/configmap.go
```diff
@@ -20,16 +20,17 @@ func (c *Client) WatchConfigMap(ctx context.Context, namespace string, name stri
...
-			if evt.Object != nil {
-				configMap, ok := evt.Object.(*v1.ConfigMap)
-				if !ok {
-					return fmt.Errorf("expected a ConfigMap but received %#v", evt.Object)
-				}
+		case evt, ok := <-w.ResultChan():
+			if !ok {
+				return fmt.Errorf("watch closed")
+			}
+			configMap, ok := evt.Object.(*v1.ConfigMap)
```
addyess, line 27:
```suggestion
            if evt.Object == nil {
                // Did you need to handle me?  The case where the channel is still open but the evt.Object is now nil? 
                // This used to be handled butt he log has changed
            }
			configMap, ok := evt.Object.(*v1.ConfigMap)
```

addyess: what occurs if someone deletes the config-map?

neoaggelos: no we are still good, the `_, ok := evt.Object(*v1.ConfigMap)` just below will just fail instead, and return an error

addyess (approved):
LGTM thank you

I've confirmed this halts the busy loops

PR #341, opened 2024-04-17
Adjust systemd service definitions

neoaggelos:
### Summary

Adjust systemd service definitions in `snapcraft.yaml`:

- Add `before` and `after` to note the order in which to start the services
- Adjust the strict patch for the updates to `snap/snapcraft.yaml`
- Add a 5s timeout to apiserver restarts


berkayoz (approved):
LGTM

PR #349, opened 2024-04-17
Decouple update k8sd-config configmap from k8s set API

neoaggelos:
This is work by @bschimke95, because I clicked the wrong button in #339

This PR adds a controller that updates the k8sd-config config map of the cluster asynchronously.
It is triggered by the sending struct{} to the controllers UpdateCh (e.g. by the bootstrap hook or k8s set).

This is a first effort of making commands return fast instead of e.g. waiting for service restarts and apply changes async. Follow-up work will move the component updates to a controller as well.

Relates and works towards fixing:
https://github.com/canonical/k8s-snap/issues/321 and https://github.com/canonical/k8s-snap/issues/277


bschimke95 (approved):
LGTM 

image

PR #351, opened 2024-04-18
Support signed kubelet configmap

neoaggelos:
### Summary

Add machinery to sign and verify the kubelet configmap contents. We currently specify `nil` keys everywhere for now.

### Changes

Extend `types.KubeletFromConfigMap()` and `types.Kubelet.ToConfigMap()` to accept an optional ECDSA public and private key respectively.

If a non-nil key is passed:

- We hash the configuration contents (using `Kubelet.hash()`)
- In `Kubelet.ToConfigMap()` we use the private key and include it in the configmap as `k8sd-mac`
- In `types.KubeletFromConfigMap()` we use the public key to verify the hash and return an error otherwise.

### Notes

- `ecdsa.SignASN1` will only sign `hash[:curveBitsize/8]`. We ensure that we don't verify mismatching signatures by ensuring that the key is large enough. The hash is a sha256 sum (32 bytes long), therefore we need at least `elliptic.P256()`
- Added multiple tests to verify that any mutations to the configmap will result in it not being verified.

### DIFF
--- a/src/k8s/pkg/k8sd/controllers/node_configuration.go
+++ b/src/k8s/pkg/k8sd/controllers/node_configuration.go
@@ -67,7 +67,7 @@ func (c *NodeConfigurationController) Run(ctx context.Context) {
 }
 
 func (c *NodeConfigurationController) reconcile(ctx context.Context, configMap *v1.ConfigMap) error {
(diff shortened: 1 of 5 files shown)


bschimke95 (approved):
LGTM, awesome work

PR #356, opened 2024-04-19
Refactor built-in features

neoaggelos:
### Summary

This is preliminary work to allow moving the feature reconciliation to a separate controller. This is a huge PR but it is essentially the same work for all supported built-in features.

### Changes

- add `pkg/k8sd/features` package with, `ApplyFeature(...)` functions
- all manifest management is done internally, with the `Manager` interface, using `manager.Apply()`

### Important changes

- Clean up of unnecessary duplicate code in `Reconcile$FooComponent()` functions that handled enable vs refresh vs delete, this is now handled in `manager.Apply()` once.
- Do not require the `ApplyFeature()` functions to know whether the component was previously enabled or not. They retrieve this from the history (see the `action.NewHistory`.
- `manager.Apply()` returns a bool to note if there were changes after applying the new changes.
- Adjust component reconciliation logic on the cluster bootstrap and `putClusterConfig` handler. These will be moved to a separate controller as a next PR.
- Remove logic that directly edited service arguments, this is now handled in the node config controller


src/k8s/pkg/k8sd/features/features.go
```diff
@@ -0,0 +1,47 @@
+package features
+
+import "path"
+
+var (
+	// featureDNS is manifests for the built-in featureDNS feature, powered by CoreDNS.
```
bschimke95, line 6: `built-in DNS feature` stutters a bit less

src/k8s/pkg/k8sd/features/manager.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (h *helmManager) Apply(ctx context.Context, f feature, desired state, values map[string]any) (bool, error) {
+	cfg, err := h.newActionConfiguration(f.namespace)
+	if err != nil {
+		return false, fmt.Errorf("failed to create action configuration: %w", err)
+	}
+
+	isInstalled := true
+	var oldConfig map[string]interface{}
```
bschimke95, line 46: nit: `any`

src/k8s/pkg/k8sd/features/manager.go
```diff
@@ -0,0 +1,114 @@
...
+		if err != nil {
+			return false, fmt.Errorf("failed to load manifest for %s: %w", f.name, err)
+		}
+
+		release, err := upgrade.RunWithContext(ctx, f.name, chart, values)
+		if err != nil {
+			return false, fmt.Errorf("failed to upgrade %s: %w", f.name, err)
+		}
+
+		return !jsonEqual(oldConfig, release.Config), nil
```
bschimke95, line 95: nit: maybe add an explanation or TODO on why we can not usr `reflect.DeepEqual` here.

bschimke95 (commented):
Great work, only minor comments

bschimke95 (approved):

PR #359, opened 2024-04-20
Simplify checks for empty configs

neoaggelos:
### Summary

No functional changes, all configs now use pointers, so `.Empty()` checks can be much simpler


bschimke95 (approved):
Cool stuff!

PR #360, opened 2024-04-20
Sign and verify configmap/k8sd-config

neoaggelos:
### Summary

Sign and verify the k8sd-config configmap contents using an RSA keypair

### Changes

- Generate an RSA keypair along with the control plane certificates and store in cluster config
- Pass the public key to worker nodes
- Use the private key in the update node config controller to sign the configmap
- Use the public key in the node config controller to verify that the configmap signature matches
- Switch to RSA keys from ECDSA. RSA gives us deterministic signatures, which is useful because it makes `Apply()` a no-op when there are no config updates.


src/k8s/pkg/k8sd/controllers/node_configuration_test.go
```diff
@@ -129,7 +130,8 @@ func TestConfigPropagation(t *testing.T) {
 		return &k8s.Client{Interface: clientset}, nil
 	})
 
-	go configController.Run(ctx)
+	// TODO: add test with signing key
```
bschimke95, line 133: FYI: I created a card for this.

src/k8s/pkg/k8sd/pki/load.go
```diff
@@ -19,29 +19,59 @@ func loadCertificate(certPEM string, keyPEM string) (*x509.Certificate, *rsa.Pri
...
+	key, err := LoadRSAPrivateKey(keyPEM)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to load RSA private key: %w", err)
 	}
+
 	return cert, key, nil
 }
+
+// LoadRSAPrivateKey parses the specified PEM block and return the rsa.PrivateKey.
+func LoadRSAPrivateKey(keyPEM string) (*rsa.PrivateKey, error) {
```
bschimke95, line 35: We'll also need some tests for this. Feel free to add is as a task to the already created card.

bschimke95 (commented):
Great work, left some comments

src/k8s/pkg/k8sd/controllers/update_node_configuration.go
```diff
@@ -95,7 +96,13 @@ func (c *UpdateNodeConfigurationController) Run(ctx context.Context, getClusterC
 }
 
 func (c *UpdateNodeConfigurationController) reconcile(ctx context.Context, client *k8s.Client, config types.ClusterConfig) error {
-	cmData, err := config.Kubelet.ToConfigMap(nil)
+	keyPEM := config.Certificates.GetK8sdPrivateKey()
```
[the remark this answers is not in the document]
neoaggelos: it's okay to leave as separate steps. this allows us to avoid tying to rsa, in case we want to do something different in the future.

bschimke95 (approved):
LGTM!

PR #363, opened 2024-04-22
Expose Close() for dqlite client and use Close() where relevant

eaudetcobello:


src/k8s/pkg/client/dqlite/client.go
```diff
@@ -59,3 +59,14 @@ func NewClient(ctx context.Context, opts ClientOpts) (*Client, error) {
...
+func (c *Client) Close(ctx context.Context) error {
+	client, err := c.clientGetter(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create dqlite client: %w", err)
+	}
+	if err = client.Close(); err != nil {
+		return fmt.Errorf("failed to close dqlite client: %w", err)
+	}
+	return nil
+}
```
neoaggelos, line 72: this is just creating a new client and closing it, which is not what we want. Instead, we should be closing the dqlite clients we create in

- https://github.com/canonical/k8s-snap/blob/0602a3cc44b45bade210b9481790c338752aee03/src/k8s/pkg/client/dqlite/remove.go#L12
- https://github.com/canonical/k8s-snap/blob/95ef8737383d7914ac8d60bcbea6641febb3e69b/src/k8s/pkg/client/dqlite/list.go#L9

neoaggelos (changes requested):

PR #364, opened 2024-04-23
Use strict json decoder for API requests

bschimke95:
The current `json.NewDecoder()` parses unknown fields and invalid json which might cause silent failures. With `DisallowUnknownFields` we now only allow valid json with known fields.

### DIFF
--- a/src/k8s/pkg/k8sd/api/cluster_bootstrap.go
+++ b/src/k8s/pkg/k8sd/api/cluster_bootstrap.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"encoding/json"
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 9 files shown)


neoaggelos (approved):

neoaggelos (commented):
Should we also add this in?

- https://github.com/canonical/k8s-snap/blob/15688515506b62821e83831b50771e174e8bbe17/src/k8s/pkg/k8sd/app/hooks_bootstrap.go#L109

bschimke95:
I explicitly removed this again because of:
https://github.com/canonical/k8s-snap/actions/runs/8796749988/job/24140630753#step:7:350

neoaggelos:
Ah, I see it's the microcluster response, so yes, that's OK.

PR #365, opened 2024-04-23
Update requirements in 'getting started' guide

bschimke95:
The Juju controller will already claim 20G which causes follow-up issues if the user only created e.g. a 20G VM.


beliaev-maksim:
LGTM

neoaggelos (approved):

PR #366, opened 2024-04-23
Move `utils/database` out of the `utils` package

mateoflorido:
## Overview
This pull request refactors the database utils out of the `utils` package. Also, merges the `vals` package with `utils`.


src/k8s/api/v1/types_test.go
```diff
@@ -3,7 +3,7 @@ package v1
 import (
 	"testing"
 
-	"github.com/canonical/k8s/pkg/utils/vals"
+	"github.com/canonical/k8s/api/v1/testutils"
```
neoaggelos, line 6: Revert this, and use `package v1_test` above (such that the test code can use `pkg/utils` without the package importing types.

src/k8s/api/v1/testutils/pointer.go
```diff
@@ -0,0 +1,5 @@
+package testutils
+
+func Pointer[T any](value T) *T {
```
neoaggelos, line 3: Revert this, see other comment

neoaggelos (changes requested):
No need to introduce the testutils package. This import cycle is caused because the API tests are defined in `package v1` instead of `package v1_test`

neoaggelos (approved):

PR #370, opened 2024-04-24
Fixup unit test assertion

neoaggelos:
Looks like something slipped in #289. As written, the test does not assert anything


kwmonroe (approved):
Lgtm!

PR #371, opened 2024-04-24
Adjust trigger channels on the UpdateNodeConfigController

neoaggelos:
- triggerCh is passed during creation, as a receive-only channel
- reconciledCh is internal and there is a receive-only accessor for downstream

Also make minor naming adjustments


bschimke95 (approved):
LGTM!

PR #372, opened 2024-04-24
Add inspection script

eaudetcobello:
Based on: https://github.com/canonical/microk8s/blob/master/scripts/inspect.sh

image


bschimke95 (commented):
Going in the right direction.

Please also collect the SBOM and the k8sd database content and updated the bug template.
Also, the scripts directory is not included in the snap. Move this folder in the k8s directory or add a new entry in the `snapcraft.yaml`

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -37,4 +31,4 @@ about: Something is not working
 #### Are you interested in contributing with a fix?
 <!-- yes/no, or @mention maintainers. Community contributions are welcome. -->
 
-<!-- Thank you for making the k8s-snap better -->
+<!-- Thank you for making the k8s-snap better -->
```
neoaggelos, line 34: since we're here

```suggestion
<!-- Thank you for making k8s-snap better -->
```

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,120 @@
...
+    printf -- '\033[32m SUCCESS: \033[0m %s\n' "$1"
+}
+
+function log_info {
+	printf -- '\033[34m INFO: \033[0m %s\n' "$1"
+}
+
+function log_warning_red {
+	printf -- '\033[31m WARNING: \033[0m %s\n' "$1"
+}
```
neoaggelos, line 15: let's be consistent with indentation. also, please use 2 spaces, like we have in other places

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env bash
+
+INSPECT_DUMP=$(pwd)/inspection-report
+
+SVC_ARGS_DIR=/var/snap/k8s/common/args
+K8SD_STATE_DIR=/var/snap/k8s/common/var/lib/k8sd/state
+K8SD_BIN=/snap/k8s/current/bin/k8sd
+SBOM_FILE=/snap/k8s/current/bom.json
```
neoaggelos: Let's use some prefixes here so that we help reading the code below. Also, do not define things that are used once (`$K8S_SNAP` and `$K8S_SNAP_COMMON` can just be used there).

```
K8S_SNAP=/snap/k8s/current
K8S_SNAP_COMMON=/var/snap/k8s/common

# then 
K8S_ARGS_DIR="${K8S_SNAP_COMMON}/args"
...
```

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,120 @@
...
+
+    systemctl status "snap.$service" &> "$status_file"	
+	
+    if grep -q "active (running)" "$status_file"; then
+        log_info "Service $service is running"
+    else
+        log_info "Service $service is not running"
+    fi
+
+	journalctl -n 100000 -u "snap.$service" &> "$INSPECT_DUMP/$service/journal.log"
```
neoaggelos, line 69: we also need some way to note if services are _expected_ to run.

e.g. if we are on a control plane node, the apiserver-proxy service is not expected to run.

further, add a check for the `NRestart` property, which is non-zero if the service is constantly restarting due to errors. That's a useful metric to have

neoaggelos (changes requested):
Going in the right direction, but needs work

.github/ISSUE_TEMPLATE/bug_report.md
bschimke95, line 1: The template is now a yaml file `bug_report.yaml`. Apparently this is the new way of templating in Github.

.github/ISSUE_TEMPLATE/bug_report.yaml
neoaggelos, line 1: Why is this diff here?

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,136 @@
...
+  printf -- '\033[34m INFO: \033[0m %s\n' "$1"
+}
+
+function log_warning () {
+  printf -- '\033[33m WARNING: \033[0m %s\n' "$1"
+}
+
+function log_warning_red {
+  printf -- '\033[31m WARNING: \033[0m %s\n' "$1"
+}
```
neoaggelos, line 19: we only use `log_warning_red` ?

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,136 @@
...
+
+function collect_args {
+  log_info "Copy service args to the final report tarball"
+  cp -r --no-preserve=mode,ownership /var/snap/k8s/common/args "$INSPECT_DUMP"
+}
+
+function collect_cluster_info {
+  log_info "Copy k8s cluster-info dump to the final report tarball"
+  k8s kubectl cluster-info dump &> "$INSPECT_DUMP/cluster-info.log"
```
neoaggelos, line 32: let's do `k8s kubectl cluster-info dump --output-directory` instead, as that has a much better structure than the single file

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,136 @@
...
+
+  log_info "Copy k8s diagnostics to the final report tarball"
+  k8s kubectl version &>"$INSPECT_DUMP/k8s-version.log"
+  k8s status &>"$INSPECT_DUMP/k8s-status.log"
+  k8s get &>"$INSPECT_DUMP/k8s-get.log"
+  k8s kubectl get cm k8sd-config -n kube-system -o yaml &>"$INSPECT_DUMP/k8sd-configmap.log"
+  k8s kubectl get cm -n kube-system &>"$INSPECT_DUMP/k8s-configmaps.log"
+
+  cp --no-preserve=mode,ownership /var/snap/k8s/common/var/lib/k8s-dqlite/cluster.yaml "$INSPECT_DUMP/k8s-dqlite-cluster.yaml"
+  cp --no-preserve=mode,ownership /var/snap/k8s/common/var/lib/k8s-dqlite/info.yaml "$INSPECT_DUMP/k8s-dqlite-info.yaml"
```
neoaggelos, line 58: also do a `ls -la /var/snap/k8s/common/var/lib/k8s-dqlite &> k8s-dqlite-files.log` and similar for k8sd

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,136 @@
...
+
+  systemctl status "snap.$service" &> "$status_file"
+
+  if grep -q "active (running)" "$status_file"; then
+    log_info "Service $service is running"
+  else
+    log_info "Service $service is not running"
+  fi
+
+  printf -- "%s $(systemctl show "snap.$service" -p NRestarts)\n" "$service" >> "$INSPECT_DUMP/nrestarts.log"
```
neoaggelos, line 80: isn't it better if this is added to `$service/nrestarts` instead? Also, we need a warning in case this is not zero.

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,136 @@
...
+
+  journalctl -n 100000 -u "snap.$service" &>"$INSPECT_DUMP/$service/journal.log"
+}
```
neoaggelos, line 83: we need a warning if the service is not "healthy" but we expect it to be. e.g. worker nodes need the apiserver-proxy service to be up. control plane nodes need the control plane services to be up. all nodes need kubelet, containerd, etc

neoaggelos (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yaml
eaudetcobello: Originally it was a simple modification to bug_report.md but took the opportunity to transition to github's new .yaml bug report format.

k8s/scripts/inspect.sh
eaudetcobello: So if someone sends us their archive it'll be less convenient to have to cd into every $service directory to check if there were restarts?

bschimke95 (approved):
LGTM

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,169 @@
...
+  printf -- 'Inspection ran on a control plane node.' >"$INSPECT_DUMP/is-control-plane-node"
+else
+  printf -- 'Running inspection on a worker node\n'
+  printf -- 'Inspection ran on a worker node.' >"$INSPECT_DUMP/is-worker-node"
+fi
+
+control_plane_services=("k8s.containerd" "k8s.kube-proxy" "k8s.k8s-dqlite" "k8s.k8sd" "k8s.kube-apiserver" "k8s.kube-controller-manager" "k8s.kube-scheduler" "k8s.kubelet")
+worker_services=("k8s.containerd" "k8s.k8s-apiserver-proxy" "k8s.kubelet" "k8s.k8sd" "k8s.kube-proxy")
+
+if is_control_plane_node; then
```
bschimke95, line 126: The content of this if statement is duplicated. Just move it into a function and invoke it with either the worker services or control_plane services depending

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,169 @@
...
+
+  local status_file
+  status_file="$INSPECT_DUMP/$service/systemctl.log"
+
+  systemctl status "snap.$service" &>"$status_file"
+
+  local n_restarts
+  n_restarts=$(systemctl show "snap.$service" -p NRestarts | cut -d'=' -f2) 
+
+  printf -- "%s NRestarts=%s" "$service" "$n_restarts" >> "$INSPECT_DUMP/nrestarts.log"
```
bschimke95, line 87: nit: the file is already named `nrestarts.log`.
Let's just have something like 
```
$service -> $s restarts
```
or even 
```
$service -> $s
```

k8s/scripts/inspect.sh
```diff
@@ -0,0 +1,179 @@
...
+  fi
+
+  journalctl -n 100000 -u "snap.$service" &>"$INSPECT_DUMP/$service/journal.log"
+}
+
+function collect_network_diagnostics {
+  log_info "Copy network diagnostics to the final report tarball"
+  ip a &>"$INSPECT_DUMP/ip-a.log"
+  ip r &>"$INSPECT_DUMP/ip-r.log"
+  iptables-save &>"$INSPECT_DUMP/iptables.log"
```
neoaggelos, line 100: this might not be sufficient, we would also need `iptables-legacy-save`, and `ip6tables-save`, `ip6tables-legacy-save`.

and make sure that we do not fail if any of the commands here fail (e.g. due to missing binaries). For example, if `ip6tables-legacy-save` is not present on the machine, we don't want the inspection script to fail.

neoaggelos (approved):
LGTM! Nice work! Just a final comment

PR #375, opened 2024-04-24
Feature Controller

neoaggelos:
### Summary

Delegate management of all built-in features to a separate FeatureController


bschimke95 (approved):
LGTM

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -44,5 +47,32 @@ func (a *App) onStart(s *state.State) error {
...
+				if err := s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
+					if _, err := database.SetClusterConfig(ctx, tx, types.ClusterConfig{
+						Kubelet: types.Kubelet{ClusterDNS: utils.Pointer(dnsIP)},
+					}); err != nil {
+						return fmt.Errorf("failed to update cluster configuration for dns=%s: %w", dnsIP, err)
+					}
+					return nil
+				}); err != nil {
+					return fmt.Errorf("database transaction to update cluster configuration failed: %w", err)
+				}
```
eaudetcobello, line 67: I'm trying to make sense of this but the nesting is confusing me. How do I read what this does?

eaudetcobello (commented):
Nice work. One comment that's more of a clarification

src/k8s/pkg/k8sd/app/hooks_start.go
neoaggelos: this is a callback function that will run when the deployment of coreDNS has given us a service IP. When that happens, we start a database transaction `s.Database.Transaction(ctx, func(ctx, *tx) error { ... }) error` and update the cluster configuration accordingly. If any of that fails, we wrap the error and return to the caller

PR #380, opened 2024-04-26
Add update part versions workflow

berkayoz:
Workflow checks upstream repositories for new versions under the same major and creates a PR with the newest version.

### DIFF
--- a/.github/workflows/update-versions.yaml
+++ b/.github/workflows/update-versions.yaml
@@ -1,0 +1,53 @@
+name: Create PRs for new versions of parts
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 10 * * *"
+
+jobs:
+  update-versions:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        part: ["cni", "containerd", "helm", "runc"]
+
+    steps:
+      - name: Checking out repo
+        uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Fetch and filter upstream tags
+        id: fetch-upstream
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          sudo apt-get -qq update
+          sudo apt-get -qq install -y jq
+
+          repository="$(cat "build-scripts/components/${{ matrix.part }}/repository")"
+          current_tag="$("build-scripts/components/${{ matrix.part }}/version.sh")"
+
+          IFS='.' read -r -a current_tag_parts <<< "$current_tag"
+
+          git clone $repository part
+
+          # Get the latest tag for major version, excluding -rc etc.
+          latest_tag="$(git -C part tag --sort=-v:refname | grep "${current_tag_parts[0]}" | grep -vE "-" | head -n 1)"
+
+          rm -rf part
+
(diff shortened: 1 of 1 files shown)


bschimke95 (approved):
LGTM, just out of curiosity: can we not configure dependabot to do something like this?

berkayoz:
@bschimke95,
I am not sure about dependabot but it seems like we can write a custom manager for Renovate bot that can be used for all starcraft applications(snaps, rocks etc.). We would likely go through parts that has `source-type: git` and check for newer versions. Seemed a bit out of scope and our current build system manages the sources itself.

bschimke95:
Alright, thanks for the context.

.github/workflows/update-versions.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+          repository="$(cat "build-scripts/components/${{ matrix.part }}/repository")"
+          current_tag="$("build-scripts/components/${{ matrix.part }}/version.sh")"
+
+          IFS='.' read -r -a current_tag_parts <<< "$current_tag"
+
+          git clone $repository part
+
+          # Get the latest tag for major version, excluding -rc etc.
+          latest_tag="$(git -C part tag --sort=-v:refname | grep "${current_tag_parts[0]}" | grep -vE "-" | head -n 1)"
```
neoaggelos, line 36: I think this might need finer grained control. E.g. we might not want to jump to newer minor or major versions of components (e.g. containerd 2 vs 1.6.xx). This would create noisy PRs for us in that case.

.github/workflows/update-versions.yaml
```diff
@@ -0,0 +1,53 @@
...
+          export DEBIAN_FRONTEND=noninteractive
+          sudo apt-get -qq update
+          sudo apt-get -qq install -y jq
+
+          repository="$(cat "build-scripts/components/${{ matrix.part }}/repository")"
+          current_tag="$("build-scripts/components/${{ matrix.part }}/version.sh")"
+
+          IFS='.' read -r -a current_tag_parts <<< "$current_tag"
+
+          git clone $repository part
```
neoaggelos, line 33: is `--depth 1` faster here?

.github/workflows/update-versions.yaml
neoaggelos: Also, this works for any git repo, but given that we source everything from github, perhaps we could use the GitHub API to fetch the releases instead?

berkayoz: I also though about the api to fetch releases but ended up sticking to tags so it can work with any repository.

For the finer grained control, I'm open to suggestions. I tried sticking to semver and went for latest minor version so no major skips from `1.x.x` to `2.x.x` happens. Doing only patch level felt like defeating the purpose of the action. Currently this might create a lot of PRs which might get noisy, I can also keep a single PR active at one time and update that if not merged.

.github/workflows/update-versions.yaml
berkayoz: Wouldn't that not fetch the tags(shallow clone)? We can do `git clone --depth 1` and `git fetch --tags` but not sure if there are any time savings there.

addyess (approved):
Let's go with this.  if it gets too noisy, we can always adjust.  There's clearly nothing broken here

PR #384, opened 2024-04-29
Adjust snap.Interface to return Kubernetes clients

neoaggelos:
### Summary

Adjust the `snap` interface to return Kubernetes clients, such that they can be easily mocked in unit tests.

### Changes

- Move `pkg/utils/k8s` -> `pkg/client/kubernetes`
- Move helm client from `pkg/k8sd/features` -> standalone `pkg/client/helm`

### Notes

1. This reverses the package dependencies. Instead of `k8s.NewClient(snap, ...)`, we now do `snap.KubernetesClient(...)`. Also, we now do `snap.HelmClient(...)` instead of `features.newHelm(snap)`
2. Adds a mock `pkg/client/helm/mock` and a `TestApplyMetricsServer` unit test, to show how the changes allow us to write effective unit tests for more parts of the code base.
3. An example where things are simplified a lot with this change is the `pkg/k8sd/controllers`. Previously, we passed the snap as well as a Kubernetes client factory, in order to be able to mock the Kubernetes client. This is no longer required, see the relevant changes.


src/k8s/pkg/client/helm/interface.go
```diff
@@ -0,0 +1,13 @@
+package helm
+
+import "context"
+
+// Client handles the lifecycle of charts (manifests + config) on the cluster.
+type Client interface {
+	// Apply ensures the state of a InstallableChart on the cluster.
+	// When state is StatePresent, Apply will install or upgrade the chart using the specified values as configuration. Apply returns true if the chart was not installed, or any values were changed.
```
bschimke95, line 8: Apply will install the chart if it is not present and state is StatePresent.
So it should return true if the chart was installed, no?

bschimke95 (approved):
LGTM, great work!
Can you create follow up cards for the other unittest and the new TODO?

src/k8s/pkg/client/helm/interface.go
neoaggelos: yes, it will.

- if the chart was not installed: https://github.com/canonical/k8s-snap/blob/b71e82d1bb06a57927a6d52fe7ade1d8c14b140d/src/k8s/pkg/client/helm/client.go#L87
- if the chart was installed, we return true if any changes were made to the values https://github.com/canonical/k8s-snap/blob/b71e82d1bb06a57927a6d52fe7ade1d8c14b140d/src/k8s/pkg/client/helm/client.go#L105

bschimke95: Ah so it returns true if the chart was not installed *before* apply was executed. Maybe adding that to the comment makes it more obvious.

On another note: are we missing `isInstalled && desired == StatePresent` in the switch statement? It should basically just return `false,nil` as the default case but the default case states that it should never happen.

neoaggelos: That's https://github.com/canonical/k8s-snap/pull/384/files#diff-5c4fa01064f7687e37712f85a6a929c0ec362e8c453bb7eb460cdca083f14341R88

PR #385, opened 2024-04-29
Harden CI

louiseschmidtgen:
## Description
This PR hardens CI by:

- [x] restricting permissions of the gh token
- [x] installing a [security agent on the Github-hosted runner](https://github.com/step-security/harden-runner/tree/v2/). We can potentially add a rule to this security agent that only allows outbound traffic to a list of endpoints.

Note: I've also verified our chron-jobs work with my adjustemnts in this [commit](https://github.com/canonical/k8s-snap/pull/385/commits/e4e88cf60f35c0a257217d0db02bb513f12df14e).


bschimke95:
I think the integration tests fail because of the limited permissions you give in the GH workflow (the tests pass locally). Try to give more permissions to the `integration-tests` workflow.

.github/workflows/scorecard.yaml
```diff
@@ -31,6 +31,10 @@ jobs:
       # actions: read
```
eaudetcobello, line 31: This file has lots of extraneous comments, let's remove them.

.github/workflows/scorecard.yaml
```diff
@@ -31,6 +31,10 @@ jobs:
...
 
     steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@v2
+        with:
+          egress-policy: audit
       - name: "Checkout code"
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
eaudetcobello, line 39: We use checkout@v4 elsewhere, we should do the same here.

eaudetcobello (changes requested):
Nice work, couple of comments

.github/workflows/scorecard.yaml
louiseschmidtgen: I can do that, just have to silence it too 👍🏼

bschimke95 (approved):
LGTM

eaudetcobello (approved):

PR #386, opened 2024-04-30
Parse k8s set with mapstructure

neoaggelos:
### Summary

Use `mapstructure` to parse `k8s set` command arguments

### Changes

- Allows parsing string slice inputs as `val1,val2` or YAML (or JSON) formatted strings
- Add unit tests for `k8s set` arguments

### Notes

- If `mapstructure.NewDecoder` fails we panic, as all errors are programming bugs (see implementation of `NewDecoder`)
- We check for the list of known keys such that we present a meaningful user error, rather than a mapstructure error that the field is unknown.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_set.go
+++ b/src/k8s/cmd/k8s/k8s_set.go
@@ -2,13 +2,12 @@ package k8s
 
 import (
 	"fmt"
-	"github.com/canonical/k8s/pkg/utils"
-	"strconv"
 	"strings"
-	"unicode"
 
 	apiv1 "github.com/canonical/k8s/api/v1"
 	cmdutil "github.com/canonical/k8s/cmd/util"
+	"github.com/canonical/k8s/pkg/utils"
+	"github.com/mitchellh/mapstructure"
 	"github.com/spf13/cobra"
 )
 
@@ -34,7 +33,7 @@ func newSetCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 			config := apiv1.UserFacingClusterConfig{}
 
 			for _, arg := range args {
-				if err := updateConfig(&config, arg); err != nil {
+				if err := updateConfigMapstructure(&config, arg); err != nil {
 					cmd.PrintErrf("Error: Invalid option %q.\n\nThe error was: %v\n", arg, err)
 					env.Exit(1)
 				}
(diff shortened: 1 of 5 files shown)


src/k8s/cmd/k8s/k8s_set.go
```diff
@@ -66,119 +65,69 @@ func newSetCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
+		return fmt.Errorf("unknown option key %q", key)
+	}
+
+	if err := decoder.Decode(toRecursiveMap(key, value)); err != nil {
+		return fmt.Errorf("invalid option %q: %w", arg, err)
 	}
 	return nil
 }
+
+func toRecursiveMap(key, value string) map[string]any {
```
bschimke95, line 127: nit: does it make sense to move it to `util/mapstructure.go`?

src/k8s/pkg/utils/mapstructure.go
```diff
@@ -0,0 +1,44 @@
...
+	"reflect"
+	"strings"
+	"unicode"
+
+	"github.com/mitchellh/mapstructure"
+	"gopkg.in/yaml.v2"
+)
+
+// YAMLToStringSliceHookFunc returns a mapstructure.DecodeHookFunc that converts string to []string by parsing YAML.
+func YAMLToStringSliceHookFunc(f reflect.Kind, t reflect.Kind, data interface{}) (interface{}, error) {
```
bschimke95, line 13: nit: `any` instead of `interface{}`

bschimke95 (approved):
LGTM

src/k8s/cmd/k8s/k8s_set.go
```diff
@@ -66,119 +65,69 @@ func newSetCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
-	case "metrics-server.enabled":
-		v, err := strconv.ParseBool(value)
-		if err != nil {
-			return fmt.Errorf("invalid boolean value for metrics-server.enabled: %w", err)
-		}
-		config.MetricsServer.Enabled = &v
-	default:
-		return fmt.Errorf("unknown config key %q", key)
+	if _, ok := knownSetKeys[key]; !ok {
+		return fmt.Errorf("unknown option key %q", key)
```
louiseschmidtgen, line 118: Should we add "consult `k8s get` command to see configurable options to the error message?

src/k8s/cmd/k8s/k8s_set.go
neoaggelos: let's keep it in mind when we need in a second place in the code base.

src/k8s/pkg/utils/mapstructure.go
neoaggelos: going by https://github.com/mitchellh/mapstructure/blob/main/mapstructure.go#L185-L197

PR #389, opened 2024-05-01
Add proxy config docs

evilnick:
Adds a reference and how to page for proxy configuration


docs/src/snap/howto/proxy.md
```diff
@@ -0,0 +1,31 @@
...
+```
+HTTPS_PROXY=http://squid.internal:3128
+HTTP_PROXY=http://squid.internal:3128
+NO_PROXY=10.0.0.0/8,192.168.0.0/16,127.0.0.1,172.16.0.0/12
+https_proxy=http://squid.internal:3128
+http_proxy=http://squid.internal:3128
+no_proxy=10.0.0.0/8,192.168.0.0/16,127.0.0.1,172.16.0.0/12
+```
+
+Note that you may need to restart for these settings to take effect.
```
addyess, line 31: any need to add how this is done via the charm/juju config?

```bash
juju model-config \
      juju-http-proxy=http://squid.internal:3128 \
     juju-https-proxy=http://squid.internal:3128 \
     juju-no-proxy=10.0.8.0/24,192.168.0.0/16,127.0.0.1,10.152.183.0/24
```

It's worth pointing out `10.152.183.0/24` needs to be in the list because it's the service-cidr.  Without this -- pods that need to reach the cluster's kubernetes-api cannot b/c they'd reach out through the proxy

docs/src/snap/howto/proxy.md
```diff
@@ -0,0 +1,31 @@
...
+## Adding proxy configuration
+
+Edit the `/etc/environment` file and add the relevant URLs
+
+```{note} It is important to add whatever address ranges are used by the
+ cluster itself to the `NO_PROXY` and `no_proxy` variables.
+```
+
+For example, assume we have a proxy running at `http://squid.internal:3128` and
+we are using the networks `10.0.0.0/8`,`192.168.0.0/16` and `172.16.0.0/12`. We
```
addyess, line 19: it could be nice to have a run-down of what these CIDRs are.  
Usually you need no-proxies for:

* service-cidr
* pod-cidr
* local network cidr

addyess: the defaults for those in the snap are [here](https://github.com/canonical/k8s-snap/blob/af5c83997c1f98d21ced6b17a43a4b4ab8690558/src/k8s/pkg/k8sd/types/cluster_config_defaults.go#L12-L17)

addyess (commented):
Couple of notes regarding the CIDRs, and Juju config of the same bits

docs/src/snap/howto/proxy.md
evilnick: Okay, I created a different howto for the charm and cross referenced everything. The reference page in the charm docs is the same as the snap one, so i linked to both Howtos from there.

addyess (approved):
Really beautiful

docs/src/snap/howto/proxy.md
evilnick: I added a not for this to both Howto guides and explained it slightly more in the reference page

PR #390, opened 2024-05-01
Sync registry.k8s.io/pause and ghcr.io/canonical/k8s-snap/pause

eaudetcobello:


eaudetcobello:
See result: https://github.com/canonical/k8s-snap/pkgs/container/pause/

.github/workflows/images.yaml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Skopeo
+        run: |
+          docker run quay.io/skopeo/stable:v1.15 sync \
+            --src docker \
+            --dest docker \
+            --all registry.k8s.io/pause ghcr.io/canonical \
```
neoaggelos, line 17: let's not mirror all images, only the ones we care about. we'll need similar things for CSI images, where syncing everything would be unfeasible.

eaudetcobello: Ok, added a YAML file that allows selection by semver. Is >= 3.0 good?

.github/workflows/images.yaml
```diff
@@ -0,0 +1,27 @@
+on:
+  push:
+    branches:
+      - "eaudetcobello/KU-617"
```
bschimke95, line 4: Reminder: update before merge

bschimke95 (approved):
LGTM

.github/workflows/images.yaml
```diff
@@ -0,0 +1,27 @@
...
+        run: |
+          cat <<EOF > $HOME/registry.k8s.io-config.yaml
+          registry.k8s.io:
+            images-by-semver:
+              pause: "= 3.7"
+          EOF
+          docker run -v $HOME:$HOME quay.io/skopeo/stable:v1.15 sync \
+            --src yaml \
+            --dest docker \
+            $HOME/registry.k8s.io-config.yaml ghcr.io/canonical \
```
neoaggelos, line 25: where is this file?

neoaggelos (changes requested):

.github/workflows/images.yaml
eaudetcobello: It's created just above. https://github.com/canonical/k8s-snap/pull/390/files/97066968126db06dc133097a1bffa0472197d239#diff-d0a3d6684c78a148cbf0725d5fe8b5aab6431da05b698a82c9e015516f3020baR17

.github/workflows/sync-registry.yaml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  publish:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Sync registry.k8s.io to ghcr.io/canonical
+        run: |
+          ./${GITHUB_WORKSPACE}/.github/scripts/sync-registry.sh ${{ github.actor }} ${{ secrets.GITHUB_TOKEN }} ${GITHUB_WORKSPACE}
```
addyess, line 14: is there ANY chance you could have your shell script read the actor and token from an ENV variable? rather than a command line argument?  I'm not fond of passing tokens on the commandline

.github/data/sync-registry-config.yaml
```diff
@@ -0,0 +1,3 @@
+k8s.gcr.io/sig-storage:
+  images-by-semver:
+    pause: "= 3.7"
```
neoaggelos, line 3: newline at EOF

.github/scripts/sync-registry.sh
```diff
@@ -0,0 +1,12 @@
...
+actor=${ACTOR}
+token=${TOKEN}
+dir=${DIR}
+
+docker run -v "$dir":/tmp/ quay.io/skopeo/stable:v1.15 sync \
+  --src yaml \
+  --dest docker \
+  /tmp/.github/data/sync-registry-config.yaml ghcr.io/canonical \
+  --format oci \
+  --dest-creds "$actor":"$token"
```
neoaggelos, line 12: newline at EOF. also, this can just be `"${ACTOR}:${TOKEN}"`

.github/scripts/sync-registry.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+actor=${ACTOR}
+token=${TOKEN}
+dir=${DIR}
```
neoaggelos, line 5: we already have `build-scripts/hack` for these scripts, let's add it there

.github/scripts/sync-registry.sh
```diff
@@ -0,0 +1,12 @@
...
+
+docker run -v "$dir":/tmp/ quay.io/skopeo/stable:v1.15 sync \
```
neoaggelos, line 7: avoid mounting the whole directory here. instead, you can do:

```bash
DIR=`realpath $(dirname "${0}")`

docker run -v "${DIR}/registry-k8s-io.yaml":/config.yaml quay.io/skopeo/stable:v1.15 sync \
  --src yaml \
  --dest docker \
  /config.yaml ghcr.io/canonical \
  --format oci \
  --dest-creds "$actor":"$token"
```

.github/workflows/sync-registry.yaml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Sync registry.k8s.io to ghcr.io/canonical
+        env:
+          ACTOR: ${{ github.actor }}
+          TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DIR: ${GITHUB_WORKSPACE}
```
neoaggelos, line 16: no need to set DIR, see comment on bash script. the script can figure out its base directory on its own.

.github/workflows/sync-registry.yaml
```diff
@@ -0,0 +1,18 @@
...
+        run: |
+          ./${GITHUB_WORKSPACE}/.github/scripts/sync-registry.sh
```
neoaggelos, line 18: newline at EOF

neoaggelos (changes requested):
mostly there, minor changes and suggestions

build-scripts/hack/sync-registry.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+DIR=$(realpath $(dirname "${0}"))
+
+docker run -v "${DIR}/registry-k8s-io.yaml":/config.yaml quay.io/skopeo/stable:v1.15 sync \
+  --src yaml \
+  --dest docker \
+  /config.yaml ghcr.io/canonical \
+  --format oci \
+  --dest-creds "${ACTOR}":"${TOKEN}"
```
neoaggelos, line 9: Let's make this a `$REGISTRY_CREDS` instead, so that it matches the scopeo arg

```suggestion
  --dest-creds "${DEST_CREDS}"
```

.github/workflows/sync-registry.yaml
```diff
@@ -0,0 +1,17 @@
...
+  publish:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Sync registry.k8s.io to ghcr.io/canonical
+        env:
+          ACTOR: ${{ github.actor }}
+          TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
neoaggelos, line 15: This can be 
```
DEST_CREDS: ${{ github.actor }}:${{ services.GITHUB_TOKEN }}
```

.github/workflows/sync-registry.yaml
```diff
@@ -0,0 +1,17 @@
...
+        run: |
+          ./${GITHUB_WORKSPACE}/build-scripts/hack/sync-registry.sh
```
neoaggelos, line 17: newline at EOF

neoaggelos (commented):
Almost there, looks a lot better! Final comments

build-scripts/hack/sync-registry.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+DIR=$(realpath $(dirname "${0}"))
+
+docker run -v "${DIR}/registry-k8s-io.yaml":/config.yaml quay.io/skopeo/stable:v1.15 sync \
```
neoaggelos, line 5: where is `registry-k8s-io.yaml`?

.github/data/sync-registry-config.yaml
neoaggelos, line 3: probably need to rename this?

eaudetcobello: Yeah, don't know how that ended up here.

build-scripts/hack/sync-registry.sh
eaudetcobello: I accepted your github suggestion and missed that.

build-scripts/hack/sync-registry.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+docker run -v "${GITHUB_WORKSPACE}/.github/data/sync-registry-config.yaml":/config.yaml quay.io/skopeo/stable:v1.15 sync \
```
neoaggelos, line 3: We do not maintain this docker image, and that can cause troubles in the future. We could use `skopeo` from the ubuntu archive instead.

.github/workflows/sync-registry.yaml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Sync registry.k8s.io to ghcr.io/canonical
+        env:
+          DEST_CREDS: ${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}
+        run: |
+          chmod +x ${GITHUB_WORKSPACE}/build-scripts/hack/sync-registry.sh
```
neoaggelos, line 16: we should not chmod +x during CI, we should just chmod +x the file and commit it.

build-scripts/hack/sync-registry.sh
neoaggelos, line 3: This uses `${GITHUB_WORKSPACE}`. If run on a local machine, it would try to find `/.github/data/sync-registry-config.yaml` which is probably not what we want.

.github/data/sync-registry-config.yaml
neoaggelos, line 1: As already discussed, this file is not `.github` specific so has no reason to be in `.github/data`. It should be in `build-scripts` instead, next to all other files with a similar purpose.

.github/workflows/sync-registry.yaml
```diff
@@ -0,0 +1,17 @@
+on:
+  schedule:
+    - cron: "0 0 * * *" # Runs every midnight
```
neoaggelos, line 3: semantically, i would keep this as a `workflow_dispatch`. These images will not be changing every midnight, so no need to run this on a cron.

neoaggelos (changes requested):
This still needs work it seems.

The `skopeo` command only gets the amd64 image, but we need all architectures. Compare `docker manifest inspect registry.k8s.io/pause:3.7` with `docker manifest inspect ghcr.io/canonical/pause:3.7`

Please see https://github.com/kubernetes/registry.k8s.io/blob/main/docs/mirroring/README.md for mirroring. `crane` looks like a good candidate. `crane` also seems to only need Go (e.g. `go install github.com/google/go-containerregistry/cmd/crane@v0.19.1`, so it would be simpler than having to mess with Docker images.

neoaggelos (approved):

PR #392, opened 2024-05-02
Reuse k8s instance across feature tests

bschimke95:
Sets up an instance that enables all features and persists over the test session. All features (dns, gateway, ...) run against this instance to reduce setup/teardown time.

This reduces the integration test time by around ~18 min (58min -> 30min)

### DIFF
--- a/tests/integration/tests/conftest.py
+++ b/tests/integration/tests/conftest.py
@@ -86,7 +86,7 @@ def instances(
     snap_path = (tmp_path / "k8s.snap").as_posix()
 
     LOG.info(f"Creating {node_count} instances")
-    instances: List[util.Instance] = []
+    instances: List[harness.Instance] = []
 
     for _ in range(node_count):
         # Create <node_count> instances and setup the k8s snap in each.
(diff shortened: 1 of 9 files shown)


src/k8s/pkg/client/kubernetes/status.go
```diff
@@ -9,14 +9,14 @@ import (
...
+// WaitKubernetesEndpointAvailable waits until the kubernetes endpoint becomes available.
+func (c *Client) WaitKubernetesEndpointAvailable(ctx context.Context) error {
 	return control.WaitUntilReady(ctx, func() (bool, error) {
 		// TODO: use the /readyz endpoint instead
 		// We want to retry if an error occurs (=API server not ready)
 		// returning the error would abort, thus checking for nil
-		_, err := c.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
-		return err == nil, nil
+		endpoint, err := c.CoreV1().Endpoints("default").Get(ctx, "kubernetes", metav1.GetOptions{})
+		return err == nil && endpoint != nil, nil
```
addyess, line 19: snuck this fix in -- nice

tests/integration/tests/conftest.py
```diff
@@ -100,7 +100,40 @@ def instances(
...
     yield instances
 
-    _harness_clean(h)
+    if not config.SKIP_CLEANUP:
+        # Cleanup after each test.
+        # We cannot execute _harness_clean() here as this would also
+        # remove the session_instance. The harness ensures that everything is cleaned up
+        # at the end of the test session.
+        for instance in instances:
+            h.delete_instance(instance.id)
```
addyess, line 109: the double negative really broke my brain for a bit.  

```suggestion
    if config.SKIP_CLEANUP:
        LOG.warning("Skipping clean-up of instances, delete them on your own"
        return
    # Cleanup after each test.
    # We cannot execute _harness_clean() here as this would also
    # remove the session_instance. The harness ensures that everything is cleaned up
    # at the end of the test session.
    for instance in instances:
        h.delete_instance(instance.id)
```

addyess: is `_harness_clean` no longer used?  Can it be deleted

addyess (approved):
🐍 python approved

src/k8s/pkg/client/kubernetes/status.go
bschimke95: I rebased this PR on this fix and forgot to retarget this PR. Updated

tests/integration/tests/conftest.py
bschimke95: Thanks, removed the double-negative.

`_harness_clean` is still used when the harness is cleaned up at the end of the session. https://github.com/canonical/k8s-snap/blob/63498cfc1c7ba75cf15ce9d5db3adb94df09b316/tests/integration/tests/conftest.py#L46

It is only removed from the `instances` as this would also cleanup the session instance.

PR #393, opened 2024-05-02
Add Reference for bootstrap configuration file

berkayoz:
Add Reference for bootstrap configuration file


evilnick:
Do we need so many different files, I think it would be more useful to just put them in one.

If you are generating them from something else you could use the trick we did for the command reference  and include the generated files in one:

https://raw.githubusercontent.com/canonical/k8s-snap/main/docs/src/snap/reference/commands.md

docs/src/snap/reference/bootstrap-config-reference.md
```diff
@@ -0,0 +1,422 @@
+# Bootstrap configuration file reference
+
+A YAML file can be supplied to the `k8s bootstrap` command to configure and customize the cluster.
+This Reference section provides the format of this file by listing all available options and their details.
+
+## Format Specification
+
+### cluster-config.network
+**Type:** `object` <br>
+**Required:** `No`
```
bschimke95, line 10: I think this is a bit unnecessary. 
Just add a remark "required if bgp_enabled" in the description of the few options where this is required.

berkayoz: I tried to match with docs from other teams, it's true that we mostly don't have anything required but I also feel like explicitly stating these is worth in a Reference page. I'll leave it up to @evilnick if we should keep these

evilnick: I think there are a number of things we could probably do to improve this. i will make a card to re-evaluate it but for now I think it just needs to exist

evilnick (approved):
Thanks for your work on this @berkayoz. I will make a card to re-evaluate this in the next cycle.
I have made a few tweaks for formatting and added a link from the install docs. Feel free to merge if you are happy with these.

PR #394, opened 2024-05-02
Add tutorial for charm deployment

evilnick:
Some caveats:

- It's difficult to achieve  anything without storage ( I don't want to hack local-storage)
- It currently ends rather abruptly - suggestions welcome on how to finish it off

### DIFF
--- a/docs/src/charm/tutorial/getting-started.md
+++ b/docs/src/charm/tutorial/getting-started.md
@@ -103,7 +103,7 @@ application won't settle into a 'Ready' status, because it requires a
 connection to the control plane. This is handled in Juju by integrating the
 charms so they can communicate using a standard interface. The charm info we
 fetched earlier also includes a list of the relations possible, and from this
-we can see that the k8s-worker requires "cluster: k8s-cluster"
+we can see that the k8s-worker requires "cluster: k8s-cluster".
 
 To connect these charms and effectively add the worker to our cluster, we use
 the 'integrate' command, adding the interface we wish to connect
--- a/docs/src/charm/tutorial/getting-started.md
+++ b/docs/src/charm/tutorial/getting-started.md
@@ -59,7 +59,7 @@ argument, for example `--channel=latest/edge`.
 
 To make sure that Juju creates an instance which has enough resources to
 actually run Kubernetes, we will make use of 'constraints'. These specify the
-minimums required. For the Kubernetes control plane, the recommendation is two
+minimums required. For the Kubernetes control plane (k8s charm), the recommendation is two
 CPU cores, 16GB of memory and 40GB of disk space. Now we can go ahead and
 create a cluster:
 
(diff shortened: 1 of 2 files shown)


louiseschmidtgen (commented):
Thank you for your wonderful work on this! I've added some minor tweaks and suggestions.

docs/src/charm/tutorial/getting-started.md
```diff
@@ -1 +1,222 @@
...
+At this point Juju will fetch the charm from Charmhub, create a new instance
+according to your specification and configure and install the Kubernetes
+components (i.e. the `k8s` snap ). This may take a few minutes depending on
+your cloud. You can monitor progress by watching the Juju status output:
+
+```
+juju status --watch 2s
+```
+
+When the status reports that K8s is "idle/ready"
```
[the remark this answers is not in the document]
evilnick: ah! thanks!

docs/src/charm/tutorial/getting-started.md
```diff
@@ -1 +1,222 @@
...
+   your current config.
+```
+
+We can pipe this output directly to a config file which will just require a bit of editing:
+
+```
+juju run k8s/0 get-kubeconfig >> ~/.kube/config
+```
+
+The output includes the root of the YAML, `kubeconfig: |`, so we can just use an editor to remove that line:
```
addyess, line 166: I feel that what you've got here is more clear that having the user pipe the output through jq and writing it to a file. 

but here's the nonsense anyway
```bash
juju run k8s/0 get-kubeconfig | yq '.kubeconfig' -r > .kube/config
```

evilnick: Ah yes, cool, thansk for this. Perhaps we could get the action to do that and then we wouldn't need to do either :joy:

addyess: UGH. i wish. I don't think a juju action gives us the ability to write a key output to a file like this

docs/src/charm/tutorial/getting-started.md
```diff
@@ -1 +1,243 @@
...
+```
+nano ~/.kube/config
+```
+
+Please use the editor of your choice to delete the first line and save the file.
+
+Alternatively, if you are a `yq` user, the same can be achieved with:
+
+```
+juju run k8s/0 get-kubeconfig | yq '.kubeconfig' -r >> ~/.kube/config
```
addyess, line 182: any concern about the `>>` append to an existing `~/.kube/config`?

addyess (approved):
Very nice. Question on the `>>` redirect but otherwise fine

docs/src/charm/tutorial/getting-started.md
evilnick: No, because I would rather it got corrupted  but salvageable than overwrite an existing valid config

evilnick: I feel we have given people enough pointers. If we want to avoid having to explain in detail how to manually construct a kubeconfig then we should do some work on managing it properly.

addyess: Agree.. Nicely Done.  Thank you.

PR #395, opened 2024-05-02
Add lxd charm howto

evilnick:
I've added some instructions for managing the LXD profile when deploying the charm 

I have covered the case for deploying to localhost

Currently as a draft because:

- I'm not sure of the best way of tackling the case e.g. of bundles. How can we edit the profiles on a machine which hasn't been created yet?

### DIFF
--- a/docs/src/charm/howto/charm.md
+++ b/docs/src/charm/howto/charm.md
@@ -12,7 +12,7 @@ This guide assumes the following:
 - If you still need to do this, please take a look at the quickstart
   instructions, or, for custom clouds (OpenStack, MAAS), please consult the
   [Juju documentation][juju].
-- you are not using the Juju 'localhost' cloud (see [localhost
+- You are not using the Juju 'localhost' cloud (see [localhost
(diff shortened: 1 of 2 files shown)


addyess:
@evilnick i don't think there's anything we can do to tackle bundles.   Juju doesn't let one set a per-app lxd-profile.  Its either the model entirely -- or nothing.  Every app in the model will use the same lxd profiles.

You can however prepare the profile before the machines are created.  As you've defined in this doc, having the `juju-<model>` profile is the correct way to setup the profile before the machines are started.

addyess:
I've also created a new task card to have the charm (heck -- maybe the snap) detect the virtual machine type and to warn users when they may not have their lxd-profiles correct instead of the snap just not working, and requiring people to dig through kubelet logs trying to find out why cgroups aren't correct or loop devices don't exist.

evilnick:
Don't worry about the moonray failure, this branch is from the before-times

addyess (approved):
Let's go! Nice Nick

PR #396, opened 2024-05-02
Fix issue with finding release branch matches

addyess:
Don't match exactly with branch, but use bash regex matching

# needs backport to `release-1.30`


bschimke95 (approved):
LGTM

neoaggelos (approved):

PR #399, opened 2024-05-03
Allow changing cluster configuration through `snap set k8s`

neoaggelos:
### Summary

Allow using `snap set k8s ...` to update the cluster configuration

### How to test this PR

```
# bootstrap cluster, perhaps add more cluster nodes
$ k8s bootstrap

# set 'meta.orb=k8sd' to enable k8sd-snapd configuration sync (disabled by default)
$ snap set k8s meta.orb=k8sd

$ snap set k8s ingress.enabled=true

$ k8s get
```


src/k8s/cmd/k8s/k8s_x_snapd_config.go
```diff
@@ -0,0 +1,99 @@
...
+				if !empty {
+					cmd.PrintErrf("Error: failed to parse meta configuration: %v\n", err)
+					env.Exit(1)
+					return
+				}
+
+				cmd.PrintErrf("Warning: failed to parse meta configuration: %v\n", err)
+				cmd.PrintErrf("Warning: ignoring further errors to prevent infinite loop\n")
+				if setErr := snapdconfig.SetMeta(cmd.Context(), env.Snap, snapdconfig.Meta{
+					APIVersion: "1.30",
```
eaudetcobello, line 36: When we bump the kubernetes version, will we have to remember to change this value?

src/k8s/cmd/k8s/k8s_x_snapd_config.go
```diff
@@ -0,0 +1,99 @@
...
+					Error:      err.Error(),
+				}); setErr != nil {
+					cmd.PrintErrf("Warning: failed to set meta configuration to safe defaults: %v\n", setErr)
+				}
+				return
+			}
+
+			switch mode.Orb {
+			case "none":
+				cmd.PrintErrln("Warning: meta.orb is none, do not do anything")
```
eaudetcobello, line 47: Who should not do anything? Can this be rephrased to "no configuration changes have taken place"?

src/k8s/cmd/k8s/k8s_x_snapd_config.go
neoaggelos: This is the "API version" of the snapd configuration format. It is named like this to note that it originates with the 1.30 release. We do not need to bump this on each release. Rather, this should be bumped whenever breaking changes happen (e.g. a configuration field changes, or a new version of snapd comes around).

PR #409, opened 2024-05-06
Adjustments so launchpad builders pick up arm64 requirement

addyess:
According to [snapcraft io docs](https://snapcraft.io/docs/architectures#core20-4) the `architectures` tag for `core20` snaps, this replacement is identical.  Launchpad builds however seem to be having difficulty seeing that this snap should be built for multiple architectures without this change


bschimke95 (approved):
LGTM

PR #410, opened 2024-05-07
Only run code actions on code changes

eaudetcobello:
This gets the ball rolling... can probably be improved. Any ideas?

Rationale is that right now integration tests, unit tests, builds occur on documentation updates. Also, go code coverage bot comments on PR's that don't contain go code.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -7,8 +7,12 @@ on:
       - autoupdate/strict
       - 'release-[0-9]+.[0-9]+'
       - 'release-[0-9]+.[0-9]+-strict'
+    paths:
+      - '**/*.go'
 
   pull_request:
+    paths:
+      - '**/*.go'
 
 jobs:
   test:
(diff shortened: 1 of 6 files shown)


addyess (approved):
Well i think it's a very nice start

.github/workflows/go.yaml
```diff
@@ -7,8 +7,12 @@ on:
       - autoupdate/strict
       - 'release-[0-9]+.[0-9]+'
       - 'release-[0-9]+.[0-9]+-strict'
+    paths:
+      - '**/*.go'
```
neoaggelos, line 11: careful, this might miss things. we also need paths like .github workflows (e.g. in case the workflow changes)

neoaggelos (commented):
Not a huge fan of this, it probably adds more friction than it helps (e.g. this could very easily become out of date, and it would be very easy to miss, or start to figure out what is going on).

Please at least address my one comment. We definitely need to include the github workflows in the list of paths

eaudetcobello:
@neoaggelos Do you have another solution in mind?

louiseschmidtgen (commented):
Would it be possible to reverse `paths` so list everything that we do not need the workflows to run on, such as the docs directory?

PR #411, opened 2024-05-08
use `autoupdate/release-1.xx-strict` branches for strict

neoaggelos:
### Summary

Use `autoupdate/release-1.xx-strict` for strict release branches, to avoid having the auto promotion jobs failing due to branch protection rules

This also makes the strict branches fall in line with `autoupdate/strict`, to make it clear that these branches are not to be interacted with.

### DIFF
--- a/.github/ISSUE_TEMPLATE/create_release_branch.md
+++ b/.github/ISSUE_TEMPLATE/create_release_branch.md
@@ -42,11 +42,6 @@ The steps are to be followed in-order, each task must be completed by the person
 
       Thank you, $name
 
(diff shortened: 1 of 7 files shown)


addyess:
Can you link to the failing auto update jobs?  Is there anyway to adjust the branch-protection rules instead?

neoaggelos:
An example can be seen in https://github.com/canonical/k8s-snap/actions/runs/8986030620/job/24681222255#step:4:5, where the job currently succeeds precisely because of manually bypassing the branch protection rules.

addyess (approved):
NICE LGTM

PR #412, opened 2024-05-08
Update RTD build conditions

evilnick:
Updating the RTD build script to auto-cancel creating docs for PRs with no doc changes

N.B., the webhook will still run, it will just finish quickly without building any docs and saving some server power.

### DIFF
--- a/docs/tools/.readthedocs.yaml
+++ b/docs/tools/.readthedocs.yaml
@@ -14,6 +14,7 @@ build:
     post_checkout:
       # Cancel building pull requests when there aren't changed in the docs directory or YAML file.
       #
+      # https://docs.readthedocs.io/en/latest/build-customization.html#cancel-build-based-on-a-condition
       # If there are no changes (git diff exits with 0) we force the command to return with 183.
       # This is a special exit code on Read the Docs that will cancel the build immediately.
       - |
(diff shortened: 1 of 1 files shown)


docs/tools/.readthedocs.yaml
```diff
@@ -9,8 +9,18 @@ version: 2
...
   os: ubuntu-22.04
   tools:
-    python: "3.11"
+    python: "3.12"
   jobs:
+    post_checkout:
+      # Cancel building pull requests when there aren't changed in the docs directory or YAML file.
+      #
+      # If there are no changes (git diff exits with 0) we force the command to return with 183.
+      # This is a special exit code on Read the Docs that will cancel the build immediately.
```
addyess, line 18: The reasoning behind the exit code is super nerdy. Maybe a link to this doc?

```suggestion
      # https://docs.readthedocs.io/en/latest/build-customization.html#cancel-build-based-on-a-condition
      # If there are no changes (git diff exits with 0) we force the command to return with 183.
      # This is a special exit code on Read the Docs that will cancel the build immediately.
```

addyess:
```python
>>> sum(list("skip".encode("ascii")))
439
>>> 439 % 256
183
```

evilnick: It really **is** super nerdy, but Okay

addyess (approved):
lgtm

PR #417, opened 2024-05-16
Group feature implementations in a swappable interface

neoaggelos:
### Summary

Add machinery to swap feature implementations


src/k8s/pkg/k8sd/features/interface.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"context"
+
+	"github.com/canonical/k8s/pkg/k8sd/types"
+	"github.com/canonical/k8s/pkg/snap"
+)
+
+// Interface abstracts the management of built-in Canonical Kubernetes features.
+type Interface interface {
```
bschimke95, line 11: Go best-practice would probably be to call this `Featurer` but I agree that this is weird. `feature.Interface` is fine.

src/k8s/pkg/k8sd/features/localpv/chart.go
```diff
@@ -0,0 +1,22 @@
+package localpv
+
+import (
+	"path"
+
+	"github.com/canonical/k8s/pkg/client/helm"
+)
+
+var (
+	// chart is manifests for the built-in local storage feature, powered by Rawfile LocalPV CSI.
```
bschimke95, line 10: nit: here and elsewhere "is" sounds weird. Maybe `chart represents manifests ...`

src/k8s/pkg/k8sd/features/metrics-server/metrics_server_test.go
```diff
@@ -1,12 +1,12 @@
...
+package metrics_server_test
 
 import (
 	"context"
 	"testing"
 
 	"github.com/canonical/k8s/pkg/client/helm"
 	helmmock "github.com/canonical/k8s/pkg/client/helm/mock"
-	"github.com/canonical/k8s/pkg/k8sd/features"
+	metrics_server "github.com/canonical/k8s/pkg/k8sd/features/metrics-server"
```
bschimke95, line 9: The package is called `metrics_server`. How does this work with `-`?

bschimke95 (commented):
Great work. Left some comments

For future work: I guess eventually we need some README documenting how this works and how the Go tags are used for the different flavours.

src/k8s/pkg/k8sd/features/metrics-server/metrics_server_test.go
neoaggelos: files in `pkg/my-package` get a `package my_package` header, and then Go imports it as `my_package ".../pkg/my-package"`

PR #425, opened 2024-05-16
Copy control-plane taint from bootstrap config

addyess:
Complete the register control-planes with taints task and test that taints are adequately applied on bootstrap

### DIFF
--- a/src/k8s/pkg/k8sd/types/cluster_config_convert.go
+++ b/src/k8s/pkg/k8sd/types/cluster_config_convert.go
@@ -65,6 +65,7 @@ func ClusterConfigFromBootstrapConfig(b apiv1.BootstrapConfig) (ClusterConfig, e
 
 	// Kubelet
 	config.Kubelet.CloudProvider = b.ClusterConfig.CloudProvider
+	config.Kubelet.ControlPlaneTaints = &b.ControlPlaneTaints
 
 	return config, nil
 }
--- a/src/k8s/pkg/k8sd/types/cluster_config_convert.go
+++ b/src/k8s/pkg/k8sd/types/cluster_config_convert.go
@@ -65,7 +65,7 @@ func ClusterConfigFromBootstrapConfig(b apiv1.BootstrapConfig) (ClusterConfig, e
 
 	// Kubelet
 	config.Kubelet.CloudProvider = b.ClusterConfig.CloudProvider
-	if b.ControlPlaneTaints != nil {
+	if len(b.ControlPlaneTaints) != 0 {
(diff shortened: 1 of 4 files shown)


src/k8s/pkg/k8sd/types/cluster_config_convert.go
```diff
@@ -65,6 +65,7 @@ func ClusterConfigFromBootstrapConfig(b apiv1.BootstrapConfig) (ClusterConfig, e
 
 	// Kubelet
 	config.Kubelet.CloudProvider = b.ClusterConfig.CloudProvider
+	config.Kubelet.ControlPlaneTaints = &b.ControlPlaneTaints
```
bschimke95, line 68: please also update the `cluster_config_convert_test`

tests/integration/tests/test_control_plane_taints.py
```diff
@@ -0,0 +1,38 @@
...
+        {"control-plane-taints": ["node-role.kubernetes.io/control-plane:NoSchedule"]}
+    )
+
+    k8s_instance.exec(
+        ["dd", "of=/root/config.yaml"],
+        input=str.encode(bootstrap_conf),
+    )
+
+    k8s_instance.exec(["k8s", "bootstrap", "--file", "/root/config.yaml"])
+    while not (nodes := util.get_nodes(k8s_instance)):
```
bschimke95, line 30: this will run forever in case something goes wrong, doesn't it?

bschimke95 (commented):
Great work, two comments

tests/integration/tests/test_control_plane_taints.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+
+    k8s_instance.exec(
+        ["dd", "of=/root/config.yaml"],
+        input=str.encode(bootstrap_conf),
+    )
+
+    k8s_instance.exec(["k8s", "bootstrap", "--file", "/root/config.yaml"])
+    retries = 10
+    while retries and not (nodes := util.get_nodes(k8s_instance)):
```
bschimke95, line 31: nit: Would be cool, if we could reuse the existing retry mechanisms here but I see how this would require a refactoring of the `util.get_nodes` function.

bschimke95 (approved):
LGTM

tests/integration/tests/test_control_plane_taints.py
addyess: right, i looked at this too but didn't find a good way in this PR to accomplish this.

bschimke95: something for another time...

PR #438, opened 2024-05-22
Add track promotion script

bschimke95:
Automatically promote snap revisions to next risk level after a certain amount of time.
A revision will not be promoted to stable unless there is already a stable release for this track.

WIP: Currently try to figure out the best way to inject the authentication token


scripts/requirements.txt
```diff
@@ -0,0 +1 @@
+requests==2.25.1
```
neoaggelos, line 1: we are doing only simple GET and POST requests, let's use the standard library for that. then, we only need a Python 3 environment without any other dependencies to run this.

scripts/promote-tracks.py
```diff
@@ -0,0 +1,84 @@
...
+        risk = channel["risk"]
+        next_risk = RISK_LEVELS[RISK_LEVELS.index(risk)+1]
+        revision = channel_info["revision"]
+
+        if track in IGNORE_TRACKS:
+            continue
+
+        now = datetime.datetime.now(datetime.timezone.utc)
+
+        if ((now - parser.parse(channel["released-at"])).days > DAYS_TO_STAY_IN_RISK[risk]
```
neoaggelos, line 70: `dateutil` is not standard library either. instead of relying on that, can we use `datetime.strptime` instead?

scripts/promote-tracks.py
```diff
@@ -0,0 +1,84 @@
+import requests
```
neoaggelos, line 1: This is missing, and is the reason for the cryptic error message you see on the CI.

```suggestion
#!/usr/bin/env python3

import requests
```

neoaggelos (commented):
As a generic note, this is currently missing:

- Needs a dangerous SNAPSTORE_AUTH_TOKEN secret, which we would rather not have in the whole repo.
- Running end to end tests, spread tests, etc. In case this needs to run on public clouds, this means more powerful secrets will be added in the repo.

Perhaps we could move this (and the dangerous secrets) in a separate `k8s-snap-release-ci` repo?

bschimke95:
Closing this in favour of: https://github.com/canonical/canonical-kubernetes-release-ci/pull/1

PR #443, opened 2024-05-24
Explicitly set timeout in each CLI command

bschimke95:
Right now, the context deadline is initially set at root level when the command is created.This causes problems for interactive commands as the deadline may b
e exceeded while the user inputs the data.
To prevent this, we remove the global timeout and specifically set 
it in each command right before the RPC calls.


berkayoz (commented):
Looks good to me overall, I was wondering if it is possible to keep the flag on root and just pass the timeout value down to the sub/child commands to create the context there instead in the `PersistentPreRun`. Maybe by having a global variable root opts, or something for just the timeout?

bschimke95:
Good point, that's what I initially thought doing that as well. But, a global root opts variable is not a good idea. We made the opts function-scoped on purpose to make them easily testable and isolated. Adding global state would break that.

berkayoz (approved):
LGTM

PR #444, opened 2024-05-24
Sync snapd from k8sd and set `meta.orb` by default after bootstrap/join

berkayoz:
Sync snapd from k8sd and set `meta.orb` by default after bootstrap/join

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -4,6 +4,4 @@
 
 k8s::common::setup_env
 
-if [ -e "$SNAP_COMMON/var/lib/k8sd/state/control.socket" ]; then
-  k8s::cmd::k8s x-snapd-config reconcile
-fi
+k8s::cmd::k8s x-snapd-config reconcile
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -4,4 +4,6 @@
 
 k8s::common::setup_env
 
-k8s::cmd::k8s x-snapd-config reconcile
+if [ -e "$SNAP_COMMON/var/lib/k8sd/state/control.socket" ]; then
+  k8s::cmd::k8s x-snapd-config reconcile
+fi
--- a/src/k8s/cmd/k8s/k8s_x_snapd_config.go
+++ b/src/k8s/cmd/k8s/k8s_x_snapd_config.go
@@ -78,7 +78,7 @@ func newXSnapdConfigCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				}
 			}
 
-			mode.Orb = "snapd"
+			mode.Orb = "k8sd"
 			if err := snapdconfig.SetMeta(cmd.Context(), env.Snap, mode); err != nil {
 				cmd.PrintErrf("Error: failed to set snapd configuration: %v\n", err)
 				env.Exit(1)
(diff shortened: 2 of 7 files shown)


src/k8s/pkg/utils/experimental/snapdconfig/k8s_to_snapd.go
```diff
@@ -13,7 +13,7 @@ import (
 func SetSnapdFromK8sd(ctx context.Context, config apiv1.UserFacingClusterConfig, snap snap.Snap) error {
 	var sets []string
 	for key, cfg := range map[string]any{
-		"meta":          Meta{Orb: "snapd", APIVersion: "1.30"},
+		"meta":          Meta{Orb: "k8sd", APIVersion: "1.30"},
```
berkayoz, line 16: AFAIK, this function is called on `x_snapd_config reconcile` when `meta.orb=k8sd`, however after reconciling we set it back to `meta.orb=snapd` which will override the initial option. Not sure if this is a bug or this is intended, I need more context on this

src/k8s/cmd/k8s/k8s_x_snapd_config.go
```diff
@@ -78,7 +78,7 @@ func newXSnapdConfigCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				}
 			}
 
-			mode.Orb = "snapd"
+			mode.Orb = "k8sd"
```
neoaggelos, line 81: revert

src/k8s/pkg/utils/experimental/snapdconfig/k8s_to_snapd.go
neoaggelos: revert

snap/hooks/configure
```diff
@@ -4,4 +4,6 @@
 
 k8s::common::setup_env
 
-k8s::cmd::k8s x-snapd-config reconcile
+if [ -e "$SNAP_COMMON/var/lib/k8sd/state/control.socket" ]; then
```
neoaggelos, line 7: best to revert this, as it is a 95% solution, and 50% problem. if you `snap set` without k8sd running, your configs will be silently dropped. also, the socket existing does not mean that k8sd is ready to accept RPCs, or that cluster quorum is up for things to succeed.

in this case, the safest thing and least surprising thing to do is to just fail, so this check is unnecessary.

src/k8s/pkg/utils/experimental/snapdconfig/k8s_to_snapd.go
neoaggelos: `meta.orb` is used to denote which source of truth to use. when k8sd updates the configs, we set it to `snapd`, so that any `snap set` afterwards are applied

src/k8s/pkg/utils/experimental/snapdconfig/meta.go
```diff
@@ -33,9 +33,9 @@ func ParseMeta(ctx context.Context, s snap.Snap) (Meta, bool, error) {
 		return Meta{}, false, fmt.Errorf("failed to parse snap config mode: %w", err)
 	}
 
-	// default meta.orb is none
+	// default meta.orb is k8sd
 	if parse.Meta.Orb == "" {
-		parse.Meta.Orb = "none"
+		parse.Meta.Orb = "k8sd"
```
neoaggelos, line 38: that's all you really need, the controller will use this to configure snapd configs after the node is bootstrapped (or joins a cluster).

neoaggelos (changes requested):

src/k8s/pkg/utils/experimental/snapdconfig/meta.go
neoaggelos, line 38: i believe this stay as `none`

neoaggelos (approved):

PR #445, opened 2024-05-27
Enable dualstack and add test files

bschimke95:
# Summary
This PR contains the relevant changes to make dualstack (IPv4/IPv6) work on the k8s-snap. 

# Changes 
The current implementation for Kubernetes works out of the box as the only requirements from there is to set the proper `PodCIDR` and `ServiceCIDR` as `IPv4CIDR,IPv6CIDR`. A tiny unittest verifies that the kube-apiserver CIDR is properly set with both CIDRs.

For Cilium, IPv6 is disabled by default. Hence, we need to explicitly enable it if the IPv6 CIDR is set. 

Lastly, the `nginx-dualstack.yaml` was added to manually verify dualstack and will later be used in the integration tests.

### DIFF
--- a/src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
+++ b/src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
@@ -132,6 +132,19 @@ func TestKubeAPIServer(t *testing.T) {
 		g.Expect(len(args)).To(Equal(len(tests)))
 	})
 
+	t.Run("ArgsDualstack", func(t *testing.T) {
+		g := NewWithT(t)
+
+		s := mustSetupSnapAndDirectories(t, setKubeAPIServerMock)
+
+		// Setup without proxy to simplify argument list
+		g.Expect(setup.KubeAPIServer(s, "10.0.0.0/24,fd01::/64", "https://auth-webhook.url", false, types.Datastore{Type: utils.Pointer("external"), ExternalServers: utils.Pointer([]string{"datastoreurl1", "datastoreurl2"})}, "Node,RBAC")).To(BeNil())
+
+		g.Expect(snaputil.GetServiceArgument(s, "kube-apiserver", "--service-cluster-ip-range")).To(Equal("10.0.0.0/24,fd01::/64"))
+		_, err := utils.ParseArgumentFile(path.Join(s.Mock.ServiceArgumentsDir, "kube-apiserver"))
+		g.Expect(err).ToNot(HaveOccurred())
+	})
+
 	t.Run("ArgsExternalDatastore", func(t *testing.T) {
 		g := NewWithT(t)
 
--- a/tests/integration/templates/nginx-dualstack.yaml
+++ b/tests/integration/templates/nginx-dualstack.yaml
@@ -1,0 +1,37 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nginxdualstack
+spec:
+  selector:
+    matchLabels:
+      run: nginxdualstack
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        run: nginxdualstack
+    spec:
+      containers:
+      - name: nginxdualstack
+        image: rocks.canonical.com/cdk/diverdane/nginxdualstack:1.0.0
+        ports:
+        - containerPort: 80
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: nginx6
+  labels:
+    run: nginxdualstack
+spec:
+  type: NodePort
+  ipFamilies:
+  - IPv4
+  - IPv6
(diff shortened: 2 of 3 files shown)


src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
```diff
@@ -132,6 +132,19 @@ func TestKubeAPIServer(t *testing.T) {
 		g.Expect(len(args)).To(Equal(len(tests)))
 	})
 
+	t.Run("ArgsDualstack", func(t *testing.T) {
+		g := NewWithT(t)
+
+		s := mustSetupSnapAndDirectories(t, setKubeAPIServerMock)
+
+		// Setup without proxy to simplify argument list
+		g.Expect(setup.KubeAPIServer(s, "10.0.0.0/24,fdXY:IJKL:MNOP:15::/64", "https://auth-webhook.url", false, types.Datastore{Type: utils.Pointer("external"), ExternalServers: utils.Pointer([]string{"datastoreurl1", "datastoreurl2"})}, "Node,RBAC")).To(BeNil())
```
neoaggelos, line 141: let's make sure this is an actual IPv6 CIDR :)

```suggestion
		g.Expect(setup.KubeAPIServer(s, "10.0.0.0/24,fd01::/64", "https://auth-webhook.url", false, types.Datastore{Type: utils.Pointer("external"), ExternalServers: utils.Pointer([]string{"datastoreurl1", "datastoreurl2"})}, "Node,RBAC")).To(BeNil())
```

tests/integration/templates/nginx-dualstack.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: apps/v1
```
neoaggelos, line 1: i imagine we'll be writing a test that uses this? do we need this nginxdualstack image or does a simple `nginx` also work?

neoaggelos (commented):
Nice work so far!

tests/integration/templates/nginx-dualstack.yaml
bschimke95: It seems like you need to explicitly enable IPv6 in nginx. 
If you deploy the default nginx image and try to connect to it then it fails with:
```
root@k8s-dualstack:~# curl http://[fd98::89f]/
curl: (7) Failed to connect to fd98::89f port 80 after 0 ms: Connection refused
```

See also: https://ma.ttias.be/enable-ipv6-connectivity-for-nginx/

bschimke95: I don't have strong opinions here - this was really just used for a first manual test - and I mainly added it so that @eaudetcobello can easily follow the steps I described in the documentation. Microk8s uses the same image and I just stole it from there.

Let's keep it like this for now. We may still update the image when writing the integration test.

neoaggelos: sounds good

tests/integration/templates/nginx-dualstack.yaml
```diff
@@ -0,0 +1,36 @@
...
+kind: Service
+metadata:
+  name: nginx6
+  labels:
+    run: nginxdualstack
+spec:
+  type: NodePort
+  ipFamilies:
+  - IPv6
+  ipFamilyPolicy: RequireDualStack
```
eaudetcobello, line 31: Should this be SingleStack since we specify ipFamilies as only IPv6?

bschimke95: nope we want dualstack

I added the ipv4 familiy to address this.

src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
```diff
@@ -132,6 +132,19 @@ func TestKubeAPIServer(t *testing.T) {
 		g.Expect(len(args)).To(Equal(len(tests)))
 	})
 
+	t.Run("ArgsDualstack", func(t *testing.T) {
```
louiseschmidtgen, line 135: I know this PR is addressing dualstack, but does ipv6 only also work? If so can we add a test case here?

louiseschmidtgen (commented):
Great work, little question.

src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
bschimke95: So, IPv6-only does not work with Cilium Loadbalancer so we don't use it (for now).
Apart from that IPv6-only should work. In this testcase we only check if the string-concatenated CIDRs are correctly passed down. Single CIDR is tested above (there is no difference between IPv6 and IPv4 CIDR since we just pass them down as strings.

PR #450, opened 2024-05-29
Switch to x509 auth

neoaggelos:
### Summary

Use a separate CA for client certificates, and switch to x509 auth for all kubeconfigs

### Changes

- Add `clusterConfig.Certificates.ClientCA{Cert,Key}`
- Add `bootstrapConfig.ClientCA{Cert,Key}`
- Store the client ca (and admin client) certs in the cluster config
- Generate client certificates for kube-proxy and kubelet when joining worker nodes
- Generate client certificates for kubernetes components
- Use x509 cert auth for all generated configs
- Pass pre-generated client certificates when bootstrapping the node or joining a cluster

### Notes

- If the client ca cert is not set in the cluster config, we return the cluster CA instead (since they used to be the same)
- We do not yet disable the auth token webhook to ensure a proper migration path for existing clusters


src/k8s/api/v1/bootstrap_config.go
```diff
@@ -29,6 +29,8 @@ type BootstrapConfig struct {
 	// Seed configuration for external certificates
 	CACert                     *string `json:"ca-crt,omitempty" yaml:"ca-crt,omitempty"`
 	CAKey                      *string `json:"ca-key,omitempty" yaml:"ca-key,omitempty"`
+	ClientCACert               *string `json:"client-ca-crt,omitempty" yaml:"client-ca-crt,omitempty"`
```
bschimke95, line 32: nit: consider adding this new config option here: https://github.com/canonical/k8s-snap/blob/f4fd0e35ac7ee369c9c40213c9cf820b8a2ebe1d/src/k8s/cmd/k8s/k8s_bootstrap_test.go#L35

src/k8s/cmd/example/main.go
bschimke95, line 1: what is this file for?

bschimke95 (approved):
LGTM (after tests are passing), great work - only minor nits

src/k8s/cmd/example/main.go
neoaggelos: leftover from other testing, thanks for catching it

src/k8s/api/v1/bootstrap_config.go
neoaggelos: i'd avoid that for now, just because there are more fields missing from that test

PR #452, opened 2024-05-30
Make `k8s get-join-token $hostname` optional

neoaggelos:
### Summary

Allow `k8s get-join-token` to return tokens without specifying a hostname (and match any name specified during `k8s join-cluster`)

### Changes

- Add a separate `worker_tokens` database table
- `database.CheckWorkerNodeToken()` checks against the token itself, not the node name
- When a token has an empty node name, any joining name is accepted

### DIFF
--- a/src/k8s/pkg/k8sd/database/sql/queries/worker-tokens/delete-by-token.sql
+++ b/src/k8s/pkg/k8sd/database/sql/queries/worker-tokens/delete-by-token.sql
@@ -1,0 +1,4 @@
+DELETE FROM
+    worker_tokens AS t
+WHERE
+    t.token = ?
--- a/src/k8s/pkg/k8sd/database/sql/queries/worker-tokens/insert.sql
+++ b/src/k8s/pkg/k8sd/database/sql/queries/worker-tokens/insert.sql
@@ -1,0 +1,4 @@
+INSERT INTO
+    worker_tokens(name, token)
+VALUES
+    ( ?, ? )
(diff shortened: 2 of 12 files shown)


src/k8s/pkg/k8sd/database/sql/queries/worker-tokens/insert.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO
```
louiseschmidtgen, line 1: Nice, and we don't need to handle conflicts on the same node name anymore 👍🏼

src/k8s/pkg/k8sd/database/sql/queries/worker-tokens/delete-by-token.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM
+    worker_tokens AS t
+WHERE
+    t.token = ?
```
louiseschmidtgen, line 4: What was the reasoning behind deleting by token and not the id (primary key)?

louiseschmidtgen (approved):
LGTM, please add a little explanation to my one question.

src/k8s/pkg/k8sd/database/sql/queries/worker-tokens/delete-by-token.sql
neoaggelos: This will delete the database row that matches the token. The code that deletes the token does not know the id.

PR #456, opened 2024-05-31
Docs linter on Markdown files only

louiseschmidtgen:
## Docs Linter on Markdown only 
Follow-up on https://github.com/canonical/k8s-snap/pull/451

Inspired by example 2 of https://github.com/tj-actions/changed-files

Note: added top-level gh permissions

### DIFF
--- a/.github/workflows/markdown.yaml
+++ b/.github/workflows/markdown.yaml
@@ -6,20 +6,17 @@ jobs:
   markdown-lint:
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v4
-      with:
-        fetch-depth: 0
-    - uses: tj-actions/changed-files@v44
-      id: changed-files
-      with:
-        files: |
-          docs/src/snap/**
-          docs/src/charm/**
-    - uses: DavidAnson/markdownlint-cli2-action@v16
-      if: steps.changed-files.outputs.any_changed == 'true'
-      with:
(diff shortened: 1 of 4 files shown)


evilnick (approved):

docs/src/snap/explanation/about.md
```diff
@@ -30,7 +31,8 @@ In addition to the upstream Kubernetes services, Canonical Kubernetes also inclu
 
 The Canonical Kubernetes snap can be installed on a Linux OS, wherever it may
 be: run it in several local containers or VMs for example, or use it on
-public/private cloud instances. For deploying with [Juju], a machine [charm] to deploy
+public/private cloud instances. 
+For deploying with [Juju], a machine [charm] to deploy
 the snap is also available.
```
evilnick, line 36: :) this is the file I used to test it

.github/workflows/markdown.yaml
```diff
@@ -2,24 +2,24 @@ name: "Docs: Markdown lint"
...
-
-
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: tj-actions/changed-files@v44
+        id: changed-markdown-files
+        with:
+          files: |
+            docs/**/*.md
```
evilnick, line 19: can we add a 
!docs/src/_parts/**
or whatever. Some of those files will fail the rules because they are bits of md rather than complete files.

PR #459, opened 2024-06-03
Cluster config annotations

neoaggelos:
### Summary

Support annotations for cluster config to provide implementation specific config without cluttering the API.


src/k8s/cmd/k8s/k8s_set_test.go
```diff
@@ -79,6 +79,47 @@ func generateMapstructureTestCasesStringSlice(keyName string, fieldName string)
...
+		{
+			val:        fmt.Sprintf("%s=k1=", keyName),
+			assertions: []types.GomegaMatcher{HaveField(fieldName, map[string]string{"k1": ""})},
+		},
+		{
+			val:        fmt.Sprintf("%s=k1=,k2=test", keyName),
+			assertions: []types.GomegaMatcher{HaveField(fieldName, map[string]string{"k1": "", "k2": "test"})},
+		},
+		{
+			val:        fmt.Sprintf("%s=k1=v1", keyName),
```
bschimke95, line 101: interesting that this format is valid

src/k8s/pkg/utils/mapstructure.go
```diff
@@ -42,3 +42,45 @@ func StringToFieldsSliceHookFunc(r rune) mapstructure.DecodeHookFunc {
...
+	if raw == "" {
+		return map[string]string{}, nil
+	}
+
+	fields := strings.FieldsFunc(raw, func(this rune) bool { return this == ',' || unicode.IsSpace(this) })
+	result := make(map[string]string, len(fields))
+	for _, kv := range strings.FieldsFunc(raw, func(this rune) bool { return this == ',' || unicode.IsSpace(this) }) {
+		parts := strings.SplitN(kv, "=", 2)
+		if len(parts) < 2 {
+			return data, nil
```
bschimke95, line 80: I'm a bit confused about this function. Why do you return the data here? 
For `key1 value1, key2=value2` this will return `key1 value1, key2=value2` without any error.
Also, we seem to allow spaces (line 75) but this doesn't work with this check (e.g. ` key1 = value1 , key2 = value2 ` returns ` key1 = value1 , key2 = value2 `).
Please add some unittests to cover this function and the one above.
You can take my tests from here: https://go.dev/play/p/49a0buG-qEj

bschimke95 (commented):
Great work. A comment around the `mapstructure` implementation.

src/k8s/pkg/utils/mapstructure.go
neoaggelos: when parsing arbitrary `k1=v1, k2=v2` string inputs here, `kv` will be `k1=v1` and `k2=v2`.

then, we split into at most 2 parts on `=`. The result of that is:

```
strings.SplitN("k1", "=", 2) ==> []string{"k1"}
strings.SplitN("k1=v1", "=", 2) ==> []string{"k1", "v1"}
strings.SplitN("k1=v1=v2", "=", 2) ==> []string{"k1", "v1=v2"} 
```

the check for `len(parts) < 2` handles the first case (i.e. there is no `=` in the string. in that case, the decode hook has failed, therefore we `return data, nil` so that we fallthrough to the next hook (if any).

src/k8s/cmd/k8s/k8s_set_test.go
neoaggelos: this is what we enable parsing by specifying the `utils.StringToStringMapHookFunc()` hook.

bschimke95 (approved):
LGTM

src/k8s/pkg/utils/mapstructure.go
louiseschmidtgen: I know it's not pretty but can we please include this example in a comment right there?

louiseschmidtgen (approved):
Great work!

src/k8s/pkg/utils/mapstructure.go
neoaggelos: the test cases should account for this https://github.com/canonical/k8s-snap/pull/459/files#diff-9f869847ae70a71fcb8c1ca0ad5adeb8caec20026961223a709bc6361a179ba4R82

PR #466, opened 2024-06-03
Support MetalLB

eaudetcobello:
**Testing**

* l2-mode is covered by our integration tests and has been tested manually by deploying an nginx service.
* bgp mode is not covered by our integration tests and has been tested manually by deploying a bgp router using bird on a multipass vm and verifying that an nginx service on a node can obtain an ip using the bgp protocol.

**Example generated templates from values.yaml**

Metallb `values.yaml`:
```
driver: metallb

l2:
  enabled: true
  # interfaces:
  # - "^eth[0-9]+"
  interfaces:
    - eth1

ipPool:
  # cidrs:
  # - cidr: "10.42.254.176/28"
  cidrs:
    - cidr: "192.168.10.0/24"
    - start: "192.168.11.1"
      end: "192.168.11.254"

bgp:
  enabled: true
  localASN: 64512
  # neighbors:
  # - peerAddress: '10.0.0.60/24'
  #   peerASN: 65100
  #   peerPort: 179
  neighbors:
    - peerAddress: "10.0.0.60"
      peerASN: 65100
      peerPort: 179
```

Metallb template output:
```
---
# Source: ck-loadbalancer/templates/metallb/bgp-policy.yaml
apiVersion: "metallb.io/v1beta1"
kind: BGPAdvertisement
metadata:
  name: release-name-ck-loadbalancer
  labels:
    helm.sh/chart: ck-loadbalancer-0.1.1
    app.kubernetes.io/name: ck-loadbalancer
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "0.1.1"
    app.kubernetes.io/managed-by: Helm
spec:
   ipAddressPools:
     - release-name-ck-loadbalancer
---
# Source: ck-loadbalancer/templates/metallb/bgp-policy.yaml
apiVersion: "metallb.io/v1beta2"
kind: BGPPeer
metadata:
  name: release-name-ck-loadbalancer
  labels:
    helm.sh/chart: ck-loadbalancer-0.1.1
    app.kubernetes.io/name: ck-loadbalancer
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "0.1.1"
    app.kubernetes.io/managed-by: Helm
spec:
  myASN: 64512
  neighbors:
    - peerASN: 65100
      peerAddress: 10.0.0.60
      peerPort: 179
---
# Source: ck-loadbalancer/templates/metallb/lb-ip-pool.yaml
apiVersion: "metallb.io/v1beta1"
kind: IPAddressPool
metadata:
  name: release-name-ck-loadbalancer
  labels:
    helm.sh/chart: ck-loadbalancer-0.1.1
    app.kubernetes.io/name: ck-loadbalancer
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "0.1.1"
    app.kubernetes.io/managed-by: Helm
spec:
  addresses:
    - 192.168.10.0/24
---
# Source: ck-loadbalancer/templates/metallb/l2-policy.yaml
apiVersion: "metallb.io/v1beta1"
kind: L2Advertisement
metadata:
  name: release-name-ck-loadbalancer
  labels:
    helm.sh/chart: ck-loadbalancer-0.1.1
    app.kubernetes.io/name: ck-loadbalancer
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "0.1.1"
    app.kubernetes.io/managed-by: Helm
spec:
  ipAddressPools:
    - release-name-ck-loadbalancer
  interfaces:
    - eth1
```

**Potential improvements**

When a user gets an IP for a service using bgp-mode=true, a BGPPeer and a BGPAdvertisement resource are created. If the user then sets bgp-mode=false, the BGPPeer and BGPAdvertisement resources will continue existing. When the user sets l2-mode=true, the service will successfully receive a IP using l2 mode. 

From my testing this does not cause problems in terms of bugs or functionality but it could be nice to clean up when the config is changed.


src/k8s/pkg/k8sd/features/metallb/loadbalancer.go
```diff
@@ -0,0 +1,147 @@
...
+				},
+			},
+		},
+	}
+
+	if _, err := m.Apply(ctx, chartMetalLBLoadBalancer, helm.StatePresent, values); err != nil {
+		return fmt.Errorf("failed to apply LoadBalancer configuration: %w", err)
+	}
+
+	if err := rolloutRestartMetalLB(ctx, snap, 3); err != nil {
```
neoaggelos, line 84: this is not required, metallb picks up the configuration dynamically

k8s/manifests/charts/ck-loadbalancer/templates/cilium/lb-ip-pool.yaml
```diff
@@ -1,4 +1,6 @@
-{{- if .Values.ipPool.cidrs }}
+{{- if .Values.l2.enabled }}
```
addyess, line 1: Why'd we go from `ipPool.cidrs` to `l2.enabled` ?

k8s/manifests/charts/ck-loadbalancer/templates/metallb/lb-ip-pool.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: IPAddressPool
+metadata:
+  name: {{ include "ck-loadbalancer.fullname" . }}
+  labels:
+    {{- include "ck-loadbalancer.labels" . | nindent 4 }}
+spec:
+  addresses:
+    {{- range .Values.ipPool.cidrs }}
+    - {{ . }}
+    {{- end }}
```
addyess, line 14: I learned the hard-way on this.  The valid formats are pretty specific here.  this can be any of these:

`a.b.c.d/<mask>`
`aaaa:bbbb:cccc:dddd::/<mask>`
`a.b.c.d-e.f.g.h`

but definitely not `a.b.c.d`.  The metallb validating webhook will deny you with a 403 Forbidden when trying to apply.  Nothing more confusing than getting a 403 when that's not quite the right return code for invalid format.

src/k8s/pkg/k8sd/features/metallb/loadbalancer.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := waitForRequiredLoadBalancerCRDs(ctx, snap, loadbalancer.GetBGPMode()); err != nil {
+		return fmt.Errorf("failed to wait for required MetalLB CRDs: %w", err)
+	}
+
+	cidrs := []string{}
+	cidrs = append(cidrs, *loadbalancer.CIDRs...)
+
+	for _, ipRange := range loadbalancer.GetIPRanges() {
+		cidrs = append(cidrs, fmt.Sprintf("%s-%s", ipRange.Start, ipRange.Stop))
+	}
```
addyess, line 56: aight -- nice...looks like we're protected here from wild values

tests/integration/tests/test_cilium_e2e.py
```diff
@@ -1,54 +0,0 @@
...
-CILIUM_CLI_TAR_GZ = f"https://github.com/cilium/cilium-cli/releases/download/{CILIUM_CLI_VERSION}/cilium-linux-{CILIUM_CLI_ARCH_MAP.get(ARCH)}.tar.gz"  # noqa
-
-
-@pytest.mark.skipif(
-    ARCH not in CILIUM_CLI_ARCH_MAP, reason=f"Platform {ARCH} not supported"
-)
-def test_cilium_e2e(instances: List[harness.Instance]):
-    instance = instances[0]
-    instance.exec(["bash", "-c", "mkdir -p ~/.kube"])
-    instance.exec(["bash", "-c", "k8s config > ~/.kube/config"])
```
addyess, line 25: Sad to see this not get turned into a `test_loadbalancers_e2e` or some such.

addyess (commented):
Provided some thoughts to chew on.  I think there's a bug in the `cilium/l2-policy.yaml`

k8s/manifests/charts/ck-loadbalancer/templates/metallb/lb-ip-pool.yaml
eaudetcobello: Good to know, thx for the comment

tests/integration/tests/test_cilium_e2e.py
neoaggelos, line 1: revert

k8s/manifests/charts/ck-loadbalancer/templates/cilium/l2-policy.yaml
```diff
@@ -1,4 +1,6 @@
 {{- if .Values.l2.enabled }}
+{{- if (eq .Values.driver "cilium") }}
```
neoaggelos, line 2: note: it reads better to have the "driver" check first

k8s/manifests/charts/ck-loadbalancer/templates/metallb/l2-policy.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+apiVersion: "metallb.io/v1beta1"
+kind: L2Advertisement
+metadata:
+  name: {{ include "ck-loadbalancer.fullname" . }}
+  labels:
+    {{- include "ck-loadbalancer.labels" . | nindent 4 }}
+spec:
+  ipAddressPools:
+  - {{ include "ck-loadbalancer.fullname" . }}
```
neoaggelos, line 12: let's also quote here

k8s/manifests/charts/ck-loadbalancer/templates/metallb/bgp-policy.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.bgp.enabled }}
+{{- if (eq .Values.driver "metallb") }}
+
+{{- range .Values.bgp.neighbors }}
+apiVersion: "metallb.io/v1beta2"
+kind: BGPPeer
+metadata:
+  name: {{ include "ck-loadbalancer.fullname" $ }}
```
neoaggelos, line 8: be consistent with the way you render this name. also, make sure to quote all string values (`| quote`) to be on the safe side

k8s/manifests/charts/metallb-0.14.5.tgz
neoaggelos, line 1: we need to know how this file comes to exist. add an entry for this in the `build-scripts/hack/update-components.py` script (with appropriate METALLB_CHART configs next to the other configs). It's OK to pin an exact charm version to start with.

neoaggelos (changes requested):

k8s/manifests/charts/ck-loadbalancer/templates/cilium/lb-ip-pool.yaml
eaudetcobello: Fixed

eaudetcobello:
Let me know if you have any ideas on what to do with gateway-api. Leave the script in hacks/ and run it from update-component-versions.py?

k8s/manifests/charts/ck-loadbalancer/values.schema.json
```diff
@@ -22,10 +25,15 @@
...
+          "type": [
+            "array"
+          ],
           "items": [
             {
-              "type": "object",
+              "type": [
+                "object",
+                "string"
+              ],
```
neoaggelos, line 36: no, do not make this either an object or a string. this will simply lead to confusion in the future, where people will be confused why the schema is sometimes an object and sometimes a string (and why would either be preferred).

we control the schema, we should pass objects here, just as we are already doing

k8s/manifests/charts/ck-loadbalancer/values.yaml
```diff
@@ -1,3 +1,4 @@
+driver: cilium
```
neoaggelos, line 1: this should rather be set on the helm chart values in cilium, just as we set this on metallb as well.

build-scripts/hack/util.py
```diff
@@ -48,3 +49,9 @@ def read_file(path: Path) -> str:
 
 def read_url(url: str) -> str:
     return urlopen(url).read().decode().strip()
+
+def helm_pull(chart_name: str, repo_url: str, version: str, destination: Path) -> None:
+    parse_output(["helm", "repo", "add", chart_name, repo_url])
+    parse_output(["helm", "pull", f"{chart_name}/{chart_name}", "--version", version, "--destination", destination])
```
neoaggelos, line 55: please stay aware for conflicts with https://github.com/canonical/k8s-snap/pull/468/files/e52a1b2ce1fd4207e9772a40d86b018d8f93abb8#diff-c6cdac9381e357f73ccfe7a568a0468768bb5eadb2f32c4ba86db4d05d722c4fR52

also, please use `helm pull --repo $repo ...` instead of changing the repositories of the local helm client, as this is a recipe for unintended side-effects

src/k8s/pkg/k8sd/features/metallb/loadbalancer.go
neoaggelos, line 56: this is quite problematic for two reasons:
- we always dereference `loadbalancer.CIDRs`, so we'll panic if that field is nil
- we mix data of different schemas in the same place. this is something that will be very confusing for future debugging efforts, and how this is handled in the chart / code / tests etc.

Instead, please do the same thing we do elsewhere, https://github.com/canonical/k8s-snap/blob/f026b18c18a4e80d5dd0b6147ef6a638127d415d/src/k8s/pkg/k8sd/features/cilium/loadbalancer.go#L93-L99

neoaggelos (changes requested):

neoaggelos (changes requested):
In addition, please add a comment with all the metallb images that are used.

```
k8s kubectl get node -o template='{{ range .items }}{{ .metadata.name }}{{":"}}{{ range .status.images }}{{ "\n- " }}{{ index .names 1 }}{{ end }}{{"\n"}}{{ end }}'
```

can help with this

k8s/manifests/charts/ck-loadbalancer/values.yaml
eaudetcobello: This is just a default value, it's overridden in the Go code.

neoaggelos: yes. implementations (cilium or metallb) must specify this, let's not hide this requirement with implicit defaults

eaudetcobello:
MetalLB images:

- quay.io/metallb/speaker:v0.14.5
- quay.io/metallb/controller:v0.14.5

eaudetcobello:
I added a description with an example values.yaml and the generated template.

build-scripts/hack/util.py
```diff
@@ -48,3 +49,26 @@ def read_file(path: Path) -> str:
...
     return urlopen(url).read().decode().strip()
+
+
+def helm_pull(chart_name: str, repo_url: str, version: str, destination: Path) -> None:
+    parse_output(
+        [
+            "helm",
+            "pull",
+            f"{repo_url}",
+            f"{chart_name}/{chart_name}",
```
louiseschmidtgen, line 60: I think this won't work for example for contour it's: 
`sudo k8s helm pull oci://registry-1.docker.io/bitnamicharts/contour bitnami/contour --version 17.0.4 --destination ./k8s/manifests/charts/`
So something like f"{chart_maintainer}/{chart_name}", would be better.

neoaggelos, line 60: Please update this command, this is not the right invocation. It might work locally because you already have the repo, but it is not right. See:

```
$ helm pull https://metallb.github.io/metallb metallb/metallb --version 0.14.5
Error: repo metallb not found
```

The correct way to write this command is

```
$ helm pull metallb --repo https://metallb.github.io/metallb --version 0.14.5
```

neoaggelos (changes requested):

k8s/manifests/charts/ck-loadbalancer/values.schema.json
neoaggelos, line 1: revert these changes

build-scripts/hack/update-component-versions.py
```diff
@@ -36,6 +37,14 @@
 # Helm release branch to track. The most recent tag in the branch will be used.
 HELM_RELEASE_BRANCH = "release-3.14"
 
+# MetalLB Helm repository and chart version
+METALLB_REPO = "https://metallb.github.io/metallb"
+METALLB_CHART_VERSION = "0.14.5"
+
+# Metrics Server Helm repository
+METRICS_REPO = "https://kubernetes-sigs.github.io/metrics-server"
```
neoaggelos, line 45: NOTE: these are not relevant for this PR.

build-scripts/hack/util.py
neoaggelos: this is not addressed

k8s/manifests/charts/ck-loadbalancer/values.schema.json
neoaggelos: this is not addressed

build-scripts/hack/update-component-versions.py
neoaggelos: Please revert this, and also remove the `metrics-server-3.12.1.tgz` chart. This PR is related to adding MetalLB support, not changing metrics-server.

neoaggelos (changes requested):

build-scripts/hack/util.py
```diff
@@ -3,6 +3,7 @@
 import tempfile
 import subprocess
 import logging
+import yaml
```
louiseschmidtgen, line 6: Do we use the yaml import?

k8s/manifests/charts/ck-loadbalancer/values.yaml
```diff
@@ -1,3 +1,4 @@
+driver:
```
louiseschmidtgen, line 1: I think we could set "cilium" as our default driver here to avoid leaving a value undefined.

louiseschmidtgen (commented):
Great work Etienne!

One last thing, please add all the images in use by metallb in the PR.
```
k8s kubectl get node -o template='{{ range .items }}{{ .metadata.name }}{{":"}}{{ range .status.images }}{{ "\n- " }}{{ index .names 1 }}{{ end }}{{"\n"}}{{ end }}'
```
Just add a comment or add it to your PR description please.
Additionally, please add a var with the image repo and the tag similar to [here](https://github.com/canonical/k8s-snap/blob/48be3da4cc7ec3035f8eefd83ce0a9eb61ee73ad/src/k8s/pkg/k8sd/features/contour/chart.go#L37).

louiseschmidtgen (commented):
Another last comment :) 
Please write a card to add some bgp integration tests in the future.

build-scripts/hack/update-component-versions.py
```diff
@@ -71,7 +75,9 @@ def get_containerd_version() -> str:
     """Update containerd version using latest tag of specified branch"""
     containerd_repo = util.read_file(COMPONENTS / "containerd/repository")
 
-    with util.git_repo(containerd_repo, CONTAINERD_RELEASE_BRANCH, shallow=False) as dir:
+    with util.git_repo(
+        containerd_repo, CONTAINERD_RELEASE_BRANCH, shallow=False
+    ) as dir:
```
neoaggelos, line 80: revert

build-scripts/hack/update-component-versions.py
```diff
@@ -93,6 +99,16 @@ def get_helm_version() -> str:
...
+def pull_metallb_chart() -> None:
+    LOG.info("Pulling MetalLB chart @ %s", METALLB_CHART_VERSION)
+    util.helm_pull("metallb", METALLB_REPO, METALLB_CHART_VERSION, CHARTS)
+
+
+def pull_metrics_server_chart() -> None:
+    LOG.info("Pulling Metrics Server chart @ %s", METRICS_CHART_VERSION)
+    util.helm_pull("metrics-server", METRICS_REPO, METRICS_CHART_VERSION, CHARTS)
+
+
```
neoaggelos, line 111: revert

build-scripts/hack/util.py
neoaggelos, line 6: revert

build-scripts/hack/util.py
```diff
@@ -65,3 +66,4 @@ def helm_pull(chart, repo_url: str, version: str, destination: Path) -> None:
     )
 
     LOG.info("Pulled helm chart %s @ %s as %s to %s", chart, version, repo_url, destination)
+
```
neoaggelos, line 69: remove extraneous newline

k8s/manifests/charts/ck-loadbalancer/values.yaml
eaudetcobello: This has been brought up before. Conclusion is to not hide implicit defaults and let the implementation define it.

eaudetcobello:
> Great work Etienne!
> 
> One last thing, please add all the images in use by metallb in the PR.
> 
> ```
> k8s kubectl get node -o template='{{ range .items }}{{ .metadata.name }}{{":"}}{{ range .status.images }}{{ "\n- " }}{{ index .names 1 }}{{ end }}{{"\n"}}{{ end }}'
> ```
> 
> Just add a comment or add it to your PR description please. Additionally, please add a var with the image repo and the tag similar to [here](https://github.com/canonical/k8s-snap/blob/48be3da4cc7ec3035f8eefd83ce0a9eb61ee73ad/src/k8s/pkg/k8sd/features/contour/chart.go#L37).

Already done.

build-scripts/hack/update-component-versions.py
```diff
@@ -13,7 +13,6 @@
 import argparse
 import logging
 import sys
-import yaml
```
neoaggelos, line 16: `import yaml` is used in this file

neoaggelos (approved):

PR #468, opened 2024-06-04
Contour Ingress & Gateway

louiseschmidtgen:
### Ingress
When we `k8s enable ingress` we get a static deployment of contour:
```
pod/ck-ingress-contour-contour-5df8788b9d-rbq89   1/1     Running   0          4m56s
pod/ck-ingress-contour-envoy-jll5b                2/2     Running   0          4m56s
```
And we get this IngressClass:
```
sudo k8s kubectl get ingressclass
NAME         CONTROLLER                                                    PARAMETERS   AGE
ck-ingress   projectcontour.io/projectcontour/ck-ingress-contour-contour   <none>       108s
```
When you disable ingress, those will get removed. 

When you `k8s set default-tls-secret=bananas` you will get a TLSCertificateDelegation resource (from the mini /ck-ingress-tls chart) which you can use to point to your bananas secret. See [here](https://github.com/projectcontour/contour/issues/2488)
```
sudo k8s set ingress.default-tls-secret=bananas
sudo k8s kubectl get TLSCertificateDelegation ck-ingress-tls-delegation -o yaml -n projectcontour-root
apiVersion: projectcontour.io/v1
kind: TLSCertificateDelegation
metadata:
<shortened>
spec:
  delegations:
  - secretName: bananas
    targetNamespaces:
    - '*'
  ```
Now for `k8s set enable-proxy-protocol=true` this sets a config arg for the static deployment of contour. Contour gets restarted and voila there is an extra Arg:
```
sudo k8s kubectl describe pod ck-ingress-contour-contour -n projectcontour | grep proxy
      --use-proxy-protocol
```

### Gateway
Now when you do `k8s enable gateway` you get this dynamic gateway provisioner.
```
projectcontour     contour-gateway-provisioner-6dbcd4fbb-pd6kc   1/1     Running   0          8m51s
```
This helm chart lives in ck-gateway. I also added a gateway class to the chart with which people can create their gateway resource.

Additionally, another chart ck-gateway-contour installs:
```
sudo k8s kubectl get gatewayclass
NAME         CONTROLLER                             ACCEPTED   AGE
ck-gateway   projectcontour.io/gateway-controller   True       108m
```
You can `k8s disable gateway` and it will remove all gateway resources.

### Shared CRDS
Gateway and Ingress share some common contour CRDs. These are moved to their own chart which lives in ck-contour-common.
```
sudo k8s kubectl get crd | grep projectcontour
contourconfigurations.projectcontour.io               2024-06-12T08:47:49Z
contourdeployments.projectcontour.io                  2024-06-12T08:47:49Z
extensionservices.projectcontour.io                   2024-06-12T08:47:49Z
httpproxies.projectcontour.io                         2024-06-12T08:47:49Z
tlscertificatedelegations.projectcontour.io           2024-06-12T08:47:49Z
```

## Changes Cilium
Deploy a second gatewayclass named ck-gateway.
```
sudo k8s kubectl get gatewayclass
NAME         CONTROLLER                     ACCEPTED   AGE
cilium       io.cilium/gateway-controller   True       3m52s
ck-gateway   io.cilium/gateway-controller   True       3m53s
```

## List of images (Moonray):
```
sudo k8s kubectl get node -o template='{{ range .items }}{{ .metadata.name }}{{":"}}{{ range .status.images }}{{ "\n- " }}{{ index .names 1 }}{{ end }}{{"\n"}}{{ end }}'
louise-inspiron-5580:
- docker.io/calico/node:v3.28.0
- docker.io/calico/cni:v3.28.0
- ghcr.io/canonical/rawfile-localpv:0.8.0-ck5
- docker.io/bitnami/envoy:1.28.2-debian-12-r0
- docker.io/bitnami/contour:1.28.2-debian-12-r4
- docker.io/calico/apiserver:v3.28.0
- docker.io/calico/kube-controllers:v3.28.0
- docker.io/calico/typha:v3.28.0
- k8s.gcr.io/sig-storage/csi-provisioner:v3.4.1
- k8s.gcr.io/sig-storage/csi-resizer:v1.7.0
- k8s.gcr.io/sig-storage/csi-snapshotter:v6.2.1
- ghcr.io/canonical/metrics-server:0.7.0-ck0
- quay.io/tigera/operator:v1.34.0
- ghcr.io/canonical/coredns:1.11.1-ck4
- ghcr.io/projectcontour/contour:v1.28.2
- docker.io/calico/node-driver-registrar:v3.28.0
- k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.10.0
- docker.io/calico/csi:v3.28.0
- docker.io/calico/pod2daemon-flexvol:v3.28.0
- ghcr.io/canonical/k8s-snap/pause:3.10
```

### CODE REVIEW

File: k8s/manifests/charts/ck-gateway/templates/_helpers.tpl [hunk 1]
```diff
@@ -0,0 +1,62 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "ck-gateway.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
```
eaudetcobello, line 7: We should find a way to not repeat these helpers because they are also in the ck-loadbalancer chart.

louiseschmidtgen: Yes, you are right. I can probably remove this, it came with creating the initial Chart 👍🏼

build-scripts/hack/update-component-versions.py
```diff
@@ -59,6 +63,9 @@ def get_cni_version() -> str:
 
         raise Exception(f"Failed to find cni dependency in {deps_file}")
 
+def pull_contour_chart() -> None:
+    LOG.info("Pulling Contour Helm chart from %s with version %s", CONTOUR_HELM_REPO, CONTOUR_CHART_VERSION)
+    util.helm_pull("contour", CONTOUR_HELM_REPO, CONTOUR_CHART_VERSION, CHARTS)
```
neoaggelos, line 68: we seem to have more charts than that, as you have taken out some CRDs into a separate helm chart. This would have to be updated as well.


- please either document or write a script that updates these charts as well, as needed. for example, if we pull in a different `ck-contour`, we need to adjust `ck-contour-common` and `ck-contour-gateway` accordingly.
- please compress the big charts you have added, and add a version, e.g. `ck-contour-common-17.0.4-ck1.tgz`

neoaggelos (changes requested):

build-scripts/hack/update-component-versions.py
louiseschmidtgen: I chose a script and compressed the charts 👍🏼

tests/integration/tests/test_ingress.py
```diff
@@ -14,23 +15,35 @@
...
-    ingress_http_port = p.stdout.decode().replace("'", "")
+
+    ingress_http_port = None
+    services = json.loads(p.stdout.decode())
+    for svc in services["items"]:
+        if "ingress" in svc["metadata"]["name"]:
+            LOG.info(f"Found service {svc['metadata']['name']}")
+            for port in svc["spec"]["ports"]:
+                if port["port"] == 80:
+                    ingress_http_port = port.get("nodePort")
```
eaudetcobello, line 40: Is there a reason for using .get here instead of []? Is a port not guaranteed to have a nodePort? Not that this is logically wrong, but I'm asking for the sake of consistency with the rest of the dict access code you have.

tests/integration/tests/test_ingress.py
```diff
@@ -14,23 +15,35 @@
...
+            "-o=json",
         ],
         capture_output=True,
     )
-    ingress_http_port = p.stdout.decode().replace("'", "")
+
+    ingress_http_port = None
+    services = json.loads(p.stdout.decode())
+    for svc in services["items"]:
+        if "ingress" in svc["metadata"]["name"]:
```
eaudetcobello, line 36: Invert if

tests/integration/tests/test_ingress.py
```diff
@@ -14,23 +15,35 @@
...
+            LOG.info(f"Found service {svc['metadata']['name']}")
```
eaudetcobello, line 37: Is logging this useful?

tests/integration/tests/test_ingress.py
```diff
@@ -14,23 +15,35 @@
...
+            for port in svc["spec"]["ports"]:
+                if port["port"] == 80:
+                    ingress_http_port = port.get("nodePort")
+                    break
+            if ingress_http_port:
+                break
+
+    assert ingress_http_port is not None, "No ingress nodePort found."
+    LOG.info(f"Found http port: {ingress_http_port}")
```
eaudetcobello, line 46: Is logging this useful?

tests/integration/tests/test_gateway.py
```diff
@@ -47,13 +50,26 @@ def test_gateway(session_instance: harness.Instance):
...
+            for port in svc["spec"]["ports"]:
+                if port["port"] == 80:
+                    gateway_http_port = port["nodePort"]
+                    break
+            if gateway_http_port:
+                break
+
+    assert gateway_http_port is not None, "No ingress nodePort found."
+
+    LOG.info(f"Gateway http port is {gateway_http_port}")
```
eaudetcobello, line 72: Same as other logs.

src/k8s/pkg/k8sd/features/contour/ingress.go
```diff
@@ -0,0 +1,118 @@
...
+	m := snap.HelmClient()
+	if enabled {
+		if _, err := m.Apply(ctx, chartCommonContourCRDS, helm.StatePresent, nil); err != nil {
+			return fmt.Errorf("failed to install common CRDS: %w", err)
+		}
+	} else {
+		if _, err := m.Apply(ctx, chartCommonContourCRDS, helm.StateDeleted, nil); err != nil {
+			return fmt.Errorf("failed to uninstall common CRDS: %w", err)
+		}
+	}
```
eaudetcobello, line 96: What do you think of this? Often, when an if's only other branch is else, the else can be removed.

```suggestion
	if enabled {
		if _, err := m.Apply(ctx, chartCommonContourCRDS, helm.StatePresent, nil); err != nil {
			return fmt.Errorf("failed to install common CRDS: %w", err)
		}
		return nil
	}
	
    if _, err := m.Apply(ctx, chartCommonContourCRDS, helm.StateDeleted, nil); err != nil {
	    return fmt.Errorf("failed to uninstall common CRDS: %w", err)
    }
    
	return nil
```

src/k8s/pkg/k8sd/features/contour/ingress.go
```diff
@@ -0,0 +1,118 @@
...
+	// and gets set via k8s set defaultTLSSecret=bananas.
+	if ingress.GetDefaultTLSSecret() != "" {
+		values = map[string]any{
+			"defaultTLSSecret": ingress.GetDefaultTLSSecret(),
+		}
+		if _, err := m.Apply(ctx, chartDefaultTLS, helm.StatePresent, values); err != nil {
+			return fmt.Errorf("failed to install the delegation resource for default TLS secret: %w", err)
+		}
+	} else {
+		if _, err := m.Apply(ctx, chartDefaultTLS, helm.StateDeleted, nil); err != nil {
```
eaudetcobello, line 76: Consider the comment about if else here as well.

tests/integration/tests/test_gateway.py
```diff
@@ -47,13 +50,26 @@ def test_gateway(session_instance: harness.Instance):
             "kubectl",
             "get",
             "service",
-            "cilium-gateway-my-gateway",
-            "-o=jsonpath='{.spec.ports[?(@.name==\"port-80\")].nodePort}'",
```
eaudetcobello, line 51: Does replacing this with Python (below) give any advantages? If we keep it as jsonpath, the code is 10 lines shorter.

tests/integration/tests/test_ingress.py
```diff
@@ -14,23 +15,35 @@
...
     p = session_instance.exec(
         [
             "k8s",
             "kubectl",
             "get",
             "service",
-            "-n",
-            "kube-system",
-            "cilium-ingress",
-            "-o=jsonpath='{.spec.ports[?(@.name==\"http\")].nodePort}'",
```
eaudetcobello, line 29: Same as the other jsonpath comment. 1 line in jsonpath vs 10 in python.

src/k8s/pkg/k8sd/features/cilium/gateway.go
```diff
@@ -15,12 +15,21 @@ import (
 // ApplyGateway will rollout restart the Cilium pods in case any Cilium configuration was changed.
 // ApplyGateway returns an error if anything fails.
 func ApplyGateway(ctx context.Context, snap snap.Snap, gateway types.Gateway, network types.Network, _ types.Annotations) error {
+	if !network.GetEnabled() && gateway.GetEnabled() {
+		return fmt.Errorf("cilium gateway requires network to be enabled")
+	}
```
eaudetcobello, line 20: Interesting. I don't think I considered this in my PR 🤔

src/k8s/pkg/k8sd/features/contour/ingress.go
```diff
@@ -0,0 +1,118 @@
...
+// ApplyIngress returns an error if anything fails.
+// Contour CRDS are applied through a ck-contour common chart (Overlap with gateway)
+func ApplyIngress(ctx context.Context, snap snap.Snap, ingress types.Ingress, _ types.Network, _ types.Annotations) error {
+	m := snap.HelmClient()
+
+	if !ingress.GetEnabled() {
+		if _, err := m.Apply(ctx, chartContour, helm.StateDeleted, nil); err != nil {
+			return fmt.Errorf("failed to uninstall ingress: %w", err)
+		}
+	}
```
eaudetcobello, line 27: My PR might need this too 🤔

eaudetcobello (changes requested):
Well done, left a few comments. Also, in many places you use:
```
if thing {
  if _, err := ...
    return error
} else {
  if _, err := ...
    return error
}
return nil
```

Often an if statement whose only other branch is an else can be simplified:

```
if thing {
  if _, err := ...
    return error
  return nil
}
 
if _, err := ...
    return error
}

return nil
```

tests/integration/tests/test_gateway.py
louiseschmidtgen: We've got to grab all the services and check for one with "my-gateway" in the name to make it more general for cilum/contour...

build-scripts/hack/util.py
```diff
@@ -48,3 +48,9 @@ def read_file(path: Path) -> str:
 
 def read_url(url: str) -> str:
     return urlopen(url).read().decode().strip()
+
+def helm_pull(chart_name: str, repo_url: str, version: str, destination: Path) -> None:
+    parse_output(["helm", "repo", "add", chart_name, repo_url])
+    parse_output(["helm", "pull", f"{repo_url}/{chart_name}", "--version", version, "--destination", destination])
```
neoaggelos, line 54: this was not addressed, use `helm pull $chart --repo $repo`

build-scripts/hack/util.py
```diff
@@ -48,3 +48,9 @@ def read_file(path: Path) -> str:
...
+
+    LOG.info("Pulled helm repository %s @ %s as %s to %s", repo_url, version, chart_name, destination)
```
neoaggelos, line 56: newline at EOF

neoaggelos: also, message should read like `LOG.info("Pulled helm chart %s @ %s from %s at %s", chart_name, version, repo_url destination)`

k8s/manifests/charts/ck-ingress-tls/templates/tests/tlscertificatedelegation.yaml
neoaggelos, line 1: This is only going to be used by something like `helm test`, which we do not do. Do we need this for something else?

src/k8s/pkg/k8sd/features/cilium/gateway.go
neoaggelos: This is irrelevant for the context of this PR, and is also currently handled elsewhere https://github.com/canonical/k8s-snap/blob/5f34bb08c02c59d8df6287aef8a151ab96f45f86/src/k8s/pkg/k8sd/types/cluster_config_validate.go#L33-L45

I would recommend not doing the same check in multiple places around the code, because that would be incredibly hard to then track down and adjust the code in the future.

For now, please revert this

tests/integration/tests/test_ingress.py
```diff
@@ -14,23 +15,35 @@
...
+            "-o=json",
         ],
         capture_output=True,
     )
-    ingress_http_port = p.stdout.decode().replace("'", "")
+
+    ingress_http_port = None
+    services = json.loads(p.stdout.decode())
+
+    ingress_services = [svc for svc in services["items"] if "ingress" in svc["metadata"]["name"]]
```
neoaggelos, line 36: this is dangerous and could very well lead in a future where this test passes due to false positives. We know the resource we have deployed, so we should check for that.

louiseschmidtgen: What would you suggest here? I added this to check for *ingress or in the other test *my-gateway because cilium seems to add their name in front of the resources it creates... So I can't check for the exact resource name with contour and cilium...

src/k8s/pkg/k8sd/features/cilium/gateway.go
louiseschmidtgen: Yes, I wanted to move this to the feature implementation because it's not applicable for contour...

k8s/manifests/charts/ck-ingress-tls/templates/tests/tlscertificatedelegation.yaml
louiseschmidtgen: Apologies, I didn't get to testing the default-tls-flag as thouroughly as I wanted to. I moved it to templates this resource get's created since there is no flag to set for it like in cilium. So we do something like this: [here](https://github.com/projectcontour/contour/issues/2488)

build-scripts/hack/util.py
louiseschmidtgen: Removed the helm repo add.

src/k8s/pkg/k8sd/features/cilium/gateway.go
louiseschmidtgen: Ok, I will remove it from this PR 👍🏼

build-scripts/hack/util.py
neoaggelos: you also need `--repo $repo_url`, otherwise this will not work

tests/integration/tests/test_gateway.py
```diff
@@ -47,12 +50,26 @@ def test_gateway(session_instance: harness.Instance):
...
         ],
         capture_output=True,
     )
-    gateway_http_port = p.stdout.decode().replace("'", "")
+
+    services = json.loads(p.stdout.decode())
+
+    ingress_services = [
+        svc for svc in services["items"] if "my-gateway" in svc["metadata"]["name"]
+    ]
```
neoaggelos, line 62: left the same comment but i don't it has been addressed. first of all, this lists `ingress_services` but the tes is about gateway.

secondly, this check is not accurate, as it's quite easy to get false positives by other resources that exist on the cluster. this would be very hard to track down, and result in bugs finding their way in the code without being found.

We know what is the resource that we just deployed, please check directly for that resource

neoaggelos (changes requested):
Also, please make sure that there is a list of _all_ images that are required by contour with your changes.

This can be done by deploying everything, and then checking the list of images reported by the container runtime with a command like

```
k8s kubectl get node -o template='{{ range .items }}{{ .metadata.name }}{{":"}}{{ range .status.images }}{{ "\n- " }}{{ index .names 1 }}{{ end }}{{"\n"}}{{ end }}'
```

Please add the list of images as a comment on this PR

tests/integration/tests/test_ingress.py
neoaggelos: in that case, do not try to guess around implementation details. rather, start from the gateway object and track down the linked service.

eaudetcobello:
I think this needs a waitForRequiredCRDs as well like Cilium and Metallb have.

src/k8s/pkg/k8sd/features/contour/ingress.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if changed {
+		if err := rolloutRestartContour(ctx, snap, 3); err != nil {
+			return fmt.Errorf("failed to rollout restart contour to apply ingress: %w", err)
+		}
+	}
+
+	// Install the delegation resource for the default TLS secret.
+	// The default TLS secret is created by the user created,
+	// and gets set via k8s set defaultTLSSecret=bananas.
```
eaudetcobello, line 72:
```suggestion
	// The default TLS secret is created by the user
	// and gets set via k8s set defaultTLSSecret=bananas.
```

src/k8s/pkg/k8sd/features/contour/gateway.go
```diff
@@ -0,0 +1,68 @@
...
+
+	return control.WaitUntilReady(ctx, func() (bool, error) {
+		resources, err := client.ListResourcesForGroupVersion("apiextensions.k8s.io/v1")
+		if err != nil {
+			// This error is expected if the group version is not yet deployed.
+			return false, nil
+		}
+
+		requiredCRDs := map[string]bool{
+			"customresourcedefinitions": true,
```
eaudetcobello, line 57: This doesn't seem right. Maybe `gatewayclasses.gateway.networking.k8s.io`?

eaudetcobello (changes requested):
Nice work! Two comments and then lgtm

src/k8s/pkg/k8sd/features/contour/gateway.go
louiseschmidtgen: I was a little bit confused here too but it's the equivalent of doing:
```
sudo k8s kubectl api-resources --api-group=apiextensions.k8s.io 
NAME                        SHORTNAMES   APIVERSION                NAMESPACED   KIND
customresourcedefinitions   crd,crds     apiextensions.k8s.io/v1   false        CustomResourceDefinition
```

src/k8s/pkg/k8sd/features/contour/gateway.go
```diff
@@ -46,27 +46,48 @@ func waitForRequiredContourCommonCRDs(ctx context.Context, snap snap.Snap) error
...
 			// This error is expected if the group version is not yet deployed.
 			return false, nil
 		}
 
-		requiredCRDs := map[string]bool{
-			"contourconfigurations.projectcontour.io":     true,
-			"contourdeployments.projectcontour.io":        true,
-			"extensionservices.projectcontour.io":         true,
-			"httpproxies.projectcontour.io":               true,
-			"tlscertificatedelegations.projectcontour.io": true,
```
neoaggelos: i see why we are doing this, but the changed code is now doing lots of memory allocations, and that's going to happen often, so let's try to avoid it.

You could keep the same code, by using `$group:$name` as your keys, e.g. `projectcontour.io/v1:tlscertificategelegations.projectcontour.io`.

src/k8s/pkg/k8sd/features/contour/gateway.go
```diff
@@ -46,27 +46,48 @@ func waitForRequiredContourCommonCRDs(ctx context.Context, snap snap.Snap) error
...
+		requiredMap := make(map[string]bool)
+		for _, crd := range required {
+			requiredMap[crd] = true
 		}
 
-		requiredCount := len(requiredCRDs)
 		for _, resource := range resources.APIResources {
-			if _, exists := requiredCRDs[resource.Name]; exists {
-				requiredCount--
+			delete(requiredMap, resource.Name)
```
neoaggelos, line 75: then this becomes `if _, exists := requiredCRDs[fmt.Sprintf("%s:%s", resource.Group, resource.Name)]`

src/k8s/pkg/k8sd/features/contour/gateway.go
louiseschmidtgen: Great idea! Thanks :)

build-scripts/hack/util.py
```diff
@@ -48,3 +48,19 @@ def read_file(path: Path) -> str:
 
 def read_url(url: str) -> str:
     return urlopen(url).read().decode().strip()
+
+def helm_pull(chart, repo_url: str, version: str, destination: Path) -> None:
+    parse_output(
+        [
+            "helm",
+            "pull",
+            f"{repo_url}",
```
neoaggelos, line 57: Please see https://github.com/canonical/k8s-snap/pull/466#discussion_r1647141166

neoaggelos (commented):
Almost there! Nice work here @louiseschmidtgen !

louiseschmidtgen:
Ah interesting the api-group seems to be simply projectcontour.io:
```
10:09:48 louise@louise-Inspiron-5580 k8s-snap ±|KU-864/contour ✗|→ sudo k8s kubectl api-resources --api-group=projectcontour.io
NAME                        SHORTNAMES                           APIVERSION                   NAMESPACED   KIND
contourconfigurations       contourconfig                        projectcontour.io/v1alpha1   true         ContourConfiguration
contourdeployments          contourdeploy                        projectcontour.io/v1alpha1   true         ContourDeployment
extensionservices           extensionservice,extensionservices   projectcontour.io/v1alpha1   true         ExtensionService
httpproxies                 proxy,proxies                        projectcontour.io/v1         true         HTTPProxy
tlscertificatedelegations   tlscerts                             projectcontour.io/v1         true         TLSCertificateDelegation
10:10:02 louise@louise-Inspiron-5580 k8s-snap ±|KU-864/contour ✗|→ sudo k8s kubectl api-resources --api-group=projectcontour.io/v1
NAME   SHORTNAMES   APIVERSION   NAMESPACED   KIND
10:10:17 louise@louise-Inspiron-5580 k8s-snap ±|KU-864/contour ✗|→ sudo k8s kubectl api-resources --api-group=projectcontour.io/v1alpha1 
NAME   SHORTNAMES   APIVERSION   NAMESPACED   KIND
```

src/k8s/pkg/k8sd/features/contour/gateway.go
louiseschmidtgen: resources.Group and resource.Version are optional and don't get passed. I'm only going to check the resource.Name.

tests/integration/tests/test_gateway.py
```diff
@@ -37,8 +38,10 @@ def test_gateway(session_instance: harness.Instance):
         ]
     )
 
+    # Get gateway node port
+    gateway_http_port = None
     util.stubbornly(retries=5, delay_s=2).on(session_instance).until(
-        lambda p: "cilium-gateway-my-gateway" in p.stdout.decode()
+        lambda p: "my-gateway" in p.stdout.decode()
     ).exec(["k8s", "kubectl", "get", "service", "-o", "json"])
```
neoaggelos, line 45: this code here can be dangerous. we repeatedly get the list of services until `my-gateway` appears on the text. this is right, as the service might take some time to appear (perhaps the 5 retries with 2 seconds backoff is also very short).

However, we then retrieve them again to parse the service info and retrieve the nodePort. Note that this nodePort might not be available upon service creation, so there is a race here: If the test retrieves the service before the ports are configured, then our test will flake.

Instead, we should do all of the logic inside the `util.stubbornly`. Currently, it's using a lambda, but we should probably write this as a proper function instead, then do the retries until we get the nodePort we were after.

If we don't get the nodePort in time, that's OK, the timeout here will tell us about the failure

tests/integration/tests/test_ingress.py
```diff
@@ -14,23 +15,44 @@
 def test_ingress(session_instance: List[harness.Instance]):
 
     util.stubbornly(retries=5, delay_s=2).on(session_instance).until(
-        lambda p: "cilium-ingress" in p.stdout.decode()
-    ).exec(["k8s", "kubectl", "get", "service", "-n", "kube-system", "-o", "json"])
+        lambda p: "ingress" in p.stdout.decode()
+    ).exec(["k8s", "kubectl", "get", "service", "-A", "-o", "json"])
```
neoaggelos, line 19: same here, we fetch services until we see some arbitrary output, then fetch them again. this should all be done inside the `util.stubbornly` here

neoaggelos (approved):
Approved due to pressure of time, but we need to revisit the tests, as they are a possible source of flakes in the future.

PR #472, opened 2024-06-05
Remove deadcode checker

bschimke95:
The deadcode checker doesn't work well with our features setup as it detects non-main feature modules (that are used e.g. in moonray) as unused and blocks the CI.
That checker hasn't really contributed anything useful anyway, so let's just remove it.


louiseschmidtgen (approved):
LGTM

bschimke95:
Not necessary, we can use Go tags instead: https://github.com/canonical/k8s-snap/pull/473

PR #473, opened 2024-06-05
Add moonray implementation and go ignore tags

bschimke95:
Add moonray implementation that (will) contain the features for moonray. This is ignored by default and only enabled by the moonray patch (which in turn also disables the `default_implementation`)


berkayoz (approved):
LGTM

PR #474, opened 2024-06-06
Improve Network Integration Tests

mateoflorido:
## Overview
Make the network tests more generic. This will allow us to test on any implementation of the network feature.

## Changes
- Removed Cilium bits from the Network integration tests
- Drive-by fixes for the `x-wait-for` error messages and improvements to the StatusInterface docs.

### DIFF
--- a/tests/integration/tests/test_network.py
+++ b/tests/integration/tests/test_network.py
@@ -51,13 +51,6 @@ def test_network(session_instance: harness.Instance):
     assert len(out["items"]) > 0, "No NGINX pod found"
     podIP = out["items"][0]["status"]["podIP"]
 
-    p = session_instance.exec(
-        [
-            "curl",
-            "-s",
-            f"http://{podIP}",
-        ],
-        capture_output=True,
-    )
-
-    assert "Welcome to nginx!" in p.stdout.decode()
+    util.stubbornly(retries=5, delay_s=5).on(session_instance).until(
+        lambda p: "Welcome to nginx!" in p.stdout.decode()
+    ).exec(["curl", "-s", f"http://{podIP}"])
--- a/tests/integration/tests/test_network.py
+++ b/tests/integration/tests/test_network.py
@@ -12,48 +12,6 @@ LOG = logging.getLogger(__name__)
 
 
 def test_network(session_instance: harness.Instance):
-    p = session_instance.exec(
-        [
-            "k8s",
-            "kubectl",
-            "get",
-            "pod",
-            "-n",
-            "kube-system",
-            "-l",
-            "k8s-app=cilium",
-            "-o",
-            "json",
-        ],
-        capture_output=True,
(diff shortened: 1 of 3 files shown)


tests/integration/tests/test_network.py
```diff
@@ -78,20 +36,28 @@ def test_network(session_instance: harness.Instance):
...
+    p = session_instance.exec(
+        [
+            "curl",
+            "-s",
+            f"http://{podIP}",
+        ],
+        capture_output=True,
+    )
+
+    assert "Welcome to nginx!" in p.stdout.decode()
```
louiseschmidtgen, line 63:
```suggestion
    assert "Welcome to nginx!" in p.stdout.decode(), "No NGINX welcome message found"
```

louiseschmidtgen (approved):
Little suggestion, LGTM

tests/integration/tests/test_network.py
```diff
@@ -78,20 +36,28 @@ def test_network(session_instance: harness.Instance):
...
             "json",
         ],
         capture_output=True,
     )
-    assert "nginx" in p.stdout.decode().strip()
+
+    out = json.loads(p.stdout.decode())
+
+    assert len(out["items"]) > 0, "No NGINX pod found"
+    podIP = out["items"][0]["status"]["podIP"]
```
neoaggelos, line 52: this needs some sort of retry in case the pod is not yet ready?

bschimke95 (commented):
Great work Mateo. 
Two things:
* Please rebase on https://github.com/canonical/k8s-snap/pull/475 on ce this is merged, so that we can check if this works for all flavours.
* This covers `test_network`, we also need to update the gateway, ingress and loadbalancer tests. Should we add this changes here or just create follow-up cards?

tests/integration/tests/test_network.py
mateoflorido: Sure, lines 11-33 should wait until the pod reaches `condition=ready`. Do you want to retry the other parts of the test as well?

louiseschmidtgen:
> Great work Mateo. Two things:
> 
>     * Please rebase on [Add moonray integration tests CI #475](https://github.com/canonical/k8s-snap/pull/475) on ce this is merged, so that we can check if this works for all flavours.
> 
>     * This covers `test_network`, we also need to update the gateway, ingress and loadbalancer tests. Should we add this changes here or just create follow-up cards?

The gateway, ingress and loadbalancer tests are assigned to Etienne and I 👍🏼

tests/integration/tests/test_network.py
bschimke95: The pod should always be ready at this point. As Mateo already said, we wait for the pod to come up further above.

bschimke95 (approved):
LGTM, great stuff @mateoflorido

PR #476, opened 2024-06-07
Add CI to handle auto test and promote to strict and moonray branches

neoaggelos:
### Summary

Adjust CI jobs for the informing strict integration tests and the auto-update jobs to handle the moonray patch as well.


.github/workflows/integration-informing.yaml
```diff
@@ -1,4 +1,4 @@
-name: Integration Tests (Strict)
+name: Informing Integration Tests
```
eaudetcobello, line 1: What does 'Informating integration tests' mean? I think this name should be changed

eaudetcobello (approved):
Good stuff. One comment.

src/k8s/pkg/k8sd/features/calico/status.go
```diff
@@ -6,42 +6,56 @@ import (
...
 		}
-		if !isReady {
-			return false, nil
+		if len(pods) == 0 {
+			return false, fmt.Errorf("no %v pods exist on the cluster", check.name)
+		}
+
+		for _, pod := range pods {
+			if !podIsReady(pod) {
+				return false, fmt.Errorf("%v pod %q not ready", check.name, pod.Name)
```
bschimke95, line 57: If we return an error if the pods are not ready in all `CheckX` functions, why do we have the `bool` at all?

src/k8s/pkg/k8sd/features/calico/status.go
```diff
@@ -6,42 +6,56 @@ import (
...
+			return false, fmt.Errorf("failed to get %v pods: %w", check.name, err)
 		}
-		if !isReady {
-			return false, nil
+		if len(pods) == 0 {
+			return false, fmt.Errorf("no %v pods exist on the cluster", check.name)
+		}
+
+		for _, pod := range pods {
+			if !podIsReady(pod) {
```
bschimke95, line 56: nit: instead of having this local function, we could refactor the `client.IsPodReady` or have a `client.PodsAreReady(ctx, namespace, ListOptions)` function. 
Feels a bit weird to have this generic pod checker in here.
Up to you

bschimke95 (approved):
LGTM, minor comments

src/k8s/pkg/k8sd/features/calico/status.go
neoaggelos: yes, exactly, there is followup work to change `(bool, error)` to simply `error`. Kept this out of the PR for brevity.

src/k8s/pkg/k8sd/features/calico/status.go
neoaggelos: yes, exactly, there is followup work to change how `client.ListPods` and `client.IsPodReady` to address the issues we discuss offline. Kept this out of the PR for brevity.

.github/workflows/update-branches.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
+  update:
+    name: "${{ matrix.patch }}"
+    permissions:
+      contents: write  # for Git to git push
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        patch: ["strict", "moonray"]
```
louiseschmidtgen, line 20: Could you please add the patch matrix to our nightly run as well or add a follow-up card?

louiseschmidtgen (approved):
Thank you @neoaggelos! 
LGTM, let's make sure we have cards for the follow-up work.

PR #478, opened 2024-06-10
Add optional pebble support

neoaggelos:
### Summary

Support using pebble as service manager for k8sd (k8sd decides at runtime by means of an environment variable). How this environment is set is outside the scope of this PR.


src/k8s/pkg/snap/pebble.go
```diff
@@ -0,0 +1,71 @@
...
+func (s *pebble) RestartService(ctx context.Context, name string) error {
+	return s.runCommand(ctx, []string{filepath.Join(s.snapDir, "bin", "pebble"), "restart", name})
+}
+
+func (s *pebble) Strict() bool {
+	return false
+}
+
+func (s *pebble) OnLXD(ctx context.Context) (bool, error) {
+	return true, nil
```
bschimke95, line 60: Why is this always true?

bschimke95 (approved):
LGTM (once CI succeeds), only two comments for context.

src/k8s/pkg/snap/pebble.go
neoaggelos: Using pebble is for environments where we do not control the `/sys` in any way, so it's safe to just do this. Note that this does not (currently) have any false positives, where valid deployments that can mount things automatically will break otherwise.

PR #485, opened 2024-06-11
Add extra arguments for each service

bschimke95:
Provide a way to pass to provide arbitrary extra arguments to each `k8sd` service. Those parameters are applied on a node-level and not shared across the cluster. 

We want to support power users to supply custom configuration to the individual services, to allow management of setups like setting up kube-apiserver for encryption at rest, kubelet topology configuration, etc.

Extra node config files will be addressed in a follow-up PR. This one is already big enough.


src/k8s/pkg/k8sd/setup/util_extra_files.go
```diff
@@ -0,0 +1,45 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to create file %s: %w", filePath, err)
+		}
+		defer file.Close()
+
+		// Write the content to the file
+		_, err = file.WriteString(content)
+		if err != nil {
+			return fmt.Errorf("failed to write to file %s: %w", filePath, err)
+		}
```
neoaggelos, line 33: why not `os.WriteFile`?

src/k8s/pkg/snap/util/services.go
```diff
@@ -75,3 +75,20 @@ func StopK8sDqliteServices(ctx context.Context, snap snap.Snap) error {
 	}
 	return nil
 }
+
+// ServiceArgsFromMap processes a map of string pointers and categorizes them into update and delete lists.
+// - If the value pointer is nil, it adds the argument name to the delete list.
+// - If the value pointer is not nil, it adds the argument and its value to the update map.
+func ServiceArgsFromMap(args map[string]*string) (map[string]string, []string) {
```
neoaggelos, line 82: `snaputil` is reserved for utilities that interact with snaps. This is not one of them, so let's move it to `pkg/utils`

tests/integration/tests/test_smoke.py
```diff
@@ -17,3 +17,25 @@ def test_smoke(session_instance: harness.Instance):
     config = result.stdout.decode()
     assert len(config) > 0
     assert "server: https://192.168.210.41" in config
+
```
neoaggelos, line 20: looks like we would be better using a separate instance in `test_smoke`. That way we can keep the bootstrap config (the extra configs) closer to the test code here. Because right now, they are very far away, and it is not obvious how the bootstrap config reaches this place.

tests/integration/tests/test_util/util.py
```diff
@@ -223,8 +223,8 @@ def get_join_token(
 
 
 # Join an existing cluster.
-def join_cluster(instance: harness.Instance, join_token: str):
-    instance.exec(["k8s", "join-cluster", join_token])
+def join_cluster(instance: harness.Instance, join_token: str, *args: str):
+    instance.exec(["k8s", "join-cluster", join_token, *args])
```
neoaggelos, line 227: is this used yet? let's revert

bschimke95: leftover from extra test. removed

tests/integration/tests/test_smoke.py
bschimke95: done

tests/integration/tests/test_smoke.py
```diff
@@ -2,18 +2,52 @@
...
+    # For each service, verify that the extra arg was written to the args file.
+    for service, value in {
+        "kube-apiserver": "--request-timeout=2m",
+        "kube-controller-manager": "--leader-elect-retry-period=3s",
+        "kube-scheduler": "--authorization-webhook-cache-authorized-ttl=11s",
+        "kube-proxy": "--config-sync-period=14m",
+        "kubelet": "--authentication-token-webhook-cache-ttl=3m",
+        "containerd": "--log-level=debug",
+        "k8s-dqlite": "--watch-storage-available-size-interval=6s",
+    }.items():
```
neoaggelos, line 49: hmm, as discussed earlier, this is a unit test in disguise. This is important because we would not like this test to break if any of the flags here have a side-effect on the respective service.

But I see why this is here given the situation.

neoaggelos (commented):
Please also update the docs reference accordingly

tests/integration/tests/test_smoke.py
bschimke95: I partially agree. However, we currently don't have a good way to unittest the logic in e.g. `hook_bootstrap.go`. I much rather have this unittest in disguise than silently breaking the extra-args feature which can go unnoticed for quite a while. 
I don't think the changed arguments will affect this test (**famous last words...**). All of them are very close to the default and only differ by a second.

docs/src/snap/reference/bootstrap-config-reference.md
```diff
@@ -437,6 +443,79 @@ If omitted defaults to an auto generated certificate.
 The key to be used for the kubelet.
 If omitted defaults to an auto generated key.
 
+### extra-node-config-files:
```
neoaggelos, line 446: other sections don't have a `:`

neoaggelos (approved):
final nit

PR #486, opened 2024-06-11
Update external-datastore.md

eaudetcobello:
Just so we don't have to check if it goes under the cluster-config key or not.

### DIFF
--- a/docs/src/snap/howto/external-datastore.md
+++ b/docs/src/snap/howto/external-datastore.md
@@ -23,8 +23,8 @@ There is no migration path between the bundled dqlite and the external datastore
 
 To use an external datastore, a configuration file that contains the required
 datastore parameters needs to be provided to the bootstrap command.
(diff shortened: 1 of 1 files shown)


bschimke95 (commented):
Out of curiosity: Where does this change come from, did anyone complain or did this wrong? 
You basically just rephrase what is implicitly there already hence I'm asking.

Before approval, please make the linter happy.

eaudetcobello:
> Out of curiosity: Where does this change come from, did anyone complain or did this wrong? 
> 
> You basically just rephrase what is implicitly there already hence I'm asking.
> 
> 
> 
> Before approval, please make the linter happy.

I don't think it's simply a rephrase, there is information added. The original "Insert the content below" does not make it obvious if the data should go under the cluster-config key or at the root level.

bschimke95:
You are probably right, and I just am too much exposed to this codebase already.

PR #487, opened 2024-06-11
Update bootstrap-config-reference.md

eaudetcobello:
It can be confusing that the datastore address needs the protocol and if it doesn't have it bootstrapping fails.

### DIFF
--- a/docs/src/snap/reference/bootstrap-config-reference.md
+++ b/docs/src/snap/reference/bootstrap-config-reference.md
@@ -302,7 +302,8 @@ Can be used to point to an external datastore like etcd.
 **Type:** `list[string]` <br>
 **Required:** `No` <br>
 
-The server addresses to be used when `datastore-type` is set to `external`.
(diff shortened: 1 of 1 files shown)


docs/src/snap/reference/bootstrap-config-reference.md
```diff
@@ -302,7 +302,8 @@ Can be used to point to an external datastore like etcd.
 **Type:** `list[string]` <br>
 **Required:** `No` <br>
 
-The server addresses to be used when `datastore-type` is set to `external`.
+The server addresses, in URL form,
```
bschimke95, line 305: The url form does not require the protocol: https://stackoverflow.com/questions/59600332/is-a-protocol-eg-http-or-https-required-for-a-url-to-be-valid

Maybe write the explicit format or an example instead.

bschimke95 (commented):
Small comment, please also resolve the line length issues.

PR #490, opened 2024-06-12
Add extra config files parameters

bschimke95:
Extends the `Bootstrap/JoinConfigs` to allow additional files to be loaded when joining the cluster. Those can then e.g. be pointed to by `ExtraNodeXArgs`.

Also, added an integration test that verifies all extra service configs are correctly set and extra files are uploaded to the correct place on the node.


bschimke95:
Closed and merged into https://github.com/canonical/k8s-snap/actions/runs/9484275968/job/26164829138?pr=485 to have this feature contained in one PR.

PR #491, opened 2024-06-12
Add cleanup interface, move cilium cleanup and add calico cleanup

berkayoz:
Add cleanup interface, move cilium cleanup from bash and add calico cleanup


src/k8s/pkg/k8sd/features/calico/cleanup.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"os/exec"
+	"regexp"
+	"strings"
+
+	"github.com/canonical/k8s/pkg/snap"
+)
+
+func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
+	output, err := exec.CommandContext(ctx, "ip", "-j", "link", "show").Output()
```
neoaggelos, line 15: use `net.Interfaces` instead

neoaggelos: that way we use golang's library and do not rely on the arguments passed to an external tool

src/k8s/pkg/k8sd/features/calico/cleanup.go
```diff
@@ -0,0 +1,88 @@
...
+				return fmt.Errorf("failed to delete interface %s: %w", iface.Name, err)
+			}
+		}
+	}
+
+	// List network namespaces in JSON format
+	nsOutput, err := exec.CommandContext(ctx, "ip", "-j", "netns", "list").Output()
+	if err != nil {
+		return fmt.Errorf("failed to list network namespaces: %w", err)
+	}
```
neoaggelos, line 49: i would greatly recommend listing `/run/netns` or through the `netns` filesystem mounts

src/k8s/pkg/k8sd/features/calico/cleanup.go
```diff
@@ -0,0 +1,88 @@
...
+		return fmt.Errorf("failed to parse network namespace JSON: %w", err)
+	}
+
+	// Delete the namespaces that start with "cali-"
+	for _, ns := range namespaces {
+		if strings.HasPrefix(ns.Name, "cali-") {
+			// Delete the namespace
+			if _, err := exec.CommandContext(ctx, "ip", "netns", "delete", ns.Name).CombinedOutput(); err != nil {
+				return fmt.Errorf("failed to delete network namespace %s: %w", ns.Name, err)
+			}
```
neoaggelos, line 67: i would greatly recommend a umount, see https://github.com/vishvananda/netns/blob/v0.0.4/netns_linux.go#L83C1-L92C2

src/k8s/pkg/k8sd/features/calico/cleanup.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+
+	if _, err := exec.Command("bash", "-c", "iptables-legacy-save | grep -iv cali | iptables-legacy-restore").CombinedOutput(); err != nil {
```
neoaggelos, line 71: this needs the equivalent of `set -o pipefail`. an easier way to approach this is drop the bash pipes, parse the output of the first command, remove matching lines and then call `iptables-legacy-restore` with the right stdin.

same below

neoaggelos (commented):
Great work, some comments

neoaggelos:
Don't forget https://github.com/canonical/k8s-snap/blob/6772673c4c7981a953d53cb080ae37ac5126efff/build-scripts/patches/moonray/apply#L18

src/k8s/pkg/k8sd/features/calico/cleanup.go
```diff
@@ -0,0 +1,81 @@
...
+func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return fmt.Errorf("failed to list network interfaces: %w", err)
+	}
+
+	// Find the interfaces created by Calico
+	for _, iface := range interfaces {
+		// Check if the interface name matches the regex pattern
+		match, err := regexp.MatchString("^vxlan[-v6]*.calico|cali[a-f0-9]*$", iface.Name)
```
bschimke95, line 27: Why do you use a more advanced regexp for the interfaces but for iptables grepping for "cali" is sufficient?

bschimke95 (approved):
Great work, only a small question.

src/k8s/pkg/k8sd/features/calico/cleanup.go
```diff
@@ -0,0 +1,81 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to read iptables rules: %w", err)
+		}
+
+		grep := exec.Command("grep", "-iv", "cali")
+		grep.Stdin = bytes.NewReader(out)
+		grepOut, err := grep.Output()
+		if err != nil {
+			return fmt.Errorf("failed to filter calico iptables rules: %w", err)
+		}
```
neoaggelos, line 71: this should not be a command. just iterate over the output lines and drop the ones that do not match

src/k8s/pkg/k8sd/features/cilium/cleanup.go
```diff
@@ -0,0 +1,26 @@
...
+	if err := os.Remove("/var/run/cilium/cilium.pid"); err != nil {
+		return fmt.Errorf("failed to remove cilium pid file: %w", err)
+	}
+
+	if _, err := os.Stat("/opt/cni/bin/cilium-dbg"); err == nil {
+		if err := exec.CommandContext(ctx, "/opt/cni/bin/cilium-dbg", "cleanup", "--all-state", "--force").Run(); err != nil {
+			return fmt.Errorf("cilium-dbg cleanup failed: %w", err)
+		}
+	} else {
+		return fmt.Errorf("cilium-dbg can't be accessed: %w", err)
```
neoaggelos, line 22: cleanup is done on a best-effort basis. we have the `k8s .. cleanup || true` at the top-level, true. but this here could be problematic.

e.g: if the pid file does not exist, that's not a problem, it doesn't exist so we can just proceed. etc etc.

src/k8s/pkg/k8sd/features/calico/cleanup.go
neoaggelos: Also, does this catch the interfaces in case of bird (e.g. `tunl0`)?

berkayoz: Calico operates on multiple modes, can create vxlan, bird and various interfaces which does not always contain a keyword. I've added `tunl` interfaces to the list.

src/k8s/pkg/k8sd/features/cilium/cleanup.go
```diff
@@ -0,0 +1,22 @@
...
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/canonical/k8s/pkg/snap"
+)
+
+func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
+	os.Remove("/var/run/cilium/cilium.pid")
```
neoaggelos, line 13: it is preferred to either log the error, or be explicit that you ignore it, by means of `_ = os.Remove("/var/run/cilium/cilium.pid")`. but we don't have good logging practices in place, so that's acceptable as is

src/k8s/cmd/k8s/k8s_x_cleanup.go
```diff
@@ -0,0 +1,41 @@
...
+			ctx, cancel := context.WithTimeout(cmd.Context(), opts.timeout)
+			defer cancel()
+
+			if err := features.Cleanup.CleanupNetwork(ctx, env.Snap); err != nil {
+				cmd.PrintErrf("Error: failed to cleanup network: %v\n", err)
+				env.Exit(1)
+			}
+		},
+	}
+	cleanupNetworkCmd.Flags().DurationVar(&opts.timeout, "timeout", 5*time.Minute, "maximum time to wait")
```
eaudetcobello, line 30: Other files say "the max time to wait for the command to execute". This should be changed for consistency. It should also be changed in k8s_x_wait_for.go

```suggestion
	cleanupNetworkCmd.Flags().DurationVar(&opts.timeout, "timeout", 5*time.Minute, "the max time to wait for the command to execute")
```

src/k8s/pkg/k8sd/features/calico/cleanup.go
```diff
@@ -0,0 +1,80 @@
...
+func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return fmt.Errorf("failed to list network interfaces: %w", err)
+	}
+
+	// Find the interfaces created by Calico
+	for _, iface := range interfaces {
+		// Check if the interface name matches the regex pattern
+		match, err := regexp.MatchString("^vxlan[-v6]*.calico|cali[a-f0-9]*|tunl[0-9]*$", iface.Name)
```
eaudetcobello, line 26: Can you put a comment for the source of this regex pattern?

eaudetcobello (changes requested):

src/k8s/pkg/k8sd/features/cilium/cleanup.go
berkayoz: Agreed, I wasn't sure how we do logging across the code so open to suggestions there if not I can make it explicit the way you've described.

eaudetcobello (approved):

PR #494, opened 2024-06-13
add arguments to improve cis compliance and adjust tests accordingly

eaudetcobello:
This adds default arguments to kubelet, api server, controller manager services that improve our compliance with CIS guidelines. Also modifies our tests to reflect these changes.


bschimke95 (approved):
LGTM

PR #497, opened 2024-06-14
Add YAML to Markdown conversion script and generated Markdown files

eaudetcobello:
This contains every control in CIS adapted to Canonical K8s in markdown form. There is also the python script and jinja template.

Note this doesn't have the metadata for webpages yet.


k8s/scripts/cis-template.jinja2
```diff
@@ -0,0 +1,34 @@
+<!-- markdownlint-disable -->
+## {{ title | trim }}
+{% for group in groups %}
+### {{group.text}}
+{% for check in group.checks %}
+#### Control {{ check.id }}
+
+{% if check.text %}
+Description: `{{ check.text | trim }}`
```
evilnick, line 9:
```suggestion
Description: `{{ check.text | trim | wordwrap(64) }}`
```

k8s/scripts/cis-template.jinja2
```diff
@@ -0,0 +1,34 @@
+<!-- markdownlint-disable -->
```
evilnick, line 1: I would rather work on making the text compliant:
 -  It is easier to fix here than after the md is generated
 - the mdlint failing can be a good indicator that we have generated garbage

k8s/scripts/cis-template.jinja2
evilnick, line 9:
```suggestion
Description: `{{ check.text | trim | wordwrap(64) }}`
```
I think this is likely the only one likely to break line length

k8s/scripts/cis-template.jinja2
```diff
@@ -0,0 +1,34 @@
...
+{% for check in group.checks %}
+#### Control {{ check.id }}
+
+{% if check.text %}
+Description: `{{ check.text | trim }}`
+{% endif %}
+{% if check.audit %}
+
+Audit:
+```
```
evilnick, line 14: space code fences

k8s/scripts/cis-template.jinja2
```diff
@@ -0,0 +1,34 @@
...
+{{ check.audit | trim }}
+```
+{% endif %}
+
+{% if check.remediation %}
+Remediation:
+```
```
evilnick, line 21: space

k8s/scripts/cis-template.jinja2
```diff
@@ -0,0 +1,34 @@
...
+{{ check.remediation | trim }}
+```
+{% endif %}
+
+{% if check.tests %}
+Expected output:
+```
```
evilnick, line 28: space

k8s/scripts/cis-yaml-to-md.py
```diff
@@ -0,0 +1,115 @@
...
+USAGE = """
+python3 cis-yaml-to-md.py --input-directory=INPUT_DIR --output-directory=OUTPUT_DIR
+"""
+
+DESCRIPTION = """
+Parse the YAML files in the input directory and generate
+Markdown files in the output directory. The Markdown files are generated using the
+Jinja2 template file.
+
+It is expected that the input directory contains a config.yaml file. This file
```
evilnick, line 17: it might be handy to suggest where this comes from (I assume it is the cis config.yaml from https://raw.githubusercontent.com/canonical/kube-bench/ck8s/cfg/config.yaml)

evilnick (changes requested):
Nice so far! Questions and comments:

- The output files should not go directly to docs, that isn't how we do it. Create a 'cis' directory in '_parts' and put them in there
- As far as I understand, this is a list of the various tests and remediations which apply which we do need, but we also need/want a list of the ones which Fail without user steps. Are we going to autogenerate that too?
- i made a few suggestions on the script. I would rather try and generate something that is expected to pass linting because then if linting fails we know something has gone wrong
 - i will also a few notes on some of the generated output on things we can improve.

docs/src/snap/howto/cis-hardening/etcd.md
```diff
@@ -0,0 +1,198 @@
+<!-- markdownlint-disable -->
+## Etcd Node Configuration
+### Etcd Node Configuration
```
evilnick, line 3: why is this title a repeat of the above?

docs/src/snap/howto/cis-hardening/controlplane.md
```diff
@@ -0,0 +1,52 @@
+<!-- markdownlint-disable -->
+## Control Plane Configuration
+### Authentication and Authorization
```
evilnick, line 3: what options do we have for inserting text between these? run on headings  look terrible and are difficult to parse.

docs/src/snap/howto/cis-hardening/etcd.md
```diff
@@ -0,0 +1,198 @@
...
+Description: `Ensure that the --cert-file and --key-file arguments are set as appropriate (Automated)`
+
+Audit:
+```
+/bin/ps -ef | /bin/grep etcd | /bin/grep -v grep
+```
+
+Remediation:
+```
+Follow the etcd service documentation and configure TLS encryption.
```
evilnick, line 15: I know text like this comes from the actual config. What scope do we have to change that upstream to make these work better

k8s/scripts/cis-template.jinja2
eaudetcobello: Done

docs/src/snap/howto/cis-hardening/etcd.md
eaudetcobello: The subtitle is sometimes the same as the main title. I added an if statement that doesn't render the subtitle if it's the case.

evilnick:
Aren't the outputs still in the wrong order? I thought it should go:
- Audit
- Expected outcome
- Remediation
@ktsakalozos ?

Other than that, as none of this is actually visible I suggest we merge so:
- I can consume it in docs PR
- wecan further imrpove without blocking other work

docs/src/snap/howto/cis-hardening/controlplane.md
eaudetcobello: I added a new line between them for the moment. Let's plan work to add text between them next.

docs/src/snap/howto/cis-hardening/etcd.md
eaudetcobello: We can change them. Let's schedule work for this next.

evilnick (approved):

PR #498, opened 2024-06-14
Implement CLI API Endpoints for CAPI Clustering

mateoflorido:
## Overview
This pull request introduces two API endpoints for enabling support for CAPI clustering.

## Rationale
We need to provide a mechanism to issue tokens using a static token for ClusterAPI clustering. This pull request adds the `set-auth-token` sub-command (under the `x-capi` command), which allows the user to set a static token for requesting a join token for a node. Additionally, it includes the `GenerateJoinToken` endpoint, which is accessible through the API and will verify the token header against the one set in the database via the previously mentioned CLI command.

## Testing
This pull request has been manually tested.
```
root@kcp-1:~# k8s bootstrap
Bootstrapping the cluster. This may take a few seconds, please wait.
Bootstrapped a new Kubernetes cluster with node address "10.45.248.119:6400".
The node will be 'Ready' to host workloads after the CNI is deployed successfully.

root@kcp-1:~# k8s x-capi set-auth-token foo
root@kcp-1:~# curl --unix-socket /var/snap/k8s/common/var/lib/k8sd/state/control.socket localhost/1.0/x/capi/generate-join-token -H "token: foo" --request POST -H "Content-Type: application/json" --data '{"name": "kcp-2", "worker": false}'
{"type":"sync","status":"Success","status_code":200,"operation":"","error_code":0,"error":"","metadata":{"token":"eyJuYW1lIjoia2NwLTIiLCJzZWNyZXQiOiJiYWRmOWNkNjk0MDNmYWI2YTMzN2ZmZjk0ZmRmODk4ODE3MTVlMmY4ZTdmNjg0ZTNkZjFhMjEyMzY0OTU4YjgwIiwiZmluZ2VycHJp
bnQiOiJiZTc0YWI4MmYxYmNkOTU3NGJjNWQxZTk3MDlhMmFlMjdmZjdkZDJkOTk3MGFiNzVlNjJkOGQzZjhiMjI0MDI5Iiwiam9pbl9hZGRyZXNzZXMiOlsiMTAuNDUuMjQ4LjExOTo2NDAwIl19"}}
root@kcp-1:~# curl --unix-socket /var/snap/k8s/common/var/lib/k8sd/state/control.socket localhost/1.0/x/capi/generate-join-token -H "token: foo" --request POST -H "Content-Type: application/json" --data '{"name": "", "worker": true}'
{"type":"sync","status":"Success","status_code":200,"operation":"","error_code":0,"error":"","metadata":{"token":"eyJ0b2tlbiI6IiIsInNlY3JldCI6Indvcmtlcjo6ZGZmNzZlZmRkZDYyZTRlMmI2NDk2MjRmMDMzYzQ0MzM3ZDVlMjI0OCIsImpvaW5fYWRkcmVzc2VzIjpbIjEwLjQ1LjI0OC4x
MTk6NjQwMCJdLCJmaW5nZXJwcmludCI6ImJlNzRhYjgyZjFiY2Q5NTc0YmM1ZDFlOTcwOWEyYWUyN2ZmN2RkMmQ5OTcwYWI3NWU2MmQ4ZDNmOGIyMjQwMjkiLCJfIjoibSEhIn0"}}
root@kcp-1:~# curl --unix-socket /var/snap/k8s/common/var/lib/k8sd/state/control.socket localhost/1.0/x/capi/generate-join-token -H "token: bad-token" --request POST -H "Content-Type: application/json" --data '{"name": "", "worker": true}'
{"type":"error","status":"","status_code":0,"operation":"","error_code":401,"error":"invalid token","metadata":null}
```


src/k8s/pkg/k8sd/api/capi_auth.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/state"
+)
+
+func (e *Endpoints) postSetAuthToken(s *state.State, r *http.Request) response.Response {
+	request := apiv1.SetAuthTokenRequest{}
+	if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
+		return response.BadRequest(fmt.Errorf("failed to parse request: %w", err))
+	}
+	fmt.Println("request: ", request)
```
eaudetcobello, line 21: Is this intentional?

mateoflorido: whoops. Nice catch!

src/k8s/pkg/k8sd/database/capi_auth.go
```diff
@@ -0,0 +1,48 @@
...
+
+var (
+	capiTokenStmts = map[string]int{
+		"insert-token": MustPrepareStatement("capi-auth", "insert-token.sql"),
+		"select-token": MustPrepareStatement("capi-auth", "select.sql"),
+	}
+)
+
+// CheckAuthToken returns true if the specified token matches the ClusterAPI token.
+func CheckAuthToken(ctx context.Context, tx *sql.Tx, token string) (bool, error) {
```
neoaggelos, line 21: These will be used as `database.CheckAuthToken` and `database.SetAuthToken`. These names are very generic and do not communicate their purpose well-enough, without the caller having to lookup the implementation details to figure what they are used for. Further, looking for `database.CheckAuthToken()` can be very easily confused with different tokens (e.g. kubernetes tokens, join tokens, etc), which this thing is not.

Instead, consider names like `database.SetClusterAPIToken()` and `database.ValidateClusterAPIToken`, which clearly convey the purpose without callers having to double-check the implementation.

src/k8s/pkg/k8sd/database/capi_auth.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+
+	"github.com/canonical/microcluster/cluster"
+)
+
+var (
+	capiTokenStmts = map[string]int{
+		"insert-token": MustPrepareStatement("capi-auth", "insert-token.sql"),
+		"select-token": MustPrepareStatement("capi-auth", "select.sql"),
+	}
```
neoaggelos, line 17: There is no need to define a separate table for which only one row is ever used, and where we always check for the row ID to be 1. It is very "esoteric", and the error messages will not be useful for users to understand what is happening.

For simplicity, I would instead recommend using the cluster_configs table, with a config key like `clusterapi::auth-token` or something similar. This is functionally equivalent with what you already have, but re-uses existing constructs in the code more effectively.

src/k8s/pkg/k8sd/database/schema.go
```diff
@@ -13,6 +13,7 @@ import (
 
 var (
 	SchemaExtensions = []schema.Update{
+		schemaApplyMigration("capi-auth", "000-create.sql"),
```
neoaggelos, line 16: You do not need to add any new tables, see the other comment. Further, any database migrations are always "append-only", which means that new migrations are only supposed to be added at the tail of the existing list.

For some context: This is because applied migrations are typically tracked using the index of the most recent migration that was applied. So _prepending_ migrations can easily mess up this logic.

neoaggelos (changes requested):

src/k8s/pkg/k8sd/api/endpoints.go
```diff
@@ -89,5 +89,16 @@ func (e *Endpoints) Endpoints() []rest.Endpoint {
 			Path: "kubernetes/auth/webhook",
 			Post: rest.EndpointAction{Handler: e.postKubernetesAuthWebhook, AllowUntrusted: true},
 		},
+		// ClusterAPI management endpoints.
+		{
+			Name: "GenerateJoinToken",
+			Path: "x/capi/generate-join-token",
+			Post: rest.EndpointAction{Handler: e.postClusterJoinTokens, AccessHandler: ValidateCAPIAuthTokenAccessHandler("token")},
```
bschimke95, line 96: Just for my own understanding: We probably need to update the k8sd-proxy later to only forward requests that target the `x/capi/...` endpoints, right?

Should we also give the token a slightly more verbose name, e.g. `capi-auth-token`?

mateoflorido: Sounds good! I've changed the token header name

tests/integration/tests/test_smoke.py
```diff
@@ -54,3 +55,39 @@ def test_smoke(instances: List[harness.Instance]):
             ["cat", f"/var/snap/k8s/common/args/{service}"], capture_output=True
         )
         assert value in args.stdout.decode()
+
+
+def test_capi_auth(session_instance: harness.Instance):
```
bschimke95, line 60: please don't use the session_instance here. This instance is meant to test the features and should not alter configuration.

Just move that test into the above smoke test and use the instance there.

src/k8s/pkg/k8sd/database/capi_auth.go
```diff
@@ -0,0 +1,42 @@
+package database
+
+import (
+	"context"
+	"crypto/subtle"
+	"database/sql"
+	_ "embed"
```
bschimke95, line 7: This looks unused.

bschimke95 (commented):
Great work, some final comments

bschimke95 (approved):
LGTM, great stuff

src/k8s/api/v1/capi_config.go
```diff
@@ -0,0 +1,6 @@
+package v1
+
+// SetAuthTokenRequest is used to request to set the auth token for ClusterAPI.
+type SetAuthTokenRequest struct {
```
neoaggelos, line 4: `SetClusterAPIAuthToken`, and adjust any API endpoints, handlers, etc

src/k8s/pkg/k8s/client/capi_auth.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"context"
+	"fmt"
+
+	apiv1 "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/lxd/shared/api"
+)
+
+// SetAuthToken calls "POST 1.0/x/capi/set-auth-token".
+func (c *k8sdClient) SetAuthToken(ctx context.Context, request apiv1.SetAuthTokenRequest) error {
```
neoaggelos, line 12: adjust

src/k8s/pkg/k8s/client/interface.go
```diff
@@ -34,6 +34,8 @@ type Client interface {
 	UpdateClusterConfig(ctx context.Context, request apiv1.UpdateClusterConfigRequest) error
 	// GetClusterConfig retrieves configuration of the cluster.
 	GetClusterConfig(ctx context.Context, request apiv1.GetClusterConfigRequest) (apiv1.UserFacingClusterConfig, error)
+	// SetAuthToken sets the auth token for the CAPI provider.
+	SetAuthToken(ctx context.Context, request apiv1.SetAuthTokenRequest) error
```
neoaggelos, line 38: adjust

src/k8s/pkg/k8sd/api/capi_auth.go
```diff
@@ -0,0 +1,29 @@
...
+	"fmt"
+	"net/http"
+
+	apiv1 "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/k8s/pkg/k8sd/database"
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/state"
+)
+
+func (e *Endpoints) postSetAuthToken(s *state.State, r *http.Request) response.Response {
```
neoaggelos, line 16: adjust

src/k8s/pkg/k8sd/api/endpoints.go
```diff
@@ -89,5 +89,16 @@ func (e *Endpoints) Endpoints() []rest.Endpoint {
 			Path: "kubernetes/auth/webhook",
 			Post: rest.EndpointAction{Handler: e.postKubernetesAuthWebhook, AllowUntrusted: true},
 		},
+		// ClusterAPI management endpoints.
+		{
+			Name: "GenerateJoinToken",
```
neoaggelos, line 94: adjust, you can use `ClusterAPI/GenerateJoinToken`

src/k8s/pkg/k8sd/api/endpoints.go
```diff
@@ -89,5 +89,16 @@ func (e *Endpoints) Endpoints() []rest.Endpoint {
...
+			Path: "x/capi/generate-join-token",
+			Post: rest.EndpointAction{Handler: e.postClusterJoinTokens, AccessHandler: ValidateCAPIAuthTokenAccessHandler("capi-auth-token"), AllowUntrusted: true},
+		},
+		{
+			Name: "SetAuthToken",
```
neoaggelos, line 99: adjust, you can use `ClusterAPI/SetAuthToken`

src/k8s/pkg/k8sd/database/capi_auth.go
```diff
@@ -0,0 +1,50 @@
...
+		return false, fmt.Errorf("failed to prepare select statement: %w", err)
+	}
+
+	var storedToken string
+	err = selectTxStmt.QueryRowContext(ctx).Scan(&storedToken)
+	if err != nil {
+		return false, fmt.Errorf("failed to query ClusterAPI token: %w", err)
+	}
+
+	return subtle.ConstantTimeCompare([]byte(token), []byte(storedToken)) == 1, nil
```
neoaggelos, line 49: Change this so that the query is `SELECT ... WHERE value == "$tokenByUser AND key == "capi-token"`

So we do not fetch the key in memory then compare. We just check against the key and the row, and then verify whether we got a row back (== token matches)

This needs adjustment both here and in your SQL

neoaggelos (changes requested):

neoaggelos (approved):
Nice!

PR #499, opened 2024-06-17
Do not set --hostname-override on kubelet and kube-proxy if hostname matches

neoaggelos:
### Summary

When configuring kubelet and kube-proxy, only specify the `--hostname-override` flag if the node name does not match.

### Changes

- Add a `snap.Snap.Hostname() string` method, which returns `os.Hostname()` or falls back to `"dev"`
- When configuring kubelet and kube-proxy, if the node name (coming from microcluster config) does not match the hostname (from snap.Hostname()`, only then configure `--hostname-override`


ktsakalozos (approved):
LGTM +1

PR #500, opened 2024-06-17
Initial implementation for embedded datastore

neoaggelos:
### Summary

Initial implementation for etcd datastore option.

Requires https://github.com/canonical/k8s-dqlite/pull/105 

### Changes

- Add `pkg/client/etcd` to interact with the embedded cluster
- Add new `etcd` datastore type
- Add new bootstrap configuration options to optionally seed k8s-snap with external certificates for the etcd cluster

### Notes for reviewers

- Start by reviewing the accompanying docs, as they tell most of the story.


docs/src/snap/howto/embedded-datastore.md
```diff
@@ -0,0 +1,138 @@
+# How to use the embedded datastore
+
+Canonical Kubernetes supports using an embedded datastore such as etcd
```
evilnick, line 3: I think this could lead to some level of confusion. It confused me. Mainly because this implies that there may be other options for an 'embedded' datastore (and slightly because it is etcd which may be slightly confusing as thats the example we use for an external datastore too).
From what I can see this means that we are building in our own etcd alongside dqlite and either can be used at deploy-time. It isn't entirely clear what each will be used for. 
So in short, I think we need a much more detailed explanation document which covers the datastore, which we can then refer to from here and keep the explanations where they need to be, freeing this up to be a more straightforward howto

bschimke95 (commented):
Great work (most questions resolved offline).

As a general note: why do we call this `embedded`? At least I found it a bit confusing as it does not really convey that etcd is used. Can we not just call it etcd (at least for user facing configs)? Then we have `datastore-type: k8s-dqlite|etcd|external` which would make sense to me.

docs/src/snap/howto/embedded-datastore.md
ktsakalozos: Indeed embedded seems unclear to end users. dqlite is also positioned as an embedded database so if the user selects embedded as a datastore  does it mean he wants dqlite or etcd? I think the intention here is to let users slect the option of a "managed-etcd". With "managed-etcd" it is clear you do not want "managed-dqlite" or "external-etcd".

Also the word "embedded" leads you to believe that etcd is embedded into k8s whereas in reality we keep it in its own service. The end user probably does not know that the correct term used among engineers doing the implementation of that that service is that they use an embedded etcd.

docs/src/snap/explanation/datastore/etcd.md
```diff
@@ -0,0 +1,78 @@
...
+When using the managed etcd datastore, all nodes participate equally in the
+raft quorum. That means an odd number of **2k + 1** nodes is needed to maintain
+a fault tolerance of **k** nodes (such that the rest **k + 1** nodes maintain
+an active quorum).
+
+## etcd configuration and data directory
+
+The etcd configuration and data directories to be aware of are:
+
+- `/var/snap/k8s/common/var/lib/k8s-dqlite/etcd.yaml`: YAML file with etcd
```
bschimke95, line 69: nit: I think we need to somewhere mention why we have `k8s-dqlite` as the path. 
In the future, we could also think about renaming this folder entirely to something like `datastore`.

docs/src/snap/explanation/datastore/external.md
```diff
@@ -0,0 +1,49 @@
+# external datastore
```
bschimke95, line 1: Here and elsewhere: Is there a guideline/convention to start with capital letters for a headline? @evilnick

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,137 @@
...
+	Hostname          string
+	DNSSANs           []string
+	IPSANs            []net.IP
+	Years             int
+	AllowSelfSignedCA bool
+}
+
+func NewEtcdPKI(opts EtcdPKIOpts) *EtcdPKI {
+	if opts.Years == 0 {
+		opts.Years = 10
```
louiseschmidtgen, line 40: What happens when the self-signed CA expires? Do you have to manually create new ones?

docs/src/snap/explanation/datastore/k8s-dqlite.md
```diff
@@ -0,0 +1,43 @@
+# k8s-dqlite datastore
+
+Canonical Kubernetes supports using a managed dqlite cluster as the underlying
+datastore of the cluster. This is the default option when no configuration is
+specified.
+
+This page explains the behaviour of the managed dqlite cluster. See How-To
+[Configure Canonical Kubernetes with dqlite][how-to-dqlite] for steps to
+deploy Canonical Kubernetes with a managed etcd datastore.
```
bschimke95, line 9: typo: with a managed dqlite datastore

docs/src/snap/howto/datastore/external.md
```diff
@@ -0,0 +1,122 @@
...
+        "type": "external",
+        "servers": "https://10.0.0.11:2379,https://10.0.0.12:2379,https://10.0.0.13:2379",
+        "ca-crt": "------BEGIN CERTIFICATE------\n.....\n-----END CERTIFICATE-----",
+        "client-crt": "------BEGIN CERTIFICATE------\n.....\n-----END CERTIFICATE-----",
+        "client-key": "------BEGIN RSA PRIVATE KEY------\n.....\n-----END RSA PRIVATE KEY-----",
+      }
+    }
+    ```
+
+2.  Apply the new configuration using the k8sd API directly. You must run this
```
bschimke95, line 100: nit: maybe we should have a hidden command for that in the future.

bschimke95 (approved):
Great work, some minor nits then it is good to ship.

PR #501, opened 2024-06-17
Add command to retrieve images used by k8s-snap

neoaggelos:
### Summary

Add a hidden command that can be used to retrieve the list of images a particular build of `k8s-snap` will use.

### Changes

- Add `pkg/k8sd/images` package. This exports two functions: `Registrer` and `Images`
- Individual packages "register" their images by using a `func init() { images.Register(...) }`
- The list-images command simply retrieves the list using `images.Images()`.

### Notes

- This way, individual implementations (e.g. calico or cilium) simply have an `init()` that registers the images. `init()` does not run for packages that are not imported by the code.
- We currently implicitly depend on more images than just the ones in our configs. Added TODOs to resolve them separately.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_x_list_images.go
+++ b/src/k8s/cmd/k8s/k8s_x_list_images.go
@@ -1,0 +1,21 @@
+package k8s
+
+import (
+	"strings"
+
+	cmdutil "github.com/canonical/k8s/cmd/util"
+	"github.com/canonical/k8s/pkg/k8sd/images"
+	"github.com/spf13/cobra"
+)
+
+func newXListImagesCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
+	cmd := &cobra.Command{
+		Hidden: true,
+		Use:    "x-list-images",
+		Short:  "List all images used by the current version of k8s-snap",
+		Run: func(cmd *cobra.Command, args []string) {
+			cmd.PrintErrln(strings.Join(images.Images(), "\n"))
+		},
+	}
+	return cmd
+}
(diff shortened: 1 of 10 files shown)


src/k8s/cmd/k8s/k8s_x_list_images.go
```diff
@@ -0,0 +1,21 @@
...
+
+	cmdutil "github.com/canonical/k8s/cmd/util"
+	"github.com/canonical/k8s/pkg/k8sd/images"
+	"github.com/spf13/cobra"
+)
+
+func newXListImagesCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
+	cmd := &cobra.Command{
+		Hidden: true,
+		Use:    "x-list-images",
```
ktsakalozos, line 14: Why is this called "x-....". Sorry I am not familiar with this concept.

neoaggelos: we can also call this `k8s list-images`. it is a hidden command either way, so it would not matter

src/k8s/pkg/k8sd/images/images.go
```diff
@@ -0,0 +1,22 @@
+package images
+
+import "slices"
+
+var registeredImages []string
+
+// Images returns the list of images registered by individual components.
+func Images() []string {
+	if registeredImages == nil {
+		return nil
```
bschimke95, line 10: Any specific reason why you return nil here instead of an empty list?

bschimke95 (approved):
LGTM

src/k8s/pkg/k8sd/images/images.go
neoaggelos: it is customary to return `nil` instead of an empty list when the result is empty. Go will handle it in the same way (range over nil does not do anything)

PR #502, opened 2024-06-17
Use a containerd plugin for side-loading images

neoaggelos:
### Summary

Add a containerd plugin that sideloads images when containerd starts (and, optionally, at a refresh interval)

The plugin can be configured by a section like this in the `config.toml` file of containerd (or in a separate file):

```toml
[plugins."io.containerd.service.v1.sideload-images"]
  disabled = true 
  sources = ["path1", "path2"]
  namespace = "k8s.io"
```

The default behaviour is to attempt to sideload any `tar` files found in `/var/snap/k8s/common/images` when containerd starts.

### Changes

- Add patch for containerd plugin. This plugin is based on https://github.com/canonical/microk8s/blob/7a788d6d308672b0bde609e777727ded4255c134/build-scripts/components/containerd/patches/default/0001-microk8s-sideload-images-plugin.patch
- Also `regctl` under `tools`, such that we can easily run commands like `regctl image export $image > image.tar`


bschimke95 (approved):
LGTM

PR #507, opened 2024-06-19
Sync CSI images to `ghcr.io`

mateoflorido:
## Overview
Sync CSI images from upstream to the ghcr.io registry.
## Rationale
We have noticed that we are using CSI images from a legacy repository. This pull request mirrors the images we are using from the upstream repository, allowing the k8s snap to pull them from there instead of the legacy Kubernetes repository.

The Helm chart for rawfile was updated to allow configuring the CSI images.


addyess (approved):
Nice job @mateoflorido

src/k8s/pkg/k8sd/features/localpv/chart.go
```diff
@@ -11,11 +11,21 @@ var (
...
 		Namespace:    "kube-system",
-		ManifestPath: path.Join("charts", "rawfile-csi-0.8.0.tgz"),
+		ManifestPath: path.Join("charts", "rawfile-csi-0.9.0.tgz"),
 	}
 
-	// imageRepo is the image to use for Rawfile LocalPV CSI.
+	// imageRepo is the repository to use for Rawfile LocalPV CSI.
 	imageRepo = "ghcr.io/canonical/rawfile-localpv"
+	// csiImageRepo is the repository to use for the CSI images.
+	csiImageRepo = "ghcr.io/canonical/k8s-snap/sig-storage"
```
neoaggelos, line 20: its ok to not have this as a separate arg, so that individual mages can be changed easily without much trouble

neoaggelos (approved):

PR #509, opened 2024-06-20
Add CAPI node-remove endpoint

bschimke95:
## Summary

Add a new endpoint `x/capi/remove-node` that is protected by the CAPI access handler.

## Rationale

This is a prerequisite for the removal of machines from a CAPI workload cluster. Similar to the `x/capi/set-auth-token` we cannot simply reuse the existing removal endpoint (`k8sd/cluster/remove`) as the CAPI management cluster lacks a proper auth mechanism for this.
See #498 for more on this.


berkayoz (approved):
LGTM

PR #514, opened 2024-06-26
Switch default encapsulation to VXLAN, add annotations for autodetection

berkayoz:
* Changed default encapsulation to VXLAN for both IPv4 and IPv6 pools
* Disabled Calico api server deployment by default
* Added annotations for node address auto detection

### DIFF
--- a/src/k8s/pkg/k8sd/features/calico/chart.go
+++ b/src/k8s/pkg/k8sd/features/calico/chart.go
@@ -36,9 +36,9 @@ var (
 	// calicoCtlTag represents the tag to use for the calicoctl image.
 	calicoCtlTag = "v3.28.0"
 
-	// encapsulation represents the default encapsulation method to use for Calico.
-	encapsulation = "VXLAN"
+	// defaultEncapsulation represents the default defaultEncapsulation method to use for Calico.
+	defaultEncapsulation = "VXLAN"
 
-	// apiServerEnabled determines if the Calico API server should be enabled.
-	apiServerEnabled = false
+	// defaultApiServerEnabled determines if the Calico API server should be enabled.
+	defaultApiServerEnabled = false
 )
--- a/src/k8s/pkg/k8sd/features/calico/chart.go
+++ b/src/k8s/pkg/k8sd/features/calico/chart.go
@@ -39,6 +39,6 @@ var (
 	// defaultEncapsulation represents the default defaultEncapsulation method to use for Calico.
 	defaultEncapsulation = "VXLAN"
 
-	// defaultApiServerEnabled determines if the Calico API server should be enabled.
-	defaultApiServerEnabled = false
+	// defaultAPIServerEnabled determines if the Calico API server should be enabled.
+	defaultAPIServerEnabled = false
 )
--- a/src/k8s/pkg/k8sd/features/calico/chart.go
+++ b/src/k8s/pkg/k8sd/features/calico/chart.go
@@ -35,4 +35,10 @@ var (
 	calicoCtlImage = "ghcr.io/canonical/k8s-snap/calico/ctl"
 	// calicoCtlTag represents the tag to use for the calicoctl image.
 	calicoCtlTag = "v3.28.0"
+
+	// encapsulation represents the default encapsulation method to use for Calico.
+	encapsulation = "VXLAN"
+
+	// apiServerEnabled determines if the Calico API server should be enabled.
+	apiServerEnabled = false
 )
--- a/src/k8s/pkg/k8sd/features/calico/cleanup.go
+++ b/src/k8s/pkg/k8sd/features/calico/cleanup.go
@@ -20,15 +20,18 @@ func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
 		return fmt.Errorf("failed to list network interfaces: %w", err)
 	}
 
+	// Compile the regular expression outside the loop
+	regex, err := regexp.Compile("^vxlan[-v6]*.calico|cali[a-f0-9]*|tunl[0-9]*$")
+	if err != nil {
+		return fmt.Errorf("failed to compile regex pattern: %w", err)
+	}
+
 	// Find the interfaces created by Calico
 	for _, iface := range interfaces {
 		// Check if the interface name matches the regex pattern
 		// Adapted from MicroK8s' link removal hook:
 		// https://github.com/canonical/microk8s/blob/dff3627959d4774198000795a0a0afcaa003324b/microk8s-resources/default-hooks/remove.d/10-cni-link#L15
-		match, err := regexp.MatchString("^vxlan[-v6]*.calico|cali[a-f0-9]*|tunl[0-9]*$", iface.Name)
-		if err != nil {
-			return fmt.Errorf("failed to match regex pattern: %w", err)
-		}
+		match := regex.MatchString(iface.Name)
 		if match {
 			// Perform cleanup for Calico interface
 			if err := exec.CommandContext(ctx, "ip", "link", "delete", iface.Name).Run(); err != nil {
--- a/src/k8s/pkg/k8sd/features/calico/cleanup.go
+++ b/src/k8s/pkg/k8sd/features/calico/cleanup.go
@@ -14,16 +14,12 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+var reDefaultCalicoInterface = regexp.MustCompile("^vxlan[-v6]*.calico|cali[a-f0-9]*|tunl[0-9]*$")
+
 func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
 	interfaces, err := net.Interfaces()
 	if err != nil {
 		return fmt.Errorf("failed to list network interfaces: %w", err)
-	}
-
-	// Compile the regular expression outside the loop
-	regex, err := regexp.Compile("^vxlan[-v6]*.calico|cali[a-f0-9]*|tunl[0-9]*$")
-	if err != nil {
-		return fmt.Errorf("failed to compile regex pattern: %w", err)
 	}
 
 	// Find the interfaces created by Calico
@@ -31,7 +27,7 @@ func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
 		// Check if the interface name matches the regex pattern
 		// Adapted from MicroK8s' link removal hook:
 		// https://github.com/canonical/microk8s/blob/dff3627959d4774198000795a0a0afcaa003324b/microk8s-resources/default-hooks/remove.d/10-cni-link#L15
-		match := regex.MatchString(iface.Name)
+		match := reDefaultCalicoInterface.MatchString(iface.Name)
 		if match {
 			// Perform cleanup for Calico interface
 			if err := exec.CommandContext(ctx, "ip", "link", "delete", iface.Name).Run(); err != nil {
--- a/src/k8s/pkg/k8sd/features/calico/internal.go
+++ b/src/k8s/pkg/k8sd/features/calico/internal.go
@@ -41,136 +41,100 @@ func checkEncapsulation(v string) error {
 	return fmt.Errorf("unsupported encapsulation type: %s", v)
 }
 
+func parseAutodetectionAnnotations(annotations types.Annotations, autodetectionMap map[string]string) (map[string]any, error) {
+	var autodetectionAnnotations []string
+	var autodetectionKey string
+	var autodetectionValue any
+
+	for annotation, key := range autodetectionMap {
+		if v, ok := annotations.Get(annotation); ok {
+			autodetectionAnnotations = append(autodetectionAnnotations, annotation)
+			autodetectionKey = key
+			autodetectionValue = v
+		}
+	}
+
+	if len(autodetectionAnnotations) > 1 {
+		return nil, fmt.Errorf("multiple annotations found: %s", strings.Join(autodetectionAnnotations, ", "))
+	}
+
+	// If any annotation is set, return the map otherwise it's left nil
+	if autodetectionKey != "" {
+		switch autodetectionKey {
+		case "firstFound":
+			autodetectionValue = autodetectionValue == "true"
+		case "cidrs":
+			autodetectionValue = strings.Split(autodetectionValue.(string), ",")
+		}
+
+		return map[string]any{
+			autodetectionKey: autodetectionValue,
+		}, nil
+	}
+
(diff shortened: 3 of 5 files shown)


src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,172 @@
...
+
+type config struct {
+	encapsulationV4  string
+	encapsulationV6  string
+	apiServerEnabled bool
+	autodetectionV4  map[string]any
+	autodetectionV6  map[string]any
+}
+
+func validateEncapsulation(encapsulation string) bool {
```
neoaggelos, line 36: `func checkEncapsulation(v string) error`

and return an error message if the value is not valid, e.g. `fmt.Errorf("unsupported encapsulation type %q", v)`

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	if v, ok := annotations.Get(annotationAPIServerEnabled); ok {
+		config.apiServerEnabled = v == "true"
+	}
+
+	if v, ok := annotations.Get(annotationEncapsulationV4); ok {
+		if !validateEncapsulation(v) {
+			return config, fmt.Errorf("invalid encapsulation-v4 annotation: %s", v)
+		}
```
neoaggelos, line 62: get an error (see previous comment) and wrap it here

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,172 @@
...
+		config.encapsulationV4 = v
+	}
+
+	if v, ok := annotations.Get(annotationEncapsulationV6); ok {
+		if !validateEncapsulation(v) {
```
neoaggelos, line 67: same here

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,172 @@
...
+	annotationAutodetectionV6CanReach      = "k8sd/v1alpha1/calico/autodetection-v6/canReach"
+	annotationAutodetectionV6Cidrs         = "k8sd/v1alpha1/calico/autodetection-v6/cidrs"
+)
+
+type config struct {
+	encapsulationV4  string
+	encapsulationV6  string
+	apiServerEnabled bool
+	autodetectionV4  map[string]any
+	autodetectionV6  map[string]any
```
neoaggelos, line 33: can we make this a struct with json tags and `omitempty`?

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,172 @@
...
+		if !validateEncapsulation(v) {
+			return config, fmt.Errorf("invalid encapsulation-v6 annotation: %s", v)
+		}
+		config.encapsulationV6 = v
+	}
+
+	if v, ok := annotations.Get(annotationAutodetectionV4Firstfound); ok {
+		if config.autodetectionV4 != nil {
+			return config, fmt.Errorf("multiple autodetection-v4 annotations found")
+		}
```
neoaggelos, line 76: then this becomes easier. also, calico does not seem to have trouble with multiple autodetection methods, should we allow that as well?

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,172 @@
...
+		config.autodetectionV4 = map[string]any{
+			"firstFound": v == "true",
+		}
+	}
+	if v, ok := annotations.Get(annotationAutodetectionV4Kubernetes); ok {
+		if config.autodetectionV4 != nil {
+			return config, fmt.Errorf("multiple autodetection-v4 annotations found")
```
neoaggelos, line 83: for the error messages here, include the annotation key itself, such that it's easier to track down the error (in case there is one)

neoaggelos (commented):
Looks great, style comments and some possible improvements

src/k8s/pkg/k8sd/features/calico/chart.go
```diff
@@ -35,4 +35,10 @@ var (
 	calicoCtlImage = "ghcr.io/canonical/k8s-snap/calico/ctl"
 	// calicoCtlTag represents the tag to use for the calicoctl image.
 	calicoCtlTag = "v3.28.0"
+
+	// encapsulation represents the default encapsulation method to use for Calico.
+	encapsulation = "VXLAN"
```
neoaggelos, line 40: just so it reads better in places in the code, let's call this `defaultEncapsulation`

same below, `defaultAPIServerEnabled`

src/k8s/pkg/k8sd/features/calico/network.go
```diff
@@ -65,10 +72,15 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, _ type
 		},
 		"installation": map[string]any{
 			"calicoNetwork": map[string]any{
-				"ipPools": podIpPools,
+				"ipPools":                    podIpPools,
+				"nodeAddressAutodetectionV4": config.autodetectionV4,
+				"nodeAddressAutodetectionV6": config.autodetectionV6,
```
neoaggelos, line 77: careful as helm values still need to be `map[string]any` at the end of the day. perhaps `mapstructure` can help us here.

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,177 @@
...
+	CanReach           string   `json:"canReach,omitempty"`
+	CIDRs              []string `json:"cidrs,omitempty"`
+}
+
+type config struct {
+	encapsulationV4  string
+	encapsulationV6  string
+	apiServerEnabled bool
+	autodetectionV4  *autodetection
+	autodetectionV6  *autodetection
```
neoaggelos, line 42: okay, these does multiple nil checks and multiple allocations at the end of the day.

let's instead go for `autodetectionV4Key, autodetectionV4Value string`

then, below, we check for an empty string as needed.

src/k8s/pkg/k8sd/features/calico/network.go
neoaggelos, line 77: then here, we pass nil (if nothing is set) or a map based on the key/value of the autodetection config

neoaggelos (changes requested):
Final comment, let's avoid an extra step of structs, since helm values need to be `map[string]any` at the end of the day

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,176 @@
...
+		autodetectionV4Key = "canReach"
+		autodetectionV4Value = v
+	}
+	if v, ok := annotations.Get(annotationAutodetectionV4Cidrs); ok {
+		if autodetectionV4Key != "" {
+			return config, fmt.Errorf("multiple autodetection-v4 annotations found: %s", annotationAutodetectionV4Cidrs)
+		}
+		autodetectionV4Key = "cidrs"
+		autodetectionV4Value = strings.Split(v, ",")
+	}
```
neoaggelos, line 113: let's loop over a `map[string]string{annotationAutodetectionV4CanReach: "cidrs", ...}`?

src/k8s/pkg/k8sd/features/calico/chart.go
```diff
@@ -35,4 +35,10 @@ var (
 	calicoCtlImage = "ghcr.io/canonical/k8s-snap/calico/ctl"
 	// calicoCtlTag represents the tag to use for the calicoctl image.
 	calicoCtlTag = "v3.28.0"
+
+	// defaultEncapsulation represents the default defaultEncapsulation method to use for Calico.
+	defaultEncapsulation = "VXLAN"
+
+	// defaultApiServerEnabled determines if the Calico API server should be enabled.
+	defaultApiServerEnabled = false
```
neoaggelos, line 43: `defaultAPIServerEnabled`

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,176 @@
...
+const (
+	annotationAPIServerEnabled             = "k8sd/v1alpha1/calico/apiserver-enabled"
+	annotationEncapsulationV4              = "k8sd/v1alpha1/calico/encapsulation-v4"
+	annotationEncapsulationV6              = "k8sd/v1alpha1/calico/encapsulation-v6"
+	annotationAutodetectionV4Firstfound    = "k8sd/v1alpha1/calico/autodetection-v4/firstFound"
+	annotationAutodetectionV4Kubernetes    = "k8sd/v1alpha1/calico/autodetection-v4/kubernetes"
+	annotationAutodetectionV4Interface     = "k8sd/v1alpha1/calico/autodetection-v4/interface"
+	annotationAutodetectionV4SkipInterface = "k8sd/v1alpha1/calico/autodetection-v4/skipInterface"
+	annotationAutodetectionV4CanReach      = "k8sd/v1alpha1/calico/autodetection-v4/canReach"
+	annotationAutodetectionV4Cidrs         = "k8sd/v1alpha1/calico/autodetection-v4/cidrs"
```
neoaggelos, line 19: `annotationAutodetectionV4CIDRs`, also elsewhere

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,176 @@
...
+	"strings"
+
+	"github.com/canonical/k8s/pkg/k8sd/types"
+)
+
+const (
+	annotationAPIServerEnabled             = "k8sd/v1alpha1/calico/apiserver-enabled"
+	annotationEncapsulationV4              = "k8sd/v1alpha1/calico/encapsulation-v4"
+	annotationEncapsulationV6              = "k8sd/v1alpha1/calico/encapsulation-v6"
+	annotationAutodetectionV4Firstfound    = "k8sd/v1alpha1/calico/autodetection-v4/firstFound"
```
neoaggelos, line 14: `annotationAutodetectionV4FirstFound`, also elsewhere

src/k8s/pkg/k8sd/features/calico/network.go
neoaggelos, line 77: i'm worried if passing `nodeAddressAutodetectionV4: null` in the values afffects or unsets any defaults that may come from the chart, breaking deployments. can you make sure that this is handled?

neoaggelos (commented):
Looks great, mostly style comments

src/k8s/pkg/k8sd/features/calico/network.go
neoaggelos: I would make the `config` carry the `autodetectionV4{Key,Value}`, then set these fields only if both are set.

src/k8s/pkg/k8sd/features/calico/internal.go
berkayoz: I have also wanted to do this, I'm not sure how to handle `cidrs` and `firstFound` options as they expect values other than string.

src/k8s/pkg/k8sd/features/calico/cleanup.go
```diff
@@ -20,15 +20,18 @@ func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
 		return fmt.Errorf("failed to list network interfaces: %w", err)
 	}
 
+	// Compile the regular expression outside the loop
+	regex, err := regexp.Compile("^vxlan[-v6]*.calico|cali[a-f0-9]*|tunl[0-9]*$")
```
neoaggelos, line 24: use a `defaultCalicoInterfaceRe = regexp.MustCompile(....)`

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,176 @@
...
+			autodetectionV4Key: autodetectionV4Value,
+		}
+	}
+
+	var autodetectionV6Key string
+	var autodetectionV6Value any
+
+	if v, ok := annotations.Get(annotationAutodetectionV6FirstFound); ok {
+		if autodetectionV6Key != "" {
+			return config, fmt.Errorf("multiple autodetection-v6 annotations found: %s", annotationAutodetectionV6FirstFound)
```
neoaggelos, line 127: return zero values in case of errors

src/k8s/pkg/k8sd/features/calico/internal.go
```diff
@@ -0,0 +1,176 @@
...
+	if v, ok := annotations.Get(annotationAutodetectionV6Interface); ok {
+		if autodetectionV6Key != "" {
+			return config, fmt.Errorf("multiple autodetection-v6 annotations found: %s", annotationAutodetectionV6Interface)
+		}
+		autodetectionV6Key = "interface"
+		autodetectionV6Value = v
+	}
+	if v, ok := annotations.Get(annotationAutodetectionV6SkipInterface); ok {
+		if autodetectionV6Key != "" {
+			return config, fmt.Errorf("multiple autodetection-v6 annotations found: %s", annotationAutodetectionV6SkipInterface)
```
neoaggelos, line 148: this error is not _too_ useful, as it does not say what other annotations are set (therefore is not easy to take action).

the autodetectionValue can be a `any`

src/k8s/pkg/k8sd/features/calico/internal.go
neoaggelos: the value can be a `any`, so it would hold a string, a bool, or a []string just fine.

src/k8s/pkg/k8sd/features/calico/network.go
```diff
@@ -69,9 +76,20 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, _ type
...
 			"registry": imageRepo,
 		},
+		"apiServer": map[string]any{
+			"enabled": config.apiServerEnabled,
+		},
 		"serviceCIDRs": serviceCIDRs,
 	}
 
+	if config.autodetectionV4 != nil {
+		values["installation"].(map[string]any)["calicoNetwork"].(map[string]any)["nodeAddressAutodetectionV4"] = config.autodetectionV4
```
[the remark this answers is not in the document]
neoaggelos: with the final values being
```
values = map[string]any{
  "installation": map[string]any{
     "calicoNetwork": calicoNetworkValues,
   },
}
```

neoaggelos (approved):

PR #515, opened 2024-06-26
Prepare LXD images for e2e tests to improve runtimes

neoaggelos:
### Summary

Use a pre-built image and fetched OCI images for integration tests, with the aim to reduce e2e test running times and reduce the amount of fetched images.


bschimke95 (approved):
LGTM

PR #516, opened 2024-06-27
Sync Contour images

louiseschmidtgen:
## Description
Sync contour images to `ghcr.io/canonical`.
```
docker.io/envoyproxy/envoy:v1.29.2
ghcr.io/projectcontour/contour:v1.28.2
docker.io/bitnami/contour:1.28.2-debian-12-r4
docker.io/bitnami/envoy:1.28.2-debian-12-r0
```

Manipulated the pulled gateway provisioner manifests to be able to set the images via helm values:
```
    spec:
      containers:
      - args:
        - gateway-provisioner
        - --metrics-addr=127.0.0.1:8080
        - --enable-leader-election
        - --envoy-image={{ .Values.envoyproxy.image.repository }}:{{ .Values.envoyproxy.image.tag }}
        - --contour-image={{ .Values.projectcontour.image.repository }}:{{ .Values.projectcontour.image.tag }}
        command: ["contour"]
        image: "{{ .Values.projectcontour.image.repository }}:{{ .Values.projectcontour.image.tag }}"
        imagePullPolicy: IfNotPresent
        name: contour-gateway-provisioner
```
Note: Gateway provisioner uses envoyproxy image v1.29.2 instead of v1.28.2 in their release for gateway provisioner v1.28.2.

Sidenote: Gateway provisioner allows you to change the images used by setting these args [here](https://github.com/projectcontour/contour/blob/main/cmd/contour/gatewayprovisioner.go#L39C1-L39C63).


build-scripts/hack/update-contour-components.sh
```diff
@@ -41,6 +41,27 @@ cp contour-src/examples/gateway-provisioner/01-roles.yaml ck-gateway-contour/tem
...
+
+# Add image args to the gateway provisioner 
+sed -i '/^        - --enable-leader-election$/a\ \ \ \ \ \ \ \ - --envoy-image={{ .Values.envoyproxy.image.registry }}/{{ .Values.envoyproxy.image.repository }}:{{ .Values.envoyproxy.image.tag }}\n\ \ \ \ \ \ \ \ - --contour-image={{ .Values.projectcontour.image.registry }}/{{ .Values.projectcontour.image.repository }}:{{ .Values.projectcontour.image.tag }}' ck-gateway-contour/templates/03-gateway-provisioner.yaml
+
+# Add values.yaml
+cat <<EOF >ck-gateway-contour/values.yaml
+projectcontour:
+  image:
+    registry: ghcr.io
+    repository: projectcontour/contour
```
neoaggelos, line 55: this is unnecessary, just make it `repository` (merge registry/repository) and `tag`. it simplifies your configs as well.

neoaggelos (commented):
Remember to update the images that contour reports.

src/k8s/pkg/k8sd/features/contour/chart.go
```diff
@@ -33,20 +33,32 @@ var (
 		Namespace:    "projectcontour",
 		ManifestPath: path.Join("charts", "ck-contour-common-1.28.2.tgz"),
 	}
+
+	// registry represents the registry (ghcr.io/canonical) to use for Contour images.
+	registry = "ghcr.io"
```
neoaggelos, line 38: not a huge fan of having this here, then have it be hardcoded in some places and not others. when looking at a `$fooImageRepo` in this package, it is not clear if it should have the `ghcr.io` prefix or not.

neoaggelos (commented):
We are also missing the updated list of images in `register.go`

src/k8s/pkg/k8sd/features/contour/register.go
```diff
@@ -11,5 +11,6 @@ func init() {
 		fmt.Sprintf("%s:%s", envoyImageRepo, envoyImageTag),
```
neoaggelos, line 11: envoyImageRepo is now just `canonical/k8s-snap/bitnami/envoy`, which is not a valid image.

src/k8s/pkg/k8sd/features/contour/ingress.go
```diff
@@ -40,13 +40,25 @@ func ApplyIngress(ctx context.Context, snap snap.Snap, ingress types.Ingress, _
 	values = map[string]any{
 		"envoy-service-namespace": "projectcontour",
 		"envoy-service-name":      "envoy",
+		"envoy": map[string]any{
+			"image": map[string]any{
+				"registry":   registry,
```
neoaggelos, line 45: can you check if we can avoid setting `registry`, then we can just have `envoyImageRepo` be the full image ?

neoaggelos (changes requested):

src/k8s/pkg/k8sd/features/contour/ingress.go
louiseschmidtgen: Hm... Not sure if there is a nice way here.
Registry has a default value. We could try pass "" to registry and pack it in the repository but that feels a little wrong?
```
# values.yaml
  image:
    registry: docker.io
    repository: bitnami/contour
    tag: 1.28.2-debian-12-r4
```
Alternatively, there is a global image registry we could set:
```
global:
  imageRegistry: ""
```

neoaggelos: My suggestion is to try to see if it works. If it does not, then take the registry out of all images, and adjust the code accordingly

louiseschmidtgen: Alright, setting this works:
```
  image:
    registry: ""
    repository: ghcr.io/canonical/k8s-snap/bitnami/contour
    tag: 1.28.2-debian-12-r4
    ```

File: src/k8s/pkg/k8sd/features/contour/chart.go [hunk 5]
```diff
@@ -33,20 +33,29 @@ var (
...
+	// to follow the upstream configuration for the contour gateway provisioner.
+	// envoyProxyImageTag is the tag to use for for envoy in the gateway.
+	envoyProxyImageTag = "v1.29.2"
+
 	// envoyImageRepo represents the image to use for the Contour Envoy proxy.
-	envoyImageRepo = "docker.io/bitnami/envoy"
+	envoyImageRepo = "ghcr.io/canonical/k8s-snap/bitnami/envoy"
 
 	// envoyImageTag is the tag to use for the Contour Envoy proxy image.
 	envoyImageTag = "1.28.2-debian-12-r0"
```
neoaggelos, line 49: It is quite easy to get lost with this, since these basically refer to the same image, without any distiction between them.

When updating the chart in the future, how do we know that we have not missed or mixed these images?

Use a specific name for each, in order to make it harder to miss, e.g. `contourGatewayProvisionerEnvoyImage....`, `contourIngressEnvoyImage...`

neoaggelos (approved):
Almost there!

PR #518, opened 2024-06-28
Rework microcluster k8sd.Client package

neoaggelos:
### Summary

- Move `pkg/k8s/client.Client` to `pkg/client/k8sd.Client`
- Add a `snap.Snap.K8sdClient(address string)` method (empty address for local clients, non-empty for remote. currently unused)

### Changes

- Drop `k8s generate-auth-token` and `k8s revoke-auth-token` hidden commands, since we do not use tokens any more
- Define `k8sd.Client` interface, which only focuses on interacting with a k8sd API. It does not include any logic that requires the local node

### Differences between `client.Client` -> `k8sd.Client`

- There is no `IsBootstrapped` method, clients should use `NodeStatus()` instead and check for errors


bschimke95 (approved):
LGTM

PR #519, opened 2024-06-28
Add SR-IOV CNI Helm Chart

HomayoonAlimohammadi:
# Summary
Add SR-IOV CNI Helm Chart

# How to test
```shell
cd k8s/manifests/charts/sr-iov
helm template --dry-run .
```
and compare the results with [reference](https://github.com/k8snetworkplumbingwg/sriov-cni/blob/master/images/sriov-cni-daemonset.yaml)

### DIFF
--- a/k8s/manifests/charts/sr-iov/values.yaml
+++ b/k8s/manifests/charts/sr-iov/values.yaml
@@ -1,0 +1,24 @@
+# Default values for sr-iov.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+daemonset:
+  name: sriov-cni
+  fullName: kube-sriov-cni-ds
+  containerName: kube-sriov-cni
+  namespace: kube-system
+  labels:
+    tier: node
+  image: 
+    repository: ghcr.io/k8snetworkplumbingwg/sriov-cni
+    tag: latest
+    pullPolicy: IfNotPresent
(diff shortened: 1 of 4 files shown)


k8s/manifests/charts/sr-iov/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+    repository: ghcr.io/k8snetworkplumbingwg/sriov-cni
+    tag: latest
+    pullPolicy: IfNotPresent
+  resources:
+    requests:
+      cpu: "100m"
+      memory: "50Mi"
+    limits:
+      cpu: "100mu"
+      memory: "50Mi"
```
eaudetcobello, line 22: I'd propose this chart doesn't *need* a values.yaml. All these values can be hardcoded in the daemonset template. These are not values that we'd be changing dynamically.

HomayoonAlimohammadi: that's a valid statement imo. I was thinking the same while creating the chart. what do you think @neoaggelos ?

HomayoonAlimohammadi: IIn fact, I really agree with you @eaudetcobello . I will remove the .values file and push the new changes to https://github.com/canonical/sriov-cni-charts/pull/1

PR #520, opened 2024-06-30
Microcluster timeouts during `k8s bootstrap` and `k8s join-cluster`

neoaggelos:
### Summary

This is a pre-requisite for fixing how k8sd cleans up when a bootstrap or join cluster operation fails.

Merge after #518 

### Changes

Handle context cancellations during the bootstrap and join hooks. Without this change:

1. microcluster always applies a 30-second timeout if the context has no deadline set, effectively ignoring the timeout passed by the client
2. context deadline exceeded now cancels the bootstrap or join hook. Previously, this was using `s.Context` (from microcluster daemon), and therefore the hooks would keep running.

### DIFF
--- a/src/k8s/api/v1/cluster.go
+++ b/src/k8s/api/v1/cluster.go
@@ -1,5 +1,7 @@
 package apiv1
 
+import "time"
+
 // GetClusterStatusRequest is used to request the current status of the cluster.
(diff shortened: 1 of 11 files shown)


bschimke95 (approved):
LGTM

PR #521, opened 2024-06-30
Cleanup on failed `k8s bootstrap` an `k8s join-cluster` attempts

neoaggelos:
### Summary

Merge after #520 

### Changes

- Bootstrap control plane, bootstrap worker and join control plane hooks are refactored. We always defer a function that checks the result of the hook.
- In the case of preRemove, we simply log the error and proceed, otherwise the node is removed by microcluster peers but not the underlying dqlite database, breaking the cluster
- In the case of `k8s bootstrap`, remove all configs, stop control plane services, then use `ResetClusterMember`. This resets the microcluster state. Note that this runs automatically by k8sd, no manual action from the client is required.
- In the case of `k8s join-cluster` for worker nodes, similarly revert configs, then use `ResetClusterMember`
- In the case of `k8s join-cluster` for control plane nodes: When the `postJoin` hook runs, the node has already joined microcluster. Therefore, we need to revert configs, but also make sure to use `DeleteClusterMember`, such that the failed node is removed from the cluster before resetting.

### Notes

- The wait for a node to be not Pending before removing has been moved to the `k8s remove-node` command, instead of delaying the completion of the `k8s join-cluster` command.

### DIFF
--- a/src/k8s/pkg/client/k8sd/cluster.go
+++ b/src/k8s/pkg/client/k8sd/cluster.go
@@ -45,6 +45,11 @@ func (c *k8sd) JoinCluster(ctx context.Context, request apiv1.JoinClusterRequest
 }
 
 func (c *k8sd) RemoveNode(ctx context.Context, request apiv1.RemoveNodeRequest) error {
+	// NOTE(neoaggelos): microcluster adds an arbitrary 30 second timeout in case no context deadline is set.
+	// Configure a client deadline for timeout + 30 seconds (the timeout will come from the server)
+	ctx, cancel := context.WithTimeout(ctx, request.Timeout+30*time.Second)
+	defer cancel()
+
 	if err := c.client.Query(ctx, "POST", api.NewURL().Path("k8sd", "cluster", "remove"), request, nil); err != nil {
 		return fmt.Errorf("failed to POST /k8sd/cluster/remove: %w", err)
 	}
--- a/src/k8s/pkg/client/k8sd/cluster.go
+++ b/src/k8s/pkg/client/k8sd/cluster.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	apiv1 "github.com/canonical/k8s/api/v1"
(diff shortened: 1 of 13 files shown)


src/k8s/pkg/client/k8sd/cluster.go
```diff
@@ -42,23 +41,15 @@ func (c *k8sd) JoinCluster(ctx context.Context, request apiv1.JoinClusterRequest
...
-	})
-
 	return nil
 }
 
 func (c *k8sd) RemoveNode(ctx context.Context, request apiv1.RemoveNodeRequest) error {
+	// NOTE(neoaggelos): microcluster adds an arbitrary 30 second timeout in case no context deadline is set.
+	// Configure a client deadline for timeout + 30 seconds (the timeout will come from the server)
+	ctx, cancel := context.WithTimeout(ctx, request.Timeout+30*time.Second)
+	defer cancel()
```
bschimke95, line 51: nit: this could be part of #520

src/k8s/pkg/client/kubernetes/node.go
```diff
@@ -4,16 +4,17 @@ import (
...
 	"fmt"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
 )
 
 // DeleteNode will remove a node from the kubernetes cluster.
 // DeleteNode will retry if there is a conflict on the resource.
+// DeleteNode will not fail if the node does not
```
bschimke95, line 14: nit: comment incomplete

bschimke95 (approved):
LGTM

src/k8s/pkg/client/kubernetes/node.go
neoaggelos: found a couple more things like this, will round it with a separate fix typos PR, to avoid unnecessary conflicts across #518, #520, #521

PR #530, opened 2024-07-04
Fix markdownlinter: _ not -

evilnick:
It was a typo in the config file :) @bschimke95 @louiseschmidtgen


bschimke95 (approved):
Ah a classic :)

PR #533, opened 2024-07-08
Remove extra bookeeping for worker nodes

neoaggelos:
### Summary

Remove extraneous worker nodes table.

### Changes

This table is used as bookkeeping for cluster worker nodes, but only causes issues right now:

- If a worker node has already left the cluster, it is impossible to remove it, as the `client.DeleteNode()` will fail.
- If a worker node fails to join, then it still gets an entry on the worker_nodes table, therefore is unable to join with that name (with a not useful error message that the node already exists).

Also, we begin marking nodes with `k8sd.io/role=control-plane` or `k8sd.io/role=worker`, such that we know these nodes are managed by k8sd


bschimke95 (approved):
LGTM, so in essence you moved the "bookkeeping" from k8sd to the workers themselves, right?

PR #538, opened 2024-07-10
Proposal 001: Expose feature status in `k8s status` command

neoaggelos:
### Summary

Add formal feature proposal template, and proposal for improving the output of `k8s status` command.

### DIFF
--- a/docs/proposals/001-k8s-status.md
+++ b/docs/proposals/001-k8s-status.md
@@ -135,7 +135,6 @@ more fields are shown:
 
 ```bash
 $ k8s status
-ready: true
 status: ready
 high-availability: no
 datastore:
--- a/docs/proposals/001-k8s-status.md
+++ b/docs/proposals/001-k8s-status.md
@@ -54,7 +54,7 @@ Instead, we want the `k8s status` command to return meaningful information about
 the current _deployment status_ of each feature. We want to replace the
 following output (taking the `dns` feature as an example):
 
-```bash
+```
 $ k8s status
 ...
 dns:
@@ -65,7 +65,7 @@ dns:
 
 With something like this:
 
-```bash
+```
 $ k8s status
 ...
 dns:
@@ -87,7 +87,7 @@ The output format of `k8s status` will change.
 
 #### Old output
 
-```bash
+```
 $ k8s status
 status: ready
 high-availability: no
@@ -133,7 +133,7 @@ gateway:
 With `--output-format=plain`, a simple overview is shown, with `yaml` or `json`
 more fields are shown:
 
-```bash
+```
 $ k8s status
 status: ready
 high-availability: no
@@ -160,7 +160,7 @@ gateway:
 
 When the output format is `json` or `yaml`, more fields may be shown:
 
-```bash
+```
 $ k8s status --output-format=yaml
 ...
 network:
--- a/docs/proposals/001-k8s-status.md
+++ b/docs/proposals/001-k8s-status.md
@@ -135,27 +135,16 @@ more fields are shown:
 
 ```
 $ k8s status
-status: ready
-high-availability: no
-datastore:
-  type: k8s-dqlite
-  voter-nodes:
-    - 10.87.24.119:6400
-  standby-nodes: none
-  spare-nodes: none
-network:
-  enabled: true
-dns:
-  enabled: true
-ingress:
-  enabled: false
-load-balancer:
-  enabled: false
-local-storage:
-  enabled: true
-gateway:
-  enabled: false
-  message: Cilium GatewayAPI support could not be enabled, error was "..."
+cluster status:         ready/not ready
+control plane nodes:    10.0.0.10 (voter), 10.0.0.20 (voter), 10.0.0.30 (standby)
+high availability:      yes/no
+datastore:              k8s-dqlite running/k8s-dqlite not running/external
+network:                enabled/disabled
+dns:                    enabled/disabled at 10.152.183.10
+ingress:                enabled/disabled
+load-balancer:          enabled/disabled, L2/BGP mode
+local-storage:          enabled/disabled at /var/snap/k8s/common/rawfile-storage
+gateway:                enabled/disabled
 ```
 
 When the output format is `json` or `yaml`, more fields may be shown:
@@ -165,7 +154,7 @@ $ k8s status --output-format=yaml
 ...
 network:
   enabled: true
-  message: Cilium is deployed
+  message: enabled
   timestamp: 2020-01-01 10:00:00
   version: v1.15.2
 ...
--- a/docs/proposals/001-k8s-status.md
+++ b/docs/proposals/001-k8s-status.md
@@ -145,19 +145,14 @@ datastore:
   spare-nodes: none
 network:
   enabled: true
-  message: Cilium is deployed
 dns:
   enabled: true
-  message: CoreDNS is deployed
 ingress:
   enabled: false
-  message: ""
 load-balancer:
-  enabled: true
-  message: Cilium configured for LoadBalancer services
+  enabled: false
 local-storage:
   enabled: true
-  message: Rawfile LocalPV is deployed
 gateway:
   enabled: false
   message: Cilium GatewayAPI support could not be enabled, error was "..."
@@ -202,6 +197,9 @@ The `message` field is the only information that will be presented to the user.
 We should be explicit that this field is only meant to be informative, users
 should not try to programmatically parse it to extract meaningful information
 about the feature.
+
+In case things are working as expected, the message could be empty (such that
+the output of `k8s status` remains concise).
 
 It is out-of-scope for this proposal to provide any sort of _structured_
 fields into the output of `k8s status` command.
--- a/docs/proposals/001-k8s-status.md
+++ b/docs/proposals/001-k8s-status.md
@@ -5,7 +5,7 @@
 - **Index**: 001
 
 <!-- Status -->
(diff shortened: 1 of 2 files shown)


docs/proposals/001-k8s-status.md
```diff
@@ -0,0 +1,433 @@
...
+
+When the output format is `json` or `yaml`, more fields may be shown:
+
+```bash
+$ k8s status --output-format=yaml
+...
+network:
+  enabled: true
+  message: Cilium is deployed
+  timestamp: 2020-01-01 10:00:00
```
louiseschmidtgen, line 175: Why do we need the timestamp?

louiseschmidtgen (commented):
Lovely template! I particularly like the way you separate the changes into api, cli, db etc.

docs/proposals/001-k8s-status.md
neoaggelos: As a means of explaining _when_ the last update was done. It is different to see a recent timestamp, vs a timestamp 3 months ago

docs/proposals/001-k8s-status.md
```diff
@@ -0,0 +1,433 @@
...
+These changes will affect the output of the `k8s status` command, as explained
+in the section above.
+
+## Database Changes
+<!--
+This section MUST mention any changes required in the k8sd database schema or
+internal types.
+-->
+
+We need to define a new table (changes go in [pkg/k8sd/database]):
```
HomayoonAlimohammadi, line 297: I'm a bit confused about the database that we're adding this table to. AFAIK, there's `k8sd dqlite` which stores information about the nodes and does not really know anything about the actual k8s implementation on top of it (which nodes are master/worker, what is installed on each node or the cluster etc.) and there's this `k8s dqlite` which stores the actual cluster info known to k8s. we're changing the `k8sd/database` here which seems a bit odd to me. shouldn't this info be stored in the k8s database rather than the "node manager guy (k8sd)" database?

p.s. I can see that there is a `cluster_configs` table in the k8sd dqlite, but again, is k8sd (which afaik is a wrapper around MicroCluster) really supposed to know about the details of the actual k8s on top of it?

docs/proposals/001-k8s-status.md
```diff
@@ -0,0 +1,433 @@
...
+```
+
+#### New output
+
+With `--output-format=plain`, a simple overview is shown, with `yaml` or `json`
+more fields are shown:
+
+```bash
+$ k8s status
+ready: true
```
perk, line 138: What does the `ready:true` here mean?
It seems repeated by the `status: ready` one line below.

neoaggelos: ah, let me revert. while we are at it, i thought of changing `status: ready`/`status: not ready` to `ready: true/false`

but let me take this out, it's irrelevant for this proposal

docs/proposals/001-k8s-status.md
neoaggelos: So, `k8sd` is the orchestrator layer, exactly the place where we want this data to be stored. the kubernetes cluster itself cannot see this data, or interact with it.

docs/proposals/001-k8s-status.md
```diff
@@ -0,0 +1,432 @@
...
+  local-path: /var/snap/k8s/common/rawfile-storage
+  reclaim-policy: Delete
+  default: true
+gateway:
+  enabled: true
+```
+
+#### New output
+
+With `--output-format=plain`, a simple overview is shown, with `yaml` or `json`
```
evilnick, line 133: is this the deafult?

docs/proposals/001-k8s-status.md
```diff
@@ -0,0 +1,432 @@
...
+$ k8s status
+status: ready
+high-availability: no
+datastore:
+  type: k8s-dqlite
+  voter-nodes:
+    - 10.87.24.119:6400
+  standby-nodes: none
+  spare-nodes: none
+network:
```
evilnick, line 146: counterpoint - I think this ouput is still too long and I'm not sure that most people care to see all this every time they are checking whether k8s is running or not. Would be interesting to get some actual user feedback

docs/proposals/001-k8s-status.md
```diff
@@ -0,0 +1,432 @@
...
+  enabled: true
+  message: Cilium is deployed
+dns:
+  enabled: true
+  message: CoreDNS is deployed
+ingress:
```
evilnick, line 152: I think the status of most of these components could easily be reduced to one line if we wanted to be brief

docs/proposals/001-k8s-status.md
```diff
@@ -0,0 +1,432 @@
...
+  type: k8s-dqlite
+  voter-nodes:
+    - 10.87.24.119:6400
+  standby-nodes: none
+  spare-nodes: none
+network:
+  enabled: true
+  message: Cilium is deployed
+dns:
+  enabled: true
```
bschimke95, line 150: Right now, The plain output looks rather similar to the `yaml` output. The goal of the `plain` format should be to be as human-readable as possible.

What about using icons instead? People love icons!
```
✅ network
✅ dns
❌ ingress: Cilium IngressAPI support could not be enabled, error was ...
```

docs/proposals/001-k8s-status.md
```diff
@@ -0,0 +1,432 @@
...
+of how to solve, so we explicitly do not tackle them until we have more
+information.
+
+This section is very useful to help guide the implementation details section
+below, or serve as reference for future proposals.
+-->
+
+### `message` field
+
+The `message` field is the only information that will be presented to the user.
```
bschimke95, line 201: I'm +1 on this for errors but I don't think it provides value if we would e.g. add an additional "Cilium deployed" message if all is good.

bschimke95 (commented):
Great proposal, I added my thoughts around how to visualize the status of the features.

docs/proposals/001-k8s-status.md
neoaggelos: yes

docs/proposals/001-k8s-status.md
neoaggelos: we could drop the `message` if status is healthy?

docs/proposals/001-k8s-status.md
neoaggelos: should we do something like? (for the plain output)
```
network: enabled/disabled
ingress: enabled/disabled
```

docs/proposals/001-k8s-status.md
neoaggelos: not a bad idea, but, needs a lot of effort to make it like this, and it's not part of other microcluster apps either.

docs/proposals/001-k8s-status.md
neoaggelos: my fear is that this prevents us from expanding on this output in the future, if more info is needed for each feature.

docs/proposals/001-k8s-status.md
neoaggelos: but i still think `enabled` and `message` are two ok messages to have, doesn't feel too verbose to me. and dropping them might even make matters more complex for the implementation (when to show vs not show the message), or the users (why is a message there only sometimes?)

docs/proposals/001-k8s-status.md
neoaggelos: in all cases, we reserve the right to update the `plain` output as we prefer. scripts should use `json` or `yaml`

docs/proposals/001-k8s-status.md
evilnick: I was thinking something that gives some relevant info where useful. Just as an example:

e.g.
```
 sudo k8s status
cluster status:	   ready at 10.152.183.1
connected nodes:   10.152.183.16, 10.152.183.18,...[6 other nodes]
high-availability: no
datastore:         k8s-dqlite running
network:           enabled
dns: 		   enabled for cluster.local
ingress:           disabled
load-balancer:     disabled
local-storage:     enabled at /var/snap/k8s/common/rawfile-storage
gateway:           enabled
```
I don't really see the point of 9 lines of yaml config for the load balancer when it isn't even on.  If you want the full dump of everything that can be an option, and we should tell people that in a line at the end.

I am coming from a position of "what would the user find useful", not "what can we tell them" and I think the starting point should be that - why are you running 'status'?

neoaggelos: this reads like a great suggestion to me, more than happy to adjust accordingly. then it's up to the implementation to pick the level of detail for each component

neoaggelos: By the way, a few more thoughts on this: This subject touches on another peeve of mine about the current `k8s status` output, where the k8sd nodes appear as the datastore nodes but that is not really true (the microcluster endponts are not the cluster datastore, and are probably not even useful info)

I greatly see the value of adjusting `k8s status` to look like this, but the proposal started with the aim of iterating on it, one step at a time. as you said, there is no reason to show 9 lines of load balancer configs when people run `k8s status`, so we can start improving from this, in small iterations.

But I like your suggestion, it will be more involving for the implementation, but it will make `k8s status` a lot more useful than it currently is. Let me update the document accordingly, thanks!

docs/proposals/001-k8s-status.md
perk: If that's the latest update, can we call it `updated:` or sth along those lines?

perk (approved):
LGTM, thank you! :rocket:

PR #541, opened 2024-07-12
Move pki utils to separate pkiutil package

neoaggelos:
### Summary

Move utility functions that deal with x509 certificates and public/private keys into a reusable `pkiutil` package

This is to reduce diff with incoming changes around implementing support for certificate refreshes.


bschimke95 (approved):
LGTM

PR #542, opened 2024-07-14
clamp timeout to minimum in k8s_set.go

eaudetcobello:
other commands clamp timeout to minimum but this one did not


bschimke95 (approved):
LGTM, nice catch

bschimke95:
@eaudetcobello what's left to do here?

PR #546, opened 2024-07-15
do not fail-fast on integration tests

neoaggelos:
### Summary

We are at a point where integration tests fail more often because of flakes rather than an actual bug, therefore do not `fail-fast` when a flake occurs


bschimke95 (approved):
LGTM

PR #548, opened 2024-07-15
Add csrsigning controller for worker node certificate refreshes

neoaggelos:
### Summary

The following CSRs are signed:

- `k8sd.io/kubelet-serving`: used to sign certificates for serving kubelet TLS on the worker nodes
- `k8sd.io/kubelet-client`: used to sign certificates used by the kubelet kubeconfig on the worker nodes
- `k8sd.io/kube-proxy-client`: used to sign certificates used by the kube-proxy kubeconfig on the worker nodes

CSRs will not be signed unless they have previously been approved. The controller can optionally auto-approve valid CSRs (validation not implemented fully yet), but this behaviour is disabled by default and controlled by a command-line flag (for now).

### TODOs

- validate CSRs before accepting
- handle certificate TTL (currently hardcoded)

### Notes

`csrsigning` controller uses controller-runtime instead of the existing client-go tools. Reasons for this include:
- we need clean logic that allows us to retry requests (where watching for the resource is not sufficient), e.g. in case of CSRs not yet approved
- we need logic to retry (with a backoff on errors) reconcile loops
- we need leader election (so that only one k8sd instance manages certificates on the cluster, not all


src/k8s/pkg/k8sd/app/app.go
```diff
@@ -69,6 +72,9 @@ type App struct {
 	triggerFeatureControllerMetricsServerCh chan struct{}
 	triggerFeatureControllerDNSCh           chan struct{}
 	featureController                       *controllers.FeatureController
+
+	// csrsigningController
+	csrsigningController *csrsigning.Controller
```
eaudetcobello, line 77: Comment can be removed?

src/k8s/pkg/k8sd/controllers/csrsigning/controller.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get Kubernetes REST config: %w", err)
+	}
+
+	// TODO(neoaggelos): In case of more controllers, a single manager object should be created
+	// and passed here as configuration.
+	mgr, err := manager.New(config, manager.Options{
+		Logger:                  log.FromContext(ctx),
+		LeaderElection:          c.leaderElection,
+		LeaderElectionID:        "a27980c4.k8sd-csrsigning-controller",
```
eaudetcobello, line 76: Add a comment explaining how this ID came about?

src/k8s/pkg/k8sd/controllers/csrsigning/reconcile.go
```diff
@@ -0,0 +1,186 @@
...
+			return ctrl.Result{}, err
+		}
+		crtPEM = pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
+		if crtPEM == nil {
+			log.Info("Failed to encode signed certificate to memory")
+			return ctrl.Result{RequeueAfter: requeueAfterSigningFailure}, nil
+		}
+	default:
+		// NOTE(neoaggelos): this should never happen
+		return ctrl.Result{}, nil
```
eaudetcobello, line 175: Would it be useful to log an error here?

neoaggelos: This really only comes in code because Go lacks proper sum types. Essentially, we have already limited the signerName values of this switch with https://github.com/canonical/k8s-snap/blob/3fcd43c2f7f7473cd75cb1fa8b18d7af353c812a/src/k8s/pkg/k8sd/controllers/csrsigning/setup.go#L31-L39

Perhaps the one thing to do here would be to log a warning, that "hey, things have really gone south here"

src/k8s/pkg/k8sd/controllers/csrsigning/controller.go
neoaggelos: This ID is used to identify the lease and avoid conflicts with other controllers. `k8sd-csrsigning-controller` would also suffice, but let's take an extra step of safety. `a27980c4` is just 4 random bytes

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -151,6 +155,17 @@ func New(cfg Config) (*App, error) {
 	} else {
 		log.L().Info("feature-controller disabled via config")
 	}
+
+	if !cfg.DisableCSRSigningController {
+		app.csrsigningController = csrsigning.New(csrsigning.Options{
+			Snap:           cfg.Snap,
+			WaitReady:      app.readyWg.Wait,
+			LeaderElection: true,
```
mateoflorido, line 163: More than a suggestion, this is a question. This is a Kubernetes controller running outside the cluster, right? How does the leader election happen here in the cluster? Does it interact with the other K8s nodes to elect a controller leader?

neoaggelos: Right. Essentially, this maps to a `lease` object. See the "name" and "holder" columns

```
root@t1:~# k8s kubectl get leases -A
NAMESPACE         NAME                                   HOLDER                                                                      AGE
kube-node-lease   t1                                     t1                                                                          26h
kube-node-lease   t2                                     t2                                                                          26h
kube-system       a27980c4.k8sd-csrsigning-controller    t1_0fdf508a-72bf-40de-90a1-d967b34f3086                                     26h
kube-system       apiserver-tfvf6sfbh5ge4ia3ni5rdnxjoq   apiserver-tfvf6sfbh5ge4ia3ni5rdnxjoq_5612db5b-ece7-41fd-aafd-75e6d5fa833c   26h
kube-system       cilium-operator-resource-lock          t2-znc8m94nv6                                                               26h
kube-system       kube-controller-manager                t1_cbe995ef-cceb-4f9b-af4c-6a6cb25991cd                                     26h
kube-system       kube-scheduler                         t1_75da548f-bcca-4142-83c1-e9b76a738bd7                                     26h
```

What we want to avoid here, is having a 3-node control plane where the controller is running 3 times. In that case, any certificate would be signed thrice (with the 2 first getting discarded). Instead, we create a lease, and only let the owner of the lease run the code of the controller itself. Other controllers will stay dormant and watch for changes to this lease. If anything happens to the leader, the lease will expire, and someone else will take it and run the controller.

This is also exactly what the Kubernetes components themselves do (see e.g. the kube-scheduler and kube-controller-manager leases)

mateoflorido: This is amazing. Thank you for the great explanation!

mateoflorido (approved):
Great work, LGTM!

PR #551, opened 2024-07-16
Do not upload Tiobe tarball

ktsakalozos:
TICSQServer uploads the scan artifacts. We do not need to do anything more than that.


louiseschmidtgen (approved):
LGTM, one little nit: We should be able to remove `security-events: write ` from the job's permissions now.

PR #554, opened 2024-07-17
drop unpinned `docker/pkg/reexec` dependency from cni plugins

neoaggelos:
### Summary

When building cni plugins, we pull in `pkg/reexec` and vendor before running the build. This could result in an unpredictable source of external dependencies.

We avoid this by dropping the dependency and doing a simple switch-case, similar to https://github.com/canonical/k8s-snap/blob/a3d112ad4a74797e8b1bd6849442bb3069a039a9/src/k8s/cmd/main.go#L33-L41


berkayoz (approved):
LGTM

PR #555, opened 2024-07-17
Consistently use path/filepath for file paths

neoaggelos:
### Summary

From `go doc path`

```
package path // import "path"

Package path implements utility routines for manipulating slash-separated paths.

The path package should only be used for paths separated by forward slashes,
such as the paths in URLs. This package does not deal with Windows paths with
drive letters or backslashes; to manipulate operating system paths, use the
path/filepath package.
```


berkayoz (approved):
LGTM

PR #556, opened 2024-07-17
Proposal 002: Certificate refreshes

neoaggelos:
### Summary

Add proposal for implementing certificate refreshes for control plane and worker nodes for k8s-snap

### DIFF
--- a/docs/proposals/002-refresh-certificates.md
+++ b/docs/proposals/002-refresh-certificates.md
@@ -332,6 +332,17 @@ updated (e.g. command outputs).
 This section MUST explain how the new feature will be tested.
 -->
 
+1. An e2e test with 1 control plane and 1 worker node. We should be able to
+   refresh their certificates and see that the cluster is up and running. For
+   the needs of the test, we can enable auto-approval.
+2. Unit tests for csrsigning controller reconcile loop, where we test how it
+   handles individual CSR objects.
+3. (stretch) Investigate and add integration tests for the csrsigning controller
+   (perhaps using controller-runtime's envtest package).
+4. Unit tests for CSR object validations.
+5. Unit tests for CSR object creation from the worker nodes.
+6. Unit tests for any helper utility we need to add for signing certificates.
+
 ## Considerations for backwards compatibility
 <!--
 In this section, you MUST mention any breaking changes that are introduced by
--- a/docs/proposals/002-refresh-certificates.md
+++ b/docs/proposals/002-refresh-certificates.md
@@ -9,7 +9,7 @@ fill out the sections below.
 - **Index**: 002
 
 <!-- Status -->
-- **Status**: **DRAFTING** <!-- **DRAFTING**/**ACCEPTED**/**REJECTED** -->
+- **Status**: **ACCEPTED** <!-- **DRAFTING**/**ACCEPTED**/**REJECTED** -->
 
 <!-- Short description for the feature -->
 - **Name**: Refresh certificates
--- a/docs/proposals/002-refresh-certificates.md
+++ b/docs/proposals/002-refresh-certificates.md
@@ -1,0 +1,269 @@
+<!--
+To start a new proposal, create a copy of this template on this directory and
+fill out the sections below.
+-->
+
+# Proposal information
+
+<!-- Index number -->
+- **Index**: 002
+
+<!-- Status -->
+- **Status**: **DRAFTING** <!-- **DRAFTING**/**ACCEPTED**/**REJECTED** -->
+
+<!-- Short description for the feature -->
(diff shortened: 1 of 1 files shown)


docs/proposals/002-refresh-certificates.md
```diff
@@ -0,0 +1,371 @@
...
+<!--
+This section MUST mention any new configuration options or service arguments
+that are introduced.
+-->
+
+The changes include a `csrsigning` controller, which watches the cluster for
+`CertificateSigningRequest` objects coming from the worker nodes. After they are
+approved, it will sign the certificates.
+
+### k8sd `--disable-csrsigning-controller`
```
bschimke95, line 293: if we go with the `CSRSigning` cluster config, we could also provide this as a bootstrap/config option. If a user really don't want this feature, it would feeld cumbersome to change this on every node. We would still have this flag if the user wants to disable the controller just on one node.

docs/proposals/002-refresh-certificates.md
```diff
@@ -0,0 +1,371 @@
...
+
+In line with the rest of the `--disable-$foo-controller` flags, this flag can be
+used to completely disable the csrsigning controller on a node.
+
+### annotation `k8sd/v1alpha1/csrsigning/auto-approve=true`
```
bschimke95, line 298: As discussed, I'd go with a cluster config option here.

bschimke95 (commented):
Great work, added some comment. 
Will review the code once we finalized this doc.

docs/proposals/002-refresh-certificates.md
```diff
@@ -0,0 +1,371 @@
...
+type RefreshCertificatesRunRequest struct {
+  // Seed must match the seed from the received RefreshCertificatesInitResponse.
+  Seed int
+  // ExpirationDuration is the duration of the requested certificates.
+  ExpirationDuration time.Duration
+}
+
+type RefreshCertificatesRunResponse struct {
+  // ExpirationDuration is the duration of the new certificates.
+  ExpirationDuration time.Duration
```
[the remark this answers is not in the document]
neoaggelos: yes it can. someone might request for certificates for 10 years, we might not want to do that. instead, we will give them the maximum TTL we want, then include it here

docs/proposals/002-refresh-certificates.md
neoaggelos: annotations can help buy us some time before we know all the configs we might need/want or care about. there are more, e.g. `max-ttl`. i'd start with annotations, and we can move them to stable fields in the cluster config as we iterate on this feature

berkayoz: I'd say we should go with annotations first especially while we figure things out. Once something like this gets into the actual config people might expect it to be "officially" supported. I believe we are still in an early stage for this, I'd be up to discussing/working more on this to be official after we get the actual certificate refresh process in place.

mateoflorido: I'm +1 on annotations. I believe this could help us "graduate" new functionalities we add without messing up the internal cluster configuration, following a process similar to the upstream K8s API.

docs/proposals/002-refresh-certificates.md
```diff
@@ -0,0 +1,374 @@
...
+### `POST /refresh-certs/plan`
+
+```go
+type RefreshCertificatesPlanRequest struct {
+}
+
+type RefreshCertificatesPlanResponse struct {
+  // Seed needs to be passed to `/refresh-certs/run` to ensure that CSR
+  // names are valid.
+  Seed int
```
mateoflorido, line 188: I see that `seed` is mentioned in the CLI changes section:
```
k8s kubectl certificate approve k8sd-seed-worker-kubelet-client
```
So, I guess this is an identifier that the node knows, right? Does it need to be something specific, like a number we certainly know how to generate using available data, or a randomly generated number?

neoaggelos: Yes, CertificateSigningRequests cannot be "mutated" after they are created. Therefore, to prevent previous failures from blocking the refresh-certs process, we use this seed to differentiate the names across attempts.

This can be just a random number. The only reason why we do this, is to know the names of the created CSRs that will need to be signed (and print the relevant message)

docs/proposals/002-refresh-certificates.md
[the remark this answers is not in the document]
bschimke95: ack

docs/proposals/002-refresh-certificates.md
bschimke95: democracy - annotations it is

bschimke95 (approved):
LGTM

PR #560, opened 2024-07-18
Improve join logic to handle unreachable nodes

bschimke95:
Fixes an issue where the join process would fail if a node in the join_address list is unreachable.

When a token is created, it includes the IP addresses of all nodes in the cluster. If a node is removed between the token creation and joining, the join process previously failed because it tried to connect to the non-existent node.

Now, the client discovery logic will continue attempting to connect to other nodes in the join_address list instead of failing early. This ensures that the join process can succeed as long as at least one node is reachable.

This PR fixes this issue on `k8s-snap` level for the worker nodes and requires the microcluster upgrades from https://github.com/canonical/k8s-snap/pull/562 to actually pass the integration test.


bschimke95:
rebased on top of #562 to verify this fix works.

tests/integration/tests/test_clustering_race.py
```diff
@@ -0,0 +1,26 @@
...
+
+    new_join_token = util.get_join_token(cluster_node, instances[2])
+
+    cluster_node.exec(["k8s", "remove-node", instances[1].id])
+
+    another_join_token = util.get_join_token(cluster_node, instances[2])
+
+    # The join token should have changed after the node was removed as
+    # it contains the ip addresses of all cluster nodes.
+    assert new_join_token != another_join_token
```
louiseschmidtgen, line 25:
```suggestion
    assert new_join_token != another_join_token, "join token is not updated after node removal"
```

louiseschmidtgen (approved):
LGTM, thanks Ben!

berkayoz (approved):
LGTM

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -104,16 +70,25 @@ func (a *App) onBootstrapWorkerNode(ctx context.Context, s *state.State, encoded
...
+	// One might fail because the node is not part of the cluster anymore but was at the time the token was created.
+	var cert *x509.Certificate
+	var address string
+	var err error
+	for _, address = range token.JoinAddresses {
+		cert, err = utils.GetRemoteCertificate(address)
+		if err == nil {
+			break
+		}
+	}
```
neoaggelos, line 83: In general, try to avoid mixing dependency upgrades along with changes in the code logic, as it would be harder to backport and/or revert if required.

PR #561, opened 2024-07-18
update runc patch for runc v1.1.13

eaudetcobello:
the previous version of the patch was compatible with runc v1.1.12 but the code the patch touches was changed in v1.1.13 which is the version of runc we would like to update to

patch 1 and 2 are identical to v1.1.12, only patch 3 is modified

Merge this before https://github.com/canonical/k8s-snap/pull/559 and https://github.com/canonical/k8s-snap/pull/558


neoaggelos:
Let's make this part of #559, these changes should go together

PR #562, opened 2024-07-22
Update microcluster version and adapt to API changes

bschimke95:
## Summary

This PR upgrades Microcluster to the latest version, including the changes from [microcluster#189](https://github.com/canonical/microcluster/pull/189). This update is essential for enabling CAPI upgrades.

#### Removal of Custom Cleanup Routines

Previously, we managed cleanup functions by appending them to a cleanups list during the bootstrap/join hooks. These were executed in reverse order in a deferred, separate Go routine upon failure. This approach allowed us to provide immediate feedback to the user when an error occurred, indicating that cleanup was in progress. Users attempting another bootstrap during cleanup would encounter a "Node already part of the cluster" error because the node remained registered until the final cleanup step completed.

#### New Cleanup Strategy

With the updated Microcluster, internal cleanup is now automatically handled on failure. This means we can no longer control the exact timing of node cleanup, preventing us from using a separate Go routine for our cleanups. Instead, all cleanup processes are now performed in the PreRemove hook, which is guaranteed to run as part of the Microcluster bootstrap and join cleanups (see [microcluster#194](https://github.com/canonical/microcluster/pull/194)). All steps in this hook must be blocking, as the context is cancelled once the hook returns.

#### Impact on User Experience

This change slightly degrades user experience as we can no longer return early to inform the user that cleanup is ongoing. However, it ensures that cleanup is correctly synchronized with Microcluster’s internal logic, preventing race conditions and potential inconsistencies.

## Other notable changes

* Microcluster has changed the way it configures endpoints. Additional endpoints are now configured through `Servers`.
* The signature of access handlers has changed
* The API calls require a prefix now
* The internal microcluster context is not exposed anymore, instead the request context should be used. See also https://github.com/canonical/microcluster/pull/190 and this [internal MM discussion](https://chat.canonical.com/canonical/pl/iq4ji3ro6tgtfffx99mbudjk5e)
* The microcluster state is an interface now instead of a struct.


src/k8s/go.mod
```diff
@@ -1,19 +1,21 @@
 module github.com/canonical/k8s
 
-go 1.22.0
+go 1.22.4
+
+toolchain go1.22.5
```
neoaggelos, line 5: can we simply do `go 1.22.5`?

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -32,7 +32,7 @@ func (e *Endpoints) postClusterRemove(s *state.State, r *http.Request) response.
 		defer cancel()
 	}
 
-	log := log.FromContext(s.Context).WithValues("name", req.Name)
+	log := log.FromContext(ctx).WithValues("name", req.Name)
```
neoaggelos, line 35: if microcluster has dropped `s.Context`, let's make a `func (e *Endpoints) Context() context.Context` instead and use that.

src/k8s/pkg/k8sd/api/kubernetes_auth_tokens.go
```diff
@@ -6,22 +6,23 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	databaseutil "github.com/canonical/k8s/pkg/k8sd/database/util"
 	"net/http"
 
+	databaseutil "github.com/canonical/k8s/pkg/k8sd/database/util"
+
```
neoaggelos, line 12: no spaces between imports

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -81,6 +80,8 @@ func New(cfg Config) (*App, error) {
 		cfg.StateDir = cfg.Snap.K8sdStateDir()
 	}
 	cluster, err := microcluster.App(microcluster.Args{
+		// TODO(ben): Properly manage this new required version parameter.
+		Version:  "1.0",
```
neoaggelos, line 84: is this the same as the `apiv1.K8sdVersionPrefix`?

src/k8s/api/v1/path.go
```diff
@@ -0,0 +1,8 @@
+package apiv1
+
+import "github.com/canonical/microcluster/rest/types"
+
+const (
+	// K8sdVersionPrefix is the path prefix that will be used for the k8sd endpoints for this api version.
+	K8sdVersionPrefix types.EndpointPrefix = "1.0"
```
neoaggelos, line 7: maybe `K8sdAPIVersion`?

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -78,13 +78,13 @@ func (a *App) onBootstrapWorkerNode(ctx context.Context, s *state.State, encoded
...
 				// run cleanup functions in reverse order
-				if err := cleanups[i](s.Context); err != nil {
+				if err := cleanups[i](ctx); err != nil {
 					log.Error(err, fmt.Sprintf("Cleanup hook %d/%d failed", i, len(cleanups)))
 				}
 			}
 			log.Info("All cleanup hooks finished, resetting microcluster state")
 
-			if err := a.client.ResetClusterMember(s.Context, s.Name(), true); err != nil {
+			if err := a.client.ResetClusterMember(ctx, s.Name(), true); err != nil {
```
neoaggelos, line 87: this is not right, as the request context is cancelled after we've sent the response. see my other comment about a `.Context()` that gives us the base context.

bschimke95: https://github.com/canonical/microcluster/pull/193 updates the ctx to be the microcluster daemon context instead of the request context.

src/k8s/pkg/k8sd/api/cluster_remove.go
bschimke95: As discussed offline: This would only be required for the long-running cleanup tasks which should be addressed here https://github.com/canonical/microcluster/pull/193. The remaining uses of `s.Context` should be sufficient with `r.Context`

src/k8s/pkg/client/dqlite/remove.go
```diff
@@ -20,6 +20,10 @@ func (c *Client) RemoveNodeByAddress(ctx context.Context, address string) error
 		return fmt.Errorf("failed to retrieve cluster nodes")
 	}
 
+	if len(members) == 1 {
```
bschimke95, line 23: exit early if there is only one node in the cluster.
Otherwise, this would unnecessarily retry and block further down this function

src/k8s/pkg/k8sd/api/cluster_remove.go
bschimke95: As discussed in MM: the cleanup is now entirely moved to the remove hook. PR is updated to explain the rationale.

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -14,13 +14,13 @@ import (
 	"github.com/canonical/microcluster/state"
 )
 
-func (a *App) onStart(s *state.State) error {
+func (a *App) onStart(ctx context.Context, s state.State) error {
```
bschimke95, line 17: For the `OnStart` context, the full lifecyle context is passed. See: https://github.com/canonical/microcluster/pull/190

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
bschimke95: obsolete

src/k8s/pkg/k8sd/api/capi_access_handler.go
```diff
@@ -11,28 +11,28 @@ import (
...
+	return func(s state.State, r *http.Request) (bool, response.Response) {
 		token := r.Header.Get(tokenHeaderName)
 		if token == "" {
-			return response.Unauthorized(fmt.Errorf("missing header %q", tokenHeaderName))
+			return false, response.Unauthorized(fmt.Errorf("missing header %q", tokenHeaderName))
 		}
 
 		var tokenIsValid bool
-		if err := s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
+		if err := s.Database().Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
```
mateoflorido, line 22: I noticed we are changing many instance `s/s.Context/r.Context()`. What's the reason for this change?

bschimke95: Please see the MM link above for the full story.

Tl;dr: Microcluster does not expose `s.Context` any more and for *most* of the usages that we had, the request context is perfectly fine.

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/canonical/microcluster/state"
 )
 
-func (e *Endpoints) postClusterBootstrap(s *state.State, r *http.Request) response.Response {
+func (e *Endpoints) postClusterBootstrap(s state.State, r *http.Request) response.Response {
```
addyess, line 15: it looks like the `s state.State` is unused.  Should this arg be removed or marked in some way to show it's not used?

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -42,8 +42,8 @@ func (e *Endpoints) postClusterRemove(s *state.State, r *http.Request) response.
 		log.Info("Waiting for node to not be pending")
 		control.WaitUntilReady(ctx, func() (bool, error) {
 			var notPending bool
-			if err := s.Database.Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
-				member, err := cluster.GetInternalClusterMember(ctx, tx, req.Name)
+			if err := s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
+				member, err := cluster.GetCoreClusterMember(ctx, tx, req.Name)
```
addyess, line 46: what's this change do?

`GetInternalClusterMember` -> `GetCoreClusterMember`

src/k8s/pkg/k8sd/api/worker_access_handler.go
```diff
@@ -13,53 +13,53 @@ import (
...
+		return false, response.InternalError(fmt.Errorf("failed to check if node is a worker: %w", err))
 	}
 
 	if isWorker {
-		return response.Forbidden(fmt.Errorf("this action is restricted on workers"))
+		return false, response.Forbidden(fmt.Errorf("this action is restricted on workers"))
 	}
 
-	return response.EmptySyncResponse
+	return true, nil
```
addyess, line 28: again, do we need the `EmptySyncResponse`

src/k8s/pkg/k8sd/api/worker_access_handler.go
```diff
@@ -13,53 +13,53 @@ import (
...
-			return response.InternalError(fmt.Errorf("check token database transaction failed: %w", err))
+			return false, response.InternalError(fmt.Errorf("check token database transaction failed: %w", err))
 		}
 		if !tokenIsValid {
-			return response.Unauthorized(fmt.Errorf("invalid token"))
+			return false, response.Unauthorized(fmt.Errorf("invalid token"))
 		}
 
-		return response.EmptySyncResponse
+		return true, nil
```
addyess, line 63: and here?

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/canonical/k8s/pkg/k8sd/setup"
+	"github.com/canonical/k8s/pkg/log"
+	snaputil "github.com/canonical/k8s/pkg/snap/util"
+	"github.com/canonical/k8s/pkg/utils/control"
+	"github.com/canonical/microcluster/cluster"
+	"github.com/canonical/microcluster/state"
+)
+
+// NOTE(ben): the pre-remove performs a series of cleanup steps on a best-effort basis.
+// If any of the steps fail, the error is logged and the cleanup continues, with depdent tasks being skipped.
```
mateoflorido, line 21:
```suggestion
// If any step fails, the error is logged, and the cleanup continues, skipping dependent tasks.
```
Wdyt?

addyess (commented):
Just a few comments.  It looks like you've put a ton of good thought into this

src/k8s/pkg/k8sd/app/hooks_remove.go
bschimke95: sounds good, thanks!

src/k8s/pkg/k8sd/api/capi_access_handler.go
```diff
@@ -11,28 +11,28 @@ import (
...
-			return response.InternalError(fmt.Errorf("check CAPI auth token database transaction failed: %w", err))
+			return false, response.InternalError(fmt.Errorf("check CAPI auth token database transaction failed: %w", err))
 		}
 		if !tokenIsValid {
-			return response.Unauthorized(fmt.Errorf("invalid token"))
+			return false, response.Unauthorized(fmt.Errorf("invalid token"))
 		}
 
-		return response.EmptySyncResponse
+		return true, nil
```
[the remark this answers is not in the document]
bschimke95: The microcluster team changed the signatures for access handlers. Previously a response object was returned, e.g. `EmptySyncResponse` for "access granted` and `response.Unauthorized` if the user may not access that resource.
Now, the return values are `hasAccess, error`.

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
bschimke95: good point - fixed

src/k8s/pkg/k8sd/api/capi_access_handler.go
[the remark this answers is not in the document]
addyess: 👍🏻

src/k8s/pkg/k8sd/api/cluster_remove.go
bschimke95: The microcluster team renamed their core implementation from `internal` to `core`. This function is provided by microcluster and just got renamed as well.
No functional changes.

src/k8s/pkg/k8sd/api/worker_access_handler.go
bschimke95: Actually, I think both work, returning `nil` and `EmptySyncResponse`. 
Just followed along with how they do it now in their default access handler: 
https://github.com/canonical/microcluster/blob/main/rest/access/handlers.go#L32

src/k8s/api/v1/path.go
```diff
@@ -0,0 +1,8 @@
+package apiv1
```
louiseschmidtgen, line 1: Can we rename the filename please? Maybe something like `api_version` or just `version`?

src/k8s/api/v1/path.go
```diff
@@ -0,0 +1,8 @@
...
+
+import "github.com/canonical/microcluster/rest/types"
+
+const (
+	// K8sdAPIVersion is the path prefix that will be used for the k8sd endpoints for this api version.
+	K8sdAPIVersion types.EndpointPrefix = "1.0"
```
louiseschmidtgen, line 7: Is it possible to grab this version from microcluster directly? Maybe from here: https://github.com/canonical/microcluster/blob/main/internal/db/update/schema.go

bschimke95: This is not the microcluster version but *our* API version

louiseschmidtgen (approved):
LGTM, thanks for updating our microcluster version!

berkayoz (approved):
Great work, wishing for a more stable upstream API 🙂

PR #563, opened 2024-07-23
Improve k8s status

HomayoonAlimohammadi:
Reworked `k8s status` to comply with the [new proposal](https://github.com/canonical/k8s-snap/blob/main/docs/proposals/001-k8s-status.md)


src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -69,65 +69,96 @@ func NewFeatureController(opts FeatureControllerOpts) *FeatureController {
...
+	apply func(cfg types.ClusterConfig) (types.FeatureStatus, error),
+	setFeatureStatus func(ctx context.Context, name string, status types.FeatureStatus) error,
+) {
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-triggerCh:
-			if err := c.reconcile(ctx, getClusterConfig, apply); err != nil {
+			if err := c.reconcile(ctx, getClusterConfig, apply, componentName, setFeatureStatus); err != nil {
```
neoaggelos, line 161: as a generic note: each `reconcile` loop here does not _need_ access to the full `setFeatureStatus`, only a partial with the specified componentName.

instead of passing both `componentName` and `setFeatureStatus`, consider doing something like

```
if err := c.reconcile(ctx, getClusterConfig, apply, func(ctx context.Context, status types.FeatureStatus) error {
  return setFeatureStatus(ctx, componentName, status)
}); err != nil {
  // ...
}
```

So `reconcile` only sees a ["partial"](https://stackoverflow.com/questions/58339040/what-exactly-is-meant-by-partial-function-in-functional-programming), and does not need to concern itself with passing through the componentName to the setFeatureStatus call

src/k8s/pkg/k8sd/database/feature_status.go
```diff
@@ -0,0 +1,79 @@
...
+		typ := types.FeatureStatus{}
+
+		if err := rows.Scan(&name, &typ.Message, &typ.Version, &ts, &typ.Enabled); err != nil {
+			return nil, fmt.Errorf("failed to scan row: %w", err)
+		}
+
+		typ.UpdatedAt, err = time.Parse(time.RFC3339, ts)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse timestamp: %w", err)
+		}
```
neoaggelos, line 69: an invalid timestamp should be probably be a warning, not a fatal error to retrieve the status

src/k8s/pkg/k8sd/database/sql/queries/feature-status/select.sql
```diff
@@ -0,0 +1,4 @@
+SELECT
+    name, message, version, timestamp, enabled
+FROM
+    feature_status
```
neoaggelos, line 4: newline at EOF

src/k8s/pkg/k8sd/database/feature_status.go
```diff
@@ -0,0 +1,79 @@
...
+	upsertTxStmt, err := cluster.Stmt(tx, featureStatusStmts.upsert_)
+	if err != nil {
+		return fmt.Errorf("failed to prepare upsert statement: %w", err)
+	}
+
+	if _, err := upsertTxStmt.ExecContext(ctx,
+		name,
+		status.Message,
+		status.Version,
+		status.UpdatedAt.Format(time.RFC3339),
```
neoaggelos, line 32: i think it should be possible to pass (and retrieve) `time.Time` objects directly

src/k8s/pkg/k8sd/database/feature_status.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/canonical/microcluster/cluster"
+)
+
+var featureStatusStmts = struct {
+	select_ int
+	upsert_ int
+}{
+	select_: MustPrepareStatement("feature-status", "select.sql"),
+	upsert_: MustPrepareStatement("feature-status", "upsert.sql"),
+}
```
neoaggelos, line 19: let's to a `map[string]int` like in other places, to avoid the duplicate lines here

src/k8s/pkg/k8sd/database/feature_status.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	fsMap := make(map[string]types.FeatureStatus)
+
+	for rows.Next() {
+		var (
+			name string
+			ts   string
+		)
+		typ := types.FeatureStatus{}
```
neoaggelos, line 60: let's call this just `status`

src/k8s/pkg/k8sd/database/feature_status.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare select statement: %w", err)
+	}
+
+	rows, err := selectTxStmt.QueryContext(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute select statement: %w", err)
+	}
+
+	fsMap := make(map[string]types.FeatureStatus)
```
neoaggelos, line 53: let's call this `result`

src/k8s/pkg/k8sd/api/cluster.go
```diff
@@ -45,6 +69,13 @@ func (e *Endpoints) getClusterStatus(s *state.State, r *http.Request) response.R
 				Type:    config.Datastore.GetType(),
 				Servers: config.Datastore.GetExternalServers(),
 			},
+			DNS:           featureStatuses["dns"],
```
neoaggelos, line 72: you can just retrieve the `statuses`, then do `statuses["dns"].ToAPI()`

src/k8s/pkg/k8sd/types/feature_status.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func FeatureStatusFromAPI(apiFS apiv1.FeatureStatus) (FeatureStatus, error) {
+	return FeatureStatus{
+		Enabled:   apiFS.Enabled,
+		Message:   apiFS.Message,
+		Version:   apiFS.Version,
+		UpdatedAt: apiFS.UpdatedAt,
+	}, nil
+}
```
neoaggelos, line 37: these cannot fail, remove the optional error (also helps with the `.ToAPI()` conversions in the cluster status RPC

src/k8s/pkg/k8sd/types/feature_status_test.go
```diff
@@ -0,0 +1,44 @@
...
+func TestK8sdFeatureStatusToAPI(t *testing.T) {
+	k8sdFS := types.FeatureStatus{
+		Enabled:   true,
+		Message:   "message",
+		Version:   "version",
+		UpdatedAt: time.Now(),
+	}
+
+	apiFS, err := k8sdFS.ToAPI()
+	require.NoError(t, err)
```
neoaggelos, line 23: use gomega, like we do in all other tests

src/k8s/go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/onsi/gomega v1.32.0
 	github.com/pelletier/go-toml v1.9.5
 	github.com/spf13/cobra v1.8.0
+	github.com/stretchr/testify v1.9.0
```
neoaggelos, line 15: we already use `gomega`, let's not add another testing framework

src/k8s/api/v1/types.go
```diff
@@ -69,74 +88,29 @@ func (c ClusterStatus) HaClusterFormed() bool {
...
+	// Control Plane Nodes
+	result.WriteString("control plane nodes: ")
+	addrMap := c.getCPNodeAddrToRoleMap()
+	nodes := make([]string, len(addrMap))
+	i := 0
+	for addr, role := range addrMap {
+		nodes[i] = fmt.Sprintf("%s (%s)", addr, role)
+		i++
 	}
+	result.WriteString(strings.Join(nodes, ", "))
```
neoaggelos, line 113: you can just loop over the `c.Members`, and `result.WriteString` them here? you have two extra loops that come with a lot of allocations each

src/k8s/api/v1/types.go
```diff
@@ -69,74 +88,29 @@ func (c ClusterStatus) HaClusterFormed() bool {
...
+	result.WriteString("\n")
+
+	// Control Plane Nodes
+	result.WriteString("control plane nodes: ")
+	addrMap := c.getCPNodeAddrToRoleMap()
+	nodes := make([]string, len(addrMap))
+	i := 0
+	for addr, role := range addrMap {
+		nodes[i] = fmt.Sprintf("%s (%s)", addr, role)
+		i++
```
neoaggelos, line 111: the more idiomatic way to write this is:

```
nodes := make([]string, 0, len(knownSize))
for ... {
  nodes = append(nodes, item)
}
```

neoaggelos (commented):
Really nice work @HomayoonAlimohammadi ! Did a first pass

src/k8s/pkg/k8sd/controllers/feature.go
HomayoonAlimohammadi: right, amazing comment, so we're closing over the `componentName` here (forming a closure), right?

src/k8s/api/v1/types_test.go
```diff
@@ -125,19 +120,25 @@ dns:
...
+gateway                   disabled
 `,
 		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			g := NewWithT(t)
+			fmt.Println("######################## cluster status:")
+			fmt.Println(tc.clusterStatus.String())
```
neoaggelos, line 141: make sure thus is reverted

HomayoonAlimohammadi: oops my bad

src/k8s/pkg/k8sd/database/feature_status.go
HomayoonAlimohammadi: tbh I really like every bit of type safety that I can get :D

src/k8s/pkg/k8sd/api/cluster.go
HomayoonAlimohammadi: `ToAPI()` returns an `error`, we can remove it but I guess it's not a bad thing to have it there (for future)

src/k8s/pkg/k8sd/types/feature_status.go
HomayoonAlimohammadi: ah sorry saw the one above first, ok

src/k8s/api/v1/types.go
HomayoonAlimohammadi: i did this because of the "," between IPs. 
this:
```go
for _, m := range c.Members {
    result.WriteString("%s (%s),", m.Address, m.DatastoreRole)
}
```
will result in something like:
`control plane nodes: 1.2.3.4 (voter), 2,3,4,5 (stand-by),` (notice the extra "," at the end)

we will have an extra separator or we can have none (separate them by spaces), correct?

neoaggelos: ok, how about something like this, then:

```go
members := make([]string, 0, len(c.Members))
for _, member := range c.Members {
  members = append(members, fmt.Sprintf("%s (%s)", member.Address, member.Role))
}

result.WriteString(strings.Join(members, ", ")
```

HomayoonAlimohammadi: or we can just separate them with 2 spaces and call it a day 
`192.168.0.1 (voter)  192.168.0.2 (voter)  192.168.0.3 (stand-by)`

src/k8s/pkg/k8sd/database/feature_status.go
HomayoonAlimohammadi: looks like it fails with
`failed to scan row: sql: Scan error on column index 3, name "timestamp": unsupported Scan, storing driver.Value type string into type *time.Time`

src/k8s/api/v1/types.go
neoaggelos: > or maybe in case the underlying array has enough space, append does not reallocate?

exactly, that's what `make([]T, 0, len)` does. `append` here will not do extra allocations

src/k8s/api/v1/types.go
```diff
@@ -116,15 +116,12 @@ func (c ClusterStatus) String() string {
...
-		i++
-	}
-	if len(nodes) > 0 {
-		result.WriteString(strings.Join(nodes, ", "))
+	if len(c.Members) > 0 {
+		for _, m := range c.Members {
+			if m.ClusterRole == ClusterRoleControlPlane {
+				result.WriteString(fmt.Sprintf("%s (%s)  ", m.Address, m.DatastoreRole))
+			}
+		}
```
neoaggelos, line 124: put this into an array and `strings.Join(parts, ", ")` at the end, let's not add whitespace at end of lines

src/k8s/api/v1/types.go
```diff
@@ -69,94 +98,72 @@ func (c ClusterStatus) HaClusterFormed() bool {
...
+	// longer than the longest key (eye-balled), make the output left aligned
+	maxLen := 25
+
 	// Status
 	if c.Ready {
-		result.WriteString("status: ready")
+		result.WriteString(fmt.Sprintf("%-*s %s", maxLen, "cluster status:", "ready"))
+	} else {
+		result.WriteString(fmt.Sprintf("%-*s %s", maxLen, "cluster status:", "not ready"))
+	}
```
neoaggelos, line 114: i think this would read better as `fmt.Sprintf("%25s %s", "cluster status:", "ready")` (it's ok to repeat 25 a few times in the code here)

src/k8s/api/v1/types.go
```diff
@@ -69,94 +98,72 @@ func (c ClusterStatus) HaClusterFormed() bool {
...
+	result.WriteString("\n")
+
+	// Control Plane Nodes
+	result.WriteString(fmt.Sprintf("%-*s ", maxLen, "control plane nodes:"))
+	if len(c.Members) > 0 {
+		members := make([]string, 0, len(c.Members))
+		for _, m := range c.Members {
+			if m.ClusterRole == ClusterRoleControlPlane {
+				members = append(members, fmt.Sprintf("%s (%s)", m.Address, m.DatastoreRole))
+			}
```
neoaggelos, line 124: any reason not to include pending nodes ?

src/k8s/api/v1/types.go
```diff
@@ -69,94 +98,72 @@ func (c ClusterStatus) HaClusterFormed() bool {
...
+
+	// Load Balancer
+	result.WriteString(fmt.Sprintf("%-*s %s\n", maxLen, "load-balancer:", c.LoadBalancer.GetMessage()))
+
+	// Local Storage
+	result.WriteString(fmt.Sprintf("%-*s %s\n", maxLen, "local-storage:", c.LocalStorage.GetMessage()))
+
+	// Gateway
+	result.WriteString(fmt.Sprintf("%-*s %s\n", maxLen, "gateway", c.Gateway.GetMessage()))
+
```
neoaggelos, line 166: this can just be 6 lines, again i think it's OK to repeat 25 a few times:

```
result.WriteString(fmt.Sprintf("%25s %s", "network:", c.Network.GetMessage()))
result.WriteString(fmt.Sprintf("%25s %s", "dns:", c.DNS.GetMessage()))
....
```

Also, you can change `GetMessage() string` to a `String()`, then use `fmt.Sprintf("%v", c.Network))`

src/k8s/api/v1/types_test.go
```diff
@@ -125,19 +120,24 @@ dns:
...
+local-storage:            disabled
+gateway                   disabled
 `,
 		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			g := NewWithT(t)
+			fmt.Println(tc.clusterStatus.String())
```
neoaggelos, line 140: note

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -69,65 +69,97 @@ func NewFeatureController(opts FeatureControllerOpts) *FeatureController {
...
 
-	if err := apply(cfg); err != nil {
+	featureStatus, err := apply(cfg)
+	if err != nil {
 		return fmt.Errorf("failed to apply configuration: %w", err)
 	}
+
+	if err := updateFeatureStatus(ctx, featureStatus); err != nil {
+		return fmt.Errorf("failed to update feature status: %w", err)
+	}
```
neoaggelos, line 141: another way to write this (to avoid err being referenced by accident further down)

```go
   if status, err := apply(cfg); err != nil {
   ...
   } else if err := updateFeatureStatus(ctx, status); err != nil {
   ...
   }
```

src/k8s/pkg/k8sd/database/feature_status.go
```diff
@@ -0,0 +1,77 @@
...
+		)
+		status := types.FeatureStatus{}
+
+		if err := rows.Scan(&name, &status.Message, &status.Version, &ts, &status.Enabled); err != nil {
+			return nil, fmt.Errorf("failed to scan row: %w", err)
+		}
+
+		status.UpdatedAt, err = time.Parse(time.RFC3339, ts)
+		if err != nil {
+			log.L().Error(err, "failed to parse time", "original", ts)
```
neoaggelos, line 66: We have a context, so we should use `log.FromContext(ctx)` (this will include logger fields from the caller)

src/k8s/pkg/k8sd/database/feature_status.go
```diff
@@ -0,0 +1,77 @@
...
+			ts   string
+		)
+		status := types.FeatureStatus{}
+
+		if err := rows.Scan(&name, &status.Message, &status.Version, &ts, &status.Enabled); err != nil {
+			return nil, fmt.Errorf("failed to scan row: %w", err)
+		}
+
+		status.UpdatedAt, err = time.Parse(time.RFC3339, ts)
+		if err != nil {
```
neoaggelos, line 65: prefer inline

```go
if status.UpdateAt, err = time.Parse(..); err != nil {
```

src/k8s/pkg/k8sd/database/feature_status.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	result := make(map[string]types.FeatureStatus)
+
+	for rows.Next() {
+		var (
+			name string
+			ts   string
+		)
+		status := types.FeatureStatus{}
```
neoaggelos, line 58: since you have a var block:

```
var (
   timestamp, name string
   status types.FeatureStatus
)
```

src/k8s/pkg/k8sd/database/feature_status_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/canonical/k8s/pkg/k8sd/database"
+	"github.com/canonical/k8s/pkg/k8sd/types"
+)
+
+func TestFeatureStatus(t *testing.T) {
+	WithDB(t, func(ctx context.Context, db DB) {
+		_ = db.Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
+			g := NewWithT(t)
+			t0, _ := time.Parse(time.RFC3339, time.Now().Format(time.RFC3339))
+			// initial get should return nothing
```
neoaggelos, line 20: instead of comments, let's make these subtests with `t.Run(...)`.

See the existing tests for ideas

neoaggelos (commented):
Nice!

src/k8s/api/v1/types.go
neoaggelos: ^ most importantly, the comments themselves do not add much here, it's ok to remove them

src/k8s/pkg/k8sd/database/feature_status_test.go
```diff
@@ -68,38 +40,76 @@ func TestFeatureStatus(t *testing.T) {
...
+				ss, err := database.GetFeatureStatuses(ctx, tx)
+				g.Expect(err).To(BeNil())
+				g.Expect(len(ss)).To(Equal(0))
+
+			})
+
+			t.Run("SettingNewStatus", func(t *testing.T) {
+				g := NewWithT(t)
+
+				err := database.SetFeatureStatus(ctx, tx, "network", networkS)
```
neoaggelos, line 55: `networkStatus` is OK, try to avoid names that do not read easily.

HomayoonAlimohammadi:
@neoaggelos something that's bothering me is that we're not guaranteeing any consistency between `.Enabled` and the contents of the `.Message` field. we can have `Enabled: true` and `Message: Failed to ...`. In fact I'm noticing some of these issues with this PR that I'm fixing, but overall, I feel like this is going to cause problems. 

I can think of two ways:
1. `Enabled` is determined from `Message` field (whenever this field changes, we change `Enabled` accordingly)
2. we delete `Enabled` in which case we're throwing away a lot of usefulness of this feature, since there's no way that it can be programmatically parsed and used (`Message` is not reliable)

p.s. We can't really infer `Message` from `Enabled`  

what do you think?

HomayoonAlimohammadi:
Another question would be:

In the case below, should be we the `Enabled` field to `true`? because deletion was not successful so it might mean that the deployment is still there (or maybe the chart is uninstalled but the `m.Apply(...)` returned error for any other reason (timeout after chart deletion etc. in which case, we may want to decide based on the returned error)
```go
if !cfg.GetEnabled() {
		if _, err := m.Apply(ctx, chartCalico, helm.StateDeleted, nil); err != nil {
			applyErr := fmt.Errorf("failed to uninstall network: %w", err)
			status.Message = fmt.Sprintf(deleteFailedMsgTmpl, applyErr)
                        status.Enabled = true // <-------- SHOULD BE HERE??
			return status, applyErr
		}
		status.Version = ""
		status.Message = disabledMsg
		return status, nil
	}
```

neoaggelos:
> @neoaggelos something that's bothering me is that we're not guaranteeing any consistency between .Enabled and the contents of the .Message field. we can have Enabled: true and Message: Failed to .... In fact I'm noticing some of these issues with this PR that I'm fixing, but overall, I feel like this is going to cause problems.

> what do you think?

So, we do not want to have people depending on `if something["message"] == "enabled"` to check for that. the `message` field is by design _unstructured_ and is supposed to only be informative.

The `.Enabled` field should be set to true iff (1) the feature is enabled (2) the apply succeeded. All other cases should set it to `false`. Then, there is a stable `enabled` boolean field that people can query which tells them that "this thing is supposed to be running".

tests/integration/tests/test_smoke.py
```diff
@@ -23,7 +24,7 @@ def test_smoke(instances: List[harness.Instance]):
     )
 
     instance.exec(["k8s", "bootstrap", "--file", bootstrap_smoke_config_path])
-    util.wait_until_k8s_ready(instance, [instance])
+    util.wait_until_k8s_ready(instance, [instance], retries=120, delay_s=10)
```
bschimke95, line 27: any reason for this large amount of retries?
The cluster should not need 20min to get up, even if it has to pull everything.
This would only take that long if something is broken and then we don't want the runner to execute the test for that long.
Save the planet! :)

bschimke95 (approved):
LGTM, only minor comment

tests/integration/tests/test_smoke.py
HomayoonAlimohammadi: you're absolutely right :D 
The problem was that during my tests the node sometimes took +10 minutes to get ready, don't know why. I'll change this back to the original values.

bschimke95: Hm, the node is marked as ready once the CNI (Cilium) is deployed. This takes around 2min on my local machine.
However, it pulls the images so this heavily depends on your internet connection.

HomayoonAlimohammadi: yea sometimes the VMs on my machine can not use the full bandwidth, they reach <1MBps at best.

.gitignore
```diff
@@ -1,6 +1,8 @@
 /parts/
 /stage/
 /prime/
+.vscode/
+env/
```
neoaggelos, line 5: these are unrelated to the PR, let's keep them out

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -69,65 +69,95 @@ func NewFeatureController(opts FeatureControllerOpts) *FeatureController {
...
+	updateFeatureStatus func(context.Context, types.FeatureStatus) error,
+) error {
 	cfg, err := getClusterConfig(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to retrieve cluster configuration: %w", err)
 	}
 
-	if err := apply(cfg); err != nil {
+	featureStatus, err := apply(cfg)
+	if err != nil {
```
neoaggelos, line 135: make this inline
```suggestion
	if featureStatus, err := apply(cfg); err != nil {
```

src/k8s/pkg/k8sd/features/calico/network.go
```diff
@@ -10,28 +10,53 @@ import (
...
-func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, annotations types.Annotations) error {
+// ApplyNetwork will always return a FeatureStatus indicating the current status of the
+// deployment.
+// ApplyNetwork returns an error if anything fails. The error is also wrapped in the .Message field of the
+// returned FeatureStatus.
+func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, annotations types.Annotations) (types.FeatureStatus, error) {
+	status := types.FeatureStatus{
+		Version: calicoTag,
+		Enabled: cfg.GetEnabled(),
+	}
```
neoaggelos, line 30: avoid creating a `status` here that is mutated in place afterwards (as it can be easy to forget about fields in the future).

src/k8s/pkg/k8sd/features/calico/network.go
```diff
@@ -10,28 +10,53 @@ import (
...
+
 	m := snap.HelmClient()
 
 	if !cfg.GetEnabled() {
 		if _, err := m.Apply(ctx, chartCalico, helm.StateDeleted, nil); err != nil {
-			return fmt.Errorf("failed to uninstall network: %w", err)
+			applyErr := fmt.Errorf("failed to uninstall network: %w", err)
+			status.Message = fmt.Sprintf(deleteFailedMsgTmpl, applyErr)
```
neoaggelos, line 37: much better to inline the messages here. when you then have an error message, it's easier to grep the code. also, avoid reusing the same error message in multiple places. e.g. this could be

```
return FeatureStatus{Message: fmt.Sprintf("failed to delete Calico: %v", err)}, err
```

src/k8s/pkg/k8sd/features/calico/network.go
```diff
@@ -10,28 +10,53 @@ import (
...
+			return status, applyErr
 		}
-		return nil
+		status.Version = ""
+		status.Message = disabledMsg
+		return status, nil
```
neoaggelos, line 42: return `FeatureStatus{Message: "disabled"}, nil`

src/k8s/pkg/k8sd/features/calico/network.go
```diff
@@ -10,28 +10,53 @@ import (
...
 	}
 
 	config, err := internalConfig(annotations)
 	if err != nil {
-		return fmt.Errorf("failed to parse annotations: %w", err)
+		cfgErr := fmt.Errorf("failed to parse annotations: %w", err)
+		status.Message = fmt.Sprintf(deployFailedMsgTmpl, cfgErr)
+		status.Enabled = false
+		return status, cfgErr
```
neoaggelos, line 50:
```
err = fmt.Errorf("invalid calico annotations: %w", err)
return FeatureStatus{Message: fmt.Sprintf("failed to parse Calico config: %v", err)}, err
```

neoaggelos (commented):
machinery looks good, please have a look at the comments about how to raise the errors

src/k8s/pkg/k8sd/features/calico/network.go
neoaggelos: Otherwise, it's a lot harder to track down exactly what status is returned

src/k8s/pkg/k8sd/features/calico/network.go
HomayoonAlimohammadi: I guess this and the following comments are a good amount of work (~0.5). will start changing them after the meetings.

tests/integration/tests/test_smoke.py
```diff
@@ -92,3 +94,41 @@ def test_smoke(instances: List[harness.Instance]):
...
+            )
+            line, pattern = result_lines[i], patterns[i]
+            if re.search(pattern, line) is not None:
+                break
+            LOG.info(f'Waiting for "{line}" to change...')
+            time.sleep(10)
+        else:
+            assert (
+                re.search(pattern, line) is not None
+            ), f'"Wait timed out. {pattern}" not found in "{line}"'
```
neoaggelos, line 134: this should use our `stubbornly().until()`

src/k8s/pkg/k8sd/features/metallb/loadbalancer.go
```diff
@@ -10,18 +10,62 @@ import (
 	"github.com/canonical/k8s/pkg/utils/control"
 )
 
-func ApplyLoadBalancer(ctx context.Context, snap snap.Snap, loadbalancer types.LoadBalancer, network types.Network, _ types.Annotations) error {
+const (
+	enabledMsgTmpl      = "enabled, %s mode"
+	disabledMsg         = "disabled"
+	deleteFailedMsgTmpl = "Failed to delete MetalLB, the error was: %v"
+	deployFailedMsgTmpl = "Failed to deploy MetalLB, the error was: %v"
```
neoaggelos, line 17: avoid mixing case, make all of these lowercase. also elsewhere.

also, these are fine to have a few times in the code below. reading `fmt.Sprintf(deleteFailedMsgTmpl, err)` does not make it easy to track where errors are coming from.

src/k8s/pkg/k8sd/features/contour/ingress.go
```diff
@@ -10,30 +10,57 @@ import (
 	"github.com/canonical/k8s/pkg/utils/control"
 )
 
+const (
+	ingressDeleteFailedMsgTmpl = "Failed to delete Contour Ingress, the error was: %v"
+	ingressDeployFailedMsgTmpl = "Failed to deploy Contour Ingress, the error was: %v"
```
neoaggelos, line 15: same here, and other features as well

src/k8s/pkg/k8sd/features/metallb/loadbalancer.go
HomayoonAlimohammadi: I was trying to achieve the exact details in the proposal:
https://github.com/canonical/k8s-snap/blob/main/docs/proposals/001-k8s-status.md#:~:text=%22Failed%20to%20deploy%20Cilium%2C%20the%20error%20was%20%25w%22

HomayoonAlimohammadi: about the number of repetitions, I feel like having things like `deleteFailedMsgTmpl` does not hurt tracking down errors. because if I duplicate the error message in a file, not only it makes it so that changing that error is a bit harder, and they might diverge (unwantedly), when you look up for error you're faced with more than one result. I think if does more harm than good.

tests/integration/tests/test_smoke.py
HomayoonAlimohammadi: https://github.com/canonical/k8s-snap/pull/571

PR #564, opened 2024-07-24
Show correct message on NodeStatus failure

HomayoonAlimohammadi:
Make sure we don't show the `you don't have a cluster, run k8s bootstrap` in wrong scenarios.


src/k8s/cmd/k8s/node_status.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/spf13/cobra"
+
+	apiv1 "github.com/canonical/k8s/api/v1"
+	cmdutil "github.com/canonical/k8s/cmd/util"
+	"github.com/canonical/k8s/pkg/client/k8sd"
+
+	"github.com/canonical/lxd/shared/api"
+)
+
+// GetNodeStatus retrieves the NodeStatus from k8sd client.
```
bschimke95, line 16: nit: it retrieves it from the server - the client just does the underlying http request.

Also, let's not overdo it with the line breaks. The comments can be at least the same length as the code.

src/k8s/pkg/k8sd/api/node.go
```diff
@@ -14,7 +14,7 @@ func (e *Endpoints) getNodeStatus(s *state.State, r *http.Request) response.Resp
 
 	status, err := impl.GetLocalNodeStatus(r.Context(), s, snap)
 	if err != nil {
-		response.InternalError(err)
+		return response.InternalError(err)
```
bschimke95, line 17: nice catch!

src/k8s/cmd/k8s/k8s_config.go
```diff
@@ -32,11 +32,7 @@ func newKubeConfigCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
 			}
 
-			if _, err := client.NodeStatus(cmd.Context()); err != nil {
-				cmd.PrintErrln("Error: The node is not part of a Kubernetes cluster. You can bootstrap a new cluster with:\n\n  sudo k8s bootstrap")
-				env.Exit(1)
-				return
-			}
+			_ = GetNodeStatus(client, cmd, env)
```
bschimke95, line 35: This implementation hides a lot of details and is not particular easy to read. 
You would need to check the implementation of `GetNodeStatus` to understand what is going on and why we ignore everything here.

I would:
a) change the `GetNodeStatus` function to return the status and an error
b) create a new function e.g. `isBootstrapped` that wraps around the `GetNodeStatus` call, handles the error decoding logic and returns a bool
c) Then you can do something like this in the commands:
```
if isBootstrapped(...) {
   cmd.Print("Node already bootstrapped/part of a cluster etc")
   cmd.Exit(1)
}
```

src/k8s/cmd/k8s/node_status.go
bschimke95, line 1: the files in `src/k8s/cmd/k8s` follow the format `k8s_<command>_<subcommand>.go`
That way a developer can easily see what commands the CLI supports.

Therefore, please move utils like that to `src/k8s/cmd/utils`.

bschimke95 (commented):
Great work @HomayoonAlimohammadi! 
Did a first pass

src/k8s/cmd/k8s/k8s_config.go
HomayoonAlimohammadi: thanks for the amazing comment @bschimke95 . I tried to address your concern. it didn't turn out to be exactly like you suggested but I guess it's close. returning 3 things does bother me a bit but I feel it's not really that bad. it reduces duplication, it's easily testable and more readable than the previous version. the name `isBootstrapped` does not seem to work here since we need to return the `status` and an error as well.

.gitignore
```diff
@@ -1,6 +1,8 @@
 /parts/
 /stage/
 /prime/
+.vscode/
```
addyess, line 4: don't folks normally put this in their global gitignore?

https://blog.martinhujer.cz/dont-put-idea-vscode-directories-to-projects-gitignore/

I don't really know how i land on this...

addyess (approved):
Nice.  thanks @HomayoonAlimohammadi

.gitignore
HomayoonAlimohammadi: TIL global git ignore :smile:

src/k8s/cmd/util/node_status.go
```diff
@@ -0,0 +1,37 @@
...
+	"errors"
+	"net/http"
+
+	apiv1 "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/k8s/pkg/client/k8sd"
+
+	"github.com/canonical/lxd/shared/api"
+)
+
+// GetNodeStatus retrieves the NodeStatus from k8sd client. If the daemon is not initialized, it exits with an error
```
bschimke95, line 14: nit: it still receives the node status from the server, not the client :)

.gitignore
```diff
@@ -20,4 +20,4 @@ k8s_*.txt
 /docs/tools/.sphinx/warnings.txt
 /docs/tools/.sphinx/.wordlist.dic
 /docs/tools/.sphinx/.doctrees/
-/docs/tools/.sphinx/node_modules
\ No newline at end of file
+/docs/tools/.sphinx/node_modules
```
addyess, line 23: so you've added a newline here -- i like files with newlines at the end
thanks for having that here...

bschimke95 (approved):
LGTM, only minor comment left.
Great stuff

src/k8s/cmd/util/node_status.go
HomayoonAlimohammadi: oops sorry for that one! :smile:

PR #565, opened 2024-07-24
Implement Worker Certificates Renewal Endpoints

mateoflorido:
## Overview
Implement the required endpoints for refreshing certificates for Kubernetes worker nodes.
## Rationale
As stated in [Proposal 002](https://github.com/canonical/k8s-snap/blob/main/docs/proposals/002-refresh-certificates.md). We need to provide two endpoints to allow certificates refreshes in the cluster. This pull request implements the `refresh-certs/plan` and the `refresh-certs/run` for refreshing the worker nodes' certificates.
## Testing
This pull request has been tested manually. Automated testing will be added in a future pull request.
```
root@kw-1:~# curl --unix-socket /var/snap/k8s/common/var/lib/k8sd/state/control.socket http://localhost/1.0/k8sd/refresh-certs/plan -XPOST
{"type":"sync","status":"Success","status_code":200,"operation":"","error_code":0,"error":"","metadata":{"seed":1978322463,"certificates_signing_requests":["k8sd-1978322463-worker-kubelet-serving","k8sd-1978322463-worker-kubelet-client","k8sd-1978322463-worker-kube-proxy-client"]}}
root@kw-1:~# curl --unix-socket /var/snap/k8s/common/var/lib/k8sd/state/control.socket http://localhost/1.0/k8sd/refresh-certs/run -XPOST -H "Content-Type: application/json" --data '{"seed":1978322463, "expiration_seconds": 10}'
{"type":"error","status":"","status_code":0,"operation":"","error_code":500,"error":"CSR k8sd-1978322463-worker-kubelet-serving is not issued","metadata":null}
root@kw-1:~# curl --unix-socket /var/snap/k8s/common/var/lib/k8sd/state/control.socket http://localhost/1.0/k8sd/refresh-certs/run -XPOST -H "Content-Type: application/json" --data '{"seed":1978322463, "expiration_seconds": 10}'
{"type":"sync","status":"Success","status_code":200,"operation":"","error_code":0,"error":"","metadata":null}
```


src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,277 @@
...
+	result := apiv1.RefreshCertificatesPlanResponse{
+		Seed: int(seed),
+		CertificatesSigningRequests: []string{
+			fmt.Sprintf("k8sd-%d-worker-kubelet-serving", seed),
+			fmt.Sprintf("k8sd-%d-worker-kubelet-client", seed),
+			fmt.Sprintf("k8sd-%d-worker-kube-proxy-client", seed),
+		},
+	}
+
+	err = os.WriteFile(filepath.Join(snap.KubernetesPKIDir(), "kubelet.key.new"), []byte(pKeyKubeletServing), 0600)
```
bschimke95, line 133: Why are we writing those keys here? I thought the `init/plan` step should be without any side-effects etc.

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,277 @@
...
+	certificates := pki.WorkerNodePKI{}
+
+	for _, csrName := range csrNames {
+		csr, err := client.GetCertificateSigningRequest(s.Context, csrName)
+		if err != nil {
+			return response.InternalError(err)
+		}
+
+		if !isCertificateApproved(csr.Status.Conditions) {
+			return response.InternalError(fmt.Errorf("CSR %s is not issued", csrName))
```
bschimke95, line 176: `s/issued/approved/`?
I'm always confused when a certificate is issued vs. approved :)

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,277 @@
...
+
+	certificates := pki.WorkerNodePKI{}
+
+	for _, csrName := range csrNames {
+		csr, err := client.GetCertificateSigningRequest(s.Context, csrName)
+		if err != nil {
+			return response.InternalError(err)
+		}
+
+		if !isCertificateApproved(csr.Status.Conditions) {
```
bschimke95, line 175: nit: you approve a CSR, not a certificate, no?

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,277 @@
...
+			return response.InternalError(fmt.Errorf("CSR %s is not issued", csrName))
+		}
+
+		if len(csr.Status.Certificate) == 0 {
+			return response.InternalError(fmt.Errorf("CSR %s is missing certificate", csrName))
+		}
+
+		_, _, err = pkiutil.LoadCertificate(string(csr.Status.Certificate), "")
```
bschimke95, line 183: nit:
```
if _,_,err := pkiutil.LoadCertificate(string(csr.Status.Certificate), ""); err != nil {
...
}
```

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,277 @@
...
+		return response.InternalError(err)
+	}
+
+	certificates.CACert = string(ca)
+	certificates.ClientCACert = string(clientCA)
+	certificates.KubeletKey = string(bytesKubeletKey)
+	certificates.KubeletClientKey = string(bytesKubeletClientKey)
+	certificates.KubeProxyClientKey = string(bytesKubeProxyKey)
+
+	_, err = setup.EnsureWorkerPKI(snap, &certificates)
```
bschimke95, line 233: nit: `if _, err =... ; err != nil`

src/k8s/pkg/utils/pki/generate.go
```diff
@@ -122,3 +122,36 @@ func GenerateRSAKey(bits int) (string, string, error) {
 
 	return string(privPEM), string(pubPEM), nil
 }
+
+func GenerateCSR(subject pkix.Name, bits int, priv any, dnsSANs []string, ipSANs []net.IP) (string, string, error) {
```
bschimke95, line 126: nit: add godoc for public function

bschimke95 (dismissed):
Amazing work @mateoflorido 

Added a couple of comments.

bschimke95:
accidentally approved - revoked for now

src/k8s/pkg/k8sd/api/certs_refresh.go
mateoflorido: Yes, if the CSR condition is approved, we know that the certificate has been approved using kubectl. Unfortunately, there's no condition for issuance, so the only way to verify if the certificate has been issued is to check the certificate field. I'll update the errors to make them clearer.

mateoflorido: More information: https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/certificate-signing-request-v1/#CertificateSigningRequestStatus

src/k8s/pkg/k8sd/api/certs_refresh.go
mateoflorido: To generate a new CSR, it is necessary to provide a private key. In this case, I chose an approach that creates new keys for the certificates and uses them, instead of reading the existing keys from the kubeconfigs and key files in the respective directory and refreshing all the pki in the node. Since the refresh process is divided into two subprocesses (plan and run), we need to store those keys in a temporary file (hence the .new suffix). This is to ensure that if something goes wrong during the process, we have a backup to keep things running smoothly.

src/k8s/pkg/k8sd/api/certs_refresh.go
bschimke95: I see, thanks for the context. 
I think from a usage perspective, something like this would probably be more readable:
```
func isCSRApproved(csr) bool
```
as it would hide the implementation details about when a CSR is approved.

src/k8s/pkg/k8sd/api/certs_refresh.go
bschimke95: Ah that was the context that I missed. Basically, they serve as a backup in case something breaks.

Could you not just do it the other way around in the `run` step:
1. Copy existing certs to `.old`
2. Create new certificates and overwrite the existing PKIs (no need to store in temporary files) 
3. restart services 

That way, if something goes south you can *restore* the old state, something you cannot do if you only store the new certs. If something goes wrong with those certs you are in troubles.

mateoflorido: Yes, we need to provide a way to back up the existing certificates and keys. However, it is still necessary to create temporary files to save the new signer keys. The process looks like this: create new private keys, sign the CSRs, and save the new private keys in a temporary location. Then, the `run` kicks in, and we wait until the certificates have been signed. Once we have all the certificates, we can replace everything. Changing the keys at this point could potentially block the cluster, as the private keys would not match the old certificates.

src/k8s/pkg/utils/file_operation.go
```diff
@@ -0,0 +1,73 @@
...
+	SourcePath  string
+	Content     []byte
+	Permissions os.FileMode
+}
+
+// BackupFiles backs up the files in the operations slice.
+func BackupFiles(operations []FileOperations) error {
+	for _, op := range operations {
+		if err := backupFile(op.SourcePath, op.BackupPath); err != nil {
+			return fmt.Errorf("backup failed: %w", err)
```
bschimke95, line 22: it probably makes sense to include the SourcePath and BackupPath in the error message

src/k8s/pkg/utils/file_operation.go
```diff
@@ -0,0 +1,73 @@
...
+// backupFile backs up the file at sourcePath to backupPath.
+func backupFile(sourcePath, backupPath string) error {
+	err := copyFile(sourcePath, backupPath)
+	if err != nil {
+		return fmt.Errorf("failed to backup file: %w", err)
+	}
+	return nil
+}
+
+// CopyFiles copies the files in the operations slice perserving the permissions.
```
bschimke95, line 47: nit: preserving

src/k8s/pkg/utils/file_operation.go
```diff
@@ -0,0 +1,73 @@
...
+		return fmt.Errorf("failed to get source file info: %w", err)
+	}
+
+	out, err := os.Create(destinationPath)
+	if err != nil {
+		return fmt.Errorf("failed to open destination file: %w", err)
+	}
+	defer out.Close()
+
+	os.Chmod(destinationPath, sourceInfo.Mode())
```
bschimke95, line 66: error not handled

src/k8s/pkg/k8sd/api/certs_refresh.go
bschimke95: Three things:

* Do we really want the `plan` step to be mandatory before `run`? In the proposal, the `plan` step is not mandatory and purely informational.
* What happens right now if we forget the `plan` step? I guess the `run` step would break with some cryptic `kubelet.key.new not found` error? IFF we really want to require the user to run `plan` before `run` (as said, I don't thing we should), then at least we need a proper error message to communicate this. 
* The third step in my comment above was wrong. Ofc we need to approve the certificates first, apply them and then we restart the services.

src/k8s/api/v1/certificates_refresh.go
```diff
@@ -0,0 +1,15 @@
+package apiv1
+
+type RefreshCertificatesPlanResponse struct {
+	Seed                        int      `json:"seed"`
+	CertificatesSigningRequests []string `json:"certificates_signing_requests"`
+}
+
+type RefreshCertificatesRunRequest struct {
+	Seed              int `json:"seed"`
+	ExpirationSeconds int `json:"expiration_seconds"`
```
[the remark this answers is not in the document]
mateoflorido: It's just the way the struct is specified in the [proposal](https://github.com/canonical/k8s-snap/blob/main/docs/proposals/002-refresh-certificates.md#post-refresh-certsrun)

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,343 @@
...
+		fmt.Sprintf("k8sd-%d-worker-kubelet-client", req.Seed),
+		fmt.Sprintf("k8sd-%d-worker-kube-proxy-client", req.Seed),
+	}
+
+	certificates := pki.WorkerNodePKI{}
+
+	operations := []utils.FileOperations{
+		{
+			SourcePath:  filepath.Join(snap.KubernetesPKIDir(), "kubelet.crt"),
+			BackupPath:  filepath.Join(snap.KubernetesPKIDir(), "kubelet.crt.bak"),
```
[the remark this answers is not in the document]
mateoflorido: Yes, `bak` is usually a dated way to refer to the [backup](https://en.wikipedia.org/wiki/Bak_file#:~:text=bak%22%20is%20a%20filename%20extension,bak%20appended%20to%20the%20filename.) of a file. It's interchangeable with `.old` or appending a `~`. If you think `.old` is a better description of this backup behavior, I can adjust that.

src/k8s/pkg/k8sd/api/certs_refresh.go
mateoflorido: 1. The `plan` step is [mandatory](https://github.com/canonical/k8s-snap/blob/main/docs/proposals/002-refresh-certificates.md#post-refresh-certsplan) (at least for workers) as this initializes everything required for the `run` endpoint.
> To accomodate this need, while also ensuring that the k8s refresh-certs can be straightforward to implement, two API calls are introduced, which clients should call in order.
...
POST /refresh-certs/plan performs any necessary preparations for certificate refreshes on the current node.

For example, the response we receive from the `plan` endpoint has a seed that we pass later on to the `run` endpoint, allowing us to determine if the CSRs have been approved or not.

2. Yeah, it should break. We can improve the error message to communicate to the user that probably the `plan` step is a prerequisite (again for workers).

src/k8s/api/v1/certificates_refresh.go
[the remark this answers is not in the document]
bschimke95: ok, this slipped through in the review. @neoaggelos is there a reason on why you went with `int`?

src/k8s/pkg/k8sd/api/certs_refresh.go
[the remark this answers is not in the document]
bschimke95: thanks!

src/k8s/pkg/k8sd/api/certs_refresh.go
bschimke95: alright, then let's provide a proper error message and move on.

src/k8s/pkg/client/kubernetes/csr.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func (c *Client) GetCertificateSigningRequest(ctx context.Context, name string) (*certificatesv1.CertificateSigningRequest, error) {
+	csr, err := c.CertificatesV1().CertificateSigningRequests().Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get certificate signing request %s: %w", name, err)
+	}
+
+	return csr, nil
+}
```
neoaggelos, line 37: it's ok to use these directly in the code instead of defining helpers around them. we do not add any logic around the underlying code, only hiding other fields (e.g. annotations or labels).

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil {
+		return response.InternalError(err)
+	}
+
+	var seed int32
+
+	err = binary.Read(rand.Reader, binary.BigEndian, &seed)
+	if err != nil {
+		return response.InternalError(err)
+	}
```
neoaggelos, line 56: you can just use `math/rand.Intn` here. This seed has no cryptographic value, it's just a random prefix for the names.

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,346 @@
...
+		r.Context(),
+		fmt.Sprintf("k8sd-%d-worker-kubelet-serving", seed),
+		[]byte(csrKubeletServing),
+		[]v1.KeyUsage{v1.UsageDigitalSignature, v1.UsageKeyEncipherment, v1.UsageServerAuth},
+		[]string{"system:nodes"},
+		"k8sd.io/kubelet-serving",
+	)
+	if err != nil {
+		return response.InternalError(err)
+	}
```
neoaggelos, line 83: helpers are supposed to not just wrap the underlying calls but "group common operations". this whole process is such a thing, `client.CreateCertificateSigningRequest` on its own is not useful.

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,346 @@
...
+	if err := snap.RestartService(r.Context(), "kube-proxy"); err != nil {
+		return response.InternalError(err)
+	}
+
+	// Remove the new private keys
+	log.Info("Removing temporal private keys")
+	if err := os.Remove(filepath.Join(snap.KubernetesPKIDir(), "kubelet.key.tmp")); err != nil {
+		return response.InternalError(err)
+	}
```
neoaggelos, line 321: think: in this instance, we have already signed the certificates, restarted the services, and we only fail to do internal cleanup. this is not an error for the user. also, this is not required at all, see my other comment. The CSRs are to be created during `/run`

src/k8s/pkg/utils/file_operation.go
```diff
@@ -0,0 +1,75 @@
+package utils
```
neoaggelos, line 1: this is not needed, see my other comment.

neoaggelos (changes requested):
Did a first pass. Good work overall, but see the requested changes

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,346 @@
...
+			SourcePath:  filepath.Join(snap.KubernetesPKIDir(), "kubelet-client.key.tmp"),
+			Content:     []byte(pKeyKubeletClient),
+			Permissions: 0600,
+		},
+		{
+			SourcePath:  filepath.Join(snap.KubernetesPKIDir(), "kube-proxy.key.tmp"),
+			Content:     []byte(pKeyKubeProxy),
+			Permissions: 0600,
+		},
+	}
```
neoaggelos, line 155: Note: `/plan` should only return the seed. The CSR creation should happen in `/run`, and private keys do not need to be shared by writing them on the filesystem.

mateoflorido: I'm confused. In the [proposal](https://github.com/canonical/k8s-snap/blob/main/docs/proposals/002-refresh-certificates.md#post-refresh-certsplan) I see that for the `plan` endpoint it should return the CSRs list for the worker nodes:

> It returns a seed that must be passed to POST /refresh-certs/run. For worker nodes, it also returns a list of names of CertificateSigningRequest objects that will need to be approved and signed. This can be used by the CLI to print a helpful message and request that the certificates be signed.

So I believe the CSR create should happen in `/plan` instead of `/run`. Am I missing something?

mateoflorido: OH, I see. This is the heart of the matter:

> This method call is purely informative and does not have any side-effect on the cluster.

So I think the meaning here:

> POST /refresh-certs/plan performs any necessary preparations for certificate refreshes on the current node.

is that these preparations are only to generate the seed and the names for the CSRs that will be created in the future during the `run` endpoint execution.

src/k8s/api/v1/certificates_refresh.go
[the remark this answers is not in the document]
bschimke95: thanks for the context

bschimke95 (approved):
LGTM, great work

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,254 @@
...
+			}
+		})
+
+	}
+
+	if err := g.Wait(); err != nil {
+		return response.InternalError(err)
+	}
+
+	log.Info("Ensuring worker PKI")
```
neoaggelos, line 210: log messages are the user's interface to see what is happening on k8sd. this message does not help them understand what is happening. either remove it, or increase verbosity `log.V(1).Info`, or adjust it in case there is useful information to convey

neoaggelos: same for other log messages

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,254 @@
...
+					request, err := client.CertificatesV1().CertificateSigningRequests().Get(errGroupCTX, csr.name, metav1.GetOptions{})
+					if err != nil {
+						log.Error(err, fmt.Sprintf("failed to get CSR %s", csr.name))
+						continue
+					}
+
+					approved, err := isCertificateSigningRequestApproved(request)
+					if err != nil {
+						log.Error(err, fmt.Sprintf("failed to check approval status for CSR %s", csr.name))
+						return fmt.Errorf("failed to check approval status for CSR %s: %w", csr.name, err)
```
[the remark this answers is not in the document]
mateoflorido: I have updated the error message. This is how it looks like when  a certificate has been denied:
```
root@t-worker-1:~# curl --unix-socket /var/snap/k8s/common/var/lib/k8sd/state/control.socket http://localhost/1.0/k8sd/refresh-certs/run -XPOST -H "Content-Type: application/json" --data '{"seed":4606628624274574580, "expiration_seconds": 10}'
{"type":"error","status":"","status_code":0,"operation":"","error_code":500,"error":"failed to generate worker CSRs: csr is in an invalid state: CSR k8sd-4606628624274574580-worker-kube-proxy-client was denied: KubectlDeny","metadata":null}
```

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	client, err := snap.KubernetesNodeClient("")
+	if err != nil {
+		return response.InternalError(err)
+	}
+
+	certificates := pki.WorkerNodePKI{}
+	// Read the CA and client CA
+	ca, err := os.ReadFile(filepath.Join(snap.KubernetesPKIDir(), "ca.crt"))
```
neoaggelos, line 85: these should come from the cluster config instead, not by reading the files

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	clientCA, err := os.ReadFile(filepath.Join(snap.KubernetesPKIDir(), "client-ca.crt"))
+	if err != nil {
+		return response.InternalError(err)
+	}
+
+	certificates.CACert = string(ca)
+	certificates.ClientCACert = string(clientCA)
+
+	g, errGroupCTX := errgroup.WithContext(r.Context())
```
neoaggelos, line 97: g, ctx

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,265 @@
...
+					}
+
+					approved, err := isCertificateSigningRequestApproved(request)
+					if err != nil {
+						return fmt.Errorf("csr is in an invalid state: %w", err)
+					}
+
+					if approved && isCertificateSigningRequestIssued(request) {
+						if _, _, err = pkiutil.LoadCertificate(string(request.Status.Certificate), ""); err != nil {
+							log.Error(err, fmt.Sprintf("failed to load certificate for CSR %s", csr.name))
```
neoaggelos, line 195: log.WithValues("csr", csr.name).Error(err, "CertificateSigningRequest failed")

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,263 @@
...
+				},
+			}, metav1.CreateOptions{}); err != nil {
+				return fmt.Errorf("failed to create CSR for %s: %w", csr.name, err)
+			}
+			watcher, err := client.CertificatesV1().CertificateSigningRequests().Watch(ctx, metav1.SingleObject(metav1.ObjectMeta{Name: csr.name}))
+			if err != nil {
+				return fmt.Errorf("failed to watch CSR %s: %w", csr.name, err)
+			}
+
+			defer watcher.Stop()
```
neoaggelos, line 169: be careful, as this might fail because of transient errors like an apiserver restarting. this will presently fail, perhaps we would like to retry the watch until the context is cancelled.

It's OK as is, this is just an infrequent error path that we would like to avoid

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,263 @@
...
+						return nil
+					}
+				}
+			}
+		})
+
+	}
+
+	if err := g.Wait(); err != nil {
+		return response.InternalError(fmt.Errorf("failed to generate worker CSRs: %w", err))
```
neoaggelos, line 206: The `g` group does not generate CSRs, it waits for certificates to be approved and signed. a better message would probably `"failed to get worker node certificates: %w"`. Then the wrapped error will contain the cause (e.g. failed to generate, or CSR was denied/failed).

To be more precise, consider what some of the errors would look like:
- failed to generate worker CSRs: failed to generate CSR for $name <-- this is OK
- failed to generate worker CSRs: watch closed for CSR <-- not as OK, how is the watch failure related to generating the CSR?

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,263 @@
...
+	// Kubeconfigs
+	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"), "127.0.0.1:6443", certificates.CACert, certificates.KubeletClientCert, certificates.KubeletClientKey); err != nil {
+		return response.InternalError(fmt.Errorf("failed to generate kubelet kubeconfig: %w", err))
+	}
+	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "proxy.conf"), "127.0.0.1:6443", certificates.CACert, certificates.KubeProxyClientCert, certificates.KubeProxyClientKey); err != nil {
+		return response.InternalError(fmt.Errorf("failed to generate kube-proxy kubeconfig: %w", err))
+	}
+
+	// Restart the services
+	log.Info("Restarting kubelet and kube-proxy")
```
neoaggelos, line 222: this would be more fitting as ` log.V(1).WithValues("service", svcName).Info("Restarting service")` inside the `snap.RestartService(...)` instead

src/k8s/pkg/utils/pki/generate.go
```diff
@@ -122,3 +122,37 @@ func GenerateRSAKey(bits int) (string, string, error) {
...
+		return "", "", fmt.Errorf("failed to generate RSA private key: %w", err)
+	}
+	keyPEM := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(key)})
+	if keyPEM == nil {
+		return "", "", fmt.Errorf("failed to encode private key PEM")
+	}
+
+	if priv == nil {
+		priv = key
+	}
```
neoaggelos, line 139: do we use priv at all here? we use `key` directly below, let's remove it?

neoaggelos (commented):
Final comments, great work @mateoflorido !

src/k8s/pkg/k8sd/api/certs_refresh.go
mateoflorido: Looks much better, let me update the message.

src/k8s/pkg/k8sd/api/certs_refresh.go
mateoflorido: Should I introduce a more robust mechanism, such as wait.Until or wait.ExponentialBackoff for this wait? Or is a big enclosing for loop enough? Also, is it required that we retry the watcher creation even inside the watcher loop?

neoaggelos: When a watcher fails, you need to create a new one. This is not necessarily a requirement, but would give some robustness to the code.

See what we do in WatchConfigMap for this.

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,279 @@
...
+	log := log.FromContext(r.Context())
+
+	req := apiv1.RefreshCertificatesRunRequest{}
+	if err := utils.NewStrictJSONDecoder(r.Body).Decode(&req); err != nil {
+		return response.BadRequest(fmt.Errorf("failed to parse request: %w", err))
+	}
+
+	client, err := snap.KubernetesNodeClient("")
+	if err != nil {
+		return response.InternalError(err)
```
bschimke95, line 81: please wrap this error with some context

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,279 @@
...
+		})
+
+	}
+
+	if err := g.Wait(); err != nil {
+		return response.InternalError(fmt.Errorf("failed to get worker node certificates: %w", err))
+	}
+
+	if _, err = setup.EnsureWorkerPKI(snap, &certificates); err != nil {
+		return response.InternalError(err)
```
bschimke95, line 227: same here

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,279 @@
...
+	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"), "127.0.0.1:6443", certificates.CACert, certificates.KubeletClientCert, certificates.KubeletClientKey); err != nil {
+		return response.InternalError(fmt.Errorf("failed to generate kubelet kubeconfig: %w", err))
+	}
+	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "proxy.conf"), "127.0.0.1:6443", certificates.CACert, certificates.KubeProxyClientCert, certificates.KubeProxyClientKey); err != nil {
+		return response.InternalError(fmt.Errorf("failed to generate kube-proxy kubeconfig: %w", err))
+	}
+
+	// Restart the services
+	if err := snap.RestartService(r.Context(), "kubelet"); err != nil {
+		return response.InternalError(err)
```
bschimke95, line 240: and here

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	if err := snap.RestartService(r.Context(), "kube-proxy"); err != nil {
+		return response.InternalError(err)
```
bschimke95, line 243: and here

src/k8s/pkg/k8sd/api/certs_refresh.go
bschimke95, line 1: we have `api/v1/certificates_refresh.go` and `pkg/k8sd/ap/certs_refresh.go`. Let's be consistent and use `certificates_refresh.go` for both file names.

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,279 @@
...
+		{
+			name:        fmt.Sprintf("k8sd-%d-worker-kube-proxy-client", req.Seed),
+			commonName:  "system:kube-proxy",
+			usages:      []v1.KeyUsage{v1.UsageDigitalSignature, v1.UsageKeyEncipherment, v1.UsageClientAuth},
+			signerName:  "k8sd.io/kube-proxy-client",
+			certificate: &certificates.KubeProxyClientCert,
+			key:         &certificates.KubeProxyClientKey,
+		},
+	} {
+		csr := csr
```
bschimke95, line 137: what is this assignment for?

src/k8s/pkg/k8sd/api/certs_refresh.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/canonical/k8s/pkg/log"
+	"github.com/canonical/k8s/pkg/snap"
+	snaputil "github.com/canonical/k8s/pkg/snap/util"
+	"github.com/canonical/k8s/pkg/utils"
+	pkiutil "github.com/canonical/k8s/pkg/utils/pki"
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/state"
+	"golang.org/x/sync/errgroup"
+	certificatesv1 "k8s.io/api/certificates/v1"
+	v1 "k8s.io/api/certificates/v1"
```
bschimke95, line 26: you import `v1` and `certifcatesv1`. Let's only use `certificatesv1`

src/k8s/pkg/k8sd/api/certs_refresh.go
mateoflorido: This is just for binding the current value of the csr variable to the closure inside the for loop. More information here: https://go.dev/doc/faq#closures_and_goroutines

bschimke95: Thanks!

bschimke95 (approved):
LGTM!

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -0,0 +1,281 @@
...
+								*csr.certificate = string(request.Status.Certificate)
+								*csr.key = string(keyPEM)
+								return nil
+							}
+						}
+					}
+					time.Sleep(3 * time.Second)
+				}
+
+			}
```
neoaggelos, line 218: this is a bit messy code, which can lead to undesired bugs down the line. a few notes:

- avoid `time.Sleep(3 * time.Second)` as that blocks even if the context cancels in the meantime. _if you do that_, go for `select` with `<-time.After(3*.time.Second)` 
- this would be written better as a helper function that recreates the watcher if it fails. as written, you are moving code in very deep indent depths. that's not a problem by itself, but you can reduce it. for example, you could do|

```go
select {
  case <-ctx.Done():
    // exit
  default:
    // code goes here
}
```

as

```go
select {
  case <-ctx.Done():
    // exit
  default:
}

// code goes here
```

neoaggelos (commented):
Final comments about the watcher restarts. It's OK to ignore the comment to start with and just a leave a NOTE _if_ it becomes a problem in the future, rather than having huge functions, as they come with very big cyclomatic complexity

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,46 @@
...
+				return fmt.Errorf("expected a CertificateSigningRequest but received %#v", evt.Object)
+			}
+
+			valid, err := verify(csr)
+			// If the verify function returns an error, we should return it
+			if err != nil {
+				return err
+			}
+
+			// If the verify function returns true, the CSR is valid and we can return
```
neoaggelos, line 39: NOTE: these comments better be part of the godoc of this function, so that the "contract" that `verify func(..) (bool, error)` must conform to is communicated.

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,46 @@
...
+			if valid {
+				return nil
+			}
```
neoaggelos, line 42:
```go
if valid, err := verify(csr); err != nil {
  return fmt.Errorf("useful wrapped error: %w", err)
} else if valid {
  return nil
}
```

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -27,6 +28,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var (
+	// errInvalidCSR is returned when the Kubernetes CSR is in an invalid state.
+	errInvalidCSR = errors.New("csr is in an invalid state")
```
neoaggelos, line 33: Use `fmt.Errorf()`

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -165,56 +171,29 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
...
+				if err == nil {
+					return nil
+				}
+
+				// Check if error is non-recoverable
+				if errors.Is(err, errInvalidCSR) {
+					return fmt.Errorf("certificate signing request failed: %w", err)
+				}
+
+				log.WithValues("name", "k8sd").Error(err, "Failed to watch CSR")
```
neoaggelos, line 190: add all contextual values once with `log = log.WithValues(....)` once, then just do `log.Info()` later.

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -165,56 +171,29 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
 			}
 
 			for {
+				err := client.WatchCertificateSigningRequest(
+					ctx,
+					csr.name,
+					func(request *certificatesv1.CertificateSigningRequest) (bool, error) {
+						return verifyCSRAndSetPKI(request, keyPEM, csr.certificate, csr.key)
+					})
```
neoaggelos, line 179:
```
  },
)
```

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -165,56 +171,29 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
...
+
+				if err == nil {
+					return nil
+				}
+
+				// Check if error is non-recoverable
+				if errors.Is(err, errInvalidCSR) {
+					return fmt.Errorf("certificate signing request failed: %w", err)
+				}
```
neoaggelos, line 188: better to use a `retry bool, err error` signature for the result, rather than communicate this with a "typed" error. there is currently nothing that prevents someone to change the signature of the error, and then this if-check breaks.

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -275,7 +256,23 @@ func isCertificateSigningRequestApproved(csr *certificatesv1.CertificateSigningR
...
 
-// isCertificateSigningRequestIssued checks if the certificate signing request is issued.
-func isCertificateSigningRequestIssued(csr *certificatesv1.CertificateSigningRequest) bool {
-	return len(csr.Status.Certificate) > 0
+// verifyCSRAndSetPKI verifies the certificate signing request and sets the
+// certificate and key if the CSR is approved.
+func verifyCSRAndSetPKI(csr *certificatesv1.CertificateSigningRequest, keyPEM string, certificate, key *string) (bool, error) {
+	approved, err := isCertificateSigningRequestApprovedAndIssued(csr)
+	if err != nil {
+		return false, fmt.Errorf("%w: failed to validate csr: %w", errInvalidCSR, err)
```
neoaggelos, line 264: avoid wrapping two errors, also see the other comment. return a bool that specifies whether we are treating this as a retriable error or not.

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"fmt"
+
+	certificatesv1 "k8s.io/api/certificates/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// WatchCertificateSigningRequest watches a CertificateSigningRequest with the given name and calls the verify function on each event.
+func (c *Client) WatchCertificateSigningRequest(ctx context.Context, name string, verify func(csr *certificatesv1.CertificateSigningRequest) (bool, error)) error {
```
neoaggelos, line 12: Make the semantics of how CSRs are treated part of the godoc of this function. Make it clear what we expect out of `verify`, make it clear _when_ this function terminates (e.g. it terminates when the CSR is approved or failed, etc).

The thing to avoid here is: in the future, someone wants to check a CSR, sees a `kubernetes.WatchCertificateSigningRequest`, but the semantics of the function are not clear.

This is important, as the name `WatchCertificateSigningRequest` does not communicate that the watch actually stops when the verify function returns `true, nil`. That should be documented clearly. Perhaps, the function itself should return `(bool, error)` as well, or maybe the final CSR object.

neoaggelos: In general, avoid "implicit" API contracts, be very explicit with what the function expects and returns.

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,53 @@
...
+				return true, fmt.Errorf("watch closed")
+			}
+
+			csr, ok := evt.Object.(*certificatesv1.CertificateSigningRequest)
+			if !ok {
+				return true, fmt.Errorf("expected a CertificateSigningRequest but received %#v", evt.Object)
+			}
+
+			valid, err := verify(csr)
+			if err != nil {
```
neoaggelos, line 45: inline

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -0,0 +1,273 @@
...
+	if isWorker {
+		return response.SyncResponse(true, apiv1.RefreshCertificatesPlanResponse{
+			Seed: seed,
+			CertificatesSigningRequests: []string{
+				fmt.Sprintf("k8sd-%d-worker-kubelet-serving", seed),
+				fmt.Sprintf("k8sd-%d-worker-kubelet-client", seed),
+				fmt.Sprintf("k8sd-%d-worker-kube-proxy-client", seed),
+			},
+		},
+		)
```
neoaggelos, line 47: here, you would do `})` since it's in the same indentation level

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -0,0 +1,273 @@
...
+	if err := utils.NewStrictJSONDecoder(r.Body).Decode(&req); err != nil {
+		return response.BadRequest(fmt.Errorf("failed to parse request: %w", err))
+	}
+
+	client, err := snap.KubernetesNodeClient("")
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to get Kubernetes client: %w", err))
+	}
+
+	certificates := pki.WorkerNodePKI{}
```
neoaggelos, line 83:
```go
var certificates pki.WorkerNodePKI
```

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to get cluster configuration: %w", err))
+	}
+
+	if clusterConfig.Certificates.CACert == nil || clusterConfig.Certificates.ClientCACert == nil {
+		return response.InternalError(fmt.Errorf("missing CA certificates"))
+	}
+
+	certificates.CACert = *clusterConfig.Certificates.CACert
+	certificates.ClientCACert = *clusterConfig.Certificates.ClientCACert
```
neoaggelos, line 95: use `GetCACert()` and `GetClientCACert()`

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -0,0 +1,273 @@
...
+				)
+
+				if err == nil {
+					return nil
+				}
+
+				// Check if error is non-recoverable
+				if retry == false {
+					return fmt.Errorf("certificate signing request failed: %w", err)
+				}
```
neoaggelos, line 183: also here

```go
if retry, err := ...; err == nil {
  return nil
} else if !retry {
  return fmt.Errorf("...")
}
```

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -0,0 +1,273 @@
...
+
+				log.Error(err, "Failed to watch CSR")
```
neoaggelos, line 185: should this log message be before returning on a non-retriable error?

src/k8s/pkg/snap/snap.go
```diff
@@ -50,16 +51,19 @@ func NewSnap(opts SnapOpts) *snap {
 
 // StartService starts a k8s service. The name can be either prefixed or not.
 func (s *snap) StartService(ctx context.Context, name string) error {
+	log.FromContext(ctx).WithCallDepth(1).Info("Starting service", "service", name)
```
neoaggelos, line 54: Should this be a `.V(1)`  ?

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -0,0 +1,268 @@
...
+					return fmt.Errorf("certificate signing request failed: %w", err)
+				}
+
+				log.V(1).Info("Retrying to watch CSR")
+
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case <-time.After(3 * time.Second):
+				}
```
neoaggelos, line 186: i wonder if it would be best to capture this retry loop inside `client.WatchCertificateSigningRequest`. That way the `(retry bool, err error)` happens once.

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// WatchCertificateSigningRequest watches a CertificateSigningRequest with the
+// given name and calls a verify function on each event.
+// WatchCertificateSigningRequest will continue watching the CSR until the
+// verify function returns true or an non-retriable error occurs.
+// WatchCertificateSigningRequest will return true and a wrapped error if the
+// error is retriable.
+// WatchCertificateSigningRequest will return false and a wrapped error if the
+// error is not retriable.
```
neoaggelos, line 18: This does not help if I read this out of context. It would be easier if we included the retry loop in case the watcher fails, then this godoc can say: we keep repeating as long as verify returns false or "retriable" errors. Because right now the responsibility of retrying the watcher is split across two very far places in the code.

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,68 @@
...
+// (e.g., failed or denied) or the issued certificate is invalid.
+func (c *Client) WatchCertificateSigningRequest(ctx context.Context, name string, verify func(csr *certificatesv1.CertificateSigningRequest) (bool, error)) error {
+	log := log.FromContext(ctx)
+	for {
+		w, err := c.CertificatesV1().CertificateSigningRequests().Watch(ctx, metav1.SingleObject(metav1.ObjectMeta{Name: name}))
+		if err != nil {
+			log.V(1).Info("Failed to watch CSR", "error", err)
+			continue
+		}
+		defer w.Stop()
```
neoaggelos, line 32: any deferred `w.Stop()` here will only run at the end of the loop. it should probably be done at the end of the loop.

To keep the `defer`, you can split the block of each `for { }` loop in a `somethingInternalWatchCertificateSigningRequest`

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,68 @@
...
+		watchClosed := false
```
neoaggelos, line 33: having an internal function that implements the loop body will also help you with the `watchClosed`, that would then be able to go away

mateoflorido: I think offloading the loop body to another function will result in a function that returns a `(bool, error)` since we need to distinguish between retryable and non-retryable errors.

Is this acceptable for an internal function, or would you suggest another approach for this refactor?

neoaggelos: if this is internal, and callers do not need to concern themselves with the details, that is acceptable

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		if retry, err := c.watchCertificateSigningRequestEvents(ctx, w, name, verify); err != nil {
+			return fmt.Errorf("failed to watch CSR %s: %w", name, err)
+		} else if !retry {
+			return nil
+		}
+
+		w.Stop()
+		log.V(1).Info("Retrying to watch CSR")
```
neoaggelos, line 41: I do not think this log line is useful in the grand scheme of things here. If we failed, we will print the error and retry. This line, without even mentioning the CSR name, is not useful and will be just noise.

Think: in a worker node where we have run the refresh-certs command, this will just print `Retrying to watch CSR` 3 times (since we have 3 CSRs) in the logs, without much context about which CSR, why, etc.

Remember that logs are an interface for someone _not_ familiar with the code to understand what is happening. Just printing messages like this will mostly lead to goose chases. Let's only log when we want to communicate something important to the end user.

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,79 @@
...
+// The verify function should return false if the CSR is not yet valid and
+// processing should continue.
+// The verify function should return an error if the CSR is in an invalid state
+// (e.g., failed or denied) or the issued certificate is invalid.
+func (c *Client) WatchCertificateSigningRequest(ctx context.Context, name string, verify func(csr *certificatesv1.CertificateSigningRequest) (bool, error)) error {
+	log := log.FromContext(ctx)
+	for {
+		w, err := c.CertificatesV1().CertificateSigningRequests().Watch(ctx, metav1.SingleObject(metav1.ObjectMeta{Name: name}))
+		if err != nil {
+			log.V(1).Info("Failed to watch CSR", "error", err)
```
neoaggelos, line 30: `log.Error(err, "Failed to watch CSR")`

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,79 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(3 * time.Second):
+		}
+	}
+}
+
+func (c *Client) watchCertificateSigningRequestEvents(ctx context.Context, w watch.Interface, name string, verify func(csr *certificatesv1.CertificateSigningRequest) (bool, error)) (bool, error) {
+	log := log.FromContext(ctx)
```
neoaggelos, line 52: this helper should be creating the watch.Interface, then you can simply defer `w.Stop`

neoaggelos: because right now you are handling the case where the watch failed in two places (when creating the watch interface, and then here)

neoaggelos: this is important in terms of "ownership" of the watch interface. we do not want to share it with something else, so it's best to fully encapsulate its lifetime within this helper function.

src/k8s/pkg/client/kubernetes/certificate_signing_request.go
```diff
@@ -0,0 +1,76 @@
+package kubernetes
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/canonical/k8s/pkg/log"
+	certificatesv1 "k8s.io/api/certificates/v1"
```
neoaggelos, line 9: commonly imported as `certv1`

neoaggelos (approved):

PR #566, opened 2024-07-26
Disable cilium e2e on Github Actions

bschimke95:
The cilium e2e test is known to be flaky on the Github Actions runners.
We know that Cilium works in the snap and this test causes more trouble then it is useful for now.


berkayoz (approved):
LGTM

tests/integration/tests/test_cilium_e2e.py
```diff
@@ -19,6 +20,10 @@
 @pytest.mark.skipif(
     ARCH not in CILIUM_CLI_ARCH_MAP, reason=f"Platform {ARCH} not supported"
 )
+@pytest.mark.skipif(
+    os.getenv("GITHUB_ACTIONS") == "true",
+    reason="Test is known to be flaky on GitHub Actions",
+)
```
neoaggelos, line 26: Prefer using a mark like `RUN_CILIUM_E2E_TEST`, which is off default, instead of relying on _current_ status of CI infrastructure.

PR #570, opened 2024-07-30
Provide pytest marker to set custom bootstrap config

bschimke95:
# Summary

Simplify integration tests by adding new pytest markers for `dualstack` and custom `bootstrap_config`s

# Rationale

We have a bunch of tests that require custom bootstrap configuration (e.g. dualstack). The current approach is to disable the automatic bootstrapping of the first `instance` and instead do this setup manually in each test. Those setup steps only differ in the path of the bootstrap config, so we can simply pass this to the marker.

# Comments

Note that we still need the `disable_k8s_bootstrapping` marker for the `test_etcd` as this test requires the etcd cluster certs and needs to generate the bootstrap config dynamically.

### DIFF
--- a/tests/integration/tests/conftest.py
+++ b/tests/integration/tests/conftest.py
@@ -53,7 +53,7 @@ def pytest_configure(config):
         "disable_k8s_bootstrapping: By default, the first k8s node is bootstrapped. This marker disables that.\n"
         "dualstack: Support dualstack on the instances.\n"
         "etcd_count: Mark a test to specify how many etcd instance nodes need to be created (None by default)\n"
-        "node_count: Mark a test to specify how many instance nodes need to be created\n"
+        "node_count: Mark a test to specify how many instance nodes need to be created\n",
     )
 
 
--- a/tests/integration/tests/conftest.py
+++ b/tests/integration/tests/conftest.py
@@ -3,7 +3,7 @@
 #
 import logging
 from pathlib import Path
-from typing import Generator, List
+from typing import Generator, List, Union
 
 import pytest
 from test_util import config, harness, util
--- a/tests/integration/tests/conftest.py
+++ b/tests/integration/tests/conftest.py
@@ -91,7 +91,7 @@ def instances(
     node_count: int,
     tmp_path: Path,
     disable_k8s_bootstrapping: bool,
-    bootstrap_config: Path | None,
+    bootstrap_config: str | None,
     dualstack: bool,
 ) -> Generator[List[harness.Instance], None, None]:
     """Construct instances for a cluster.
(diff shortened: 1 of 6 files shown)


tests/integration/tests/test_util/util.py
```diff
@@ -139,7 +139,7 @@ def setup_k8s_snap(instance: harness.Instance, snap_path: Path):
 def wait_until_k8s_ready(
     control_node: harness.Instance,
     instances: List[harness.Instance],
-    retries: int = 15,
+    retries: int = 30,
```
bschimke95, line 142: I've ran into timeouts on my local machine. Let's increase the defaults here.

tests/integration/tests/conftest.py
```diff
@@ -65,13 +67,32 @@ def node_count(request) -> int:
 
 
 @pytest.fixture(scope="function")
-def disable_k8s_bootstrapping(request) -> int:
+def disable_k8s_bootstrapping(request) -> bool:
```
mateoflorido, line 70: 😄 Nice catch

berkayoz (approved):
I think we have linting issues but other than that LGTM

mateoflorido (approved):
LGTM

tests/integration/tests/conftest.py
bschimke95: to be fair, I also introduced it... :D

tests/integration/tests/conftest.py
```diff
@@ -65,13 +67,32 @@ def node_count(request) -> int:
...
     return bool(request.node.get_closest_marker("disable_k8s_bootstrapping"))
 
 
+@pytest.fixture(scope="function")
+def bootstrap_config(request) -> Path | None:
```
mateoflorido, line 75: I didn't notice that we are using `py3.8`. Instead, we can use:
```suggestion
def bootstrap_config(request) -> Union[Path, None]:
```

tests/integration/tests/test_control_plane_taints.py
```diff
@@ -6,28 +6,19 @@
...
-from test_util import harness, util
+from test_util import config, harness, util
 
 LOG = logging.getLogger(__name__)
 
 
 @pytest.mark.node_count(1)
-@pytest.mark.disable_k8s_bootstrapping()
+@pytest.mark.bootstrap_config(
+    config.MANIFESTS_DIR / "bootstrap-control-plane-taints.yaml"
```
neoaggelos, line 16: maybe worth making this not a path, but the contents of the config itself. then callers can do `(config.MANIFESTS_DIR / "bootstrap-control-plane-taints.yaml").read_text()` or specify the config directly

bschimke95: good point, updated

PR #577, opened 2024-07-31
Add Config Option to Control Kubernetes Node Removal

bschimke95:
## Summary

This pull request introduces a new cluster config annotation `k8sd/v1alpha/lifecycle/skip-cleanup-kubernetes-node-on-remove`. This annotations allows the user to disable the automatic removal of Kubernetes nodes during the cleanup process. When the annotation is set, only the cleanup of files and the microcluster is performed, leaving the actual Kubernetes node intact.

## Rationale

In certain scenarios, external controllers are responsible for managing the lifecycle of Kubernetes nodes. For example, in our case, the Cluster API (CAPI) handles the node lifecycle. To avoid conflicts and ensure smooth operation, we need a mechanism to bypass the automatic removal of Kubernetes nodes and only perform the necessary cleanup of additional configurations and layers we have applied. This new annotation provides the flexibility to accommodate such scenarios without interfering with external controllers. That external controllers handle the lifecycle is not the default scenario and more of an edge-case. Hence, the default behavior is to remove the node and only opt-in via the annotation.

## Changes

* Introduce a new annotation `k8sd/v1alpha/lifecycle/skip-cleanup-kubernetes-node-on-remove`
* Update the node removal logic to check for existence of this annotation. If not set, proceed with the removal of the Kubernetes node. If set, skip the node removal step and only perform the cleanup of files and microcluster configurations.

## Comment

I'm **highly** open for different naming of this option! Naming is hard...

Related CAPI PR: https://github.com/canonical/cluster-api-k8s/pull/32


src/k8s/pkg/k8sd/types/cluster_config_convert.go
```diff
@@ -60,6 +60,9 @@ func ClusterConfigFromBootstrapConfig(b apiv1.BootstrapConfig) (ClusterConfig, e
 		return ClusterConfig{}, fmt.Errorf("unknown datastore type specified in bootstrap config %q", b.GetDatastoreType())
 	}
 
+	// K8sd
+	config.K8sd.ShouldRemoveK8sNode = b.ShouldRemoveK8sNode
```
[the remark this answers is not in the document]
bschimke95: moved to annotations as discussed offline.

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -122,5 +113,21 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 		}
 	}
 
+	if _, ok := cfg.Annotations[apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove]; ok {
+		// Explicitly skip removing the node from Kubernetes.
+		log.Info("Skipping Kubernetes control-plane node removal")
```
neoaggelos, line 118: maybe make `log.WithValues("node", s.Name())` earlier when you create the logger

tests/integration/tests/test_clustering.py
```diff
@@ -64,3 +64,33 @@ def test_worker_nodes(instances: List[harness.Instance]):
...
+    assert "control-plane" in util.get_local_node_status(cluster_node)
+    assert "control-plane" in util.get_local_node_status(joining_cp)
+    assert "worker" in util.get_local_node_status(joining_worker)
+
+    nodes = util.ready_nodes(cluster_node)
+
+    cluster_node.exec(["k8s", "remove-node", joining_cp.id])
+    assert len(nodes) == 3, "cp node should not have been removed from cluster"
+    cluster_node.exec(["k8s", "remove-node", joining_worker.id])
+    assert len(nodes) == 3, "worker node should not have been removed from cluster"
```
neoaggelos, line 96: be extra careful about how much extra time a test like this takes

bschimke95: I think it makes sense to have it but maybe we can only run it as part of the nightly tests.

bschimke95: Let's keep it like this for now and eventually move this to a nightly run.

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -45,6 +46,18 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 
 	cfg, clusterConfigErr := databaseutil.GetClusterConfig(ctx, s)
 	if clusterConfigErr == nil {
+		if _, ok := cfg.Annotations[apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove]; !ok {
```
neoaggelos, line 49: `cfg, err`

neoaggelos (approved):
final nits

PR #585, opened 2024-08-06
Encrypt worker CSRs

bschimke95:
Encrypt the CSRs signature of workers with the k8sd public key. This ensures that the CSR really comes from the worker and not from a malicious actor.

Unittest is implemted in https://github.com/canonical/k8s-snap/pull/586


eaudetcobello:
Can you make sure a card is created for the follow-up unit test?

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -151,9 +154,29 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
...
+			hash := sha256.New()
+			if _, err := hash.Write([]byte(csrPEM)); err != nil {
+				return fmt.Errorf("failed to compute sha256: %w", err)
+			}
+
+			// encrypt the hash with the public cluster RSA key
+			signature, err := rsa.EncryptPKCS1v15(rand.Reader, pubKey, hash.Sum(nil))
+			if err != nil {
+				return fmt.Errorf("failed to encrypt csr signature: %w", err)
+			}
```
eaudetcobello, line 172: Move this to it's own function that returns the signature.

eaudetcobello (changes requested):

src/k8s/pkg/k8sd/api/certificates_refresh.go
bschimke95: Why would you move that into a separate function - what value do you see in doing that?

src/k8s/pkg/k8sd/controllers/csrsigning/reconcile_approve.go
```diff
@@ -5,14 +5,28 @@ import (
 	"fmt"
 
 	"github.com/canonical/k8s/pkg/log"
+	pkiutil "github.com/canonical/k8s/pkg/utils/pki"
 	certv1 "k8s.io/api/certificates/v1"
 	v1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 func (r *csrSigningReconciler) reconcileAutoApprove(ctx context.Context, log log.Logger, csr *certv1.CertificateSigningRequest) (ctrl.Result, error) {
```
neoaggelos, line 14: Accept the private key as an argument, we have already fetched the cluster config once in the reconcile loop already https://github.com/canonical/k8s-snap/blob/2994b879eab39b782556f122a2c8512bf85e9aca/src/k8s/pkg/k8sd/controllers/csrsigning/reconcile.go#L56-L69

neoaggelos: Also, consider whether we need to account for a nil key or not.

neoaggelos (changes requested):

src/k8s/pkg/k8sd/api/certificates_refresh.go
eaudetcobello: Thanks for chiming in @neoaggelos. @bschimke95, I see merit in modularizing the code but I trust you to weigh the tradeoffs and implement the solution you think is best.

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -154,7 +154,12 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
 				return fmt.Errorf("failed to generate CSR for %s: %w", csr.name, err)
 			}
 
-			pubKey, err := pkiutil.LoadRSAPublicKey(clusterConfig.Certificates.GetK8sdPublicKey())
+			k8sdPublicKey := clusterConfig.Certificates.GetK8sdPublicKey()
+			if k8sdPublicKey == "" {
+				return fmt.Errorf("k8sd public key not set")
+			}
+
+			pubKey, err := pkiutil.LoadRSAPublicKey(k8sdPublicKey)
```
neoaggelos, line 162: NOTE on style of `k8sdPublicKey` and `pubKey`: Naming wise, they appear to be entities of the same thing, but one of them is the key, the other is a PEM encoded string. 

i'd recommend having the variable name _describe_ what it contains. try `publicKeyPEM` and `publicKey` instead. That makes it easier to follow the rest of the code.

src/k8s/pkg/k8sd/controllers/csrsigning/reconcile.go
```diff
@@ -65,7 +65,7 @@ func (r *csrSigningReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		log.Info("CSR is not approved")
 		if internal.autoApprove {
 			log.V(1).Info("CSR auto-approval is enabled")
-			return r.reconcileAutoApprove(ctx, log, obj)
+			return r.reconcileAutoApprove(ctx, log, obj, config)
```
neoaggelos, line 68: if we don't need all of the config, you should only pass the key we need there.

src/k8s/pkg/k8sd/controllers/csrsigning/validate.go
```diff
@@ -9,19 +11,27 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-func validateCSR(obj *certv1.CertificateSigningRequest) error {
+func validateCSR(obj *certv1.CertificateSigningRequest, priv *rsa.PrivateKey) error {
```
neoaggelos, line 14: if you assume that `priv` is always a valid key here, document it

neoaggelos (commented):
Overall LGTM, final comments

src/k8s/pkg/k8sd/controllers/csrsigning/validate.go
```diff
@@ -9,19 +11,27 @@ import (
...
-	// signature = RSA_DECRYPT(k8sdPrivateKey, encryptedSignature)
-	// hash = SHA256(obj.Spec.Request)
-	// assert hash == signature
+	// calculate sha256 sum of CSR request
+	h := sha256.New()
+	if _, err := h.Write(obj.Spec.Request); err != nil {
+		return fmt.Errorf("failed to compute sha256: %w", err)
+	}
+
+	if !(string(h.Sum(nil)) == string(signature)) {
```
neoaggelos, line 32: this is written strangely, you can try `subtle.ConstantTimeCompare()` or just `A != B`

eaudetcobello (approved):

src/k8s/pkg/k8sd/controllers/csrsigning/validate.go
```diff
@@ -9,19 +12,30 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-func validateCSR(obj *certv1.CertificateSigningRequest) error {
+// validateCSR checks the CSR for common requirements and returns an error if it fails.
+// validateCSR is used to validate CSRs that are auto-approved.
```
neoaggelos, line 16: better to describe what the function _does_, not what are callers will use it for.

neoaggelos (approved):

src/k8s/pkg/k8sd/controllers/csrsigning/validate.go
bschimke95: fair point - fixed

PR #580, opened 2024-08-05
Update dqlite and deps

louiseschmidtgen:
## Description
This PR updates dqlite and it's dependencies, go-dqlite and sqlite.

### Dqlite updates:
Dqlite [1.16.7](https://github.com/canonical/dqlite/releases/tag/v1.16.7) has just been released (and contains a fix for a segfault which affects other PRs).
The latest dqlite is using multiple threads to serve queries.

### Dqlite dependecies
For the dqlite dependencies we choose to be in sync with the libraries of the Ubuntu Noble LTS where possible.
Example: musl `v1.2.4` ([launchpad](https://launchpad.net/ubuntu/noble/amd64/musl/1.2.4-2))

### K8s-dqlite is updated
The same updates have already been merged in k8s-dqlite: https://github.com/canonical/k8s-dqlite/pull/151 and https://github.com/canonical/k8s-dqlite/pull/153

### Arm64
The Juju team discovered segfault issues building dqlite on arm64 machines, this flag addresses the issues: `-Wl,-z,stack-size=1048576`
See: https://wiki.musl-libc.org/functional-differences-from-glibc.html#Thread-stack-size

PR #587, opened 2024-08-07
Update the microcluster version to v2

bschimke95:
V2 is an LTS so we should be good with this for a while. Interfaces where updated if needed.

Note that microcluster now supports token expiration. I hardcoded the TTL for now and created a follow-up card to properly address this later.


src/k8s/go.mod
```diff
@@ -3,9 +3,10 @@ module github.com/canonical/k8s
 go 1.22.5
 
 require (
-	github.com/canonical/go-dqlite v1.21.0
-	github.com/canonical/lxd v0.0.0-20240709081608-9df996e36bac
-	github.com/canonical/microcluster v0.0.0-20240725000112-da36ecbd845e
+	github.com/canonical/go-dqlite v1.22.0
```
neoaggelos, line 6: Perhaps wait for https://github.com/canonical/k8s-snap/pull/580 first

src/k8s/go.mod
```diff
@@ -3,9 +3,10 @@ module github.com/canonical/k8s
...
+	github.com/canonical/lxd v0.0.0-20240730172021-8e39e5d4f55f
+	github.com/canonical/microcluster v0.0.0-20240731145923-49573627997a
```
neoaggelos, line 8: this should go away, make sure to use `/v2` everywhere

src/k8s/pkg/k8sd/api/cluster_tokens.go
```diff
@@ -54,7 +55,8 @@ func getOrCreateJoinToken(ctx context.Context, m *microcluster.MicroCluster, tok
 	}
 
 	// if token does not exist, create a new one
-	token, err := m.NewJoinToken(ctx, tokenName)
+	// TODO(ben): make token expiry configurable
+	token, err := m.NewJoinToken(ctx, tokenName, time.Duration(24*time.Hour))
```
neoaggelos, line 59: 24 * time.Hour should already be a duration

src/k8s/go.mod
neoaggelos: e.g. in places like https://github.com/canonical/k8s-snap/blob/032a5836d01450b7db612f83ddcb0fb9bf34c4f0/src/k8s/pkg/k8sd/database/feature_status.go#L11

You could use `| grep github.com/canonical/microcluster` to find instances where the `/v2` is not added in the imports

neoaggelos (changes requested):

src/k8s/go.mod
```diff
@@ -3,9 +3,9 @@ module github.com/canonical/k8s
 go 1.22.5
 
 require (
-	github.com/canonical/go-dqlite v1.21.0
-	github.com/canonical/lxd v0.0.0-20240709081608-9df996e36bac
-	github.com/canonical/microcluster v0.0.0-20240725000112-da36ecbd845e
+	github.com/canonical/go-dqlite v1.22.0
+	github.com/canonical/lxd v0.0.0-20240730172021-8e39e5d4f55f
```
neoaggelos, line 7: it's a bit weird that an LTS microcluster relies on an unversioned lxd commit?

bschimke95: According to the microcluster team:
```
LXD itself doesn't use a Go compatible tag versions, because history, and also because if we were to switch to using it them (which we considered and tried just before the last LTS) we'd have to bump the Go mod version to match that would be pretty disruptive as all Go imports would need to change.

For the client and api packages these could potentially be moved into a separate repo, this too we have considered, and not ruled out, but that too would be pretty disruptive and we didn't want to make any more large changes when the license, version numbering scheme and organisation umbrella were changing at the same time :slightly_smiling_face:

One thing that is also worth keeping in mind with the LXD client and api package is that we try and keep even our LTS branches in sync with main because we expect older clients to be able to talk to newer servers and vice versa, so try pretty hard to not break compatibility at that level, and introduce new features via API extensions and new fields.

So although its an unversioned commit, that is effectively the current released version. 

[1m ago](https://chat.canonical.com/canonical/pl/hn8zjdap6pnxzyg9smxat1seto)
In the future if we split the client and api packages into their own repo I would expect us to use a v1 go mod version and then we can do release tags, but I still doubt we would have a separate LTS series for that repo.
```

src/k8s/go.mod
bschimke95: Done and rebased

neoaggelos (approved):

PR #588, opened 2024-08-07
Fix DeleteNode log

HomayoonAlimohammadi:
Fix a minor issue in the DeleteNode log

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -53,7 +53,7 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 
 			log.Info("Deleting node from Kubernetes cluster")
 			if err := c.DeleteNode(ctx, s.Name()); err != nil {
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -53,7 +53,7 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 
 			log.Info("Deleting node from Kubernetes cluster")
 			if err := c.DeleteNode(ctx, s.Name()); err != nil {
-				log.Error(err, "Failed to remove k8s node %q: %w", s.Name(), err)
+				log.Error(err, fmt.Sprintf("Failed to remove k8s node %q: %v", s.Name(), err))
```
neoaggelos, line 56: `log.Error()` already adds the error. and from line 27 we already have the node name as well in the logger. So this should be `log.Error(err, "Failed to remove Kubernetes node")`

neoaggelos (approved):

neoaggelos (approved):

PR #593, opened 2024-08-08
Generate inspection report on CI failures

bschimke95:
This PR adds steps to the CI to create and collect the inspection report if the integration tests fail. The report is uploaded as an artifact of the test run.

### DIFF
--- a/.github/workflows/integration-informing.yaml
+++ b/.github/workflows/integration-informing.yaml
@@ -80,27 +80,20 @@ jobs:
       - name: Apply ${{ matrix.patch }} patch
         run: |
           ./build-scripts/patches/${{ matrix.patch }}/apply
-      - name: Generate inspection report
-        run: |
-          sudo k8s/scripts/inspect.sh
-      - name: Upload inspection report artifact
-        uses: actions/upload-artifact@v4
-        with:
-          name: inspection-report
-          path: ./inspection-report-*.tar.gz
       - name: Run end to end tests
         run: |
           export TEST_SNAP="$PWD/build/k8s-${{ matrix.patch }}.snap"
           export TEST_SUBSTRATE=lxd
           export TEST_LXD_IMAGE=${{ matrix.os }}
           cd tests/integration && sg lxd -c 'tox -e integration'
-      - name: Generate inspection report
+      - name: Compress inspection reports
         if: failure()
         run: |
-          sudo k8s/scripts/inspect.sh
+          tar -czvf inspection-reports.tar.gz -C $HOME inspection-reports
+          ls
       - name: Upload inspection report artifact
         if: failure()
         uses: actions/upload-artifact@v4
         with:
-          name: inspection-report
-          path: ./inspection-report-*.tar.gz
+          name: inspection-reports-${{ matrix.os }}-${{ matrix.patch }}
+          path: inspection-reports.tar.gz
--- a/.github/workflows/integration-informing.yaml
+++ b/.github/workflows/integration-informing.yaml
@@ -86,12 +86,10 @@ jobs:
           export TEST_SUBSTRATE=lxd
           export TEST_LXD_IMAGE=${{ matrix.os }}
           cd tests/integration && sg lxd -c 'tox -e integration'
-      - name: Compress inspection reports
+      - name: Prepare inspection reports
         if: failure()
         run: |
           tar -czvf inspection-reports.tar.gz -C $HOME inspection-reports
-      - name: Set inspection report artifact name
-        run: |
           echo "artifact_name=inspection-reports-${{ matrix.os }}-${{ matrix.patch }}" | sed 's/:/-/g' >> $GITHUB_ENV
           echo ${{ env.artifact_name }}
       - name: Upload inspection report artifact
--- a/.github/workflows/integration-informing.yaml
+++ b/.github/workflows/integration-informing.yaml
@@ -90,10 +90,13 @@ jobs:
         if: failure()
         run: |
           tar -czvf inspection-reports.tar.gz -C $HOME inspection-reports
-          ls
+      - name: Set inspection report artifact name
+        run: |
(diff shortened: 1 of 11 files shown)


.github/workflows/integration-informing.yaml
```diff
@@ -80,9 +80,27 @@ jobs:
...
+      - name: Generate inspection report
+        if: failure()
+        run: |
+          sudo k8s/scripts/inspect.sh
+      - name: Upload inspection report artifact
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: inspection-report
+          path: ./inspection-report-*.tar.gz
```
neoaggelos, line 106: This needs to run as part of the test harness, as we need the inspection report from the LXC instance(s) (or multipass instances, or the test machines in general).

My recommendation would be: Add a config option `TEST_COLLECT_INSPECTION_REPORTS_DIR`, where we dump inspection reports from each instance on the tests (if not empty). Each inspection report should have the instance name somehow. Then, we would have a step here that `if: failure()`, would upload the directory at the end of the run.

This way, a failed run will give us output that describes what failed (the test output we already have), then we'll also have inspection reports from all instances.

neoaggelos (changes requested):

.github/workflows/integration-informing.yaml
neoaggelos: This currently runs on the github runner instance, but this is not where k8s-snap is running for our tests

tests/integration/tests/test_util/harness/local.py
```diff
@@ -22,27 +23,33 @@ class LocalHarness(Harness):
...
             raise HarnessError("failed to wait for snapd seed") from e
 
+        self.instance = self.hostname
         return Instance(self, self.hostname)
 
+
+    def get_instances(self) -> Set[str]:
+        if self.instance is None:
+            return set()
+        return set(self.instance)
```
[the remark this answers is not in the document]
neoaggelos: but i think having a `.get_instances()` is probably not something we need

.github/workflows/integration-informing.yaml
bschimke95: will do after I validated that stuff works in `integration.yaml`

tests/integration/tests/conftest.py
```diff
@@ -137,6 +163,7 @@ def instances(
     # remove the session_instance. The harness ensures that everything is cleaned up
     # at the end of the test session.
     for instance in instances:
+        _generate_inspection_reports(h)
```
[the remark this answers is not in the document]
bschimke95: yes, thanks for the catch

tests/integration/tests/test_util/harness/local.py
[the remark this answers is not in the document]
bschimke95: removed `get_instances`

tests/integration/tests/conftest.py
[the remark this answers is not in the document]
bschimke95: I also collect the output of the `inspect.sh` script and combine all steps into `_generate_inspection_report(h, instance_id)` now. I call this twice (after test cleanup and after session cleanup) so it makes sense to have this in a separate function

tests/integration/tests/conftest.py
```diff
@@ -25,6 +25,28 @@ def _harness_clean(h: harness.Harness):
...
+    inspection_path = Path(config.INSPECTION_REPORTS_DIR)
+    result = h.exec(
+        instance_id,
+        ["/snap/k8s/current/k8s/scripts/inspect.sh", "/inspection-report.tar.gz"],
+        capture_output=True,
+        text=True,
+    )
+
+    (inspection_path / instance_id).mkdir(parents=True, exist_ok=True)
+    with open(inspection_path / instance_id / "inspection_report_logs.txt", "w") as f:
```
neoaggelos, line 40: you can also use (path).write_text()

bschimke95: that's actually pretty cool - thanks TIL!

tests/integration/tests/conftest.py
```diff
@@ -37,8 +37,7 @@ def _generate_inspection_report(h: harness.Harness, instance_id: str):
     )
 
     (inspection_path / instance_id).mkdir(parents=True, exist_ok=True)
-    with open(inspection_path / instance_id / "inspection_report_logs.txt", "w") as f:
-        f.write(result.stdout)
+    Path(inspection_path / instance_id / "inspection_report_logs.txt", "w").write_text(result.stdout)
```
neoaggelos, line 40: you probably do not want the `"w"` there, also no need to add `Path()`, as `<path> / "string"` is a Path object already
```suggestion
    (inspection_path / instance_id / "inspection_report_logs.txt").write_text(result.stdout)
```

tests/integration/tests/conftest.py
```diff
@@ -25,6 +25,29 @@ def _harness_clean(h: harness.Harness):
         h.cleanup()
 
 
+def _generate_inspection_report(h: harness.Harness, instance_id: str):
+    LOG.debug("Generating inspection report for %s", instance_id)
+
+    inspection_path = Path(config.INSPECTION_REPORTS_DIR)
+    result = h.exec(
```
neoaggelos, line 32: maybe ignore failures while retrieving the inspection reports.

eg for the etcd instances, where k8s-snap is not installed. you could add maybe check=False for that

tests/integration/tests/conftest.py
```diff
@@ -25,6 +25,30 @@ def _harness_clean(h: harness.Harness):
...
+    (inspection_path / instance_id).mkdir(parents=True, exist_ok=True)
+    (inspection_path / instance_id / "inspection_report_logs.txt").write_text(
+        result.stdout
+    )
+
+    h.pull_file(
+        instance_id,
+        "/inspection-report.tar.gz",
+        (inspection_path / instance_id / "inspection_report.tar.gz").as_posix(),
+    )
```
neoaggelos, line 49: could that, too, fail if the inspection report was not created successfully?

neoaggelos (approved):

PR #599, opened 2024-08-10
Fix check for initialized nodes

neoaggelos:
### Summary

Followup for #564

### Changes

- Fix issue where `cmdutil.GetNodeStatus()` did not properly recognize 503 errors
- Make this part of `(k8sd.Client).NodeStatus()`
- Adjust CLI to use `client.NodeStatus()` 
- Removed `cmdutil.GetNodeStatus()` (it is not a command utility, it is functionality we want in the client)

### Addresses

```
$ k8s status
root@t2:~# k8s status
Error: Failed to retrieve the node status.

The error was: failed to GET /k8sd/node: Database is not yet initialized
```

With the new changes, we are back to the previous error message we had, that shows how to bootstrap a new cluster:

```
root@t1:~# k8s status
Error: The node is not part of a Kubernetes cluster. You can bootstrap a new cluster with:

  sudo k8s bootstrap
```


eaudetcobello:
I notice the isBootstrapped variables have been renamed to initialize. Can you speak a bit on the difference between the two ideas?

neoaggelos:
> I notice the isBootstrapped variables have been renamed to initialize. Can you speak a bit on the difference between the two ideas?

The first node will bootstrap a new cluster. Other nodes join the cluster. isInitialized is used to mean that the node is part of the cluster.

src/k8s/pkg/client/k8sd/status.go
HomayoonAlimohammadi, line 1: I think it's not a bad idea to add unit tests for this file just so that we make sure `http.StatusServiceUnavailable` is being handled differently. or maybe we can add that in the future or in another card.

neoaggelos: The microcluster client is used, so we would either need a mock for that (not too worth it), or spin up an HTTP server that returns this, or a mock microcluster.

We dont have unit tests for the microcluster node, so lets do that in the future, not in this PR

HomayoonAlimohammadi (approved):
Thanks a lot Angelos! LGTM

PR #601, opened 2024-08-12
Update status on failed apply

HomayoonAlimohammadi:
After trying to apply a feature configuration, make sure to update the feature status on both failure and success.

### DIFF
--- a/src/k8s/pkg/k8sd/controllers/feature.go
+++ b/src/k8s/pkg/k8sd/controllers/feature.go
@@ -131,16 +131,14 @@ func (c *FeatureController) reconcile(
 		return fmt.Errorf("failed to retrieve cluster configuration: %w", err)
 	}
 
-	featureStatus, err := apply(cfg)
-	if err != nil {
-		if err := updateFeatureStatus(ctx, featureStatus); err != nil {
-			log.FromContext(ctx).WithValues("message", featureStatus.Message).Error(err, "Failed to update feature status after failed apply")
-		}
-		return fmt.Errorf("failed to apply configuration: %w", err)
+	status, applyErr := apply(cfg)
+	if err := updateFeatureStatus(ctx, status); err != nil {
+		// NOTE (hue): status update errors are not returned but only logged. we might need some retry logic in the future.
+		log.FromContext(ctx).WithValues("message", status.Message, "applied-successfully", applyErr == nil).Error(err, "Failed to update feature status")
 	}
 
-	if err := updateFeatureStatus(ctx, featureStatus); err != nil {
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -131,10 +131,16 @@ func (c *FeatureController) reconcile(
...
+			log.FromContext(ctx).WithValues("message", featureStatus.Message).Error(err, "Failed to update feature status after failed apply")
+		}
 		return fmt.Errorf("failed to apply configuration: %w", err)
-	} else if err := updateFeatureStatus(ctx, featureStatus); err != nil {
-		return fmt.Errorf("failed to update feature status: %w", err)
+	}
+
+	if err := updateFeatureStatus(ctx, featureStatus); err != nil {
+		log.FromContext(ctx).WithValues("message", featureStatus.Message).Error(err, "Failed to update feature status after successful apply")
 	}
```
neoaggelos, line 144: this code should work, but repeats `updateFeatureStatus` calls in different "states", with the only change being the log message.

how about something like this instead:

```go
status, applyErr := apply(cfg)
if err := updateFeatureStatus(ctx, status); err != nil {
  log.FromContext(ctx).WithValues("message", status.Message, "result", applyErr == nil).Error(err, "Failed to update feature status")
}

if applyErr != nil {
  return fmt.Errorf("failed to apply configuration: %w", applyErr)
}

return nil
```

Also, should we consider failures of `updateFeatureStatus` as an error that we need to retry?

neoaggelos (changes requested):

src/k8s/pkg/k8sd/controllers/feature.go
HomayoonAlimohammadi: that's a good point. thanks.
about the retry, we can retry but we should do it synchronously (otherwise we might see stale status).
maybe we can add a `SetFeatureStatusWithRetry` [here](https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/database/feature_status.go#L20) and pass the retry args [here](https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/app/hooks_start.go#L82)?
and we can pass the retry params to `app.Config.<RetryParam>`

neoaggelos: we can defer that for later. it is not going to be a common issue, so adding such structural complexity is maybe not worth it, unless we see it becoming a problem.

I would stick to adding a `// NOTE` that mentions this for future onlookers, that errors in updating the feature status are logged and not returned

neoaggelos (approved):

PR #604, opened 2024-08-12
Centralize feature names

HomayoonAlimohammadi:
Feature names are in literal string form and scattered around, making it easy to misspell things that can lead to annoying bugs. this PR centralizes the feature names.

### DIFF
--- a/src/k8s/cmd/k8s/k8s.go
+++ b/src/k8s/cmd/k8s/k8s.go
@@ -4,11 +4,12 @@ import (
 	"time"
 
 	cmdutil "github.com/canonical/k8s/cmd/util"
+	"github.com/canonical/k8s/pkg/k8sd/features"
 	"github.com/spf13/cobra"
 )
 
 var (
-	featureList = []string{"network", "dns", "gateway", "ingress", "local-storage", "load-balancer"}
+	featureList = []string{features.Network, features.DNS, features.Gateway, features.Ingress, features.LocalStorage, features.LoadBalancer}
 
 	outputFormatter cmdutil.Formatter
 )
--- a/src/k8s/pkg/k8sd/api/cluster.go
+++ b/src/k8s/pkg/k8sd/api/cluster.go
@@ -10,6 +10,7 @@ import (
 	"github.com/canonical/k8s/pkg/k8sd/api/impl"
 	"github.com/canonical/k8s/pkg/k8sd/database"
 	databaseutil "github.com/canonical/k8s/pkg/k8sd/database/util"
+	"github.com/canonical/k8s/pkg/k8sd/features"
 	"github.com/canonical/k8s/pkg/k8sd/types"
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/microcluster/v2/state"
@@ -61,13 +62,13 @@ func (e *Endpoints) getClusterStatus(s state.State, r *http.Request) response.Re
 				Type:    config.Datastore.GetType(),
 				Servers: config.Datastore.GetExternalServers(),
 			},
-			DNS:           statuses["dns"].ToAPI(),
-			Network:       statuses["network"].ToAPI(),
-			LoadBalancer:  statuses["load-balancer"].ToAPI(),
-			Ingress:       statuses["ingress"].ToAPI(),
-			Gateway:       statuses["gateway"].ToAPI(),
-			MetricsServer: statuses["metrics-server"].ToAPI(),
-			LocalStorage:  statuses["local-storage"].ToAPI(),
+			DNS:           statuses[features.DNS].ToAPI(),
+			Network:       statuses[features.Network].ToAPI(),
+			LoadBalancer:  statuses[features.LoadBalancer].ToAPI(),
+			Ingress:       statuses[features.Ingress].ToAPI(),
+			Gateway:       statuses[features.Gateway].ToAPI(),
+			MetricsServer: statuses[features.MetricsServer].ToAPI(),
+			LocalStorage:  statuses[features.LocalStorage].ToAPI(),
 		},
 	}
 
--- a/src/k8s/pkg/k8sd/features/features.go
+++ b/src/k8s/pkg/k8sd/features/features.go
@@ -1,0 +1,11 @@
+package features
+
+const (
+	DNS           string = "dns"
+	Network       string = "network"
+	Gateway       string = "gateway"
+	Ingress       string = "ingress"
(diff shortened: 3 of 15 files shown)


src/k8s/pkg/k8sd/features/features.go
```diff
@@ -0,0 +1,11 @@
+package features
+
+const (
+	DNS           string = "dns"
```
neoaggelos, line 4: let's add a `type FeatureName string` and make these `DNS FeatureName = "dns"` etc

Then, in all functions that require the name of a feature, we use `FeatureName` as the type instead. This is not as good as an enum, but it's better to force us to think about not passing in arbitrary values.

HomayoonAlimohammadi: good point, thanks. in that case, is it ok if we use `string(features.Network)` in the places that we expect strings? 
places like this:
```go
    for _, component := range strings.FieldsFunc(components, func(r rune) bool { return unicode.IsSpace(r) || r == ',' }) {
		switch component {
		case string(features.Network):
			config.ClusterConfig.Network.Enabled = utils.Pointer(true)
		...
		}
	}
```

src/k8s/pkg/utils/experimental/snapdconfig/k8s_to_snapd.go
```diff
@@ -6,20 +6,21 @@ import (
...
 
 	apiv1 "github.com/canonical/k8s/api/v1"
+	"github.com/canonical/k8s/pkg/k8sd/features"
 	"github.com/canonical/k8s/pkg/snap"
 )
 
 // SetSnapdFromK8sd uses snapctl to update the local snapd configuration with the new k8sd cluster configuration.
 func SetSnapdFromK8sd(ctx context.Context, config apiv1.UserFacingClusterConfig, snap snap.Snap) error {
 	var sets []string
 	for key, cfg := range map[string]any{
```
neoaggelos, line 16: change map key to `FeatureName` type

src/k8s/pkg/k8sd/api/cluster.go
```diff
@@ -40,7 +41,7 @@ func (e *Endpoints) getClusterStatus(s state.State, r *http.Request) response.Re
 		return response.InternalError(fmt.Errorf("failed to check if cluster has ready nodes: %w", err))
 	}
 
-	var statuses map[string]types.FeatureStatus
+	var statuses map[features.FeatureName]types.FeatureStatus
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: For the `FeatureStatus` I was following the proposal (since it was already accepted), but for this one it just felt right to put it here. but I get what you're saying. I'd say that now I prefer the `FeatureStatus` to move into the `k8sd/feature` package, what do you think @neoaggelos?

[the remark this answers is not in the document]
neoaggelos: the `type FeatureName string` should be in `pkg/k8sd/types`, the constants should be in `pkg/k8sd/features`

src/k8s/cmd/k8s/k8s.go
```diff
@@ -4,11 +4,19 @@ import (
...
 var (
-	featureList = []string{"network", "dns", "gateway", "ingress", "local-storage", "load-balancer"}
+	featureList = []string{
+		string(features.Network),
+		string(features.DNS),
+		string(features.Gateway),
+		string(features.Ingress),
+		string(features.LocalStorage),
+		string(features.LoadBalancer),
+	}
```
neoaggelos, line 19: While we're at it, this could also be in the features package, maybe something like a `func All() []types.FeatureName` used as `features.All()`?

neoaggelos: there was some old discussion, maybe `metrics-server` would not be in that list

neoaggelos (approved):
LGTM, final comment, completely optional

src/k8s/cmd/k8s/k8s.go
HomayoonAlimohammadi: If we want to exclude `metrics-server` (so that we can use it here) the `features.All()` function would be somehow misleading. maybe something like `func All(includeMetricsServer bool) []types.FeatureName` would cover that but I don't feel good about it (seems kind of ugly). 
Or maybe we can also include the `metrics-server` in this feature list and use the `All()` function.

neoaggelos: let's do `features.Public()` instead

HomayoonAlimohammadi: Seems like we have to either go with `func Public() []string` or do something like:
```go
publicFeatures := features.Public()
featuresList := make([]string, len(publicFeatures)
for i, f := range publicFeatures {
    featuresList[i] = string(f)
}
```
cause the `featuresList` is going to be used in `strings.Join` and other places. If I have to choose, I would go with the second option.
wdyt?

neoaggelos: See https://github.com/canonical/k8s-snap/blob/2e29d066681c32ccc97392ba3ecef8d771454d86/src/k8s/pkg/k8sd/images/images.go#L7-L16

HomayoonAlimohammadi: in the example you sent it's `var registeredImages []string` and works with `copy`.
https://github.com/canonical/k8s-snap/blob/2e29d066681c32ccc97392ba3ecef8d771454d86/src/k8s/pkg/k8sd/images/images.go#L5

src/k8s/pkg/utils/experimental/snapdconfig/snapd_to_k8s.go
```diff
@@ -13,14 +13,9 @@ import (
 
 // SetK8sdFromSnapd updates the k8sd cluster configuration from the current local snapd configuration.
 func SetK8sdFromSnapd(ctx context.Context, client k8sd.Client, snap snap.Snap) error {
+	args := append([]string{"-d"}, features.Public()...)
```
eaudetcobello: Out of curiosity, what is -d for?

eaudetcobello (approved):
Nice work, this is useful

src/k8s/pkg/utils/experimental/snapdconfig/snapd_to_k8s.go
HomayoonAlimohammadi: I think this was reverted.

Issue #567, opened 2024-07-28
snap install: Cilium pod fails to start with "cp: cannot create regular file '/hostbin/cilium-mount': Permission denied"


zwindler:
### Summary

Using the snap install k8s, cilium fails to start at init state (mount-cgroup container) due to an error

```
  mount-cgroup:
    Container ID:  containerd://af532ab4be4c10d8c0f268da115b63655da1e57711daa4b9861462414785b206
    Image:         ghcr.io/canonical/cilium:1.15.2-ck1
    Image ID:      ghcr.io/canonical/cilium@sha256:5bb407edc3a8b64f45225da6b7b4aba22c0daf022e85723534df39e23327947a
    Port:          <none>
    Host Port:     <none>
    Command:
      sh
      -ec
      cp /usr/bin/cilium-mount /hostbin/cilium-mount;
      nsenter --cgroup=/hostproc/1/ns/cgroup --mount=/hostproc/1/ns/mnt "${BIN_PATH}/cilium-mount" $CGROUP_ROOT;
      rm /hostbin/cilium-mount
      
    State:       Waiting
      Reason:    CrashLoopBackOff
    Last State:  Terminated
      Reason:    Error
      Message:   cp: cannot create regular file '/hostbin/cilium-mount': Permission denied

      Exit Code:    1
      Started:      Sun, 28 Jul 2024 14:14:36 +0200
      Finished:     Sun, 28 Jul 2024 14:14:36 +0200
    Ready:          False
    Restart Count:  3
 ```

### What Should Happen Instead?

cilium should start

### Reproduction Steps

Bootstrap a new cluster 

```
$ sudo snap install k8s --channel=1.30-classic/beta --classic
$ sudo k8s bootstrap
$ sudo k8s kubectl get pods -A
NAMESPACE     NAME                                  READY   STATUS                  RESTARTS      AGE
kube-system   cilium-bq64v                          0/1     Init:CrashLoopBackOff   5 (61s ago)   4m2s
kube-system   cilium-operator-66684ccb-9v6jv        1/1     Running                 0             4m23s
kube-system   ck-storage-rawfile-csi-controller-0   2/2     Running                 0             4m25s
kube-system   ck-storage-rawfile-csi-node-k9td9     4/4     Running                 0             4m26s
kube-system   coredns-7d4dffcffd-4bcj8              1/1     Running                 0             4m25s
kube-system   metrics-server-6f66c6cc48-46zcm       1/1     Running                 0             4m25s
```

[...]

PR #606, opened 2024-08-14
Add checks for CNIBinDir ownership/permissions.

aznashwan:
When the Cilium feature is enabled, add simple user/group/permissions checks on the configured CNIBinDir (/opt/cni/bin) before installing the Cilium Helm chart.

Fixes: https://github.com/canonical/k8s-snap/issues/567

### DIFF
--- a/src/k8s/pkg/k8sd/setup/directories.go
+++ b/src/k8s/pkg/k8sd/setup/directories.go
@@ -1,16 +1,23 @@
 package setup
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
+	"syscall"
 
+	"github.com/canonical/k8s/pkg/log"
 	"github.com/canonical/k8s/pkg/snap"
 )
 
 // EnsureAllDirectories ensures all required configuration and state directories are created.
 func EnsureAllDirectories(snap snap.Snap) error {
+	if err := ensureCniBinDir(snap.CNIBinDir()); err != nil {
+		return err
+	}
+
 	for _, dir := range []string{
-		snap.CNIBinDir(),
 		snap.CNIConfDir(),
 		snap.ContainerdConfigDir(),
 		snap.ContainerdExtraConfigDir(),
@@ -30,5 +37,50 @@ func EnsureAllDirectories(snap snap.Snap) error {
 			return fmt.Errorf("failed to create required directory: %w", err)
 		}
 	}
+	return nil
+}
+
+// Ensures that the provided path is a directory with the appropriate
+// ownership/permissions for it to be used as the CNI binary directory.
+// https://github.com/canonical/k8s-snap/issues/567
+// https://github.com/cilium/cilium/issues/23838
+func ensureCniBinDir(cniBinDir string) error {
+	l := log.L().WithValues("cniBinDir", cniBinDir)
+	if cniBinDir == "" {
+		l.V(1).Info("Skipping creation of cni bin directory since it was not set")
+		return nil
+	}
+
+	var stat syscall.Stat_t
+	if err := syscall.Stat(cniBinDir, &stat); err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			return fmt.Errorf("failed to syscall.Stat(%q): %w", cniBinDir, err)
+		}
+
+		l.Info("Creating cni bin directory")
+		if err := os.MkdirAll(cniBinDir, 0o0700); err != nil {
+			return fmt.Errorf("failed to os.MkdirAll(%s): %w", cniBinDir, err)
+		}
+
+		if err := syscall.Stat(cniBinDir, &stat); err != nil {
+			return fmt.Errorf("failed to syscall.Stat(%q) newly-created cni bin dir: %w", cniBinDir, err)
+		}
+	}
+
+	if stat.Uid != 0 || stat.Gid != 0 {
+		l.Info("Ensuring ownership of cni bin directory")
+		if err := os.Chown(cniBinDir, 0, 0); err != nil {
+			return fmt.Errorf("failed to os.Chown(%q, 0, 0): %w", cniBinDir, err)
+		}
+	}
+
+	if (stat.Mode & 0o700) != 0o700 {
+		l.Info("Ensuring permissions of cni bin directory")
+		mode := os.FileMode(stat.Mode | 0o700)
+		if err := os.Chmod(cniBinDir, mode); err != nil {
+			return fmt.Errorf("failed to os.Chmod(%q, %o): %w", cniBinDir, mode, err)
+		}
+	}
+
 	return nil
 }
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -63,7 +63,9 @@ jobs:
 
       - name: go test
         working-directory: src/k8s
-        run: make go.unit
+        # NOTE: there are a handful of tests checking/setting
+        # root ownership so the tests must be run as root:
+        run: sudo make go.unit
 
   test-binary:
     name: Binaries


src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -44,6 +46,14 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, _ type
 		}, nil
 	}
 
+	if err := checkCniBinDirPermissions(snap.CNIBinDir()); err != nil {
+		return types.FeatureStatus{
+			Enabled: false,
+			Version: ciliumAgentImageTag,
+			Message: fmt.Sprintf(networkDeployFailedMsgTmpl, err),
+		}, err
```
neoaggelos, line 54: return `FeatureStatus{Version: ciliumAgentImageTag, Message: ...}, err`, as zero value is false

neoaggelos (commented):
Nice work! Some comments

neoaggelos (changes requested):
In a more general note, this does not _prevent_ or address the issue, it only raises the error message in `k8s status` rather than getting cilium deployed. 

I could see this causing more issues (e.g. gateway or ingress also failing to enable with errors). Perhaps we should just amend the required state? e.g. try to chown and chmod the directory?

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -201,3 +211,36 @@ func rolloutRestartCilium(ctx context.Context, snap snap.Snap, attempts int) err
...
+		return fmt.Errorf("CNIBinDir %q must belong to user root (UID 0), but belongs to %d. %s", cniBinDir, stat.Uid, chownMsg)
+	}
+	if stat.Gid != 0 {
+		return fmt.Errorf("CNIBinDir %q must belong to group root (GID 0), but belongs to %d. %s", cniBinDir, stat.Gid, chownMsg)
+	}
+
+	expected := uint16(0o755)
+	if stat.Mode != expected {
+		return fmt.Errorf("CNIBinDir %q must have permissions %o, but has %o. Please run: sudo chmod -R 755 %s", cniBinDir, expected, stat.Mode, cniBinDir)
+	}
```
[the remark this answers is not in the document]
aznashwan: Great catch, can confirm 0700 permissions also avoid the issue!

The Cilium/K8s snap issues explicitly mentioned `0755` and I just took it at face value...

aznashwan:
> In a more general note, this does not prevent or address the issue, it only raises the error message in k8s status rather than getting cilium deployed.

Correct, that was [what was suggested in this comment on the original issue](https://github.com/canonical/k8s-snap/issues/567#issuecomment-2260355818), so I just went with it.

> Perhaps we should just amend the required state? e.g. try to chown and chmod the directory?

I had considered it, but was unsure whether the snap should liberally change the state of files on the host system.

If you think this is something the k8s snap should/could be able to do, I can gladly update the PR to perform the chown/chmod's directly...

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -201,3 +210,33 @@ func rolloutRestartCilium(ctx context.Context, snap snap.Snap, attempts int) err
...
+// Checks that the provided directory path has correct permissions to be used
+// for mounting by the Cilium chart and storing CNI binaries within it.
+// https://github.com/canonical/k8s-snap/issues/567
+// https://github.com/cilium/cilium/issues/23838
+func checkCniBinDirPermissions(cniBinDir string) error {
+	followedLink, err := filepath.EvalSymlinks(cniBinDir)
+	if err != nil {
+		return fmt.Errorf("failed to EvalSymlinks(%q): %w", cniBinDir, err)
+	}
+	cniBinDir = followedLink
```
neoaggelos, line 223: is this necessary? `man 2 stat` says

```
       lstat() is identical to stat(), except that if pathname is a symbolic link, then it returns information about  the  link  itself,
       not the file that the link refers to.
```

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -44,6 +46,13 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, _ type
 		}, nil
 	}
 
+	if err := checkCniBinDirPermissions(snap.CNIBinDir()); err != nil {
+		return types.FeatureStatus{
+			Version: ciliumAgentImageTag,
+			Message: fmt.Sprintf(networkDeployFailedMsgTmpl, err),
+		}, err
+	}
```
neoaggelos, line 54: we _are_ managing this directory and taking it over from the host. we should not treat this as foreign. therefore, i would recommend just amending the permission bits rather than checking and raising the issue

src/k8s/pkg/k8sd/features/cilium/network.go
neoaggelos: rather, should this just be `syscall.Lstat` ?

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -201,3 +212,43 @@ func rolloutRestartCilium(ctx context.Context, snap snap.Snap, attempts int) err
...
+			if err := syscall.Stat(cniBinDir, &stat); err != nil {
+				return fmt.Errorf("failed to syscall.Stat(%q): %w", cniBinDir, err)
+			}
+		}
+		return fmt.Errorf("failed to syscall.Stat(%q): %w", cniBinDir, err)
+	}
+
+	if stat.Uid != 0 || stat.Gid != 0 {
+		msg := fmt.Sprintf("changing cniBinDir ownership from %d:%d to 0:0", stat.Uid, stat.Gid)
+		log.FromContext(ctx).WithValues("cniBinDir", cniBinDir).Info(msg)
```
neoaggelos, line 238: prefer a static log message, then add arguments as values (e.g. "from": "100:100" "to":"0:0")

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -44,6 +48,13 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, _ type
 		}, nil
 	}
 
+	if err := ensureCniBinDir(ctx, snap.CNIBinDir()); err != nil {
+		return types.FeatureStatus{
+			Version: ciliumAgentImageTag,
+			Message: fmt.Sprintf(networkDeployFailedMsgTmpl, err),
+		}, err
+	}
```
neoaggelos, line 56: As discussed elsewhere, this only runs on one of the nodes in the cluster, and therefore might leave other nodes broken. Perhaps it's best to move this as part of `setup.EnsureAllDirectories()` which runs on all cluster nodes when they join the cluster.

aznashwan:
> Perhaps it's best to move this as part of `setup.EnsureAllDirectories()` which runs on all cluster nodes when they join the cluster.

Just updated, thanks!

I initially made it part of the Cilium network setup since it was "closer" to where the Cilium chart deployment which failed due to the wrong ownership/permissions was being deployed, but it does indeed make more sense to handle this in the main k8sd setup.

src/k8s/pkg/k8sd/setup/directories.go
```diff
@@ -32,3 +39,48 @@ func EnsureAllDirectories(snap snap.Snap) error {
...
+			if err := syscall.Stat(cniBinDir, &stat); err != nil {
+				return fmt.Errorf("failed to syscall.Stat(%q): %w", cniBinDir, err)
+			}
+		}
+		return fmt.Errorf("failed to syscall.Stat(%q): %w", cniBinDir, err)
+	}
+
+	if stat.Uid != 0 || stat.Gid != 0 {
+		msg := fmt.Sprintf("changing cniBinDir ownership")
+		log.L().WithValues("cniBinDir", cniBinDir, "from", fmt.Sprintf("%d:%d", stat.Uid, stat.Gid), "to", "0:0").Info(msg)
```
neoaggelos, line 70: just `.Info("Ensure ownership of cni bin directory")`

src/k8s/pkg/k8sd/setup/directories.go
```diff
@@ -32,3 +39,48 @@ func EnsureAllDirectories(snap snap.Snap) error {
...
+		if err := os.Chown(cniBinDir, 0, 0); err != nil {
+			return fmt.Errorf("failed to os.Chown(%q, 0, 0): %w", cniBinDir, err)
+		}
+	}
+
+	if (stat.Mode & 0o700) != 0o700 {
+		mode := os.FileMode(stat.Mode | 0o700)
+		msg := fmt.Sprintf("changing cniBinDir permissions")
+		log.L().WithValues("cniBinDir", cniBinDir, "from", fmt.Sprintf("%o", stat.Mode), "to", fmt.Sprintf("%o", mode)).Info(msg)
```
neoaggelos, line 79: same here

neoaggelos (approved):
It's OK, left some comments about the logging

bschimke95 (commented):
Please fix the Go test, then we are good to merge

aznashwan:
@bschimke95 I have managed to fix the previous logic bug but the [tests are apparently being run as `1001:127`](https://github.com/aznashwan/k8s-snap/actions/runs/10577118505/job/29304369846#step:11:529) so the `chown` isn't permitted.

The only workaround I can think of is running the tests as root from the workflow (`sudo make go.unit`), but I'm open to better ideas.

bschimke95:
Hey @aznashwan 
Just talked with @berkayoz about this and we also don't really have a better idea right now. Let's just run this as root.

PR #607, opened 2024-08-14
Prevent hang and show correct error on bootstrap and join-cluster

HomayoonAlimohammadi:
# Description
On failed `bootstrap` and `join-cluster` commands, the `PreRemove` hook will be called by microcluster. In some scenarios if the Database is not initialized, we get stuck in the `control.WaitUntilReady` loop indefinitely and the command fails with `context deadline exceeded`. We set a fixed number of retries to hopefully prevent this from happening, continue with the cleanup and show the user the actual error.

# UX changes
let's say we have `cp1` which is bootstrapped, we get a token with `get-join-token` but we don't specify any node name. running `k8s join-cluster <token>` on `cp2` should result in an error according to microcluster v2 but right now, the user is hanged for a long time and after that will see this (by default after ~2 mins):
```text
root@cp2:~# k8s join-cluster <token>
Joining the cluster. This may take a few seconds, please wait.
Error: Failed to join the cluster using the provided token.

The error was: failed to POST /k8sd/cluster/join: Post "http://control.socket/1.0/k8sd/cluster/join": context deadline exceeded
```
with these changes, the the command will fail sooner and shows the correct message (default ~15 seconds) (
```text
root@cp2:~# k8s join-cluster <token>
Joining the cluster. This may take a few seconds, please wait.
Error: Failed to join the cluster using the provided token.

The error was: failed to POST /k8sd/cluster/join: failed to join k8sd cluster as control plane: 1 join attempts were unsuccessful. Last error: Joining server certificate SAN does not contain join token name
```

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -28,7 +28,13 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 	log.Info("Running preremove hook")
 
 	log.Info("Waiting for node to finish microcluster join before removing")
-	control.WaitUntilReady(ctx, func() (bool, error) {
+	// NOTE (hue): in microcluster v2, PreRemove hook is also called if something goes wrong on
+	// `bootstrap` and `join-cluster`. It is possible that we get stuck in this loop forever which causes
+	// the `bootstrap` and `join-cluster` commands to hang and finally return an uninformative `context deadline exceeded` error
+	// we optimistically stop trying after a fixed number of retries.
+	maxRetries := 10
+	var txnRetries int
+	if err := control.WaitUntilReady(ctx, func() (bool, error) {
 		var notPending bool
 		if err := s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
 			member, err := cluster.GetCoreClusterMember(ctx, tx, s.Name())
@@ -40,9 +46,18 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 			return nil
 		}); err != nil {
 			log.Error(err, "Failed database transaction to check cluster member role")
+			txnRetries++
 		}
+
+		if txnRetries >= maxRetries {
+			log.Info("Reached maximum number of retries for database transactions on pre-remove hook, continuing cleanup", "max_retries", maxRetries)
+			return true, nil
+		}
+
 		return notPending, nil
-	})
+	}); err != nil {
+		log.Error(err, "Failed to wait for node to finish microcluster join before removing. Continuing with the cleanup...")
+	}
 
 	if cfg, err := databaseutil.GetClusterConfig(ctx, s); err == nil {
 		if _, ok := cfg.Annotations[apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove]; !ok {


src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -134,7 +136,13 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
 				Timeout: opts.timeout,
 			})
-			if err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
+				// NOTE (hue): this might have failed because we're stuck (most likely indefinitely) in the pre-remove hook
+				// let's show the user a more helpful message
+				cmd.PrintErrf("Error: Failed to bootstrap the cluster.\n\nThe error was: %v\n\nIncreasing the timeout might help.\n", err)
+				env.Exit(1)
+				return
+			} else if err != nil {
```
eaudetcobello, line 145: I see where you're going with this but 99% of the time if the deadline exceeded, the problem is *not* that the default timeout is too short. Usually something is broken.

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -108,7 +110,13 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				Token:   token,
 				Config:  joinClusterConfig,
 				Timeout: opts.timeout,
-			}); err != nil {
+			}); errors.Is(err, context.DeadlineExceeded) {
+				// NOTE (hue): this might have failed because we're stuck (most likely indefinitely) in the pre-remove hook
+				// let's show the user a more helpful message
+				cmd.PrintErrf("Error: Failed to join the cluster using the provided token.\n\nThe error was: %v\n\nIncreasing the timeout might help.\n", err)
+				env.Exit(1)
+				return
```
eaudetcobello, line 118: Same

src/k8s/cmd/k8sd/k8sd.go
```diff
@@ -71,6 +73,7 @@ func NewRootCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&rootCmdOpts.disableUpdateNodeConfigController, "disable-update-node-config-controller", false, "Disable the Update Node Config Controller")
 	cmd.PersistentFlags().BoolVar(&rootCmdOpts.disableFeatureController, "disable-feature-controller", false, "Disable the Feature Controller")
 	cmd.PersistentFlags().BoolVar(&rootCmdOpts.disableCSRSigningController, "disable-csrsigning-controller", false, "Disable the CSR signing controller")
+	cmd.PersistentFlags().IntVar(&rootCmdOpts.maxPreRemoveHookDBTxnRetries, "max-remove-txn-retries", 10, "Maximum number of retries for database transactions in the pre-remove hook")
```
eaudetcobello, line 76: This doesn't seem very useful to have as an exposed flag, and exposes too much implementation detail imo. I think the timeout flag covers the vast majority of use cases.

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -28,7 +29,12 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 	log.Info("Running preremove hook")
 
 	log.Info("Waiting for node to finish microcluster join before removing")
-	control.WaitUntilReady(ctx, func() (bool, error) {
+	// NOTE (hue): in microcluster v2, PreRemvoe hook is also called if something goes wrong on
```
eaudetcobello, line 32:
```suggestion
	// NOTE (hue): in microcluster v2, PreRemove hook is also called if something goes wrong on
```

src/k8s/cmd/k8sd/k8sd.go
HomayoonAlimohammadi: so do you suggest marking this a hidden flag with `cmd.PersistentFlags().MarkHidden("max-remove-txn-retries")` or completely remove the flag and hard code the num retries?

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -28,7 +29,12 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 	log.Info("Running preremove hook")
 
 	log.Info("Waiting for node to finish microcluster join before removing")
-	control.WaitUntilReady(ctx, func() (bool, error) {
+	// NOTE (hue): in microcluster v2, PreRemove hook is also called if something goes wrong on
+	// `bootstrap` and `join-cluster`. It is possible that we get stuck in this loop forever which causes
+	// the `bootstrap` and `join-cluster` commands to hang and finally return an uninformative `context deadline exceeded` error
+	// we optimiscally stop trying after a configurable number of retries.
```
eaudetcobello, line 35:
```suggestion
	// we optimistically stop trying after a number of retries.
```

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -40,10 +46,23 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 			return nil
 		}); err != nil {
 			log.Error(err, "Failed database transaction to check cluster member role")
+			txnRetries++
 		}
+
+		if txnRetries >= a.config.MaxPreRemoveHookDBTxnRetries {
+			return false, ErrMaxRemoveTxnRetries
+		}
```
eaudetcobello, line 54: I would hardcode this to whatever you would have made the default value if it were configurable.

HomayoonAlimohammadi: I think it might. imo it's harmless to keep this around as a config. 
for others seeing this: more discussions about this one on MM

eaudetcobello (approved):

eaudetcobello (changes requested):
revoking since we're not sure what the best way forward is

berkayoz:
I believe we don't need the `control.WaitUntilReady` anymore, this is a copy/carry over from the past where the cleanup [happened in the join/bootstrap hooks ](https://github.com/canonical/k8s-snap/blob/032a5836d01450b7db612f83ddcb0fb9bf34c4f0/src/k8s/pkg/k8sd/app/hooks_join.go#L53-L96). The `reverter` defers the cleanup so should only happen after join function finishes execution.

However this still does not address getting a `context deadline exceeded`. The `NewCluster` or `JoinCluster` calls are just performing http requests which means any hangup in getting a response will result in a context deadline exceeded. Easiest approach although not as UX friendly would be to log errors and direct users to check the logs.

HomayoonAlimohammadi:
@berkayoz I think it's not safe to remove the `control.WaitUntilReady` since it's being called on `remove-node` as well. I still think the best way to go is finish retrying in case we see `force == true` flag. but I think @neoaggelos believes `force` shouldn't be used like this and for this purpose, to which I kind of agree.

bschimke95 (commented):
I agree with Angelos here. We should not use the `force` flag in this way. `force` means "kill that node no matter if anything is running" but this implementation would not do this if the database error is transient causing subtle errors down the line.

Maybe something like this makes sense?:
```
    waitForDBCtx := ctx
    deadline, ok := ctx.Deadline()
    if ok {
      // Calculate half of the remaining time
      remainingTime := time.Until(deadline)
      halfTime := remainingTime / 2

      // Create a new context with half the deadline time
      waitForDBCtx, cancel = context.WithDeadline(ctx, time.Now().Add(halfTime))
      defer cancel()
    }
```

HomayoonAlimohammadi:
context for other reviewers: 
> I agree with Angelos here. We should not use the `force` flag in this way. `force` means "kill that node no matter if anything is running" but this implementation would not do this if the database error is transient causing subtle errors down the line.
> 
> Maybe something like this makes sense?:
> 
> ```
>     waitForDBCtx := ctx
>     deadline, ok := ctx.Deadline()
>     if ok {
>       // Calculate half of the remaining time
>       remainingTime := time.Until(deadline)
>       halfTime := remainingTime / 2
> 
>       // Create a new context with half the deadline time
>       waitForDBCtx, cancel = context.WithDeadline(ctx, time.Now().Add(halfTime))
>       defer cancel()
>     }
> ```

unfortunately we don't have a `ctx.Deadline()` in the `onPreRemove` hook call.

HomayoonAlimohammadi:
switched back to the db retries as discussed

src/k8s/cmd/k8sd/k8sd.go
bschimke95, line 76: Do we really expect the user to ever change this configuration? Let's not over-engineer this and just hardcode the retries

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -40,10 +46,23 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
+
 		return notPending, nil
 	})
 
+	if errors.Is(err, ErrMaxRemoveTxnRetries) {
```
bschimke95, line 59: Is there any way for this to be an other error message? If there is an error, the query is retried, no?

Also, even if there would be another error - we don't care. This is the cleanup phase that runs on a best effort principle. Any error is just acknowledged/logged and we keep on going.

bschimke95 (commented):
Good work @HomayoonAlimohammadi but let's keep things simple. If we start to wire out every tiny configuration to the user than the amount of flags just explodes.

src/k8s/pkg/k8sd/app/hooks_remove.go
HomayoonAlimohammadi: We catch the error here to make sure if we encountered an error in future changes, we're not going to log the wrong message. I agree with the best effort and will make related changes.

bschimke95: I would just move the log from line 62 to line 54 (before returning due to max tries) and then just 
```
if err != nil {
		log.Error(err, "Failed to wait for node to finish microcluster join before removing. Continuing with the cleanup...")
	}

bschimke95 (approved):
LGTM, great work!

bschimke95:
Please rebase/fix the go test, then we are good to merge

PR #609, opened 2024-08-15
Switch to microcluster fork without join name verification

berkayoz:
Switch to microcluster v2 fork without join name verification to unblock CAPI work while we work towards a proper solution.


bschimke95 (approved):
LGTM

PR #610, opened 2024-08-16
Implement Control Plane Certificates Renewal Endpoints

mateoflorido:
## Overview
Implement the required endpoints for refreshing certificates for Kubernetes Control Plane nodes.
## Rationale
As stated in [Proposal 002](https://github.com/canonical/k8s-snap/blob/main/docs/proposals/002-refresh-certificates.md). We need to provide two endpoints to allow certificates refreshes in the cluster. This pull request implements the `refresh-certs/plan` and the `refresh-certs/run` for refreshing the control plane nodes' certificates.
## Testing
This pull request has been tested manually. Automated testing will be added in a future pull request.
```
root@t-1:~# curl --unix-socket /var/snap/k8s/common/var/lib/k8sd/state/control.socket http://localhost/1.0/k8sd/refresh-certs/run -XPOST -H "Content-Type: application/json" --data '{"seed":1978322463, "expiration-seconds": 3600}'
{"type":"sync","status":"Success","status_code":200,"operation":"","error_code":0,"error":"","metadata":{"expiration-seconds":3600}}
root@t-1:~# openssl x509 -in /etc/kubernetes/pki/apiserver.crt -noout -dates
notBefore=Aug 16 17:40:59 2024 GMT
notAfter=Aug 16 18:40:59 2024 GMT
```


src/k8s/pkg/k8sd/pki/k8sdqlite.go
```diff
@@ -24,20 +24,20 @@ type K8sDqlitePKIOpts struct {
...
 	AllowSelfSignedCA bool
 	Datastore         string
 }
 
 func NewK8sDqlitePKI(opts K8sDqlitePKIOpts) *K8sDqlitePKI {
-	if opts.Years == 0 {
-		opts.Years = 1
+	if opts.Seconds == 0 {
+		opts.Seconds = 365 * 24 * 60 * 60
 	}
```
addyess, line 35: OK, i hate to be "that guy" -- but these constants aren't always technically a "year"

addyess: I know it's wordy but at least it's exactly a year from now

```suggestion
	if opts.Seconds == 0 {
	    now := time.Now()
	    duration := now.AddDate(1, 0, 0).Sub(now)	
	    ops.Seconds = duration.Seconds()
	}
```

addyess (changes requested):
You should likely not take advice from me as a go developer

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -60,8 +60,97 @@ func (e *Endpoints) postRefreshCertsRun(s state.State, r *http.Request) response
...
+	if err := certificates.CompleteCertificates(); err != nil {
+		return response.InternalError(fmt.Errorf("failed to generate new control plane certificates: %w", err))
+	}
+
+	if _, err := setup.EnsureControlPlanePKI(snap, certificates); err != nil {
+		return response.InternalError(fmt.Errorf("failed to write control plane certificates: %w", err))
+	}
+
+	// Generate kubeconfigs
+	for _, kubeconfig := range []struct {
```
bschimke95, line 118: Let's use the existing util for this: https://github.com/canonical/k8s-snap/blob/67f8883a2fc929dcb80c8d7ed0b5437ec90b3e42/src/k8s/pkg/k8sd/app/cluster_util.go#L15

Or is there a reason not to?

src/k8s/pkg/k8sd/api/worker.go
```diff
@@ -36,7 +37,10 @@ func (e *Endpoints) postWorkerInfo(s state.State, r *http.Request) response.Resp
 		return response.InternalError(fmt.Errorf("failed to get cluster config: %w", err))
 	}
 
-	certificates := pki.NewControlPlanePKI(pki.ControlPlanePKIOpts{Years: 10})
+	// NOTE: Default certificate expiration is set to 10 years.
+	defaultExpiration := int(time.Now().AddDate(10, 0, 0).Sub(time.Now()) / time.Second)
```
bschimke95, line 41: nit: maybe this is slightly more readable: 
```
int(time.Until(time.Now().AddDate(10, 0, 0)) / time.Second)
```

src/k8s/pkg/k8sd/app/cluster_util.go
```diff
@@ -33,6 +33,27 @@ func setupKubeconfigs(s state.State, kubeConfigDir string, securePort int, pki p
 
 }
 
+func SetupControlPlaneKubeconfigs(s state.State, kubeConfigDir string, securePort int, pki pki.ControlPlanePKI) error {
```
bschimke95, line 36: this is an exact duplicate of the function `setupKubeconfigs` above. 
Also this is not used anywhere.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -280,10 +281,13 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
 
 	switch cfg.Datastore.GetType() {
 	case "k8s-dqlite":
+		// NOTE: Default certificate expiration is set to 20 years.
+		defaultDuration := int(time.Now().AddDate(20, 0, 0).Sub(time.Now()) / time.Second)
```
bschimke95, line 285: So, the main idea of this csr stuff is to implement a way to easily refresh certificates. That way, we can have those certificates expire early. 
I would expect the default to be way lower (a year?) or am I missing something here?

src/k8s/pkg/snap/util/services.go
```diff
@@ -26,6 +26,17 @@ var (
 	}
 )
 
+// RestartWorkerServices restarts the control plane services.
```
bschimke95, line 29: `RestartControlPlaneServices` - copy paste error :)

bschimke95 (commented):
great work @mateoflorido, added a couple of comments

src/k8s/pkg/k8sd/api/worker.go
neoaggelos: this will definitely pile up to off-by-one or a few seconds issues, see my other comment about how to deal with this

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
neoaggelos: `k8s-dqlite` certificate cannot be refreshed like this, as we need to coordinate this across the whole cluster.

src/k8s/pkg/k8sd/api/certificates_refresh.go
mateoflorido: This will create an import cycle. We can work around this by creating a new utils package, which will contain the utilities for creating the kubeconfigs.

src/k8s/pkg/utils/pki/generate.go
```diff
@@ -22,7 +22,7 @@ func GenerateSerialNumber() (*big.Int, error) {
 	return serialNumber, nil
 }
 
-func GenerateCertificate(subject pkix.Name, years int, ca bool, dnsSANs []string, ipSANs []net.IP) (*x509.Certificate, error) {
+func GenerateCertificate(subject pkix.Name, notAfter time.Time, ca bool, dnsSANs []string, ipSANs []net.IP) (*x509.Certificate, error) {
```
neoaggelos, line 25: the notAfter depends on the current time, but a separate time.Now() is included in this function to do this. in this case, pass both notBefore and notAfter as parameters

src/k8s/pkg/k8sd/api/certificates_refresh.go
neoaggelos: you could move this to `setup` package

src/k8s/pkg/utils/time.go
```diff
@@ -0,0 +1,51 @@
...
+	case seconds%(365*24*60*60) == 0:
+		return now.AddDate(seconds/365/24/60/60, 0, 0)
+	case seconds%(30*24*60*60) == 0:
+		return now.AddDate(0, seconds/30/24/60/60, 0)
+	case seconds%(24*60*60) == 0:
+		return now.AddDate(0, 0, seconds/24/60/60)
+	default:
+		return now.Add(time.Duration(seconds) * time.Second)
+	}
+}
```
mateoflorido, line 23: ## Note
The reasoning behind this change is to standardize our internal definition of a year. While it may not reflect the exact length of a year, the time package will handle the adjustments when we add years, months, or days. This should be passed from the `CLI --> API endpoint --> PKI`

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -82,13 +90,17 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+
 	// Certificates
 	extraIPs, extraNames := utils.SplitIPAndDNSSANs(joinConfig.ExtraSANS)
 	certificates := pki.NewControlPlanePKI(pki.ControlPlanePKIOpts{
 		Hostname:                  s.Name(),
 		IPSANs:                    append(append([]net.IP{nodeIP}, serviceIPs...), extraIPs...),
 		DNSSANs:                   extraNames,
-		Years:                     20,
+		NotBefore:                 notBefore,
+		NotAfter:                  defaultDuration,
```
neoaggelos, line 103: you can inline it here `notBefore.AddDate(20, 0, 0)` also in other places.

src/k8s/pkg/utils/time_test.go
```diff
@@ -0,0 +1,109 @@
...
+			seconds: 10 * 60,
+			want:    now.Add(time.Duration(10*60) * time.Second),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewGomegaWithT(t)
+			got := utils.SecondsToExpirationDate(now, tt.seconds)
+			g.Expect(got).To(BeTemporally("==", tt.want))
```
neoaggelos, line 45: `To(Equal(t.want))`

src/k8s/pkg/utils/time_test.go
```diff
@@ -0,0 +1,109 @@
...
+		{
+			name:    "Test Default",
+			seconds: 10 * 60,
+			want:    now.Add(time.Duration(10*60) * time.Second),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewGomegaWithT(t)
```
neoaggelos, line 43: NewGomegaWithT is deprecated, use NewWithT

src/k8s/pkg/utils/time.go
```diff
@@ -0,0 +1,51 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+// SecondsToExpirationDate returns the expiration date from now plus the given
+// number of seconds.
+func SecondsToExpirationDate(now time.Time, seconds int) time.Time {
+	switch {
+	case seconds%(365*24*60*60) == 0:
```
neoaggelos, line 14: let's make these `monthSeconds`, `daySeconds`, `yearSeconds` consts

src/k8s/pkg/utils/time.go
```diff
@@ -0,0 +1,51 @@
...
+		return now.Add(time.Duration(seconds) * time.Second)
+	}
+}
+
+// TTLToSeconds converts a TTL string to seconds. The TTL string is a number
+// followed by a unit:
+// - y: years
+// - m: months
+// - d: days
+func TTLToSeconds(ttl string) (int, error) {
```
neoaggelos, line 30: we still need to parse proper durations, e.g. 8640h. maybe fallthrough to `time.ParseDuration`, or support `h` as well

src/k8s/pkg/utils/time.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("invalid TTL number: %w", err)
+	}
+
+	unit := strings.ToLower(ttl[len(ttl)-1:])
+	switch unit {
+	case "y":
+		return value * 365 * 24 * 60 * 60, nil
+	case "m":
+		return value * 30 * 24 * 60 * 60, nil
```
neoaggelos, line 45: i would greatly recommend against this, and instead rely on `time.ParseDuration` and document usage of values like `8640h`. The reason for this is that specifying `${foo}m` as a duration, people can very easily confuse this to be minutes, (which is what the standard Go library would parse this as). Changing it to months could be surprising to people.

This is just my warning, up to you to proceed as you see fit.

mateoflorido: That makes sense. I’ve updated the function to parse months using the `mo` prefix and added tests to check this. Thanks for pointing that out!

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -60,8 +61,83 @@ func (e *Endpoints) postRefreshCertsRun(s state.State, r *http.Request) response
...
+	}
+
+	if err := setup.SetupControlPlaneKubeconfigs(snap.KubernetesConfigDir(), clusterConfig.APIServer.GetSecurePort(), *certificates); err != nil {
+		return response.InternalError(fmt.Errorf("failed to generate control plane kubeconfigs: %w", err))
+
+	}
+
+	if err := snaputil.RestartControlPlaneServices(r.Context(), snap); err != nil {
+		return response.InternalError(fmt.Errorf("failed to restart control plane services: %w", err))
+	}
```
neoaggelos, line 129: not for this PR: We need to see why this breaks cilium-operator and resolve it

neoaggelos (approved):

src/k8s/pkg/utils/time.go
```diff
@@ -0,0 +1,78 @@
+package utils
```
addyess, line 1: @mateoflorido i think this package is what i was hoping for.  I like the approximations.

addyess (approved):
Thanks for entertaining my date/time concerns

PR #611, opened 2024-08-19
Add unit tests for csrsigning controller

HomayoonAlimohammadi:
I was inspired by the [upstream kubernetes](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/volumemanager/reconciler/reconciler_test.go).

### DIFF
--- a/src/k8s/pkg/client/k8s/mock/mock.go
+++ b/src/k8s/pkg/client/k8s/mock/mock.go
@@ -1,0 +1,106 @@
+package mock
+
+import (
+	"context"
+	"testing"
+
+	certv1 "k8s.io/api/certificates/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type getArgs struct {
+	name types.NamespacedName
+	obj  client.Object
+	opts []client.GetOption
+}
+
+type getRet struct {
+	csr certv1.CertificateSigningRequest
+	err error
+}
+
+type K8sMock struct {
+	client.Client
+
+	t    *testing.T
+	srcm *subResourceClientMock
+
+	getCalledWith getArgs
+	getReturns    getRet
+}
+
+func New(
+	t *testing.T,
+	srcm *subResourceClientMock,
+	getCSR certv1.CertificateSigningRequest,
+	getErr error,
+) *K8sMock {
+	return &K8sMock{
+		t:    t,
+		srcm: srcm,
+		getReturns: getRet{
+			csr: getCSR,
+			err: getErr,
+		},
+	}
+}
+
+func (m *K8sMock) Get(ctx context.Context, name types.NamespacedName, obj client.Object, opts ...client.GetOption) error {
+	m.getCalledWith = getArgs{name, obj, opts}
+	csr, ok := obj.(*certv1.CertificateSigningRequest)
+	if !ok {
+		m.t.Fatalf("unexpected object type: %T", obj)
+	}
+	*csr = m.getReturns.csr
+	return m.getReturns.err
+}
+
+func (m *K8sMock) Status() client.SubResourceWriter {
+	return m.srcm
+}
+
+func (m *K8sMock) SubResource(subResource string) client.SubResourceClient {
+	switch subResource {
+	case "approval":
+		return m.srcm
+	default:
+		m.t.Fatalf("unexpected subResource: %s", subResource)
+	}
+	return nil
+}
+
+func (m *K8sMock) AssertUpdateCalled(t *testing.T) {
+	m.srcm.assertUpdateCalled(t)
+}
+
+type updateArgs struct {
+	obj  client.Object
+	opts []client.SubResourceUpdateOption
+}
+
+type subResourceClientMock struct {
+	client.SubResourceClient
+
+	updateCalledWith []updateArgs
+	updateErr        error
+}
+
+func NewSubResourceClientMock(
+	updateErr error,
+) *subResourceClientMock {
+	return &subResourceClientMock{
+		updateErr: updateErr,
+	}
+}
+
+func (src *subResourceClientMock) Update(ctx context.Context, obj client.Object, opts ...client.SubResourceUpdateOption) error {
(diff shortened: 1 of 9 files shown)


bschimke95:
please resolve conflicts

neoaggelos (dismissed):

src/k8s/pkg/k8sd/controllers/csrsigning/controller.go
```diff
@@ -92,7 +95,14 @@ func (c *Controller) Run(ctx context.Context, getClusterConfig func(context.Cont
...
 
-		getClusterConfig: getClusterConfig,
+		getClusterConfig:       getClusterConfig,
+		validateCSR:            validateCSR,
+		loadCertificateRequest: pkiutil.LoadCertificateRequest,
+		generateSerialNumber:   pkiutil.GenerateSerialNumber,
+		loadCertificate:        pkiutil.LoadCertificate,
+		createCertificate:      x509.CreateCertificate,
+		encodeToMemory:         pem.EncodeToMemory,
+		nowFn:                  time.Now,
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: thanks for the comment @neoaggelos. I tried to change the tests to test the behavior rather than code paths. I completely agree on not to follow the code paths in general, but here it feels like there are a lot of things to worry about. specially that we have different things happening for different signer names (kubelet-serving, kubelet-client, ...). I agree that the initial tests that I added were really brittle, but I feel like this controller itself is a bit too brittle. anyways, changed the tests, lemme know if they need further improvements.

src/k8s/pkg/k8sd/controllers/csrsigning/reconcile_approve_test.go
```diff
@@ -0,0 +1,123 @@
...
+		csr         certv1.CertificateSigningRequest
+		validateCSR func(obj *certv1.CertificateSigningRequest, priv *rsa.PrivateKey) error
+		updateErr   error
+
+		expCtrl      ctrl.Result
+		expErr       error
+		expCondition certv1.CertificateSigningRequestCondition
+	}{
+		{
+			name: "InvalidCSR--UpdateSuccessful",
```
neoaggelos, line 30: Test names are split with `/`

```suggestion
			name: "InvalidCSR/UpdateSuccessful",
```

src/k8s/pkg/k8sd/controllers/csrsigning/reconcile_approve_test.go
```diff
@@ -0,0 +1,123 @@
...
+			csr:  certv1.CertificateSigningRequest{},
+			validateCSR: func(obj *certv1.CertificateSigningRequest, priv *rsa.PrivateKey) error {
+				return errors.New("invalid")
+			},
+			updateErr: nil,
+			expCtrl:   ctrl.Result{},
+			expErr:    nil,
```
neoaggelos, line 37: zero values can be omitted

src/k8s/pkg/k8sd/controllers/csrsigning/reconcile_approve_test.go
```diff
@@ -0,0 +1,123 @@
...
+func TestAutoApprove(t *testing.T) {
+	for _, tc := range []struct {
+		name        string
+		csr         certv1.CertificateSigningRequest
+		validateCSR func(obj *certv1.CertificateSigningRequest, priv *rsa.PrivateKey) error
+		updateErr   error
+
+		expCtrl      ctrl.Result
+		expErr       error
+		expCondition certv1.CertificateSigningRequestCondition
```
neoaggelos, line 27: expectErr, expectResult, expectCondition read better

src/k8s/pkg/client/k8s/mock/mock.go
```diff
@@ -0,0 +1,106 @@
...
+) *K8sMock {
+	return &K8sMock{
+		t:    t,
+		srcm: srcm,
+		getReturns: getRet{
+			csr: getCSR,
+			err: getErr,
+		},
+	}
+}
```
neoaggelos, line 47: follow the existing idioms of what this mocks. thi sis not a mock for the k8s client package. in this case, just inline it in the test package.

src/k8s/pkg/k8sd/controllers/csrsigning/reconcile_approve.go
```diff
@@ -9,9 +9,12 @@ import (
 	certv1 "k8s.io/api/certificates/v1"
 	v1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func (r *csrSigningReconciler) reconcileAutoApprove(ctx context.Context, log log.Logger, csr *certv1.CertificateSigningRequest, priv *rsa.PrivateKey) (ctrl.Result, error) {
+func reconcileAutoApprove(ctx context.Context, log log.Logger, csr *certv1.CertificateSigningRequest,
+	priv *rsa.PrivateKey, client client.Client,
+	validateCSR func(obj *certv1.CertificateSigningRequest, priv *rsa.PrivateKey) error) (ctrl.Result, error) {
```
neoaggelos, line 17: be extra wary of situations like this, where things are either passed as arguments or available through other means. it would be very easy for future code to not use the `validateCSR` method, but rather rely on the `valiateCSR` of the package.

neoaggelos: Instead of passing these as arguments, rather replace `r.Client` in the test code.

src/k8s/pkg/client/k8s/mock/mock.go
HomayoonAlimohammadi: we're not testing the csrsigning controller from a test pacakge, it's the same package `csrsigning`, not `csrsigning_test`. is it ok to have something like `.../controllers/csrsigning/test/k8s_mock.go`? I feel like adding a `k8s_mock.go` right into the `csrsigning` package is not the best thing to do.

bschimke95 (approved):
LGTM

Issue #612, opened 2024-08-19
K8s snap leaks NFS volume mounts after removal


claudiubelu:
### Summary

If a local NFS CSI provider is used, and the k8s snap is removed afterwards, it may leak any NFS volume mounts, and also increase the snap removal time significantly.

Not only that, due to the volume leak, the ``/var/lib/kubelet`` folder is never cleaned up properly either, which means that on reinstalling and bootstrapping the cluster, the current node will not be registered to the new cluster.

### What Should Happen Instead?

Volumes should not be leaked and the ``/var/lib/kubelet`` folder should be cleaned up on snap removal.

### Reproduction Steps

1. Install k8s snap: ``sudo snap install k8s --channel=1.30-classic/beta --classic``
2. Deploy NFS server.
3. Deploy CSI provider.
4. Deploy a Pod with a PVC.
5. Uninstall k8s snap.
6. Run: ``cat /proc/mounts | grep kubelet``
7. Observe leaked NFS mount:

```
cat /proc/mounts  | grep kubelet
nfs-server.default.svc.cluster.local:/ /var/lib/kubelet/pods/130ef4eb-b053-4f27-9cf3-084ce668e8b9/volumes/kubernetes.io~csi/pv-nginx/mount nfs4 rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,clientaddr=172.19.198.120,local_lock=none,addr=10.152.183.186 0 0
```

Alternatively, the following can be run instead to replicate the issue:

```
# Clone repo.
git clone https://github.com/canonical/csi-driver-nfs-rocks
cd csi-driver-nfs-rocks/tests

# Rock image required for integration test.
[...]

PR #613, opened 2024-08-19
Forces volume umount when snap is removed

claudiubelu:
Some Kubernetes volumes may not be easily removed (e.g.: NFS volumes from a local NFS provider), leading to a significantly increased snap removal time and leaked volume mounts.

Due to the volume leak, the ``/var/lib/kubelet`` folder is never cleaned up properly either, which means that on reinstalling the k8s snap afterwards and bootstrapping the cluster, the current node will not be registered to the cluster.

This adds the ``-f`` option to ``umount``, forcing the unmount.

Fixes: https://github.com/canonical/k8s-snap/issues/612

### DIFF
--- a/k8s/lib.sh
+++ b/k8s/lib.sh
@@ -60,9 +60,17 @@ k8s::remove::containers() {
   # delete cni network namespaces
   ip netns list | cut -f1 -d' ' | grep -- "^cni-" | xargs -n1 -r -t ip netns delete || true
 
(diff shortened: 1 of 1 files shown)


bschimke95 (dismissed):
LGTM

neoaggelos (approved):
LGTM, make sure the e2e test that validates all volumes go away works

bschimke95:
Thanks @claudiubelu!

PR #614, opened 2024-08-20
Delay Gateway API CRD Removal Until After Cilium Update

bschimke95:
Previously, Gateway API CRDs were installed and removed at the start of enabling or disabling the Gateway feature.
This approach caused issues when the Gateway feature was disabled, as the CRDs were removed too early, leading to failures in Cilium's cleanup process due to the absence of these CRDs.

This commit ensures that Gateway API CRDs are only removed after Cilium has been updated, preventing cleanup failures.

### DIFF
--- a/src/k8s/pkg/k8sd/features/cilium/gateway.go
+++ b/src/k8s/pkg/k8sd/features/cilium/gateway.go
@@ -25,20 +25,13 @@ const (
 func ApplyGateway(ctx context.Context, snap snap.Snap, gateway types.Gateway, network types.Network, _ types.Annotations) (types.FeatureStatus, error) {
 	m := snap.HelmClient()
 
-	if _, err := m.Apply(ctx, chartGateway, helm.StatePresentOrDeleted(gateway.GetEnabled()), nil); err != nil {
-		if gateway.GetEnabled() {
+	if gateway.GetEnabled() {
+		if _, err := m.Apply(ctx, chartGateway, helm.StatePresent, nil); err != nil {
 			err = fmt.Errorf("failed to install Gateway API CRDs: %w", err)
 			return types.FeatureStatus{
 				Enabled: false,
 				Version: ciliumAgentImageTag,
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/k8sd/features/cilium/gateway.go
```diff
@@ -81,6 +74,19 @@ func ApplyGateway(ctx context.Context, snap snap.Snap, gateway types.Gateway, ne
...
+			err = fmt.Errorf("failed to delete Gateway API CRDs: %w", err)
+			return types.FeatureStatus{
+				Enabled: false,
+				Version: ciliumAgentImageTag,
+				Message: fmt.Sprintf(gatewayDeleteFailedMsgTmpl, err),
+			}, err
+		}
+	}
+
 	if !changed {
```
neoaggelos, line 90: if the enable/disable code paths start drifting, better move them into separate enableGateway and disableGateway functions

src/k8s/pkg/k8sd/features/cilium/gateway.go
```diff
@@ -125,3 +85,52 @@ func ApplyGateway(ctx context.Context, snap snap.Snap, gateway types.Gateway, ne
...
+			Version: ciliumAgentImageTag,
+			Message: disabledMsg,
+		}, nil
+	}
+
+	return types.FeatureStatus{
+		Enabled: false,
+		Version: ciliumAgentImageTag,
+		Message: disabledMsg,
+	}, nil
```
neoaggelos, line 135: this looks wrong. if we changed cilium, we need to rollout restart. if not, just return early

neoaggelos (changes requested):

src/k8s/pkg/k8sd/features/cilium/gateway.go
bschimke95: True, that slipped through. Thanks!

neoaggelos (approved):

PR #615, opened 2024-08-21
Add k8s endpoint check to markNodeReady

HomayoonAlimohammadi:
# Summary
`onStart` hook happens before `onBootstrap`. because of this, on non-fresh machines (non-fresh == `/etc/kubernetes/admin.conf` is available) we use invalid/old `admin.conf` kubeconfig for csrsigining controller client. this PR makes sure that we prevent running controllers until we have the correct `.conf` files and we can reach the k8s cluster.

# How to test
build and install k8s on a fresh machine, run `bootstrap` and check logs and confirm the csrsigning controller is running, e.g.:
```text
Aug 21 08:28:20 test k8s.k8sd[1642]: I0821 08:28:20.897429    1642 controller/controller.go:181] "Starting Controller" logger="k8sd.csrsigning" controller="certificatesignin
grequest" controllerGroup="certificates.k8s.io" controllerKind="CertificateSigningRequest"
Aug 21 08:28:21 test k8s.k8sd[1642]: I0821 08:28:21.005667    1642 controller/controller.go:215] "Starting workers" logger="k8sd.csrsigning" controller="certificatesigningre
quest" controllerGroup="certificates.k8s.io" controllerKind="CertificateSigningRequest" worker count=1
```
also confirm that there are kubeconfigs available in `/etc/kubernetes/`, specifically `admin.conf`.
now remove the k8s snap and reinstall k8s (same snap). Run `bootstrap` and like above confirm that csrsigning controller is started and running.

### DIFF
--- a/src/k8s/pkg/client/kubernetes/status.go
+++ b/src/k8s/pkg/client/kubernetes/status.go
@@ -20,6 +20,16 @@ func (c *Client) WaitKubernetesEndpointAvailable(ctx context.Context) error {
 	})
 }
 
+func (c *Client) CheckKubernetesEndpoint(ctx context.Context) error {
+	if endpoint, err := c.CoreV1().Endpoints("default").Get(ctx, "kubernetes", metav1.GetOptions{}); err != nil {
+		return fmt.Errorf("failed to get kubernetes endpoint: %w", err)
+	} else if endpoint == nil {
+		return fmt.Errorf("kubernetes endpoint not found")
+	}
+
+	return nil
+}
+
 // HasReadyNodes checks the status of all nodes in the Kubernetes cluster.
 // HasReadyNodes returns true if there is at least one Ready node in the cluster, false otherwise.
 func (c *Client) HasReadyNodes(ctx context.Context) (bool, error) {
--- a/src/k8s/pkg/k8sd/app/app.go
+++ b/src/k8s/pkg/k8sd/app/app.go
@@ -15,6 +15,7 @@ import (
 	"github.com/canonical/k8s/pkg/k8sd/database"
 	"github.com/canonical/k8s/pkg/log"
 	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/k8s/pkg/utils/control"
 	"github.com/canonical/microcluster/v2/client"
 	"github.com/canonical/microcluster/v2/microcluster"
 	"github.com/canonical/microcluster/v2/state"
@@ -228,17 +229,28 @@ func (a *App) Run(ctx context.Context, customHooks *state.Hooks) error {
 	return nil
 }
 
-func (a *App) markNodeReady(ctx context.Context, s state.State) {
-	for {
-		if err := s.Database().IsOpen(ctx); err == nil {
-			a.readyWg.Done()
-			return
-		}
-
-		select {
-		case <-ctx.Done():
-			return
-		case <-time.After(3 * time.Second):
-		}
+func (a *App) markNodeReady(ctx context.Context, s state.State) error {
+	// wait for the database to be open
+	if err := control.WaitUntilReady(ctx, func() (bool, error) {
+		return s.Database().IsOpen(ctx) == nil, nil
+	}); err != nil {
+		return fmt.Errorf("failed to wait for database to be open: %w", err)
 	}
+
+	// check kubernetes endpoint
+	if err := control.WaitUntilReady(ctx, func() (bool, error) {
+		client, err := a.snap.KubernetesNodeClient("")
+		if err != nil {
+			return false, nil
+		}
+		if err := client.CheckKubernetesEndpoint(ctx); err != nil {
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		return fmt.Errorf("failed to wait for kubernetes endpoint: %w", err)
+	}
+
+	a.readyWg.Done()
+	return nil
 }
--- a/src/k8s/pkg/k8sd/app/hooks_start.go
+++ b/src/k8s/pkg/k8sd/app/hooks_start.go
@@ -10,6 +10,7 @@ import (
 	"github.com/canonical/k8s/pkg/k8sd/database"
 	databaseutil "github.com/canonical/k8s/pkg/k8sd/database/util"
 	"github.com/canonical/k8s/pkg/k8sd/types"
+	"github.com/canonical/k8s/pkg/log"
 	"github.com/canonical/k8s/pkg/utils"
 	pkiutil "github.com/canonical/k8s/pkg/utils/pki"
 	"github.com/canonical/microcluster/v2/state"
@@ -17,7 +18,11 @@ import (
 
 func (a *App) onStart(ctx context.Context, s state.State) error {
 	// start a goroutine to mark the node as running
-	go a.markNodeReady(ctx, s)
+	go func() {
+		if err := a.markNodeReady(ctx, s); err != nil {
+			log.FromContext(ctx).Error(err, "Failed to mark node as ready")
+		}
+	}()
 
 	// start node config controller
 	if a.nodeConfigController != nil {


neoaggelos (changes requested):
I do not think this is what we need here. The `onStart` hook runs every time k8sd starts. Will this not remove the kubeconfigs of a running cluster?

HomayoonAlimohammadi:
right, I didn't think about that. what if we run the csrsigning controller in `onBootstrap` hook? @neoaggelos

neoaggelos:
Perhaps an alternative would be to block the csrsigning controller starting before we can validate that we have a proper kubeconfig in place

neoaggelos:
For example, we could return the rest config from here https://github.com/canonical/k8s-snap/blob/2994b879eab39b782556f122a2c8512bf85e9aca/src/k8s/pkg/k8sd/controllers/csrsigning/controller.go#L46 only if we can successfully call an endpoint (e.g. `/readyz` or a `GetNode()`), otherwise keep looping.

e.g. here is what we do on bootstrap (without the loop, as we don't want to keep using stale kubeconfigs) https://github.com/canonical/k8s-snap/blob/db5015ec88663f107e40accd0d36ed7c082991f1/src/k8s/pkg/client/kubernetes/status.go#L17-L19

src/k8s/pkg/client/kubernetes/status.go
```diff
@@ -20,6 +21,25 @@ func (c *Client) WaitKubernetesEndpointAvailable(ctx context.Context) error {
 	})
 }
 
+// IsClusterReady pings /readyz endpoint and only reports ready if status code == 200.
+func (c *Client) IsClusterReady(ctx context.Context) bool {
+	logger := log.FromContext(ctx)
+	result := c.Discovery().RESTClient().Get().AbsPath("/readyz").Do(ctx)
+	if result.Error() != nil {
+		logger.Info("Failed to check /readyz endpoint", "error", result.Error())
```
neoaggelos, line 29: this could easily flood the logs with log lines that are completely out of context, same below.

If you think this is a useful log line, maybe make this a `.V(1)` or `.V(2)`

src/k8s/pkg/client/kubernetes/status.go
```diff
@@ -20,6 +21,25 @@ func (c *Client) WaitKubernetesEndpointAvailable(ctx context.Context) error {
 	})
 }
 
+// IsClusterReady pings /readyz endpoint and only reports ready if status code == 200.
+func (c *Client) IsClusterReady(ctx context.Context) bool {
+	logger := log.FromContext(ctx)
+	result := c.Discovery().RESTClient().Get().AbsPath("/readyz").Do(ctx)
```
neoaggelos, line 27: this is hard to mock test with. we can stick with querying for the endpoints instead

neoaggelos, line 27: let's stick with just checking for the endpoints

src/k8s/pkg/client/kubernetes/status.go
```diff
@@ -20,6 +21,25 @@ func (c *Client) WaitKubernetesEndpointAvailable(ctx context.Context) error {
 	})
 }
 
+// IsClusterReady pings /readyz endpoint and only reports ready if status code == 200.
+func (c *Client) IsClusterReady(ctx context.Context) bool {
```
neoaggelos, line 25: `IsClusterReady()` is probably not the right name for this. Perhaps `CheckReadyz() error` is preferred. Return the error and let users handle it, or return `nil` for success. That can reduce the need to log the results as well.

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -242,3 +243,35 @@ func (a *App) markNodeReady(ctx context.Context, s state.State) {
 		}
 	}
 }
+
+func (a *App) waitKubernetesReady(ctx context.Context) error {
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		a.readyWg.Wait()
```
neoaggelos, line 251: we already have a `a.readyWg`, but now we add another layer on top of that. in that case, let's drop the waitgroup altogether and handle everything in one place.

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -242,3 +243,35 @@ func (a *App) markNodeReady(ctx context.Context, s state.State) {
...
+		if err != nil {
+			return false, nil
+		}
+		if !client.IsClusterReady(ctx) {
+			return false, nil
+		}
+
+		return true, nil
+	})
+	if err != nil {
```
neoaggelos, line 272: inline

```
if err := control.WaitUntilReady(...
); err != nil {
```

src/k8s/pkg/k8sd/controllers/csrsigning/controller.go
```diff
@@ -19,14 +19,14 @@ import (
 
 type Controller struct {
 	snap      snap.Snap
-	waitReady func()
+	waitReady func(context.Context) error
```
neoaggelos, line 22: rather than this, since we have the `waitKubernetesReady()` method, let's block on that and then only start the controllers after it's done. therefore, we do not need to pass the `waitReady` to each individual controller.

neoaggelos (commented):
Going in the right direction, but let's keep things simple rather than add layers on top of layers

src/k8s/pkg/client/kubernetes/status.go
HomayoonAlimohammadi: I'm not sure I understand "querying for the endpoints" correctly, are we going to do something like this? :
```go
endpoint, err := c.CoreV1().Endpoints("default").Get(ctx, "kubernetes", metav1.GetOptions{})
```
in which case the name for this function should not be `CheckReadyz`, right?

neoaggelos: right, pick a name that communicates what the function does

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -228,10 +219,10 @@ func (a *App) Run(ctx context.Context, customHooks *state.Hooks) error {
 	return nil
 }
 
-func (a *App) markNodeReady(ctx context.Context, s state.State) {
+func (a *App) waitForDBOpen(ctx context.Context, s state.State, done chan struct{}) {
 	for {
 		if err := s.Database().IsOpen(ctx); err == nil {
-			a.readyWg.Done()
+			defer close(done)
```
neoaggelos, line 225: dont defer, just close

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -242,3 +233,34 @@ func (a *App) markNodeReady(ctx context.Context, s state.State) {
...
+	dbOpen := make(chan struct{})
+	go a.waitForDBOpen(ctx, s, dbOpen)
+
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("failed to wait for db to be open: %w", ctx.Err())
+	case <-dbOpen:
+	}
+
+	logger := log.FromContext(ctx)
```
neoaggelos, line 247: lets stick with log everywhere, not mix logger and log

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -242,3 +233,34 @@ func (a *App) markNodeReady(ctx context.Context, s state.State) {
...
+
+	// also check kubernetes endpoint
+	if err := control.WaitUntilReady(ctx, func() (bool, error) {
+		client, err := a.snap.KubernetesClient("")
```
neoaggelos, line 251: caredul, this creates an admin client, which will fail on worker nodes

perhaps you can use a node client instead

neoaggelos (changes requested):

src/k8s/pkg/k8sd/app/app.go
HomayoonAlimohammadi: my bad

src/k8s/pkg/k8sd/app/app.go
HomayoonAlimohammadi: right, and I guess we should get the config for csrsigning controller from a node client as well?

HomayoonAlimohammadi:
adding a centralized wait didn't seem to be possible. in order to mark the node as ready `onStart` should finish so we couldn't wait there. I think adding the k8s check on markNodeReady is the best way to go.

bschimke95 (approved):
LGTM

neoaggelos (approved):

PR #616, opened 2024-08-21
[draft] use metallb for loadbalancers

neoaggelos:
## Summary

Try switching to metallb for loadbalancer


neoaggelos:
OK, seems like Gateway and Ingress do require Cilium socketLB, so this is not an option.

PR #620, opened 2024-08-23
Add control-plane token expiry

bschimke95:
Microcluster v2 supports token TTL. This PR exposes this configuration option to the API/CLI.

Note that is *only* applicable for the control-plane tokens right now. Workers will be done in separate work (since we don't use microcluster for that it is more involved as we need to implement the logic ourselves).

This PR requires the (backward-compatible) update in the `k8s-snap-api`. 
https://github.com/canonical/k8s-snap-api/pull/5
Once this PR is reviewed and approved, I will merge the API change, create a new tag (`1.0.3`) and update this `k8s-snap` PR to use the new API version.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_get_join_token.go
+++ b/src/k8s/cmd/k8s/k8s_get_join_token.go
@@ -9,10 +9,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const minTokenTTL = 3 * time.Second
+
 func newGetJoinTokenCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 	var opts struct {
 		worker  bool
 		timeout time.Duration
+		ttl     time.Duration
 	}
 	cmd := &cobra.Command{
 		Use:    "get-join-token <node-name>",
@@ -30,6 +33,11 @@ func newGetJoinTokenCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				opts.timeout = minTimeout
 			}
 
+			if opts.ttl < minTokenTTL {
+				cmd.PrintErrf("Token expiration time %v is less than minimum of %v. Using the minimum %v instead.\n", opts.ttl, minTokenTTL, minTokenTTL)
+				opts.ttl = minTokenTTL
+			}
+
 			client, err := env.Snap.K8sdClient("")
 			if err != nil {
 				cmd.PrintErrf("Error: Failed to create a k8sd client. Make sure that the k8sd service is running.\n\nThe error was: %v\n", err)
@@ -39,7 +47,7 @@ func newGetJoinTokenCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 
 			ctx, cancel := context.WithTimeout(cmd.Context(), opts.timeout)
 			cobra.OnFinalize(cancel)
-			token, err := client.GetJoinToken(ctx, apiv1.GetJoinTokenRequest{Name: name, Worker: opts.worker})
+			token, err := client.GetJoinToken(ctx, apiv1.GetJoinTokenRequest{Name: name, Worker: opts.worker, TTL: opts.ttl})
 			if err != nil {
 				cmd.PrintErrf("Error: Could not generate a join token for %q.\n\nThe error was: %v\n", name, err)
 				env.Exit(1)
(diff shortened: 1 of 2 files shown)


src/k8s/cmd/k8s/k8s_get_join_token.go
```diff
@@ -52,5 +60,7 @@ func newGetJoinTokenCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 
 	cmd.Flags().BoolVar(&opts.worker, "worker", false, "generate a join token for a worker node")
 	cmd.Flags().DurationVar(&opts.timeout, "timeout", 90*time.Second, "the max time to wait for the command to execute")
+	// The CLI uses verbose names for flags instead of abbreviations. Internally and for the API, the common TTL (time-to-live) name is used.
+	cmd.Flags().DurationVar(&opts.ttl, "expires-in", time.Hour, "the time until the token expires")
```
eaudetcobello, line 64: The API defines the default as 24h but here it is defined as 1 hour? Can we make these consistent?

addyess: agree here it should be 24h.  That next machine doesn't always start at the speed you want it to.  Factor in installing the snap, going through a proxy -- darn, the token expired...  I like 24h here

src/k8s/cmd/k8s/k8s_get_join_token.go
```diff
@@ -9,10 +9,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const minTokenTTL = 3 * time.Second
```
addyess, line 12: ok, so you can't make a token that is valid for 2 seconds but 4 seconds is fine?  Am i reading that right... no complaints just making sure i understand.

src/k8s/cmd/k8s/k8s_get_join_token.go
bschimke95: fair point - updated

src/k8s/cmd/k8s/k8s_get_join_token.go
bschimke95: So, I added this because we also do this for the `timeout` flag. My assumption was that this is done to avoid error if the user assumes a unit (e.g. seconds) and does not set them in the flag. However, it seems like this would just fail:
```
➜ sudo k8s get-join-token test --expires-in 10
Error: invalid argument "10" for "--expires-in" flag: time: missing unit in duration "10"
Usage:
  k8s get-join-token <node-name> [flags]

Flags:
      --expires-in duration   the time until the token expires (default 24h0m0s)
  -h, --help                  help for get-join-token
      --timeout duration      the max time to wait for the command to execute (default 1m30s)
      --worker                generate a join token for a worker node
 ```
 
So, I don't see a reason to actually have this check - removed it.

File: src/k8s/pkg/k8sd/api/cluster_tokens.go [hunk 3]
```diff
@@ -54,9 +54,11 @@ func getOrCreateJoinToken(ctx context.Context, m *microcluster.MicroCluster, tok
 		fmt.Println("No token exists yet. Creating a new token.")
 	}
 
-	// if token does not exist, create a new one
-	// TODO(ben): make token expiry configurable
-	token, err := m.NewJoinToken(ctx, tokenName, 24*time.Hour)
+	if ttl == 0 {
```
HomayoonAlimohammadi, line 57: This feels like we're having 2 defaults being set in 2 different places. Maybe we can throw an error here since we know `req.TTL` should be set (otherwise it's an error on the k8sd caller side that we need to fix)? After all k8s snap and k8sd are shipped together, we're not doing this for backwards compatibility, right?

HomayoonAlimohammadi (commented):
Thanks a lot @bschimke95! Just a minor comment.

src/k8s/pkg/k8sd/api/cluster_tokens.go
bschimke95: Good point, but the CLI is not the only way you can configure/control `k8sd`. For example, the `k8s-operator` charm (https://github.com/canonical/k8s-operator/blob/113b69cc428aaf7e1ad338c5b71b660df3e062c3/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py#L171) controls the snap directly over the API.
So, we need to provide a default for those cases.

You could argue that we don't need to set the default in the CLI. However, I'd say that people should not need to go all the way to the back-end implementation to find out the default for this value. Since it would be fine if backend default and CLI default would diverge, I guess it is fine to set this twice.

HomayoonAlimohammadi: Oh, I didn't know about the operator. Thanks for the context.

HomayoonAlimohammadi (approved):
LGTM, thanks a lot!

PR #628, opened 2024-08-27
Multiarch image support for local storage and metrics-server

addyess:
### Overview

* bumps metrics-server from 0.7.0-ck0  -> 0.7.0-ck1
* bumps rawfile-localpv from 0.8.0-ck5  -> 0.8.0-ck4

### Details
Though it looks odd, i did confirm the rawfile-localpv had its "ck*" revision drop to a smaller number, it is in fact the correct image.  We need to look at the workflow to understand how it managed to generate a lower number than the previous ck5 release.


eaudetcobello (approved):
Thanks for this!

bschimke95 (approved):
LGTM

PR #633, opened 2024-08-28
k8sd snapd config sync: Wait for k8sd to come up

bschimke95:
We have machinery in place to sync the k8sd configuration with snapd config. This is reconciled via `k8s x-snapd-config config`. The reconciliation is triggered on every config change and in the snap `configure` hook. 

This causes issues on refresh because `k8sd` may not be up yet after restart and cause:
```
buntu@juju-32b103-6:~$ sudo snap refresh k8s
2024-08-28T14:51:52Z INFO Waiting for "snap.k8s.kube-apiserver.service" to stop.
error: cannot perform the following tasks:
- Run configure hook of "k8s" snap if present (run hook "configure": Error: failed to update k8sd state: failed to update k8s configuration: failed to PUT /k8sd/cluster/config: Database is not yet initialized)
```

This PR fixes this issue by making sure that `k8sd` is up before performing the reconciliation. 
```
k8s-snap on  bschimke95/meta-orb-snapd [$!] took 17,1s …
➜ sudo snap refresh k8s --channel=latest/edge/x-snapd-config --classic
2024-08-28T17:55:18+02:00 INFO Waiting for "snap.k8s.kube-apiserver.service" to stop.
k8s (edge/x-snapd-config) v1.30.3 from Canonical✓ refreshed
```
 
Note that this is not an issue on bootstrap because we disable the sync in that time and just enabled it in the bootstrap hook.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_x_snapd_config.go
+++ b/src/k8s/cmd/k8s/k8s_x_snapd_config.go
@@ -66,7 +66,7 @@ func newXSnapdConfigCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 			if err := control.WaitUntilReady(ctx, func() (bool, error) {
 				_, partOfCluster, err := client.NodeStatus(cmd.Context())
 				if !partOfCluster {
-					cmd.PrintErrf("Node is not part of a cluster: %v\n", err.Error())
+					cmd.PrintErrf("Node is not part of a cluster: %v\n", err)
 					env.Exit(1)
 				}
 				return err == nil, nil
--- a/src/k8s/cmd/k8s/k8s_x_snapd_config.go
+++ b/src/k8s/cmd/k8s/k8s_x_snapd_config.go
@@ -1,12 +1,20 @@
 package k8s
 
 import (
+	"context"
+	"time"
+
 	cmdutil "github.com/canonical/k8s/cmd/util"
+	"github.com/canonical/k8s/pkg/utils/control"
 	"github.com/canonical/k8s/pkg/utils/experimental/snapdconfig"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


src/k8s/cmd/k8s/k8s_x_snapd_config.go
```diff
@@ -41,17 +49,34 @@ func newXSnapdConfigCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
+				env.Exit(1)
+				return
+			}
+
+			ctx, cancel := context.WithTimeout(cmd.Context(), opts.timeout)
+			defer cancel()
+			if err := control.WaitUntilReady(ctx, func() (bool, error) {
+				_, partOfCluster, err := client.NodeStatus(cmd.Context())
+				if !partOfCluster {
+					cmd.PrintErrf("Node is not part of a cluster: %v\n", err.Error())
```
mateoflorido, line 69: Is it possible that the query returns specifically this: https://github.com/canonical/k8s-snap/blob/f1c8937629405cd7e99f24215ad6e13e117092e6/src/k8s/pkg/client/k8sd/status.go#L19-L20
I know that this only occurs when the node has not been initialized, but if we ran into this, it would panic.

berkayoz (commented):
Great work LGTM overall, I just have a nit with the error messages. Maybe being more verbose and stating `k8sd` instead of generic `Node` keyword is more insightful/better.

src/k8s/cmd/k8s/k8s_x_snapd_config.go
bschimke95: Good catch @mateoflorido 
I removed the pointer deref and let `%v` just do its magic. It should automatically resolve to the error string if an error exists.

bschimke95:
Thanks @berkayoz I updated the outer error message. The inner error message should be fine as the whole node is either part of a cluster or not.

berkayoz (approved):

PR #636, opened 2024-08-29
Adds SnapRefreshRPC and node token authentication 

berkayoz:
Changes:
* Added `ValidateNodeTokenAccessHandler` to authenticate per node endpoints
* Added SnapRefreshRPC implementation
* Update `k8s-snap-api` tag

Shouldn't be merged before https://github.com/canonical/k8s-snap-api/pull/7


src/k8s/pkg/k8sd/api/snap_refresh.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	refreshOpts, err := types.RefreshOptsFromAPI(req)
+	if err != nil {
+		return response.BadRequest(fmt.Errorf("invalid refresh options: %w", err))
+	}
+	log := log.FromContext(e.Context()).WithValues("to", refreshOpts)
+
+	readyCh := make(chan error)
+	go func() {
```
bschimke95, line 28: I think this part justifies a bit of commentary. 

```
// A snap refresh requires restarting services, including `k8sd`.
// If the refresh is executed during the request, it would restart the server and terminate the request, leading to an `EOF` error on the client side.
// To prevent this, the refresh is executed in a separate goroutine that waits until the response has been sent to the client.
// This approach means that the client will receive a successful response even if the subsequent refresh fails.
```

We don't inform the user somehow that the refresh actually succeeded, do we?

src/k8s/pkg/snap/pebble.go
```diff
@@ -52,6 +55,30 @@ func (s *pebble) RestartService(ctx context.Context, name string) error {
...
 }
 
+func (s *pebble) Refresh(ctx context.Context, to types.RefreshOpts) error {
+	switch {
+	case to.Revision != "":
+		return fmt.Errorf("pebble does not support refreshing to a revision, only a local path")
+	case to.Channel != "":
+		return fmt.Errorf("pebble does not support refreshing to a channel, only a local path")
+	case to.LocalPath != "":
+		// replace the "kubernetes" binary with the new source.
```
bschimke95, line 65: Huh, `LocalPath` will only upgrade the `kubernetes` binary?

What about all the other snap content?

bschimke95 (commented):
great work, added a couple of comments

src/k8s/pkg/snap/pebble.go
berkayoz: This is for pebble only and will be used for integration tests so we can check the kubernetes version change to verify. This gives us a smaller/more minimal image compared to having 2 copies of everything. We could implement this fully but pebble/image mode is not something we support officially.

bschimke95: Hm, but `LocalPath` is somewhat part of our official API. I see people trying to refresh to a local build snap and wonder why this is not behaving as expected

berkayoz: A local built snap will work just fine, see:
https://github.com/canonical/k8s-snap/blob/5f0c724546ab80ad53aaded7cb2b5400a61d1cb4/src/k8s/pkg/snap/snap.go#L84-L85

The pebble implementation is only used in CAPI testing with CAPD.

bschimke95: My bad!

bschimke95 (approved):
LGTM

src/k8s/cmd/k8s/k8s_x_capi.go
```diff
@@ -34,6 +36,26 @@ func newXCAPICmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 			}
 		},
 	}
+	setNodeToken := &cobra.Command{
+		Use:   "set-node-token <token>",
+		Short: "Set the node token to authenticate with per-node k8sd endpoints",
+		Args:  cmdutil.ExactArgs(env, 1),
+		Run: func(cmd *cobra.Command, args []string) {
+			token := args[0]
+			if token == "" {
```
bschimke95, line 45: I don't think that can happen if you set `cmdutil.ExactArgs`, can it?

src/k8s/cmd/k8s/k8s_x_capi.go
```diff
@@ -34,6 +36,26 @@ func newXCAPICmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
+				cmd.PrintErrf("Error: The token must be provided.\n")
+				env.Exit(1)
+				return
+			}
+
+			err := os.WriteFile(env.Snap.NodeTokenFile(), []byte(token), 0600)
+			if err != nil {
```
bschimke95, line 52: nit: inline err check

src/k8s/cmd/k8s/k8s_x_capi.go
berkayoz: It checks for the argument count but not what the argument is so `k8s x-capi set-node-token ""` can create an empty file.

bschimke95: Good to know, I though this would also be prevented so that "" would not count as an argument. TIL

PR #638, opened 2024-08-30
Pass independent context to DeleteClusterMember

HomayoonAlimohammadi:
Node removal process in CAPI might fail, we figured that the context passed to`DeleteClusterMember` is somehow getting canceled but couldn't figure out why or by which component. The cancellation happens after the `RunPreRemoveHook` call and before the `DeleteCoreClusterMember` call in `clusterMemberDelete` endpoint of microcluster. This is a workaround to avoid the cancellation. keep in mind that this failure is flaky and might not happen in every run.

### DIFF
--- a/src/k8s/pkg/k8sd/api/cluster_remove.go
+++ b/src/k8s/pkg/k8sd/api/cluster_remove.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"fmt"
 	"net/http"
+	"time"
 
 	apiv1 "github.com/canonical/k8s-snap-api/api/v1"
 	databaseutil "github.com/canonical/k8s/pkg/k8sd/database/util"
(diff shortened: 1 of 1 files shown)


bschimke95 (approved):
LGTM, thanks for all the investigation work on this issue!

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -66,7 +67,15 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
 		if err != nil {
 			return response.InternalError(fmt.Errorf("failed to create client to cluster leader: %w", err))
 		}
-		if err := c.DeleteClusterMember(ctx, req.Name, req.Force); err != nil {
+
+		// NOTE(hue): node removal process in CAPI might fail, we figured that the context passed to
+		// `DeleteClusterMember` is somehow getting canceled but couldn't figure out why or by which component.
```
claudiubelu, line 72: I think this may happen if the request is to remove itself from the cluster. In that case, I think the k8sd http and socket server would have to close, causing the connection to close, and I assume its context to be cancelled. This should be alleviated / resolved with some of the fixes we've included in microcluster, allowing connections to finish before gracefully shutting down the servers.

PR #640, opened 2024-08-30
Add unit tests for cilium network

HomayoonAlimohammadi:
# Summary
This PR adds unit tests for the Cilium network k8sd feature.
Some constants and variables are exposed to reduce duplication and testability.

### DIFF
--- a/src/k8s/pkg/k8sd/features/cilium/network_test.go
+++ b/src/k8s/pkg/k8sd/features/cilium/network_test.go
@@ -21,6 +21,8 @@ import (
 
 func TestDisabled(t *testing.T) {
 	t.Run("HelmApplyFails", func(t *testing.T) {
+		g := NewWithT(t)
+
 		applyErr := errors.New("failed to apply")
 		helmM := &helmmock.Mock{
 			ApplyErr: applyErr,
@@ -36,7 +38,6 @@ func TestDisabled(t *testing.T) {
 
 		status, err := cilium.ApplyNetwork(context.Background(), snapM, cfg, nil)
 
-		g := NewWithT(t)
 		g.Expect(err).To(MatchError(applyErr))
 		g.Expect(status.Enabled).To(BeFalse())
 		g.Expect(status.Message).To(ContainSubstring(applyErr.Error()))
@@ -49,6 +50,8 @@ func TestDisabled(t *testing.T) {
 		g.Expect(callArgs.Values).To(BeNil())
 	})
 	t.Run("Success", func(t *testing.T) {
+		g := NewWithT(t)
+
 		helmM := &helmmock.Mock{}
 		snapM := &snapmock.Snap{
 			Mock: snapmock.Mock{
@@ -61,7 +64,6 @@ func TestDisabled(t *testing.T) {
 
 		status, err := cilium.ApplyNetwork(context.Background(), snapM, cfg, nil)
 
-		g := NewWithT(t)
 		g.Expect(err).ToNot(HaveOccurred())
 		g.Expect(status.Enabled).To(BeFalse())
 		g.Expect(status.Message).To(Equal(cilium.DisabledMsg))
@@ -77,6 +79,8 @@ func TestDisabled(t *testing.T) {
 
 func TestEnabled(t *testing.T) {
 	t.Run("InvalidCIDR", func(t *testing.T) {
+		g := NewWithT(t)
+
 		helmM := &helmmock.Mock{}
 		snapM := &snapmock.Snap{
 			Mock: snapmock.Mock{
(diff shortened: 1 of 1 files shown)


bschimke95 (approved):
LGTM

src/k8s/pkg/k8sd/features/cilium/network_test.go
```diff
@@ -0,0 +1,179 @@
...
+				HelmClient: helmM,
+			},
+		}
+		cfg := types.Network{
+			Enabled: ptr.To(false),
+		}
+
+		status, err := cilium.ApplyNetwork(context.Background(), snapM, cfg, nil)
+
+		g := NewWithT(t)
```
eaudetcobello, line 39: We put this after t.Run

src/k8s/pkg/k8sd/features/cilium/network_test.go
```diff
@@ -0,0 +1,179 @@
...
```
eaudetcobello, line 64: Move after t.Run

src/k8s/pkg/k8sd/features/cilium/network_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+		callArgs := helmM.ApplyCalledWith[0]
+		g.Expect(callArgs.Chart).To(Equal(cilium.ChartCilium))
+		g.Expect(callArgs.State).To(Equal(helm.StatePresent))
+		validateValues(t, callArgs.Values, cfg, snapM)
+	})
+}
+
+func validateValues(t *testing.T, values map[string]any, cfg types.Network, snap snap.Snap) {
+	t.Helper()
```
eaudetcobello, line 158: What's this for?

eaudetcobello (changes requested):

src/k8s/pkg/k8sd/features/cilium/network_test.go
HomayoonAlimohammadi: > Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped.
https://pkg.go.dev/testing#B.Helper

src/k8s/pkg/k8sd/features/cilium/network_test.go
```diff
@@ -0,0 +1,179 @@
...
+			},
+		}
+		cfg := types.Network{
+			Enabled: ptr.To(false),
+		}
+
+		status, err := cilium.ApplyNetwork(context.Background(), snapM, cfg, nil)
+
+		g := NewWithT(t)
+		g.Expect(err).To(MatchError(applyErr))
```
eaudetcobello, line 40: What's the purpose of matching the error type here? Couldn't this just be ToNot(HaveOccurred()) ?

eaudetcobello (changes requested):

src/k8s/pkg/k8sd/features/cilium/network_test.go
HomayoonAlimohammadi: right, but is it any different? I definitely can change that, but I feel like this shouldn't really matter. I mean if I have to choose I'd personally prefer keeping "gomega" things and "expects" in the same place. or does it make a real difference? I'm super welcoming to nit and repo/team specific code style comments but I feel like this might be too fine grained for that matter. wdyt?

eaudetcobello: Well it matters because these tests would be the only ones that don't stick to the (unwritten) convention we have established. I definitely see what you're going for with keeping variable declarations close to their use, but I would prefer if we stick to the style we have.

src/k8s/pkg/k8sd/features/cilium/network_test.go
HomayoonAlimohammadi: well I guess other errors might have happened. I think it's better for unit tests to cover specific parts/scenarios rather than 0/1 outcomes. in the end it's more of a personal preference I guess.

eaudetcobello: In the other test cases you only use To(HaveOccurred()), so this one needs to change or the others need to change for consistency

src/k8s/pkg/k8sd/features/cilium/network_test.go
eaudetcobello: Cool! Good to know

eaudetcobello (approved):
LGTM, above nits are at your discretion

src/k8s/pkg/k8sd/features/cilium/network_test.go
HomayoonAlimohammadi: right, changing this needs us to either return a known error from `cilium.ApplyNetwork` or `utils.ParseCIDR`. I think it's better to be accurate were possible and just leave it be were it needs unnecessary complexity.

PR #644, opened 2024-09-02
Pin k8s-dqlite version to tag

louiseschmidtgen:
## Description
This PR pins our main branch to the appropriate k8s-dqlite tag.
See: https://github.com/canonical/k8s-dqlite/blob/master/docs/versioning.md


louiseschmidtgen:
> We should not pin the version on master, only the releases

We will be doing some big changes in k8s-dqlite it will be best to pin ourselves to the most recent k8s-dqlite version

bschimke95 (approved):
LGTM, as discussed offline: pinning the version is the right decision. Sorry for the wrong call!

PR #647, opened 2024-09-03
Add unit tests for Calico network k8sd feature

HomayoonAlimohammadi:
- Add unit test for calico k8sd network feature
- Remove `t.Helper()` from cilium network tests as well since I noticed this way finding the root cause of the failed tests becomes easier.


bschimke95 (approved):
LGTM

PR #650, opened 2024-09-04
[quick fix] capi tutorial getting-started.md

maci3jka:
In clusterclt providers there is a duplicate ControlPlaneProvider type. This PR removes that duplicate

KU-1468


eaudetcobello:
Thanks @maci3jka, sorry I didn't see this before making another PR.

PR #618, opened 2024-08-22
Add command for k8sd recovery after quorum loss

petrutlucian94:
We're adding a command that can be used to recover Canonical K8s clusters after dqlite quorum loss.

The command will:

* assist the user in making the necessary k8s-dqlite and k8sd configuration changes
* create pre-recovery and post-recovery backups that can be applied to other nodes
* issue a dqlite reconfigure request, updating the raft members
* attempt to contact the other dqlite members in order to ensure that all cluster members are offline before attempting to perform the recovery steps

Based on the microcloud command, which leverages the microcluster implementation:

* https://github.com/canonical/microcloud/blob/main/cmd/microcloud/cluster_members.go
* https://github.com/canonical/microcluster/blob/42551841a56dd2d4697d577428bdb4d7d5f877d0/internal/recover/recover.go#L72

For reference, here are the microk8s and microcluster docs on cluster recovery:

* https://microk8s.io/docs/restore-quorum#backup-the-database
* https://discourse.ubuntu.com/t/microcluster-microcloud-quorum-recovery/45560

PR #651, opened 2024-09-04
Add "DEBUG_BUILD" flag

petrutlucian94:
A recent PR [1] added debug symbols by default, which wasn't intended. This increases the k8s binary size from ~74M to ~99M.

We're introducing an env variable flag for this:

  export DEBUG_BUILD=y

[1] https://github.com/canonical/k8s-snap/pull/618


bschimke95 (approved):
LGTM

PR #661, opened 2024-09-10
Change default cgroup driver to systemd

berkayoz:
Updates containerd and kubelet to use the `systemd` cgroup driver by default


bschimke95 (approved):
LGTM, please fix the unittests before merging

HomayoonAlimohammadi (approved):
Thanks a lot @berkayoz! Maybe out of scope of this card, but is it possible to make sure the `systemd` is set as the `cgroup driver` for `kubelet` and `containerd`? maybe with an integration test or something?

berkayoz:
> Thanks a lot @berkayoz! Maybe out of scope of this card, but is it possible to make sure the `systemd` is set as the `cgroup driver` for `kubelet` and `containerd`? maybe with an integration test or something?

We could do an integration test although at this point I'm not sure on how to check if the driver is changed(run-time wise) other than checking parameters which is done by unit tests already.

PR #662, opened 2024-09-11
k8sd cluster-recover: add non-interactive mode

petrutlucian94:
At the moment, the "k8sd cluster-recover" displays interactive prompts and text editors that assist the user in updating the dqlite configuration.

We need to be able to run the command non-interactively in order to automate the cluster recovery procedure.

This change adds a "--non-interactive" flag. If set, we'll no longer show confirmation prompts and we'll assume that the configuration files have already been updated, proceeding with the dqlite recovery.

### DIFF
--- a/docs/src/snap/howto/restore-quorum.md
+++ b/docs/src/snap/howto/restore-quorum.md
@@ -1,9 +1,9 @@
 # Recovering a Cluster After Quorum Loss
 
 Highly available {{product}} clusters can survive losing one or more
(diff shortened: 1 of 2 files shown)


docs/src/snap/howto/restore-quorum.md
bschimke95, line 1: Please address the linter issues.

bschimke95 (approved):
LGTM

PR #664, opened 2024-09-12
Add IPv6-only support for moonray

bschimke95:
Adds support for IPv6-only support on moonray. For the default k8s-snap this is not yet supported as Cilium loadbalancer does not support IPv6-only.

Most of the changes basically only ensure that the IPv6 address is properly put into brackets. We use different representations of IPs across the codebase (e.g. string, or net.IP or Addr by microcluster) so we cannot just simply use the same approach everywhere.

### DIFF
--- a/src/k8s/pkg/k8sd/api/certificates_refresh.go
+++ b/src/k8s/pkg/k8sd/api/certificates_refresh.go
@@ -263,10 +263,10 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
 	}
 
 	// Kubeconfigs
-	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"), "127.0.0.1:6443", certificates.CACert, certificates.KubeletClientCert, certificates.KubeletClientKey); err != nil {
+	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"), "[::1]:6443", certificates.CACert, certificates.KubeletClientCert, certificates.KubeletClientKey); err != nil {
 		return response.InternalError(fmt.Errorf("failed to generate kubelet kubeconfig: %w", err))
 	}
-	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "proxy.conf"), "127.0.0.1:6443", certificates.CACert, certificates.KubeProxyClientCert, certificates.KubeProxyClientKey); err != nil {
+	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "proxy.conf"), "[::1]:6443", certificates.CACert, certificates.KubeProxyClientCert, certificates.KubeProxyClientKey); err != nil {
 		return response.InternalError(fmt.Errorf("failed to generate kube-proxy kubeconfig: %w", err))
 	}
 
--- a/src/k8s/pkg/k8sd/api/certificates_refresh.go
+++ b/src/k8s/pkg/k8sd/api/certificates_refresh.go
@@ -126,7 +126,7 @@ func refreshCertsRunControlPlane(s state.State, r *http.Request, snap snap.Snap)
 		return response.InternalError(fmt.Errorf("failed to write control plane certificates: %w", err))
 	}
 
-	if err := setup.SetupControlPlaneKubeconfigs(snap.KubernetesConfigDir(), clusterConfig.APIServer.GetSecurePort(), *certificates, localhostAddress); err != nil {
+	if err := setup.SetupControlPlaneKubeconfigs(snap.KubernetesConfigDir(), localhostAddress, clusterConfig.APIServer.GetSecurePort(), *certificates); err != nil {
 		return response.InternalError(fmt.Errorf("failed to generate control plane kubeconfigs: %w", err))
 	}
 
--- a/src/k8s/pkg/k8sd/api/certificates_refresh.go
+++ b/src/k8s/pkg/k8sd/api/certificates_refresh.go
@@ -81,6 +81,13 @@ func refreshCertsRunControlPlane(s state.State, r *http.Request, snap snap.Snap)
 		return response.InternalError(fmt.Errorf("failed to parse node IP address %q", s.Address().Hostname()))
 	}
 
+	var localhostAddress string
+	if nodeIP.To4() == nil {
+		localhostAddress = "[::1]"
+	} else {
+		localhostAddress = "127.0.0.1"
+	}
+
 	serviceIPs, err := utils.GetKubernetesServiceIPsFromServiceCIDRs(clusterConfig.Network.GetServiceCIDR())
 	if err != nil {
 		return response.InternalError(fmt.Errorf("failed to get IP address(es) from ServiceCIDR %q: %w", clusterConfig.Network.GetServiceCIDR(), err))
@@ -119,7 +126,7 @@ func refreshCertsRunControlPlane(s state.State, r *http.Request, snap snap.Snap)
 		return response.InternalError(fmt.Errorf("failed to write control plane certificates: %w", err))
 	}
 
-	if err := setup.SetupControlPlaneKubeconfigs(snap.KubernetesConfigDir(), clusterConfig.APIServer.GetSecurePort(), *certificates); err != nil {
+	if err := setup.SetupControlPlaneKubeconfigs(snap.KubernetesConfigDir(), clusterConfig.APIServer.GetSecurePort(), *certificates, localhostAddress); err != nil {
 		return response.InternalError(fmt.Errorf("failed to generate control plane kubeconfigs: %w", err))
 	}
 
@@ -262,9 +269,16 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
 		return response.InternalError(fmt.Errorf("failed to write worker PKI: %w", err))
 	}
 
-	localhostAddress, err := utils.GetLocalhostAddress(clusterConfig.Network.GetPodCIDR(), clusterConfig.Network.GetServiceCIDR())
-	if err != nil {
-		return response.InternalError(fmt.Errorf("failed to get localhost address"))
+	nodeIP := net.ParseIP(s.Address().Hostname())
+	if nodeIP == nil {
+		return response.InternalError(fmt.Errorf("failed to parse node IP address %q", s.Address().Hostname()))
+	}
+
+	var localhostAddress string
+	if nodeIP.To4() == nil {
+		localhostAddress = "[::1]"
+	} else {
+		localhostAddress = "127.0.0.1"
 	}
 
 	// Kubeconfigs
--- a/src/k8s/pkg/k8sd/api/certificates_refresh.go
+++ b/src/k8s/pkg/k8sd/api/certificates_refresh.go
@@ -262,11 +262,16 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
 		return response.InternalError(fmt.Errorf("failed to write worker PKI: %w", err))
 	}
 
+	localhostAddress, err := utils.GetLocalhostAddress(clusterConfig.Network.GetPodCIDR(), clusterConfig.Network.GetServiceCIDR())
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to get localhost address"))
+	}
+
 	// Kubeconfigs
-	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"), "[::1]:6443", certificates.CACert, certificates.KubeletClientCert, certificates.KubeletClientKey); err != nil {
+	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"), fmt.Sprintf("%s:6443", localhostAddress), certificates.CACert, certificates.KubeletClientCert, certificates.KubeletClientKey); err != nil {
 		return response.InternalError(fmt.Errorf("failed to generate kubelet kubeconfig: %w", err))
 	}
-	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "proxy.conf"), "[::1]:6443", certificates.CACert, certificates.KubeProxyClientCert, certificates.KubeProxyClientKey); err != nil {
+	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "proxy.conf"), fmt.Sprintf("%s:6443", localhostAddress), certificates.CACert, certificates.KubeProxyClientCert, certificates.KubeProxyClientKey); err != nil {
 		return response.InternalError(fmt.Errorf("failed to generate kube-proxy kubeconfig: %w", err))
 	}
 
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -409,7 +409,7 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
 	}
 
 	// Generate kubeconfigs
-	if err := setup.SetupControlPlaneKubeconfigs(snap.KubernetesConfigDir(), cfg.APIServer.GetSecurePort(), *certificates, localhostAddress); err != nil {
+	if err := setup.SetupControlPlaneKubeconfigs(snap.KubernetesConfigDir(), localhostAddress, cfg.APIServer.GetSecurePort(), *certificates); err != nil {
 		return fmt.Errorf("failed to generate kubeconfigs: %w", err)
 	}
 
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -178,12 +178,11 @@ func (a *App) onBootstrapWorkerNode(ctx context.Context, s state.State, encodedT
 		return fmt.Errorf("failed to write worker node certificates: %w", err)
 	}
 
-	// We don't have the CIDRs on the workers, hence we check the address type of the control-plane node.
 	var localhostAddress string
-	if utils.IsIPv4(address) {
-		localhostAddress = "127.0.0.1"
-	} else {
+	if nodeIP.To4() == nil {
 		localhostAddress = "[::1]"
+	} else {
+		localhostAddress = "127.0.0.1"
 	}
 
 	// Kubeconfigs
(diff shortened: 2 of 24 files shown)


HomayoonAlimohammadi:
> We use different representations of IPs across the codebase (e.g. string, or net.IP or Addr by microcluster).

I think this might be a good candidate for improvements later on, to use uniform representations.

bschimke95:
This is not something we directly control as those come from third-parties (e.g. microcluster). Unifying them just to use them once a few lines later in a util does not seem practical to me. That being said, we should definitely do it where applicable.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -178,11 +178,19 @@ func (a *App) onBootstrapWorkerNode(ctx context.Context, s state.State, encodedT
 		return fmt.Errorf("failed to write worker node certificates: %w", err)
 	}
 
+	// We don't have the CIDRs on the workers, hence we check the address type of the control-plane node.
+	var localhostAddress string
+	if utils.IsIPv4(address) {
+		localhostAddress = "127.0.0.1"
```
HomayoonAlimohammadi, line 184: we might add consts for these:
```go
var (
    IPv4Loopback string = "127.0.0.1"
    IPv6Loopback string = fmt.Sprintf("[%s]", net.IPv6loopback)
)
```

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -158,10 +158,16 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+			if member.Address.Addr().Is6() {
+				address = fmt.Sprintf("[%s]", member.Address.Addr().String())
+			} else {
+				address = member.Address.Addr().String()
+			}
+			cluster = append(cluster, fmt.Sprintf("%s:%d", address, cfg.Datastore.GetK8sDqlitePort()))
 		}
 
-		address := fmt.Sprintf("%s:%d", nodeIP.String(), cfg.Datastore.GetK8sDqlitePort())
+		address := fmt.Sprintf("%s:%d", utils.ToIPString(nodeIP), cfg.Datastore.GetK8sDqlitePort())
```
louiseschmidtgen, line 170: We need to pass this `address` on line 195 for kubeapiserver setup as well currently it passes nodeIP. This is needed for setting the advertise-address flag.

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -158,10 +158,16 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
 		}
 		cluster := make([]string, len(members))
 		for _, member := range members {
-			cluster = append(cluster, fmt.Sprintf("%s:%d", member.Address.Addr(), cfg.Datastore.GetK8sDqlitePort()))
+			var address string
+			if member.Address.Addr().Is6() {
+				address = fmt.Sprintf("[%s]", member.Address.Addr().String())
```
HomayoonAlimohammadi, line 163: we're able to remove the `.String()` here.

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
```diff
@@ -69,7 +69,7 @@ func SetupControlPlaneKubeconfigs(kubeConfigDir string, securePort int, pki pki.
 		{file: "scheduler.conf", crt: pki.KubeSchedulerClientCert, key: pki.KubeSchedulerClientKey},
 		{file: "kubelet.conf", crt: pki.KubeletClientCert, key: pki.KubeletClientKey},
 	} {
-		if err := Kubeconfig(filepath.Join(kubeConfigDir, kubeconfig.file), fmt.Sprintf("127.0.0.1:%d", securePort), pki.CACert, kubeconfig.crt, kubeconfig.key); err != nil {
+		if err := Kubeconfig(filepath.Join(kubeConfigDir, kubeconfig.file), fmt.Sprintf("[::1]:%d", securePort), pki.CACert, kubeconfig.crt, kubeconfig.key); err != nil {
```
louiseschmidtgen, line 72: Does this break ipv4 only implementations?

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy.go
```diff
@@ -17,10 +17,15 @@ func K8sAPIServerProxy(snap snap.Snap, servers []string, extraArgs map[string]*s
 		return fmt.Errorf("failed to write proxy configuration file: %w", err)
 	}
 
+	localhostAddress := "127.0.0.1"
+	if len(servers) > 0 && !utils.IsIPv4(servers[0]) {
```
HomayoonAlimohammadi, line 21: is it possible to have a mix of ipv4 and v6?

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
HomayoonAlimohammadi, line 72: shouldn't we use a conditional localhost here as well?

tests/integration/tests/test_util/util.py
```diff
@@ -261,3 +262,24 @@ def get_default_ip(instance: harness.Instance):
         ["ip", "-o", "-4", "route", "show", "to", "default"], capture_output=True
     )
     return p.stdout.decode().split(" ")[8]
+
+
+def get_global_unicast_ipv6(instance: harness.Instance, interface="eth0") -> str:
```
louiseschmidtgen, line 267: I see a potential for flakiness here, interface eth0 is a legacy name since [Predictable Network Interface Names](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/) are being used that are more precise.

louiseschmidtgen (commented):
Great work @bschimke95, please have a look at my comments!

HomayoonAlimohammadi (commented):
Piece of art, thanks a lot Ben! Really enjoyed the thorough testing.
Left some minor comments (mostly nit)

src/k8s/pkg/utils/cidr.go
```diff
@@ -125,3 +125,51 @@ func ParseCIDRs(CIDRstring string) (string, string, error) {
 	}
 	return ipv4CIDR, ipv6CIDR, nil
 }
+
+// GetLocalhostAddress returns the localhost address based on the given pod and service CIDRs.
+// In IPv6-only mode, the IPv6 localhost address is returned. Otherwise the IPv4 localhost address is returned.
+func GetLocalhostAddress(podCIDR string, serviceCIDR string) (string, error) {
```
berkayoz, line 131: This feels a bit hacky/magic, what do you think about having a config to explicitly set the preference?

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -178,11 +178,19 @@ func (a *App) onBootstrapWorkerNode(ctx context.Context, s state.State, encodedT
 		return fmt.Errorf("failed to write worker node certificates: %w", err)
 	}
 
+	// We don't have the CIDRs on the workers, hence we check the address type of the control-plane node.
```
berkayoz, line 181: Don't we use the `response.PodCIDR` here later on to setup kubeproxy for workers?

src/k8s/pkg/k8sd/setup/kube_proxy.go
```diff
@@ -13,9 +13,14 @@ import (
...
 func KubeProxy(ctx context.Context, snap snap.Snap, hostname string, podCIDR string, extraArgs map[string]*string) error {
+	localhostAddress, err := utils.GetLocalhostAddress(podCIDR, "")
+	if err != nil {
+		return fmt.Errorf("failed to get locahost address: %w", err)
+	}
+
 	serviceArgs := map[string]string{
 		"--cluster-cidr":         podCIDR,
-		"--healthz-bind-address": "127.0.0.1",
+		"--healthz-bind-address": fmt.Sprintf("%s:10256", localhostAddress),
```
berkayoz, line 23: nit: I'd prefer this to be determined once on the hook and passed as a parameter to the setup functions.

berkayoz (commented):
Great work overall, did an initial pass and left some comments

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy.go
bschimke95: let's remove this from here altogether and get the address from the bootstrap hook.

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy.go
```diff
@@ -17,10 +17,15 @@ func K8sAPIServerProxy(snap snap.Snap, servers []string, extraArgs map[string]*s
...
+		localhostAddress = "[::1]"
+	}
+
 	if _, err := snaputil.UpdateServiceArguments(snap, "k8s-apiserver-proxy", map[string]string{
 		"--endpoints":  configFile,
 		"--kubeconfig": filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"),
-		"--listen":     "127.0.0.1:6443",
+		"--listen":     fmt.Sprintf("%s:6443", localhostAddress),
```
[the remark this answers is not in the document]
bschimke95: Let's do this as a follow-up card. This PR is already large enough.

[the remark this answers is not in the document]
bschimke95: KU-1584

src/k8s/pkg/k8sd/app/hooks_join.go
bschimke95: This address is the host:port of the k8s-dqlite endpoint while the kube-apiserver setup expects the IP of the apiserver. Or do I misunderstand your question?
Properly stringifying this IP is done now in the function itself.

src/k8s/pkg/k8sd/setup/kube_proxy.go
bschimke95: fair point - updated

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
bschimke95: Yes, thanks for the catch. Initially, I just randomly hardcoded addresses to get something running - and this one slipped through. 
I wonder why this did not break any integration tests, though...

bschimke95: yes, thanks for the catch (see Louise' comment)

src/k8s/pkg/utils/cidr.go
bschimke95: discussed offline: let's remove this hacky approach and use the `--address` flag instead

src/k8s/pkg/utils/cidr.go
```diff
@@ -125,3 +125,51 @@ func ParseCIDRs(CIDRstring string) (string, string, error) {
...
+		}
+	}
+
+	if serviceCIDR != "" {
+		serviceIPv4, _, err = ParseCIDRs(serviceCIDR)
+		if err != nil {
+			return "", fmt.Errorf("failed to parse service CIDR: %w", err)
+		}
+	}
+
```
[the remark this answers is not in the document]
bschimke95: I will throw away this function but you got a good point. Let's address this in a future PR

src/k8s/pkg/utils/cidr.go
```diff
@@ -125,3 +125,51 @@ func ParseCIDRs(CIDRstring string) (string, string, error) {
 	}
 	return ipv4CIDR, ipv6CIDR, nil
 }
+
+// GetLocalhostAddress returns the localhost address based on the given pod and service CIDRs.
+// In IPv6-only mode, the IPv6 localhost address is returned. Otherwise the IPv4 localhost address is returned.
+func GetLocalhostAddress(podCIDR string, serviceCIDR string) (string, error) {
+	if podCIDR == "" && serviceCIDR == "" {
+		return "", fmt.Errorf("both pod and service CIDRs are empty")
```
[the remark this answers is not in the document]
bschimke95: Fair point, but I throw away this function anyway :)

src/k8s/pkg/utils/cidr.go
```diff
@@ -125,3 +125,51 @@ func ParseCIDRs(CIDRstring string) (string, string, error) {
...
+	}
+
+	var podIPv4, serviceIPv4 string
+	var err error
+
+	if podCIDR != "" {
+		podIPv4, _, err = ParseCIDRs(podCIDR)
```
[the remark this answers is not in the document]
bschimke95: obsolete now.

src/k8s/pkg/utils/cidr.go
```diff
@@ -125,3 +125,51 @@ func ParseCIDRs(CIDRstring string) (string, string, error) {
...
+		if err != nil {
+			return "", fmt.Errorf("failed to parse pod CIDR: %w", err)
+		}
+	}
+
+	if serviceCIDR != "" {
+		serviceIPv4, _, err = ParseCIDRs(serviceCIDR)
```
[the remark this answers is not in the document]
bschimke95: obsolete now.

src/k8s/pkg/utils/cidr_test.go
```diff
@@ -164,3 +164,73 @@ func TestParseCIDRs(t *testing.T) {
...
+		{"fd01::/10", "fd98::/108", "[::1]", false},
+		{"fd01::/10", "", "[::1]", false},
+		{"", "fd98::/108", "[::1]", false},
+		{"", "", "", true},                      // Empty podCIDR and serviceCIDR
+		{"invalid", "10.96.0.0/12", "", true},   // Invalid podCIDR
+		{"192.168.0.0/16", "invalid", "", true}, // Invalid serviceCIDR
+	}
+
+	for _, tc := range tests {
+		g := NewWithT(t)
```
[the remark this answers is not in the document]
bschimke95: done.

tests/integration/tests/test_util/util.py
bschimke95: Not sure if I understand? What is the flake in that case?
We explicitly ask for the IPv6 address of the `eth0` interface and this interface is explicitly set by LXD.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
bschimke95: I'd say we can be explicit. Those addresses are not expected to change ever and string assembly makes it slightly harder to read.
IFF we would ever need to change that addresses then IMHO a search-replace is as fast as changing the const.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
bschimke95: yes, wrong assumption. But given that we change the way this is done now, it is obsolete anyway.

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -262,11 +262,16 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
 		return response.InternalError(fmt.Errorf("failed to write worker PKI: %w", err))
 	}
 
+	localhostAddress, err := utils.GetLocalhostAddress(clusterConfig.Network.GetPodCIDR(), clusterConfig.Network.GetServiceCIDR())
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to get localhost address"))
```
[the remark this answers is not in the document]
bschimke95: obsolete now.

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -262,11 +262,16 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
...
+	}
+
 	// Kubeconfigs
-	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"), "127.0.0.1:6443", certificates.CACert, certificates.KubeletClientCert, certificates.KubeletClientKey); err != nil {
+	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"), fmt.Sprintf("%s:6443", localhostAddress), certificates.CACert, certificates.KubeletClientCert, certificates.KubeletClientKey); err != nil {
 		return response.InternalError(fmt.Errorf("failed to generate kubelet kubeconfig: %w", err))
 	}
-	if err := setup.Kubeconfig(filepath.Join(snap.KubernetesConfigDir(), "proxy.conf"), "127.0.0.1:6443", certificates.CACert, certificates.KubeProxyClientCert, certificates.KubeProxyClientKey); err != nil {
```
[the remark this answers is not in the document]
bschimke95: This could break because:
* the address does not point to the local host address anymore
* the certifcates do not contain the new local loopback address

For a) we obtain the address directly before this call, so this cannot be outdated.
For b) we set both loopback address in the certificate SANs, so we should be good.

bschimke95:
Thanks for all the comments! I refactored this PR a bit to pass down the local address rather then determine it in every function separately.

src/k8s/pkg/utils/cidr.go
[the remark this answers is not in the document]
louiseschmidtgen: Could you make a card please?

src/k8s/pkg/k8sd/app/hooks_join.go
louiseschmidtgen: Does that string conversion work for ipv6 in kube-apiserver to pass to the advertise-address flag is my concern.

tests/integration/tests/test_util/util.py
louiseschmidtgen: Okay, then this doesn't apply. I was concerned that we might not always have eth0 as the interface name.

src/k8s/pkg/utils/cidr.go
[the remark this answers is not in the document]
bschimke95: Created as KU-1586

src/k8s/pkg/k8sd/app/hooks_join.go
bschimke95: For the record: It does not.
Kube-Apiserver expects the IPv6 to not have the brackets otherwise it fails with:
`Error: invalid argument "[fd42:1e1e:7a2f:326e:216:3eff:fe3c:db54]" for "--advertise-address" flag: failed to parse IP`.
So this was correct by accident but thanks to your comment we have double-checked that :)

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
```diff
@@ -57,7 +57,7 @@ func KubeconfigString(url string, caPEM string, crtPEM string, keyPEM string) (s
 }
 
 // SetupControlPlaneKubeconfigs writes kubeconfig files for the control plane components.
-func SetupControlPlaneKubeconfigs(kubeConfigDir string, securePort int, pki pki.ControlPlanePKI) error {
+func SetupControlPlaneKubeconfigs(kubeConfigDir string, securePort int, pki pki.ControlPlanePKI, localhostAddress string) error {
```
berkayoz, line 60:
```suggestion
func SetupControlPlaneKubeconfigs(kubeConfigDir string, localhostAddress string, securePort int, pki pki.ControlPlanePKI) error {
```
nit: I'd add the `localhostAddress` before the `securePort`.

berkayoz (approved):
LGTM

PR #675, opened 2024-09-16
Automerge every 4 hours any labeled PR with passing tests

addyess:
### Overview

Automerges any PR with an "automerge" label

### Details
* Every ~~15-min~~ 4 hours check all open PRs.  If all tests are successful on the PR and it's labeled with "automerge" then the PR will be merged by this workflow
* Set the "components upgrade" PR as an automerge PR
* update who is making the commit to use the correct address
* consistently use `DEPLOY_KEY_TO_UPDATE_STRICT_BRANCH` in order to setup the git-remote as ssh
* rename `DEPLOY_KEY_TO_UPDATE_STRICT_BRANCH` to be `BOT_SSH_KEY`


.github/workflows/update-components.yaml
```diff
@@ -51,10 +51,13 @@ jobs:
       - name: Create pull request
         uses: peter-evans/create-pull-request@v6
         with:
-          git-token: ${{ secrets.DEPLOY_KEY_TO_UPDATE_STRICT_BRANCH }}
```
addyess, line 54: it took some digging, but i believe i confirmed this wasn't doing anything because the branch on the runner was checked out using an SSH key -- the token isn't necessary here to push to the branch.  Instead it relies on the ssh key already being part of the local git config.

addyess:
If this is working right, the following PR will get auto merged
* https://github.com/canonical/k8s-snap/pull/672

.github/workflows/auto-merge-successful-prs.yaml
```diff
@@ -0,0 +1,29 @@
+name: Auto-merge Successful PRs
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "*/15 * * * *"  # Every 15 minutes
```
bschimke95, line 6: not sure if we really need to check it that often (IMHO every day would totally be enough), but I'm also fine with this.

.github/workflows/update-components.yaml
```diff
@@ -51,10 +51,13 @@ jobs:
...
           commit-message: "[${{ matrix.branch }}] Update component versions"
           title: "[${{ matrix.branch }}] Update component versions"
           body: "[${{ matrix.branch }}] Update component versions"
           branch: "autoupdate/sync/${{ matrix.branch }}"
+          labels: |
+            automerge
+            automated pr
+            component update
```
bschimke95, line 61: what are the other labels used for? Are they just informational or do you plan to couple functionality to it?
I'd keep the amount of labels to a minimum otherwise this can get messy.

build-scripts/auto-merge-successful-pr.sh
bschimke95, line 1: hm, this file feels a bit wrong here. 
IMHO this would better be suited in either `hacks/ci` or even `.github/scripts` (if this directory can contain custom stuff). The whole script is highly GH specific so I'd say the latter would fit best.

Also, we should probably use Python for every reasonable large scripts as reading bash is painful but I guess this is fine for now.

build-scripts/patches/moonray/apply
```diff
@@ -2,9 +2,12 @@
 
 DIR="$(realpath "$(dirname "${0}")")"
 
-# Configure git author
-git config user.email k8s-bot@canonical.com
-git config user.name k8s-bot
+# Configure git author if unset
```
bschimke95, line 5: thank you for this!
I committed code more than once as `k8s-bot` because I applied moonray and forgot to change back the author.

bschimke95 (commented):
some comments around where we put things but great work overall!

build-scripts/auto-merge-successful-pr.sh
addyess: yeah, i think i meant for it to be in `build-scripts/hacks` 

i considered another language..., but something is so nice about using the `gh` client here...

lemme envision this in python and see if it looks nicer

.github/workflows/update-components.yaml
addyess: sure, let's cut them down to only the necessary ones.  I was looking at future work here.

addyess:
I've confirmed that cdkbot can force push commits to `main` and `release-*` branches
https://github.com/canonical/k8s-snap/commit/22f04c6cfad9af7495c42410c63ad5e28cda843b

bschimke95 (approved):
LGTM

PR #681, opened 2024-09-18
Do not stop Kubernetes services on node removal if annotation is set.

bschimke95:
Requires https://github.com/canonical/k8s-snap-api/pull/8 - will update api version tag for the snap afterwards.

By default, all services on a leaving node are stopped. This PR adds logic to skip this step if the `k8sd/v1alpha/lifecycle/skip-stop-services-on-remove` annotation is set. 

This is required for Cluster API to properly drain nodes. See

Also adds an integration test to validate that this works as expected.

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -61,7 +61,7 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 
 	cfg, err := databaseutil.GetClusterConfig(ctx, s)
 	if err == nil {
-		if _, ok := cfg.Annotations[apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove]; !ok {
+		if _, ok := cfg.Annotations.Get(apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove); !ok {
 			c, err := snap.KubernetesClient("")
 			if err != nil {
 				log.Error(err, "Failed to create Kubernetes client", err)
@@ -130,7 +130,7 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 		log.Error(err, "failed to cleanup control plane certificates")
 	}
 
-	if _, ok := cfg.Annotations[apiv1.AnnotationSkipStopServicesOnRemove]; !ok {
+	if _, ok := cfg.Annotations.Get(apiv1.AnnotationSkipStopServicesOnRemove); !ok {
 		log.Info("Stopping worker services")
 		if err := snaputil.StopWorkerServices(ctx, snap); err != nil {
 			log.Error(err, "Failed to stop worker services")
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -59,7 +59,8 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 		log.Error(err, "Failed to wait for node to finish microcluster join before removing. Continuing with the cleanup...")
 	}
 
-	if cfg, err := databaseutil.GetClusterConfig(ctx, s); err == nil {
+	cfg, err := databaseutil.GetClusterConfig(ctx, s)
+	if err == nil {
 		if _, ok := cfg.Annotations[apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove]; !ok {
 			c, err := snap.KubernetesClient("")
 			if err != nil {
@@ -124,20 +125,22 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 		log.Error(err, "Failed to unmark node as worker")
 	}
 
-	/* 	log.Info("Stopping worker services")
-	   	if err := snaputil.StopWorkerServices(ctx, snap); err != nil {
-	   		log.Error(err, "Failed to stop worker services")
-	   	} */
-
 	log.Info("Cleaning up control plane certificates")
 	if _, err := setup.EnsureControlPlanePKI(snap, &pki.ControlPlanePKI{}); err != nil {
 		log.Error(err, "failed to cleanup control plane certificates")
 	}
 
-	/* 	log.Info("Stopping control plane services")
-	   	if err := snaputil.StopControlPlaneServices(ctx, snap); err != nil {
-	   		log.Error(err, "Failed to stop control-plane services")
-	   	} */
+	if _, ok := cfg.Annotations[apiv1.AnnotationSkipStopServicesOnRemove]; !ok {
+		log.Info("Stopping worker services")
+		if err := snaputil.StopWorkerServices(ctx, snap); err != nil {
+			log.Error(err, "Failed to stop worker services")
+		}
+
+		log.Info("Stopping control plane services")
+		if err := snaputil.StopControlPlaneServices(ctx, snap); err != nil {
+			log.Error(err, "Failed to stop control-plane services")
+		}
+	}
 
 	return nil
 }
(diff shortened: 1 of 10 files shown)


tests/integration/tests/test_clustering.py
```diff
@@ -96,6 +96,36 @@ def test_no_remove(instances: List[harness.Instance]):
...
+    join_token = util.get_join_token(cluster_node, joining_cp)
+    util.join_cluster(joining_cp, join_token)
+
+    util.wait_until_k8s_ready(cluster_node, instances)
+    nodes = util.ready_nodes(cluster_node)
+    assert len(nodes) == 2, "nodes should have joined cluster"
+
+    cluster_node.exec(["k8s", "remove-node", joining_cp.id])
+    nodes = util.ready_nodes(cluster_node)
+    assert len(nodes) == 2, "cp node should not have been removed from cluster"
```
HomayoonAlimohammadi, line 116: I don't understand something, why do we expect that the second node is not removed? the annotation is only supposed to stop the services from being stopped, right? does this mean the node won't effectively leave the cluster?

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -124,19 +125,21 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
 	if _, err := setup.EnsureControlPlanePKI(snap, &pki.ControlPlanePKI{}); err != nil {
 		log.Error(err, "failed to cleanup control plane certificates")
 	}
 
-	log.Info("Stopping control plane services")
-	if err := snaputil.StopControlPlaneServices(ctx, snap); err != nil {
-		log.Error(err, "Failed to stop control-plane services")
+	if _, ok := cfg.Annotations[apiv1.AnnotationSkipStopServicesOnRemove]; !ok {
+		log.Info("Stopping worker services")
+		if err := snaputil.StopWorkerServices(ctx, snap); err != nil {
```
HomayoonAlimohammadi, line 135: can we have some sort of a condition to see if this is a worker or a controlplane node and then stop the related services? 
not that it's super important, but I think it clarifies things a bit more.

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -124,19 +125,21 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
-
 	log.Info("Cleaning up control plane certificates")
 	if _, err := setup.EnsureControlPlanePKI(snap, &pki.ControlPlanePKI{}); err != nil {
 		log.Error(err, "failed to cleanup control plane certificates")
 	}
 
-	log.Info("Stopping control plane services")
-	if err := snaputil.StopControlPlaneServices(ctx, snap); err != nil {
-		log.Error(err, "Failed to stop control-plane services")
+	if _, ok := cfg.Annotations[apiv1.AnnotationSkipStopServicesOnRemove]; !ok {
```
HomayoonAlimohammadi, line 133: you mentioned that a stopped `kubelet` was the root cause of this, so I imagine we should be able to notice this node deletion problem on removing worker nodes (in CAPI) as well, right?

HomayoonAlimohammadi (commented):
Amazing work catching this super subtle issue! Thanks @bschimke95! left some minor comments.

tests/integration/tests/test_clustering.py
bschimke95: That's a fair point. TBH, I have not yet tested this integration tests as the build fails without https://github.com/canonical/k8s-snap-api/pull/8 (and I was too lazy yesterday evening to do a local `replace` in the `go.mod` :) )
Will update the test together with the `k8s-snap-api` version update

src/k8s/pkg/k8sd/app/hooks_remove.go
bschimke95: We explicitly don't want this. 
We check if a node is a worker by checking if the `worker` marker file exists. This can potentially fail for various IO reasons.
This is the remove hook, so everything we do is on a `best-effort` basis. If stopping the worker services fails (e.g. because the node is a CP) then this is fine and we just continue.

src/k8s/pkg/k8sd/app/hooks_remove.go
bschimke95: As discussed offline: Added worker node propagation for annotations, so that this also works on a worker. 
Added an extra test step on the integration test to verify this.

tests/integration/tests/test_clustering.py
```diff
@@ -96,6 +96,60 @@ def test_no_remove(instances: List[harness.Instance]):
...
+    worker = instances[2]
+
+    join_token = util.get_join_token(cluster_node, joining_cp)
+    util.join_cluster(joining_cp, join_token)
+
+    join_token_worker = util.get_join_token(cluster_node, worker, "--worker")
+    util.join_cluster(worker, join_token_worker)
+
+    # We don't care if the node is ready or the CNI is up.
+    util.stubbornly(retries=5, delay_s=3).until(util.get_nodes(cluster_node) == 3)
```
HomayoonAlimohammadi, line 115: not related to this PR but it's odd that the `util.get_nodes` return an integer while `util.ready_nodes` returns actual node instances.

tests/integration/tests/test_clustering.py
```diff
@@ -96,6 +96,60 @@ def test_no_remove(instances: List[harness.Instance]):
...
+
+    cluster_node.exec(["k8s", "remove-node", joining_cp.id])
+    nodes = util.ready_nodes(cluster_node)
+    assert len(nodes) == 2, "cp node should have been removed from the cluster"
+    services = joining_cp.exec(
+        ["snap", "services", "k8s"], capture_output=True, text=True
+    ).stdout.split("\n")[1:-1]
```
HomayoonAlimohammadi, line 122: I imagine trimming the first line is for the `Service  Startup  Current  Notes` line (columns). why are we trimming the last line as well? in my machine the last line is `k8s.kubelet  enabled  active  -` (which I believe should be the same in every other environment cause they look alphabetically sorted).

HomayoonAlimohammadi: wait, is this because the output has an additional `\n` in the end?

tests/integration/tests/test_clustering.py
```diff
@@ -96,6 +96,60 @@ def test_no_remove(instances: List[harness.Instance]):
...
+    print(services)
```
HomayoonAlimohammadi, line 123: maybe we can remove this?

tests/integration/tests/test_clustering.py
```diff
@@ -96,6 +96,60 @@ def test_no_remove(instances: List[harness.Instance]):
...
+        else:
+            assert " active " in service, "service should be active"
+
+    cluster_node.exec(["k8s", "remove-node", worker.id])
+    nodes = util.ready_nodes(cluster_node)
+    assert len(nodes) == 1, "worker node should have been removed from the cluster"
+    services = worker.exec(
+        ["snap", "services", "k8s"], capture_output=True, text=True
+    ).stdout.split("\n")[1:-1]
+    print(services)
```
HomayoonAlimohammadi, line 138: this one as well?

HomayoonAlimohammadi (approved):
Wonderful job, thanks a lot @bschimke95! LGTM overall, but left some minor comments.

tests/integration/tests/test_clustering.py
bschimke95: indeed, probably because this is organically grown and different people implemented different utils. We should probably align this at some point

tests/integration/tests/test_clustering.py
bschimke95: yes, there is an additional new line.

tests/integration/tests/test_clustering.py
bschimke95: I felt like this can be useful debugging output to keep.

PR #683, opened 2024-09-19
Add certificate expiry endpoint

bschimke95:
Adds an endpoint to the CAPI namespace to retrieve the earliest expiring certificate (they should be about the same though).


mateoflorido (approved):
Great work! LGTM

PR #688, opened 2024-09-20
x-snap-config: Do not check if node is part of the cluster

bschimke95:
Follow-up fix for: https://github.com/canonical/k8s-snap/pull/633

In the mentioned PR, we wait until k8sd is up to sync the config. We exit early if the node is not part of a cluster. However, a closer look at the implementation of `NodeStatus` shows that this would skip any error that we wait for:

```
func (c *k8sd) NodeStatus(ctx context.Context) (apiv1.NodeStatusResponse, bool, error) {
	response, err := query(ctx, c, "GET", apiv1.NodeStatusRPC, nil, &apiv1.NodeStatusResponse{})
	if err != nil {
		// Error 503 means the node is not initialized yet
		var statusErr api.StatusError
		if errors.As(err, &statusErr) {
			if statusErr.Status() == http.StatusServiceUnavailable {
				return apiv1.NodeStatusResponse{}, false, nil
			}
		}

		return apiv1.NodeStatusResponse{}, false, err <- Returns partOfCluster == false if an error happens. 
	}

	return response, true, nil
}
```

This PR removes the wrong check.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_x_snapd_config.go
+++ b/src/k8s/cmd/k8s/k8s_x_snapd_config.go
@@ -64,11 +64,7 @@ func newXSnapdConfigCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 			ctx, cancel := context.WithTimeout(cmd.Context(), opts.timeout)
 			defer cancel()
 			if err := control.WaitUntilReady(ctx, func() (bool, error) {
-				_, partOfCluster, err := client.NodeStatus(cmd.Context())
-				if !partOfCluster {
-					cmd.PrintErrf("Node is not part of a cluster: %v\n", err)
-					env.Exit(1)
-				}
+				_, _, err := client.NodeStatus(cmd.Context())
 				return err == nil, nil
 			}); err != nil {
 				cmd.PrintErrf("Error: k8sd did not come up in time: %v\n", err)


src/k8s/cmd/k8s/k8s_x_snapd_config.go
```diff
@@ -64,11 +64,7 @@ func newXSnapdConfigCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 			ctx, cancel := context.WithTimeout(cmd.Context(), opts.timeout)
 			defer cancel()
 			if err := control.WaitUntilReady(ctx, func() (bool, error) {
-				_, partOfCluster, err := client.NodeStatus(cmd.Context())
-				if !partOfCluster {
-					cmd.PrintErrf("Node is not part of a cluster: %v\n", err)
-					env.Exit(1)
-				}
+				_, _, err := client.NodeStatus(cmd.Context())
```
addyess, line 67: So, if `NodeStatus()` always returns `partofCluster=false` for any error why do we no longer need this exit early?  

i'm not sure i have the context from #633 to properly review this.  Maybe @mateoflorido

addyess: basically you're saying any error from `NodeStatus` at all exits early-- and that's wrong?

bschimke95: So, `WaitUntilReady` will retry if the bool return value is `false`. If something in the error return value is returned, then this exits immediately. 

`NodeStatus` second return value indicates if the node is part of a cluster or not. However, it will also return `false` for this argument if an error is returned (third return value). The current implementation would always exit with `Node is notpart of cluster` even if `NodeStatus` returned and error (which means the "is part of cluster" return value is not valid).

addyess: AHH!  i didn't look down at lines 70-71.  Thank you

addyess (approved):
LGTM, thanks for the explanation!

Issue #634, opened 2024-08-28
Can't upgrade k8s using snap refresh


mhalano:
### Summary

I tried to upgrade my snap k8s installed on a single node and got this error:
```
mhalano@skynet:~$ sudo snap refresh k8s
2024-08-28T22:51:36Z INFO Waiting for "snap.k8s.kube-apiserver.service" to stop.
error: cannot perform the following tasks:
- Run configure hook of "k8s" snap if present (run hook "configure": Node is not part of a cluster: <nil>)
```

### What Should Happen Instead?

Should complete the upgrade correctly

### Reproduction Steps

1. Install k8s in an older version. Mine is 1.30.3
2. Execute `sudo snap refresh k8s` to get the latest version. At this moment is 1.31.0
3. See the error

### System information

[inspection-report-20240828_225625.tar.gz](https://github.com/user-attachments/files/16789802/inspection-report-20240828_225625.tar.gz)

### Are you interested in contributing with a fix?

I don't know how I can fix it, but I can help troubleshooting and validate any possible solution. My cluster is not critical.

bschimke95:
Hi @mhalano 

Thanks for reporting this issue.
I *think* that I have addressed this issue with https://github.com/canonical/k8s-snap/pull/633 yesterday evening. 

For me, the upgrade to the latest version works:
```
➜ sudo snap refresh k8s --channel=1.30-classic/candidate --classic
k8s (1.30-classic/candidate) v1.30.3 from Canonical✓ refreshed

➜ sudo k8s bootstrap                                              
Bootstrapping the cluster. This may take a few seconds, please wait.
Bootstrapped a new Kubernetes cluster with node address "192.168.178.36:6400".
The node will be 'Ready' to host workloads after the CNI is deployed successfully.

➜ sudo snap refresh k8s --channel=latest/edge --classic           
2024-08-29T08:20:22+02:00 INFO Waiting for "snap.k8s.kube-apiserver.service" to stop.
k8s (edge) v1.31.0 from Canonical✓ refreshed
```

Would you mind retrying this upgrade on `latest/edge` (revision 991)? Thanks!

mhalano:
@bschimke95 I did the test you mentioned, and the upgrade process works, but it crashes my cluster after:
```
root@skynet:/var/lib/snapd# k8s kubectl get pods
Error: Failed to retrieve the node status.

[...]

PR #689, opened 2024-09-20
Restoring the Microcluster Schema Migration history

louiseschmidtgen:
# Restoring the Microcluster Migration History

This PR restores our microcluster schema migration history to the original order.

## Issue

We had an issue where the latest migration did not get applied: https://github.com/canonical/k8s-snap/issues/634

## Cause

The reason for this issue is that a [migration was removed](https://github.com/canonical/k8s-snap/commit/90d6d99c36036e6c5e7d2dd820f40b955949886c#diff-d0b33b91e3a4798c4ec5a713e4164f3e73232b327878b6c9e6273f2e100a2badR75) without dropping the schema version count. Consequently, when refreshing the snap the microcluster migration thought more migrations were already applied, leaving the latest migration un-applied.

## Fix

This PR restores the schema migration order which will need to be back-ported into all 1.31 branches.
It is not good practice to remove a migration, this does not remove the existing table. A second migration should be in place to remove the table maintaining the accurate migration history.

## Note on Prefix count 

We should have labeled these not at all in the first place. It is not good practice to change the existing migration name (in case microcluster introduces checks around migration names). To stay consistent with the current naming I am continuing the current number prefix.


src/k8s/pkg/k8sd/database/schema.go
```diff
@@ -18,9 +18,11 @@ var (
 	SchemaExtensions = []schema.Update{
 		schemaApplyMigration("kubernetes-auth-tokens", "000-create.sql"),
 		schemaApplyMigration("cluster-configs", "000-create.sql"),
+		schemaApplyMigration("worker-nodes", "000-create.sql"),
 		schemaApplyMigration("worker-tokens", "000-create.sql"),
 		schemaApplyMigration("feature-status", "000-feature-status.sql"),
 		schemaApplyMigration("worker-tokens", "001-add-expiry.sql"),
+		schemaApplyMigration("worker-nodes", "002-delete.sql"),
```
addyess, line 25: it sounds like the order and names of these migrations is important -- but we've just got bare strings identifying them.  Also, are they lexicographically ordered or numerically ordered?  I'm worried about the day we run out of numbers (999).  

```
⚡ ls -al
total 8
drwxrwxr-x  2 addyess addyess 4096 Sep 20 13:14 .
drwxrwxr-x 12 addyess addyess 4096 Sep 20 13:13 ..
-rw-rw-r--  1 addyess addyess    0 Sep 20 13:13 0001-stuff
-rw-rw-r--  1 addyess addyess    0 Sep 20 13:13 001-stuff
-rw-rw-r--  1 addyess addyess    0 Sep 20 13:14 0090-stuff
-rw-rw-r--  1 addyess addyess    0 Sep 20 13:14 009-stuff
```

addyess (commented):
Questions regarding the migration order

src/k8s/pkg/k8sd/database/schema.go
addyess: the alternative i'm hoping for:

```
⚡ ls -alv
total 8
drwxrwxr-x  2 addyess addyess 4096 Sep 20 13:14 .
drwxrwxr-x 12 addyess addyess 4096 Sep 20 13:13 ..
-rw-rw-r--  1 addyess addyess    0 Sep 20 13:13 0001-stuff
-rw-rw-r--  1 addyess addyess    0 Sep 20 13:13 001-stuff
-rw-rw-r--  1 addyess addyess    0 Sep 20 13:14 009-stuff
-rw-rw-r--  1 addyess addyess    0 Sep 20 13:14 0090-stuff
```

louiseschmidtgen: Currently, the order that microcluster uses is simply the order of schema migrations in this struct. Unfortunately, we started naming the migrations with these numbers -they don't mean anything right now. 

We can drop the numbers all-together if that makes things less messy. However, I don't like the idea of changing existing migration names- if microcluster migrations decide to use the names in the future this could get us into trouble (The migration names are stored in a table).

HomayoonAlimohammadi, line 25: I totally get what we're going for here. but since we have directories for different domains (tables), it might be confusing to share a global ordering between these domains (tables). I guess we'll soon have something like this in the `src/k8s/pkg/k8sd/database/sql/migrations` which I think is confusing and probably not necessary (?):
```
.
├── cluster-configs
│   ├── 000-create.sql
│   └── 004-something-else.sql
├── feature-status
│   ├── 000-feature-status.sql
│   └── 003-something.sql
├── kubernetes-auth-tokens
│   ├── 000-create.sql
│   └── 005-another-thing.sql
├── worker-nodes
│   ├── 000-create.sql
│   └── 002-delete.sql
└── worker-tokens
    ├── 000-create.sql
    └── 001-add-expiry.sql
```
the only thing that we need to do right now is to make sure we don't delete and re-order the migrations again and keep the orderings in the scope of the domain (table). this is more like a personal preference from me tho, nothing super important.

HomayoonAlimohammadi (commented):
Thanks a lot @louiseschmidtgen! left a minor comment for discussion.

src/k8s/pkg/k8sd/database/schema.go
louiseschmidtgen: This numbering is giving a false idea of order... You are right the sequence is confusing as well, I'll drop the number to 001 for the new migration.

addyess: thanks so much 👍🏻

berkayoz (approved):
LGTM, thanks for getting in depth with this! I feel like we should have some sort of dev documentation around this process especially on how to maintain these migrations.

PR #690, opened 2024-09-23
Return non-zero exit code in case of errors

petrutlucian94:
At the moment, k8s and k8sd return 0 even if the command fails, which is a problem especially when used inside scripts.

We'll ensure that a non-zero exit code is returned if the commands fail.

### DIFF
--- a/src/k8s/cmd/k8sd/k8sd_cluster_recover.go
+++ b/src/k8s/cmd/k8sd/k8sd_cluster_recover.go
@@ -24,6 +24,7 @@ import (
 	"golang.org/x/sys/unix"
 	"gopkg.in/yaml.v2"
 
+	cmdutil "github.com/canonical/k8s/cmd/util"
 	"github.com/canonical/k8s/pkg/log"
 	"github.com/canonical/k8s/pkg/utils"
 )
@@ -100,26 +101,28 @@ func logDebugf(format string, args ...interface{}) {
 
 }
 
-func newClusterRecoverCmd() *cobra.Command {
+func newClusterRecoverCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "cluster-recover",
 		Short: "Recover the cluster from this member if quorum is lost",
-		RunE: func(cmd *cobra.Command, args []string) error {
+		Run: func(cmd *cobra.Command, args []string) {
 			log.Configure(log.Options{
 				LogLevel:     rootCmdOpts.logLevel,
 				AddDirHeader: true,
 			})
 
 			if err := recoveryCmdPrechecks(cmd.Context()); err != nil {
-				return err
+				cmd.PrintErrf("Recovery precheck failed: %v\n", err)
+				env.Exit(1)
 			}
 
 			if clusterRecoverOpts.SkipK8sd {
-				cmd.Printf("Skipping k8sd recovery.")
+				cmd.Printf("Skipping k8sd recovery.\n")
 			} else {
 				k8sdTarballPath, err := recoverK8sd()
 				if err != nil {
-					return fmt.Errorf("failed to recover k8sd, error: %w", err)
+					cmd.PrintErrf("Failed to recover k8sd, error: %v\n", err)
+					env.Exit(1)
 				}
 				cmd.Printf("K8sd cluster changes applied.\n")
 				cmd.Printf("New database state saved to %s\n", k8sdTarballPath)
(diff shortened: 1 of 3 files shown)


bschimke95:
LGTM thanks
Please run `go fmt` to make the Go CI happy, then we are good to merge.

bschimke95 (approved):
LGTM

HomayoonAlimohammadi (commented):
I think we're using `Run` instead of `RunE` which indicates that we're relying on the `os.Exit(1)` to occur in the individual cmds. I might be wrong, but I think the `ExecuteContext` here only returns an error if the `RunE`, `PreRunE`, `PostRunE` or similar "error-returning" methods return an error which we don't have any of them (AFAIK). 
This change _is_ a valid change but if we're addressing the fact that k8s commands does not exit with code 1 when they should, this won't solve that issue.
In other words, this change _won't_ cause any behavior changes because of the current implementation.
If there's any k8s command that exits with code 0 while also facing with an error I think we should change the code for that specific cmd and call the `os.Exit(1)` there (e.g.: https://github.com/canonical/k8s-snap/blob/5fe3c27356a1a54082558b5efebf8b701224dfa0/src/k8s/cmd/k8s/k8s_status.go#L32).

petrutlucian94:
@HomayoonAlimohammadi thanks for bringing this up. I did a simple check and it seemed to work fine but I'll take a closer look.

```
w/ fix
cloudbase@ubuntuk8s2ha:~/workspace/k8s-snap/src/k8s$ bin/static/k8sd wrong-cmd
Error: unknown command "wrong-cmd" for "k8sd"
Run 'k8sd --help' for usage.
cloudbase@ubuntuk8s2ha:~/workspace/k8s-snap/src/k8s$ echo $?
1

# w/o fix
ubuntu@hatwo:~$ /snap/k8s/1211/bin/k8sd wrong-cmd
Error: unknown command "wrong-cmd" for "k8sd"
Run 'k8sd --help' for usage.
ubuntu@hatwo:~$ echo $?
0
```

bschimke95:
As discussed in the standup: You both are correct. 
We currently use the `Run` command for every k8s command except the `k8sd cluster recover` command. I overlooked that in #618. 

So,
a) @petrutlucian94 please create a PR to refactor the `k8sd cluster recover` command to use the environment, similar to what commands do.
b) I'd still merge this PR to be a safe-net in case we miss this again for other commands.

bschimke95 (approved):
LGTM. Please add a note to the error check.
e.g. `// Although k8s commands typically use Run instead of RunE and handle errors directly within the command execution, this acts as a safeguard in case any are overlooked.`

petrutlucian94:
> a) @petrutlucian94 please create a PR to refactor the k8sd cluster recover command to use the environment, similar to what commands do.

I've added another commit here, I hope that's ok.

> b) I'd still merge this PR to be a safe-net in case we miss this again for other commands.

The Run/RunE functions won't get called at all if there are argument parsing errors, so that's another reason why we need this. I've added an inline comment.

bschimke95 (approved):
LGTM, thanks for the change!

src/k8s/cmd/main.go
```diff
@@ -30,14 +30,26 @@ func main() {
...
+
+	// Although k8s commands typically use Run instead of RunE and handle
+	// errors directly within the command execution, this acts as a safeguard in
+	// case any are overlooked.
+	//
+	// Furthermore, the Cobra framework may not invoke the "Run*" entry points
+	// at all in case of argument parsing errors, in which case we *need* to
+	// handle the errors here.
+	if err != nil {
+		os.Exit(1)
```
HomayoonAlimohammadi, line 53: maybe we can call `env.Exit` here as well.

src/k8s/cmd/k8sd/k8sd_cluster_recover.go
```diff
@@ -104,22 +104,24 @@ func newClusterRecoverCmd() *cobra.Command {
...
+		Run: func(cmd *cobra.Command, args []string) {
 			log.Configure(log.Options{
 				LogLevel:     rootCmdOpts.logLevel,
 				AddDirHeader: true,
 			})
 
 			if err := recoveryCmdPrechecks(cmd.Context()); err != nil {
-				return err
+				cmd.PrintErrf("Recovery precheck failed: %v\n", err)
+				os.Exit(1)
```
HomayoonAlimohammadi, line 115: maybe it's better to use `env.Exit` here and in the other places as well.

p.s. I noticed that `env` is not passed to the `newClusterRecoverCmd`. maybe we can do this part (passing `env` and using `env.Exit`) in a follow up PR.

HomayoonAlimohammadi (approved):
LGTM, Thanks a lot @petrutlucian94! Just some very minor comments.

src/k8s/cmd/k8sd/k8sd_cluster_recover.go
petrutlucian94: Right away, to be honest I haven't noticed the ExecutionContext abstraction.

I'll update the commit.

PR #694, opened 2024-09-24
Fix TiCS action's permission error

eaudetcobello:
`make go.unit` has to be run with elevated privileges since we now change the owner of /opt/cni/bin to root. See https://github.com/canonical/k8s-snap/blob/a2c00fb6be9951f582e139d83d97ad1dceb4ec13/src/k8s/pkg/k8sd/setup/directories.go#L72 

This operation fails if `make go.unit` is run as a regular user.

### DIFF
--- a/.github/workflows/cron-jobs.yaml
+++ b/.github/workflows/cron-jobs.yaml
@@ -2,8 +2,6 @@ name: Security and quality nightly scan
 
 on:
   pull_request:
-  # schedule:
-  #   - cron: '0 10 * * *'
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/cron-jobs.yaml
```diff
@@ -1,12 +1,13 @@
 name: Security and quality nightly scan
 
 on:
-  schedule:
-    - cron: '0 10 * * *'
+  pull_request:
+  # schedule:
+  #   - cron: '0 10 * * *'
```
addyess, line 6: obviously revert this

addyess (changes requested):
Looks really good.  Don't forget the workflow changes

.github/workflows/cron-jobs.yaml
eaudetcobello: There's a new problem now: https://github.com/canonical/k8s-snap/actions/runs/11019106186/job/30604131753?pr=694#step:7:7502

addyess (approved):
Here's your 👍🏻 take it!

PR #695, opened 2024-09-24
Validate pod CIDR and service CIDR 

eaudetcobello:
We should verify that the user provides valid CIDRs when bootstrapping the cluster. 

This PR adds validation at bootstrap-time.

- The Pod CIDR and Service CIDR do not overlap

- Are within [Kubernetes CIDR limitations](https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/networking/dualstack/#cidr-size-limitations)

Examples
--

bootstrap-config:
```
pod-cidr: 192.168.1.0/24
service-cidr: 192.168.2.0/24,fe80::/64
```

```
Error: Failed to validate the CIDR configuration.                                                                           

The error was: service CIDR "fe80::/64" cannot have a prefix length of 64 or more
```

bootstrap-config:
```
pod-cidr: 192.168.1.0/24
service-cidr: 192.168.1.0/24
```

```
Error: Failed to validate the CIDR configuration.                                                                             

The error was: pod CIDR "192.168.1.0/24" and service CIDR "192.168.1.0/24" overlap
```


eaudetcobello:
Maybe the validation should could be moved to k8s/pkg/utils/cidr.go

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -127,6 +128,14 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				}
 			}
 
+			if bootstrapConfig.PodCIDR != nil && bootstrapConfig.ServiceCIDR != nil {
```
HomayoonAlimohammadi, line 131: Just out of curiosity, is it possible that only one of them is non-nil?

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -269,3 +278,55 @@ func askQuestion(stdin io.Reader, stdout io.Writer, stderr io.Writer, question s
...
 }
+
+// validateCIDROverlapAndSize checks for overlap and size constraints between pod and service CIDRs.
+// It parses the provided podCIDR and serviceCIDR strings, checks for IPv4 and IPv6 overlaps, and ensures
+// that the service IPv6 CIDR does not have a prefix length of 64 or more.
+func validateCIDROverlapAndSize(podCIDR string, serviceCIDR string) error {
+	// Parse the CIDRs
+	podIPv4CIDR, podIPv6CIDR, err := utils.ParseCIDRs(podCIDR)
+	if err != nil {
+		return err
```
HomayoonAlimohammadi, line 289: I personally almost always like to wrap errors. IMO this makes it easier to track failures to their root cause. Maybe we can wrap the errors that we have in this function as well.

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -269,3 +278,55 @@ func askQuestion(stdin io.Reader, stdout io.Writer, stderr io.Writer, question s
...
+	if podIPv6CIDR != "" && svcIPv6CIDR != "" {
+		if overlap, err := utils.CIDRsOverlap(podIPv6CIDR, svcIPv6CIDR); err != nil {
+			return err
+		} else if overlap {
+			return fmt.Errorf("pod CIDR %q and service CIDR %q overlap", podCIDR, serviceCIDR)
+		}
+	}
+
+	// Check CIDR size
+	// Ref: https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/networking/dualstack/#cidr-size-limitations
```
HomayoonAlimohammadi, line 316: > __/64 is too large for the Service CIDR__: Using a /64 CIDR for services may cause issues like failure to initialize the IPv6 allocator.

This makes me feel that the problem might not only be with the IPv6 but that IPv6 problem is just an example, WDYT? Do we need to change the docs or check for IPv4 here as well?

src/k8s/cmd/k8s/k8s_bootstrap_test.go
```diff
@@ -155,3 +155,39 @@ func TestGetConfigFromYaml_Stdin(t *testing.T) {
...
+		name        string
+		podCIDR     string
+		serviceCIDR string
+		expectErr   bool
+	}{
+		{"Empty", "", "", true},
+		{"EmptyServiceCIDR", "192.168.1.0/24", "", true},
+		{"EmptyPodCIDR", "", "192.168.1.0/24", true},
+		//IPv4
+		{"SameIPv4CIDRs", "192.168.100.0/24", "192.168.100.0/24", true},
```
HomayoonAlimohammadi, line 170: These should have been hard to come up with!

HomayoonAlimohammadi (commented):
Wonderful job @eaudetcobello! Thanks a lot. Left some minor comments.

src/k8s/cmd/k8s/k8s_bootstrap.go
bschimke95: That is a limitation of the Kubernetes controller. I understand your point, but IMHO we should describe the limitation for a user (-> use smaller CIDR size) as they normally don't care about the internals of some Kubernetes controller.

src/k8s/pkg/utils/cidr.go
```diff
@@ -142,3 +142,21 @@ func ToIPString(ip net.IP) string {
 	}
 	return "[" + ip.String() + "]"
 }
+
+// CIDRsOverlap checks if two given CIDR blocks overlap.
+// It takes two strings representing the CIDR blocks as input and returns a boolean indicating
+// whether they overlap and an error if any of the CIDR blocks are invalid.
+func CIDRsOverlap(cidr1, cidr2 string) (bool, error) {
```
bschimke95, line 149: I'd just inline this logic (90% of this function is just `net.ParseCIDR`) but I'm also fine with keeping this.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -494,3 +500,55 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
...
+			return fmt.Errorf("pod CIDR %q and service CIDR %q overlap", podCIDR, serviceCIDR)
+		}
+	}
+
+	// Check CIDR size
+	// Ref: https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/networking/dualstack/#cidr-size-limitations
+	if svcIPv6CIDR != "" {
+		_, ipv6Net, err := net.ParseCIDR(svcIPv6CIDR)
+		if err != nil {
+			// Should not happen, as we already parsed the CIDR
```
bschimke95, line 542: nit: That's an implicit assumption that you have from knowing the implementation of the CIDR Overlap function. If this function changes, this may not be accurate anymore.

bschimke95 (commented):
looking good - couple of comments

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -494,3 +500,55 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
...
+// It parses the provided podCIDR and serviceCIDR strings, checks for IPv4 and IPv6 overlaps, and ensures
+// that the service IPv6 CIDR does not have a prefix length of 64 or more.
+func validateCIDROverlapAndSize(podCIDR string, serviceCIDR string) error {
+	// Parse the CIDRs
+	podIPv4CIDR, podIPv6CIDR, err := utils.ParseCIDRs(podCIDR)
+	if err != nil {
+		return err
+	}
+
+	svcIPv4CIDR, svcIPv6CIDR, err := utils.ParseCIDRs(serviceCIDR)
```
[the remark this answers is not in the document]
eaudetcobello: I was thinking the same, will do.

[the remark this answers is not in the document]
eaudetcobello: I think it'd be even more useful if it accepted a string, and returned net.IP's

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
eaudetcobello: Good point

src/k8s/pkg/utils/cidr.go
eaudetcobello: Since this function hides 2 calls to net.Parse, and we call this function 2 times, it means I would need to add 4 calls to net.Parse and their associated error handling in validateCIDROverlap, which will bloat the function.

src/k8s/cmd/k8s/k8s_bootstrap.go
eaudetcobello: If one of them is non-nil, then we do not need to check if there is an overlap (because the other is nil), so this is not an issue.

src/k8s/cmd/k8s/k8s_bootstrap.go
eaudetcobello: Did some tests with @berkayoz and we will be going with /108. Cluster fails to bootstrap at /64 and /96.

src/k8s/pkg/k8sd/app/hooks_bootstrap_test.go
HomayoonAlimohammadi, line 1: Is this file a TODO?

src/k8s/pkg/utils/cidr.go
```diff
@@ -142,3 +142,21 @@ func ToIPString(ip net.IP) string {
...
+	_, ipNet1, err1 := net.ParseCIDR(cidr1)
+	_, ipNet2, err2 := net.ParseCIDR(cidr2)
+
+	if err1 != nil || err2 != nil {
+		return false, fmt.Errorf("invalid CIDR blocks, %v, %v", err1, err2)
```
HomayoonAlimohammadi, line 154: I think it's better to use `%w` for wrapping: https://go.dev/play/p/BZ7h5NMko5s
nit: I'd also wrap and return after each error, seeing something like `invalid CIDR blocks, nil, failed to parse CIDR` might be a bit confusing IMO

src/k8s/pkg/k8sd/types/cluster_config_validate.go
```diff
@@ -21,6 +23,62 @@ func validateCIDRs(cidrString string) error {
...
+			return fmt.Errorf("failed to check for IPv6 overlap: %w", err)
+		} else if overlap {
+			return fmt.Errorf("pod CIDR %q and service CIDR %q overlap", podCIDR, serviceCIDR)
+		}
+	}
+
+	return nil
+}
+
+// Check CIDR size ensures that the service IPv6 CIDR is not larger than /108.
```
HomayoonAlimohammadi, line 61: We might also need to change the docs. I can only see the `/64` there.

src/k8s/pkg/k8sd/types/cluster_config_validate.go
```diff
@@ -21,6 +23,62 @@ func validateCIDRs(cidrString string) error {
...
+		return fmt.Errorf("invalid CIDR: %w", err)
+	}
+
+	_, ipv6Net, err := net.ParseCIDR(svcIPv6CIDR)
+	if err != nil {
+		return fmt.Errorf("invalid CIDR: %w", err)
+	}
+
+	prefixLength, _ := ipv6Net.Mask.Size()
+	if prefixLength < 108 {
```
HomayoonAlimohammadi, line 75: nit: maybe inline?

src/k8s/pkg/k8sd/types/cluster_config_validate.go
```diff
@@ -21,6 +23,62 @@ func validateCIDRs(cidrString string) error {
 	return nil
 }
 
+// validateCIDROverlapAndSize checks for overlap and size constraints between pod and service CIDRs.
```
HomayoonAlimohammadi, line 26:
```suggestion
// validateCIDROverlap checks for overlap and size constraints between pod and service CIDRs.
```

src/k8s/pkg/k8sd/types/cluster_config_validate.go
HomayoonAlimohammadi, line 61:
```suggestion
// validateIPv6CIDRSize Check CIDR size ensures that the service IPv6 CIDR is not larger than /108.
```

src/k8s/pkg/k8sd/types/cluster_config_validate.go
```diff
@@ -31,6 +89,14 @@ func (c *ClusterConfig) Validate() error {
 		return fmt.Errorf("invalid service CIDR: %w", err)
 	}
 
+	if err := validateCIDROverlap(c.Network.GetPodCIDR(), c.Network.GetServiceCIDR()); err != nil {
+		return fmt.Errorf("invalid cidr configuration: %w", err)
```
HomayoonAlimohammadi, line 93: nit: maybe we can be more specific with this error (and the one after that)?

src/k8s/pkg/k8sd/types/cluster_config_validate_test.go
```diff
@@ -128,3 +129,46 @@ func TestValidateExternalServers(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateCIDROverlap(t *testing.T) {
```
HomayoonAlimohammadi, line 133:
```suggestion
func TestValidateCIDROverlapAndSize(t *testing.T) {
```

HomayoonAlimohammadi (commented):
Thanks a lot @eaudetcobello! Left some minor comments.

src/k8s/pkg/k8sd/app/hooks_bootstrap_test.go
eaudetcobello: I wanted to test validation there, but then I saw we have a cluster_config_validate.go already, so i moved the logic there and test in _test.

src/k8s/pkg/k8sd/types/cluster_config_validate.go
eaudetcobello: not really because we don't know the type of error that validatecidroverlap will return... could be a parsing error or a overlap error

src/k8s/pkg/k8sd/types/cluster_config_validate.go
eaudetcobello: Yep

src/k8s/pkg/k8sd/types/cluster_config_validate.go
bschimke95: The actual error is wrapped in the response so we should be fine.

bschimke95 (approved):
LGTM, great work @eaudetcobello

bschimke95:
please resolve the conflicts then this is good to merge

HomayoonAlimohammadi (approved):
LGTM, thanks a lot @eaudetcobello!

PR #696, opened 2024-09-25
[WIP] poc refactor cilium tests

maci3jka:


src/k8s/pkg/k8sd/features/cilium/network_test.go
```diff
@@ -14,67 +17,128 @@ import (
...
+	for _, tc := range []struct {
+		name   string
+		given  given
+		expect expect
+	}{
+		{
+			name: "NetworkDisabledHelmApplyFails",
+			given: given{
+				config:      types.Network{},
+				expectState: 0,
```
HomayoonAlimohammadi, line 50: maybe we can change this with `helm.Deleted` or the correct helm type for better readability.

HomayoonAlimohammadi (commented):
Thanks @maci3jka! Overall I think this is an OK change but I feel like the current implementation is easier to understand and follow. This is more advanced, complicated and smarter but I think we're not moving in the right direction with this one. WDYT?

maci3jka:
does not improve mainanace

PR #697, opened 2024-09-25
Point cilium to talk to the local apiserver or apiserver-proxy

berkayoz:
Configure cilium to use the local port of apiserver or apiserver-proxy to communicate with the apiserver instead of the kube-proxy service which should address issues with `socketLB`

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -291,6 +291,8 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
 		localhostAddress = "127.0.0.1"
 	}
 
+	cfg.Network.LocalhostAddress = utils.Pointer(localhostAddress)
+
 	// Create directories
 	if err := setup.EnsureAllDirectories(snap); err != nil {
 		return fmt.Errorf("failed to create directories: %w", err)
--- a/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
+++ b/src/k8s/pkg/k8sd/app/hooks_bootstrap.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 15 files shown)


src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -291,6 +291,8 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
 		localhostAddress = "127.0.0.1"
 	}
 
+	cfg.Network.LocalhostAddress = utils.Pointer(localhostAddress)
```
bschimke95, line 294: This will be shared across all nodes. I just thought about if this can cause any kind of race issues in dualstack environments? e.g. if one node sets `--address ipv6` and one node sets `--address myipv4`? 
I don't think this will be an issue - just a thought to consider before merging this.

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -88,9 +88,12 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, _ type
 			},
 		},
 		"nodePort": map[string]any{
-			"enabled": true,
+			"enabled":           true,
+			"enableHealthCheck": false,
```
bschimke95, line 92: could you add a comment, why we disable this?

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -88,9 +88,12 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, _ type
...
 		},
 		"disableEnvoyVersionCheck": true,
+		"k8sServiceHost":           cfg.GetLocalhostAddress(),
+		"k8sServicePort":           6443,
```
bschimke95, line 96: This should be `cfg.GetSecurePort`

src/k8s/pkg/k8sd/features/cilium/network.go
mateoflorido: Maybe we can add a link to the upstream docs:
https://docs.cilium.io/en/stable/network/kubernetes/kubeproxy-free/#kube-proxy-hybrid-modes:~:text=nodePort.enableHealthCheck%20to%20false%2C%20so%20that%20the%20Cilium%20agent%20does%20not%20start%20the%20NodePort%20health%20check%20server

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -88,9 +88,12 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, cfg types.Network, _ type
 			},
 		},
 		"nodePort": map[string]any{
-			"enabled": true,
+			"enabled":           true,
+			"enableHealthCheck": false,
 		},
 		"disableEnvoyVersionCheck": true,
+		"k8sServiceHost":           cfg.GetLocalhostAddress(),
```
HomayoonAlimohammadi, line 95: Maybe we can also test that these are passed in the unit test? Only the fields that are extracted from the config.

HomayoonAlimohammadi (commented):
Thanks a lot @berkayoz! Left a very minor comment.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
berkayoz: Well the apiserver listens to any interface and I've changed apiserver-proxy to do the same. For both cases if there is an ipv6 node, all other nodes needs to have ipv6 routing the same way for ipv4 so I believe we won't face issues in that regard. The only issue would be the old behavior on the apiserver-proxy where we gave it a certain local ip to listen to, which might've caused problems. 

I'm not sure if apiserver-proxy listening to any interface has negative effects/complications.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -184,6 +185,14 @@ func (a *App) onBootstrapWorkerNode(ctx context.Context, s state.State, encodedT
 		localhostAddress = "127.0.0.1"
 	}
 
+	var securePort = "6443"
```
bschimke95, line 188: super small nit: move this below the apiserver check. Otherwise this variable is set without usage.
Also, don't use `var` for initialization. `securePort := "6443"` is more ideomatic.

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy.go
```diff
@@ -20,7 +20,7 @@ func K8sAPIServerProxy(snap snap.Snap, servers []string, localhostAddress string
 	if _, err := snaputil.UpdateServiceArguments(snap, "k8s-apiserver-proxy", map[string]string{
 		"--endpoints":  configFile,
 		"--kubeconfig": filepath.Join(snap.KubernetesConfigDir(), "kubelet.conf"),
-		"--listen":     fmt.Sprintf("%s:6443", localhostAddress),
+		"--listen":     fmt.Sprintf(":%s", securePort),
```
bschimke95, line 23: This uses the default interface instead of the specified localhost, right?
I guess this will fail if we run a ipv6-only cluster on a cluster with both interfaces as IPv4 is preferred.

bschimke95 (commented):
Added some concerns with the localhostaddress replacement with secure port

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy.go
berkayoz: AFAIK It will listen to all interfaces.

https://pkg.go.dev/net#Listen
https://pkg.go.dev/net#Dial
Has some more information related to this.

bschimke95: Fair point, I thought this works similar to the `kubelet` `--node-ip` flag which will prefer the Ipv4 interface if you not set `::` 

But you are right, that is something that we implemented, not upstream.

bschimke95 (approved):
LGTM, great work!

PR #698, opened 2024-09-25
Add IPv6 unittests for cluster setup

bschimke95:


src/k8s/pkg/k8sd/setup/kubelet.go
```diff
@@ -81,7 +81,7 @@ func kubelet(snap snap.Snap, hostname string, nodeIP net.IP, clusterDNS string,
 		args["--cluster-domain"] = clusterDomain
 	}
 	if nodeIP != nil && !nodeIP.IsLoopback() {
-		args["--node-ip"] = utils.ToIPString(nodeIP)
+		args["--node-ip"] = nodeIP.String()
```
bschimke95, line 84: drive-by fix: brackets around the IPv6 are only required/accepted if combined with a port.

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy_test.go
```diff
@@ -130,4 +130,29 @@ func TestK8sApiServerProxy(t *testing.T) {
...
+		g.Expect(setup.K8sAPIServerProxy(s, nil, "[2001:db8::]", nil)).To(Succeed())
+
+		tests := []struct {
+			key         string
+			expectedVal string
+		}{
+			{key: "--listen", expectedVal: "[2001:db8::]:6443"},
+		}
+		for _, tc := range tests {
+			t.Run(tc.key, func(t *testing.T) {
```
HomayoonAlimohammadi, line 150: nit: Maybe we can have a `tc.name` here instead. If we decided to go with this, we might need to change the rest of the tests as well.

src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
```diff
@@ -245,4 +245,30 @@ func TestKubeAPIServer(t *testing.T) {
...
 	})
+
+	t.Run("IPv6", func(t *testing.T) {
+		g := NewWithT(t)
+
+		// Create a mock snap
+		s := mustSetupSnapAndDirectories(t, setKubeletMock)
+		s.Mock.Hostname = "dev"
+
+		g.Expect(setup.KubeAPIServer(s, net.ParseIP("2001:db8::"), "fd98::/108", "https://auth-webhook.url", false, types.Datastore{Type: utils.Pointer("k8s-dqlite")}, "Node,RBAC", nil)).To(Succeed())
```
HomayoonAlimohammadi, line 256: Just a random question. Are these IPs supposed mean anything?

src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
```diff
@@ -245,4 +245,30 @@ func TestKubeAPIServer(t *testing.T) {
...
+
+		tests := []struct {
+			key         string
+			expectedVal string
+		}{
+			{key: "--advertise-address", expectedVal: "2001:db8::"},
```
HomayoonAlimohammadi, line 262: nit: Maybe we can extract these IP strings.

src/k8s/pkg/k8sd/setup/kubelet.go
HomayoonAlimohammadi: Maybe we need to document this somewhere. I personally totally forgot about out brackets convention.

src/k8s/pkg/k8sd/setup/kube_proxy_test.go
```diff
@@ -109,4 +109,30 @@ func TestKubeProxy(t *testing.T) {
...
 	})
+
+	t.Run("IPv6", func(t *testing.T) {
+		g := NewWithT(t)
+
+		// Create a mock snap
+		s := mustSetupSnapAndDirectories(t, setKubeletMock)
+		s.Mock.Hostname = "dev"
+
+		g.Expect(setup.KubeProxy(context.Background(), s, "dev", "fd98::/108", "[::1]", nil)).To(BeNil())
```
HomayoonAlimohammadi, line 120: > drive-by fix: brackets around the IPv6 are only required/accepted if combined with a port.

Isn't `[::1]` in conflict with the above statement?

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy_test.go
```diff
@@ -130,4 +130,29 @@ func TestK8sApiServerProxy(t *testing.T) {
...
 	})
+
+	t.Run("IPv6", func(t *testing.T) {
+		g := NewWithT(t)
+
+		// Create a mock snap
+		s := mustSetupSnapAndDirectories(t, setKubeletMock)
+		s.Mock.Hostname = "dev"
+
+		g.Expect(setup.K8sAPIServerProxy(s, nil, "[2001:db8::]", nil)).To(Succeed())
```
HomayoonAlimohammadi, line 141: > drive-by fix: brackets around the IPv6 are only required/accepted if combined with a port.

Isn't `[2001:db8::]` in conflict with the above statement?

HomayoonAlimohammadi (commented):
Thanks a lot @bschimke95! Left some minor comments.

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy_test.go
bschimke95: not exactly sure what you mean here. Do you want to rename the `tc.key` to `tc.name` or do you want to add a `tc.name` field? If the latter, what value do you see in that?

src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
bschimke95: `2001:db8::` is the IP used in [docs](https://docs.oracle.com/cd/E19253-01/816-4554/ipv6-overview-170/index.html#:~:text=The%20prefix%202001%3Adb8%3A%3A%2F,has%20the%20following%20subnet%20prefix.&text=The%20subnet%20prefix%20always%20contains%2064%20bits.)
`fd98::/108` is the CIDR we use in our examples/integration tests.

src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
bschimke95: this test validates that the `--advertise-address` flag is configured with the IP string that we expect. I don't think it makes sense to have this as an IP object just to stringify it later in the test.

Or do I misunderstand your comment?

src/k8s/pkg/k8sd/setup/kube_proxy_test.go
bschimke95: For the first comment: This is not our convention but the actual IP standard.

For the second: Well, inside the KubeProxy function we assemble this with the corresponding secure port. I agree that this is not nice as it is a somewhat implicit assumption of the implementation. We should probably revisit this in the future.

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy_test.go
bschimke95: Same as above - in the implementation we add a port.

src/k8s/pkg/k8sd/setup/kubelet.go
bschimke95: This is not our convention but the actual IP standard.

berkayoz (approved):
LGTM, we should get this in before #697 since that PR changes some of these setup functions. I can rebase and adjust the unit tests there.

src/k8s/pkg/k8sd/setup/k8s_apiserver_proxy_test.go
HomayoonAlimohammadi: Hmm you're right, there's no real value in that.

src/k8s/pkg/k8sd/setup/kube_apiserver_test.go
HomayoonAlimohammadi: No no I mean that we could have variables for these IPs that we pass to both `setup.KubeAPIServer()` and `expectedVal`. It's not really important, ignore this if you don't feel like it.

HomayoonAlimohammadi (approved):
Thanks a lot @bschimke95! LGTM

PR #700, opened 2024-09-26
[Backport release-1.31] Microcluster Migration History Restoration

louiseschmidtgen:
# Microcluster Migration History Restoration
This PR backports the correction of our Microcluster migrations into 1.31.
Two PRs are backported: #689 #676. One corrects the order and the other properly drops a table instead of removing the migration.


bschimke95 (approved):
LGTM

PR #701, opened 2024-09-27
Improve Nightly Test runs to pass

addyess:
## Overview

Resolves a number of tests issues the nightly tests were having to run effectively on arm64 and amd64 units. 

## Details
* Changes to the github workflow definition 
    * Use system-wide tox installed through apt for arm64 workers to use
   * Correctly upload the right TEST_SNAP after its downloaded
   * Include inspection reports in failures
   * Add tmate session for remote debugging
* Add support for arm64 in the etcd tests (correct release url, correct systemd file)
* Address pytest syntax issue for use of `pytest.mark.xfail`
* Adjust the testing `tox.ini` to be tox 3 and 4 compatible

### DIFF
--- a/.github/workflows/nightly-test.yaml
+++ b/.github/workflows/nightly-test.yaml
@@ -61,6 +61,6 @@ jobs:
           name: ${{ env.artifact_name }}
           path: ${{ github.workspace }}/inspection-reports.tar.gz
       - name: Tmate debugging session
-        if: ${{ failure() }}
+        if: ${{ failure() && github.event_name == 'pull_request' }}
         uses: mxschmitt/action-tmate@v3
         timeout-minutes: 10
--- a/.github/workflows/nightly-test.yaml
+++ b/.github/workflows/nightly-test.yaml
@@ -3,47 +3,47 @@ name: Nightly Latest/Edge Tests
 on:
   schedule:
     - cron: '0 0 * * *' # Runs every midnight
+  pull_request:
 
 permissions:
   contents: read
 
 jobs:
   test-integration:
-    name: Integration Test ${{ matrix.os }} ${{ matrix.arch }} ${{ matrix.releases }}
+    name: Integration Test ${{ matrix.os }} ${{ matrix.arch }} ${{ matrix.release }}
     strategy:
       matrix:
         os: ["ubuntu:20.04", "ubuntu:22.04", "ubuntu:24.04"]
         arch: ["amd64", "arm64"]
-        releases: ["latest/edge"]
+        release: ["latest/edge"]
       fail-fast: false # TODO: remove once arm64 works
 
-    runs-on: ${{ matrix.arch == 'arm64' && 'Ubuntu_ARM64_4C_16G_01' || 'ubuntu-20.04' }}
+    runs-on: ${{ matrix.arch == 'arm64' && 'Ubuntu_ARM64_4C_16G_01' || 'ubuntu-22.04' }}
 
     steps:
       - name: Checking out repo
         uses: actions/checkout@v4
-      - name: Setup Python
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+          cache: 'pip'
+      - name: Install lxd and tox
         run: |
           sudo apt update
-          sudo apt install -y python3 python3-pip
-      - name: Install tox
-        run: |
-          pip3 install tox==4.13
-      - name: Install lxd
-        run: |
+          sudo apt install -y tox
           sudo snap refresh lxd --channel 5.21/stable
           sudo lxd init --auto
           sudo usermod --append --groups lxd $USER
           sg lxd -c 'lxc version'
       - name: Create build directory
         run: mkdir -p build
-      - name: Install $${ matrix.releases }} k8s snap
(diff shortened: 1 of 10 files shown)


tests/integration/tests/test_version_upgrades.py
```diff
@@ -12,7 +12,7 @@
 
 @pytest.mark.node_count(1)
 @pytest.mark.no_setup()
-@pytest.mark.xfail("cilium failures are blocking this from working")
+@pytest.mark.xfail(reason="cilium failures are blocking this from working")
```
addyess, line 15: repairs a syntax error on this xfail call

tests/integration/tests/test_version_upgrades.py
```diff
@@ -26,10 +26,7 @@ def test_version_upgrades(instances: List[harness.Instance]):
                 "'recent' requires the number of releases as second argument and the flavour as third argument"
             )
         _, num_channels, flavour = channels
-        arch = cp.exec(
-            ["dpkg", "--print-architecture"], text=True, capture_output=True
-        ).stdout.strip()
-        channels = snap.get_latest_channels(int(num_channels), flavour, arch)
+        channels = snap.get_latest_channels(int(num_channels), flavour, cp.arch)
```
addyess, line 29: move determining the arch to the Instance class

tests/integration/tests/test_util/harness/base.py
```diff
@@ -30,6 +30,12 @@ def __init__(self, h: "Harness", id: str) -> None:
     def id(self) -> str:
         return self._id
 
+    @cached_property
+    def arch(self) -> str:
+        return self.exec(
+            ["dpkg", "--print-architecture"], text=True, capture_output=True
+        ).stdout.strip()
+
```
addyess, line 38: yay! an instance can know what architecture its running on

tests/integration/tox.ini
```diff
@@ -1,47 +1,46 @@
 [tox]
-no_package = True
```
addyess, line 2: All these changes are to make this more compatible with tox 3.  Why tox 3? Because that's what you get with apt on focal and jammy. Why not pip install it?  Arm64 runners were having severe trouble loading python libraries from a virtual env on the GH runners for reasons i never was able to understand.  I think it's a combination of running tox with sg but i'm not sure.  LD_PYTHON_PATH got lost though -- using the system-wide `tox` improves this

tests/integration/templates/etcd/etcd.service
```diff
@@ -11,6 +11,7 @@
     LimitNOFILE=40000
     TimeoutStartSec=0
 
+    Environment=ETCD_UNSUPPORTED_ARCH=$ARCH
```
addyess, line 14: this is required to run on arm64
https://etcd.io/docs/v3.6/op-guide/supported-platform/

.github/workflows/nightly-test.yaml
```diff
@@ -3,47 +3,43 @@ name: Nightly Latest/Edge Tests
 on:
   schedule:
     - cron: '0 0 * * *' # Runs every midnight
+  pull_request:
```
addyess, line 6: 🚨 Don't merge with this in there -- its just for testing this PR 🚨

.github/workflows/nightly-test.yaml
```diff
@@ -52,4 +48,19 @@ jobs:
...
+        run: |
+          tar -czvf inspection-reports.tar.gz -C ${{ github.workspace }} inspection-reports
+          echo "artifact_name=inspection-reports-${{ matrix.os }}-${{ matrix.arch }}" | sed 's/:/-/g' >> $GITHUB_ENV
+      - name: Upload inspection report artifact
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.artifact_name }}
+          path: ${{ github.workspace }}/inspection-reports.tar.gz
+      - name: Tmate debugging session
```
bschimke95, line 63: IMHO this will just extend the failed test for 10min. Nobody is looking at the nightly tests until they fail. The inspection report should be sufficient to reproduce the issue, if not, we need to isolate the failed test and add the tmate session as part of a draft/test PR.

tests/integration/templates/etcd/etcd.service
bschimke95: let's just bump the version. We don't have any dependencies/reasons to run on v3.3

tests/integration/tests/test_version_upgrades.py
bschimke95: way better - thanks!

bschimke95 (commented):
Thanks for the fixes. I added some comments.

Btw, just a general comment : We eventually need to refactor this nightly test in the future since e.g. we have release independent tests (upgrade tests) and release dependent tests with various flavours.

.github/workflows/nightly-test.yaml
addyess: yep, we can take it out.  I had it in there b/c i was tinkering and trying to learn why things failed but weren't in the inspection reports

bschimke95 (approved):
LGTM, thanks!

PR #703, opened 2024-09-27
Fix IPv6 unittests

bschimke95:
I merged #697 after merging #698 without rebasing. This caused a conflict as the function signature changed.


addyess (approved):
Thanks for the fix @bschimke95

PR #704, opened 2024-09-27
Unittests for cilium k8sd feature

maci3jka:
There are lacks in unittest for cilium feature in k8sd  this pr proves test coverage form 42.9% files, 44.4% statements to  85.7% files, 86.7% statements for github.com/canonical/k8s/pkg/k8sd/features/cilium package

### DIFF
--- a/src/k8s/pkg/k8sd/features/cilium/gateway_test.go
+++ b/src/k8s/pkg/k8sd/features/cilium/gateway_test.go
@@ -3,6 +3,7 @@ package cilium_test
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/canonical/k8s/pkg/client/helm"
@@ -43,7 +44,7 @@ func TestGatewayEnabled(t *testing.T) {
 		g.Expect(err.Error()).To(ContainSubstring(applyErr.Error()))
 		g.Expect(status.Enabled).To(BeFalse())
 		g.Expect(status.Version).To(Equal(cilium.CiliumAgentImageTag))
-		g.Expect(status.Message).To(ContainSubstring("Failed to deploy Cilium Gateway"))
+		g.Expect(status.Message).To(Equal(fmt.Sprintf(cilium.GatewayDeployFailedMsgTmpl, applyErr)))
 		g.Expect(helmM.ApplyCalledWith).To(HaveLen(1))
 
 	})
@@ -101,7 +102,6 @@ func TestGatewayEnabled(t *testing.T) {
 		status, err := cilium.ApplyGateway(context.Background(), snapM, gateway, network, nil)
 
 		g.Expect(err).To(HaveOccurred())
-		g.Expect(err.Error()).To(ContainSubstring("rollout restart cilium to enable Gateway API"))
 		g.Expect(status.Enabled).To(BeFalse())
 		g.Expect(status.Version).To(Equal(cilium.CiliumAgentImageTag))
 		g.Expect(status.Message).To(ContainSubstring("Failed to deploy Cilium Gateway"))
@@ -174,7 +174,7 @@ func TestGatewayDisabled(t *testing.T) {
 		g.Expect(err.Error()).To(ContainSubstring(applyErr.Error()))
 		g.Expect(status.Enabled).To(BeFalse())
 		g.Expect(status.Version).To(Equal(cilium.CiliumAgentImageTag))
-		g.Expect(status.Message).To(ContainSubstring("Failed to delete Cilium Gateway"))
+		g.Expect(status.Message).To(Equal(fmt.Sprintf(cilium.GatewayDeleteFailedMsgTmpl, applyErr)))
 		g.Expect(helmM.ApplyCalledWith).To(HaveLen(1))
 
 	})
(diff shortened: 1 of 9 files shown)


maci3jka:
adresses also KU-1516

src/k8s/pkg/k8sd/features/cilium/gateway_test.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		network := types.Network{}
+		gateway := types.Gateway{
+			Enabled: ptr.To(true),
+		}
+
+		status, err := cilium.ApplyGateway(context.Background(), snapM, gateway, network, nil)
+
+		g.Expect(err).To(HaveOccurred())
+		g.Expect(err.Error()).To(ContainSubstring("rollout restart cilium to enable Gateway API"))
```
HomayoonAlimohammadi, line 104: I'm personally not a fan of checking for error strings, unless we're making sure we see some error that we passed in that test specific test case. Like `applyErr`s in our feature tests.

src/k8s/pkg/k8sd/features/cilium/gateway_test.go
```diff
@@ -0,0 +1,278 @@
...
+		status, err := cilium.ApplyGateway(context.Background(), snapM, gateway, network, nil)
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(status.Enabled).To(BeFalse())
+		g.Expect(status.Version).To(Equal(cilium.CiliumAgentImageTag))
+		g.Expect(status.Message).To(Equal(cilium.DisabledMsg))
+
+		helmCiliumArgs := helmM.ApplyCalledWith[1]
+		g.Expect(helmCiliumArgs.Chart).To(Equal(cilium.ChartCilium))
+		g.Expect(helmCiliumArgs.State).To(Equal(helm.StateDeleted))
+		//validateNetworkValues(g, callArgs.Values, cfg, snapM)
```
HomayoonAlimohammadi, line 206: Maybe we can remove this?

src/k8s/pkg/k8sd/features/cilium/loadbalancer_test.go
```diff
@@ -123,93 +124,120 @@ func TestLoadBalancerEnabled(t *testing.T) {
 		g.Expect(callArgs.Chart).To(Equal(cilium.ChartCilium))
 		g.Expect(callArgs.State).To(Equal(helm.StateUpgradeOnlyOrDeleted(networkCfg.GetEnabled())))
 		g.Expect(callArgs.Values["l2announcements"].(map[string]any)["enabled"]).To(Equal(lbCfg.GetL2Mode()))
-		g.Expect(callArgs.Values["bgpControlPlane"].(map[string]any)["enabled"]).To(Equal(lbCfg.GetL2Mode()))
+		g.Expect(callArgs.Values["bgpControlPlane"].(map[string]any)["enabled"]).To(Equal(lbCfg.GetBGPMode()))
```
HomayoonAlimohammadi, line 127: Nice catch! Thanks!

HomayoonAlimohammadi (commented):
Thanks a lot @maci3jka! Great work. Almost LGTM, but left some minor comments.

src/k8s/pkg/k8sd/features/cilium/ingress_test.go
```diff
@@ -0,0 +1,216 @@
...
+		status, err := cilium.ApplyIngress(context.Background(), snapM, ingress, network, nil)
+
+		g.Expect(err).To(HaveOccurred())
+		g.Expect(status.Enabled).To(BeFalse())
+		g.Expect(status.Message).To(Equal(fmt.Sprintf(cilium.IngressDeployFailedMsgTmpl,
+			fmt.Errorf("failed to rollout restart cilium to apply ingress: %v",
+				fmt.Errorf("failed to restart cilium-operator deployment after 3 attempts: %w",
+					fmt.Errorf("failed to restart cilium-operator deployment: %w",
+						fmt.Errorf("failed to get deployment cilium-operator in namespace kube-system: %w",
+							errors.New("deployments.apps \"cilium-operator\" not found"))))),
```
eaudetcobello, line 144: This is too tightly coupled to the message structure and contents. I see several instances of testing the specific error message in this PR, can those be reverted? I would much rather use a simple errors.Is, if possible.

src/k8s/pkg/k8sd/features/cilium/ingress_test.go
```diff
@@ -0,0 +1,216 @@
...
+			Enabled: ptr.To(true),
+		}
+
+		status, err := cilium.ApplyIngress(context.Background(), snapM, ingress, network, nil)
+
+		g.Expect(err).To(BeNil())
+		g.Expect(status.Enabled).To(BeTrue())
+		g.Expect(status.Message).To(Equal(cilium.EnabledMsg))
+		g.Expect(status.Version).To(Equal(cilium.CiliumAgentImageTag))
+		g.Expect(helmM.ApplyCalledWith).To(HaveLen(1))
```
eaudetcobello, line 193: I'm not a fan of the tight coupling to the implementation details here. There is a time and place for testing the arguments or the number of arguments that a function was called with, e.g. when the function doesn't return anything or isn't easily testable. In this case, we have other information that can be validated, like the status, error, and final state/outcome. Since we can validate the final state or outcome more directly, I'd suggest removing the explicit argument checking from the tests.

eaudetcobello (changes requested):
Nice work. I left 2 comments that are applicable to multiple places in the PR, could you please revisit?

src/k8s/pkg/k8sd/features/cilium/ingress_test.go
maci3jka: I misunderstood other review comments. I updated the test. Regarding errors when it makes sense i use gomega expect(err).To(ErrorMatch()) which checks if the error is in the chain of errors will that be enough?

eaudetcobello (approved):
Nice work, thanks!

PR #710, opened 2024-10-02
Add linter for Go

maci3jka:
This PR improves the development experience. It adds to go checks the linter [golangci-lint,](https://golangci-lint.run/) aggregating the number of liners and adding annotations on the PR with found issues. Additionally, it introduces `go.lint`  target to makefile which allows running the same set of linters locally.


maci3jka:
example of working https://github.com/canonical/k8s-snap/pull/722/files

golangci.yml
```diff
@@ -0,0 +1,351 @@
...
+
+  tenv:
+    # The option `all` will run against whole test files (`_test.go`) regardless of method/function signatures.
+    # Otherwise, only methods that take `*testing.T`, `*testing.B`, and `testing.TB` as arguments are checked.
+    # Default: false
+    all: true
+
+
+linters:
+  disable-all: true
```
HomayoonAlimohammadi, line 214: Overall I feel like it's better to enable all by default and selectively disable the ones that we don't need or regard as useless. But this IMHO will be a very time consuming thing since we will probably face a lot of issues by enabling new linters. I think it's good for now but something to keep in mind for later.

.github/workflows/go.yaml
```diff
@@ -71,6 +72,21 @@ jobs:
...
+          sudo apt update
+          sudo apt install dqlite-tools libdqlite-dev
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.61
+          working-directory: src/k8s
+          args: '-c ./../../golangci.yml'
+          only-new-issues: true
```
HomayoonAlimohammadi, line 88: TBH I expected us to get blown away by the sheer amount of issues that adding this linter would report, but I can see that the linter was executed without any finding any issues. Is this by any means because of the `only-new-issues` option?

HomayoonAlimohammadi (commented):
Thanks a lot @maci3jka, Almost LGTM. Just minor comments and questions.

.github/workflows/go.yaml
```diff
@@ -71,6 +72,21 @@ jobs:
...
+          sudo add-apt-repository ppa:dqlite/dev
+          sudo apt update
+          sudo apt install dqlite-tools libdqlite-dev
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.61
+          working-directory: src/k8s
+          args: '-c ./../../golangci.yml'
```
bschimke95, line 87: Move this file to the go source root, as discussed. They you don't need to do this relative path dance.

golangci.yml
bschimke95: +1 
Please enable all linters. Let's see how much errors we get.

.github/workflows/go.yaml
bschimke95: yes, we don't want that. If we enable this linter then we want the whole code to comply. 

Please disable that option.

bschimke95 (commented):
Going in the right direction.

Please enable all, see how much issues we get and post that here. Then we can create a PR that branches from this PR and fix all of those issues (if they are straightforward).

HomayoonAlimohammadi (commented):
Thanks a lot. This is wonderful. Now we have a shit ton of linting problems to fix. Merging this would cause the k8s-snap pipeline to get stuck in a failing state until everything is fixed which is not good. We can do 2 things:
1. Keep the PR open and gradually fix the issues
2. Change the base from `main` to a custom branch. 

I would go with option 1 because when we're done with the feature branch, rebasing it on main will be a huge pain and is super error prone. 
Options 1 is itself challenging tho. We will still need to make sure we constantly rebase on main to prevent a huge and problematic/conflicting rebase later on.
WDYT @bschimke95?

bschimke95:
I added a list of all linters and disabled them. Now I will gradually enable all of the linters and fix the relevant code pieces as separate PRs.

PR #716, opened 2024-10-07
Ku 1824 doc fixes

evilnick:
A variety of fixes which address:

- incorrect references
- markdown fails
- incorrect headings

all of which were generating warnings in the docs build.


addyess (approved):
Thank you Nick.  Cleaning up these warnings will be nice

PR #720, opened 2024-10-08
Create a test scenario for mixed version clustering

addyess:
### Overview
* creates a new test which clusters two control-plane nodes from different versions of the same flavor
* improves the function scoped pytest marking to support `node_count` or `node_config` when you wish to configure the  nodes differently (say with a different snap version)

### Details
* Run informing tests with a new flag `TEST_FLAVOR` to help identify which flavor is being tested
* add a new pytest marker
* improves the `test_version_upgrades` test to not need to be skipped
* **new & improved** utils methods:
  * `setup_k8s_snap` now with super powers to snap install a file, from a channel, from a track, and from a revision
  * `tracks_least_risk` method to find the least risk revision in a specific track or channel
  * `previous_track` method to find the snap track prior to a given one.

### DIFF
--- a/tests/integration/tests/conftest.py
+++ b/tests/integration/tests/conftest.py
@@ -246,7 +246,7 @@ def session_instance(
     """
     LOG.info("Setup node and enable all features")
 
-    tmp_path = str(tmp_path_factory.mktemp("data"))
+    tmp_path = tmp_path_factory.mktemp("data")
     instance = h.new_instance()
     util.setup_k8s_snap(instance, tmp_path)
 
--- a/tests/integration/tests/conftest.py
+++ b/tests/integration/tests/conftest.py
@@ -86,17 +86,73 @@ def pytest_configure(config):
         "no_setup: No setup steps (pushing snap, bootstrapping etc.) are performed on any node for this test.\n"
         "dualstack: Support dualstack on the instances.\n"
         "etcd_count: Mark a test to specify how many etcd instance nodes need to be created (None by default)\n"
-        "node_count: Mark a test to specify how many instance nodes need to be created\n",
+        "node_count: Mark a test to specify how many instance nodes need to be created\n"
+        "node_config: Mark a test to specify detail configuration for each created node\n",
     )
 
 
 @pytest.fixture(scope="function")
-def node_count(request) -> int:
+def node_config(request) -> List[dict]:
+    """Return a list of node configurations for the test.
+
+    The configuration can be specified using the `node_count` or `node_config` markers.
+    if `node_count` is used, the configuration for each node is empty, only the count is used.
+
+    example)
+    ```python
+    @pytest.mark.node_count(3)
+    def my_test(instances):
+        len(instances) == 3   # because of the node_count marker
+    ```
+
+    if `node_config` is used, the configuration returned is a list of dicts
+    where each dict represents the configuration for a node.
+
+    example)
+    ```python
+    @pytest.mark.node_config("snap", ["latest", "1.31"])
+    def my_other_test(instances):
+        len(instances) == 2   # because of the markers
+        # instances[0] has snap installed from latest track
+        # instances[1] has snap installed from 1.31 track
+    ```
+    or
+    ```python
+    @pytest.mark.node_config("snap,extra_args", [("latest", "something"), ("1.31", "something-else")])
+    def my_other_test(instances):
+        len(instances) == 2   # because of the markers
+        # instances[0] has snap installed from latest track and has extra_args="something"
+        # instances[1] has snap installed from 1.31 track and has extra_args="something-else"
+    ```
+    """
+
+    def _coerce_properties(properties, args):
+        """Coerce properties into a tuple of the correct length."""
+        arg_len = len(args)
+        if isinstance(properties, tuple) and len(properties) == arg_len:
+            coerced = properties
+        elif not isinstance(properties, tuple) and arg_len == 1:
+            coerced = (properties,)
+        assert (
+            len(coerced) == arg_len
+        ), f"Expected {arg_len} properties, got {len(coerced)}"
+        return coerced
+
+    node_config_marker = request.node.get_closest_marker("node_config")
+    if node_config_marker:
+        node_args, node_properties, *_ = node_config_marker.args
+        node_args = node_args.split(",")
+        return [
+            dict(zip(node_args, coerced))
+            for properties in node_properties
+            if (coerced := _coerce_properties(properties, node_args))
+        ]
     node_count_marker = request.node.get_closest_marker("node_count")
-    if not node_count_marker:
-        return 1
-    node_count_arg, *_ = node_count_marker.args
-    return int(node_count_arg)
+    if node_count_marker:
+        node_count_arg, *_ = node_count_marker.args
+        return [{} for _ in range(int(node_count_arg))]
+    # if no markings, create one default instance
+    return [{}]
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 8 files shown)


tests/integration/tests/conftest.py
```diff
@@ -137,23 +193,19 @@ def instances(
...
     """
-    if not config.SNAP:
-        pytest.fail("Set TEST_SNAP to the path where the snap is")
-
+    node_count = len(node_config)
     if node_count <= 0:
         pytest.xfail("Test requested 0 or fewer instances, skip this test.")
 
-    snap_path = (tmp_path / "k8s.snap").as_posix()
-
```
addyess, line 147: this is gone because we can install from all kinds of things now...track, channel, revision, or file

tests/integration/tests/test_version_upgrades.py
```diff
@@ -12,40 +12,45 @@
 
 @pytest.mark.node_count(1)
 @pytest.mark.no_setup()
-@pytest.mark.xfail(reason="cilium failures are blocking this from working")
```
addyess, line 15: @bschimke95 I did not see this to be the case any longer.  I've fixed this test again and it seems to pass??

tests/integration/tests/conftest.py
```diff
@@ -86,17 +86,73 @@ def pytest_configure(config):
...
+    ```python
+    @pytest.mark.node_config("snap", ["latest", "1.31"])
+    def my_other_test(instances):
+        len(instances) == 2   # because of the markers
+        # instances[0] has snap installed from latest track
+        # instances[1] has snap installed from 1.31 track
+    ```
+    or
+    ```python
+    @pytest.mark.node_config("snap,extra_args", [("latest", "something"), ("1.31", "something-else")])
```
bschimke95, line 121: Hm, I think this can become really messy.

Maybe a config dict makes more sense? Then we could also move long configurations to a file

tests/integration/tests/test_version_upgrades.py
bschimke95: yes, I think Berkay implemented a fix.

bschimke95 (commented):
I am a bit concerned on the node_config approach but otherwise looks great!

tests/integration/tests/conftest.py
```diff
@@ -86,17 +86,73 @@ def pytest_configure(config):
...
+    def my_other_test(instances):
+        len(instances) == 2   # because of the markers
+        # instances[0] has snap installed from latest track and has extra_args="something"
+        # instances[1] has snap installed from 1.31 track and has extra_args="something-else"
+    ```
+    """
+
+    def _coerce_properties(properties, args):
```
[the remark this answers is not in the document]
addyess: such a simple word "coerce" -- took me a while to remember it.  But it sounds like the approach of marking tests with `node_config` may be out. 

Maybe i should continue with the idea of peppering tests with markings like `snap_version` or just decorate the test with a "no_setup" and have the test install the snap itself?

tests/integration/tests/conftest.py
```diff
@@ -99,6 +101,15 @@ def node_count(request) -> int:
...
 
+def snap_versions(request) -> Iterator[Optional[str]]:
+    """An endless iterable of snap versions for each node in the test."""
+    marking = ()
+    if snap_version_marker := request.node.get_closest_marker("snap_versions"):
+        marking, *_ = snap_version_marker.args
+    # endlessly repeat of the configured snap version after exhausting the marking
+    return itertools.chain(marking, itertools.repeat(None))
+
+
```
addyess, line 112: Alrighty -- this method is simple in that is produces a steady stream of snap versions to use every time its invoked. 

just don't try to store its results into a finite sequence b/c it returns an infinite number of versions.

tests/integration/tests/test_clustering.py
```diff
@@ -33,6 +33,31 @@ def test_control_plane_nodes(instances: List[harness.Instance]):
     ), f"only {cluster_node.id} should be left in cluster"
 
 
+@pytest.mark.node_count(2)
+@pytest.mark.snap_versions([util.previous_track(config.SNAP), config.SNAP])
```
addyess, line 37: the first node uses the previous track, the second node uses the test snap

addyess:
@bschimke95 I think this should satisfy you

Merge if you give your +1.

HomayoonAlimohammadi (commented):
Thanks a lot @addyess, great work! Left some minor comments and questions.

tests/integration/tests/conftest.py
```diff
@@ -99,6 +101,15 @@ def node_count(request) -> int:
     return int(node_count_arg)
 
 
+def snap_versions(request) -> Iterator[Optional[str]]:
+    """An endless iterable of snap versions for each node in the test."""
+    marking = ()
+    if snap_version_marker := request.node.get_closest_marker("snap_versions"):
+        marking, *_ = snap_version_marker.args
+    # endlessly repeat of the configured snap version after exhausting the marking
+    return itertools.chain(marking, itertools.repeat(None))
```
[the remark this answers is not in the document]
addyess: alright -- so lets clear that up here.  The developer may mark a test with the versions they desire and the node count they desire 

```python
@pytest.mark.node_count(15)
@pytest.mark.snap_versions(["1.31/stable", "1.31/edge", "1.30/stable", ...])
def my_test(instances):
    ...
```

these two markings are not congruent, meaning you can say you want 15 nodes but only specify you'd like 3 snap versions for those instances. We know the first 3 instances should be running from specific channels but how about the other instances?

Let's look at hos `setup_k8s_snap` works:
* `setup_k8s_snap` method takes a `snap` argument.  If it's non-None, the specified version will be installed on the instance.  If it's `None`, it will install from `config.SNAP` -- an environment variable defined when running the test.

So now you see this little gem of code you've highlighted with `itertools.chain(marking, itertools.repeat(None)`

will empty the iterator provided by marking.  And when it's empty it will repeat `None` forever... 

If the test is unmarked with snap_versions, this method spits out `None` infinitely
If the test is marked with snap_versions, this method splits out all the snao versions, until they're consumed, then infinite  `None`


How does that combine with python's `zip`?  Well `zip` will keep consume the next item from all the iterators it's given as a tuple.  It's like taking 1 card off of multiple decks of card and creating a hand from them.  Once a hand is created, it will take 1 card of the decks again for another hand.  Zip will do this until the first deck is consumed and then stop.  

since `snap_versions` is an infinite deck, `zip` stops when the `range(node_count)` is reached.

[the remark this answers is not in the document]
HomayoonAlimohammadi: Got it. Wonderful. Thanks for the thorough explanation. I was confused why the remaining elements of the chain are `None`, and now I know that's because we WANT them to be (so that we can read the version from `config.SNAP`). 
Amazing. Thank you!

tests/integration/tests/test_util/util.py
```diff
@@ -294,3 +336,95 @@ def is_valid_rfc3339(date_str):
         return True
     except ValueError:
         return False
+
+
+def tracks_least_risk(track: str, arch: str) -> str:
```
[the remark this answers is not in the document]
addyess: great question.  The reason behind this is to confirm that a news snap version can be upgraded to from an older existing version.  This method is necessary to find the "old version". 

the `track_least_risk` is way of looking through the channel-map of the snapstore to find a match for a given track and arch, to determine the least risky track.  Say nothing is published to stable risk in the 1.32 track for the newly supported arch `hue64`.  `amd64` and `arm64` of course work, but we're just starting to work things out for this arch.  this method helps us find the correct revision of the snap to load on the instance for a specific arch.  It may be `1.32/edge` -- but we can confirm an upgrade from `1.32/edge` -> `/path/to/local/k8s.snap` works

bschimke95 (approved):
Great work - the new version looks really clean! 
Thanks

PR #724, opened 2024-10-09
test parallel execution

bschimke95:


.github/workflows/integration.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.8'
+          python-version: '3.10'
```
addyess, line 85: I can get behind running the tests with `3.10` or even `3.12`.  What did you run into with 3.8 that didn't work?

.github/workflows/integration.yaml
```diff
@@ -103,6 +103,7 @@ jobs:
           TEST_LXD_IMAGE: ${{ matrix.os }}
           TEST_INSPECTION_REPORTS_DIR: ${{ github.workspace }}/inspection-reports
         run: |
+          python --version
```
addyess, line 106: running pytest through `tox -e integration` should be printing the version right?

tests/integration/requirements-test.txt
```diff
@@ -1,4 +1,5 @@
 coverage[toml]==7.2.5
+pytest-xdist[psutil]==3.6.1
```
addyess, line 2: interesting -- i'm starting to see what's happening here...

addyess (commented):
Just watching what's happening here

bschimke95:
This does not work due to eBPF working on kernel level and messing up multiple deployments on the same host.

PR #728, opened 2024-10-11
Add annotations for configuring which devices cilium use for routing

### LINKED ISSUE #702: Can't access l2 loadbalancers with cilium CNI from a VM
### Summary

I have a simple setup where I'm running LXD and K8S on the same laptop. K8s is bootstrapped on the `lxdbr0` first ip address (host ip on this network), and is configured to advertised a given range on to provide loadbalancer feature.

When deploying from the k8s snap cilium based, I can't reach the LoadBalanced services exposed on k8s:
```
ubuntu@juju-5dd47c-0:~$ curl http://10.206.54.230
curl: (7) Failed to connect to 10.206.54.230 port 80 after 3073 ms: No route to host
```
When deploying with Microk8s or the moonray track, this works correctly:

```
ubuntu@juju-5dd47c-0:~$ curl http://10.206.54.230
404 page not found
```

Graph of setup:
```mermaid
architecture-beta
    group laptop(server)[Laptop]

    service nic(internet)[LXD BR] in laptop
    service vm(server)[LXD VM] in laptop
    service k8s(server)[K8S Snap] in laptop

    vm:L -- R:nic
    k8s:B -- T:nic
```

I should be able to reach l2 advertised ip addresses on my laptop from my LXD VMs

### Reproduction Steps

1. snap install k8s --channel latest/edge --classic
2. sudo k8s bootstrap --address 10.206.54.1 --file k8s.yaml
3. sudo k8s config | juju add-k8s ck8s --controller lxd
4. juju add-model ingressed ck8s
5. juju deploy traefik-k8s --channel 1.0/beta --trust
6. juju add-model compute lxd
7. juju add-machine --constraints virt-type=virtual-machine
8. juju ssh 0 curl http://10.206.54.230 (this is the chosen ip address in the loadbalancer svc)

k8s.yaml:
```yaml
cluster-config:
  network:
    enabled: true
  dns:
    enabled: true
    upstream-nameservers:
      - 10.206.54.1
  load-balancer:
    enabled: true
    cidrs:
    - 10.206.54.230-10.206.54.239
    l2-mode: true
  local-storage:
    enabled: true
    default: true
  ingress:
    enabled: false
  gateway:
    enabled: false
  metrics-server:
    enabled: true
```

### System information

PRETTY_NAME="Ubuntu 24.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.1 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu

### Can you suggest a fix?

_No response_

_No response_

berkayoz:
Fixes #702

Cilium [excludes](https://github.com/cilium/cilium/pull/17560) `bridge` devices while performing auto detection and suggests providing the selection filtermanually through the `devices` flag if auto detection is not suitable.

This PR adds the `k8sd/v1alpha1/cilium/devices` and `k8sd/v1alpha1/cilium/direct-routing-device` annotations to let users configure which devices(interfaces) can be used by cilium. 

Initially we aimed to override this auto detection and select the device based on the node internal ip however the device names could be different on multiple machines which breaks this assumption. As an alternative we could set the `devices` field with a certain default value with wildcard filters to cover common interface names, e.g. `eth+ lxdbr+ enp+`.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_bootstrap_test.go
+++ b/src/k8s/cmd/k8s/k8s_bootstrap_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	apiv1 "github.com/canonical/k8s-snap-api/api/v1"
+	apiv1_annotations "github.com/canonical/k8s-snap-api/api/v1/annotations"
 	cmdutil "github.com/canonical/k8s/cmd/util"
 	"github.com/canonical/k8s/pkg/utils"
 	. "github.com/onsi/gomega"
@@ -63,8 +64,8 @@ var testCases = []testCase{
 				},
 				CloudProvider: utils.Pointer("external"),
 				Annotations: map[string]string{
-					apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove: "true",
-					apiv1.AnnotationSkipStopServicesOnRemove:          "true",
+					apiv1_annotations.AnnotationSkipCleanupKubernetesNodeOnRemove: "true",
+					apiv1_annotations.AnnotationSkipStopServicesOnRemove:          "true",
 				},
 			},
 			ControlPlaneTaints:                 []string{"node-role.kubernetes.io/control-plane:NoSchedule"},
--- a/src/k8s/pkg/k8sd/features/cilium/internal.go
+++ b/src/k8s/pkg/k8sd/features/cilium/internal.go
@@ -1,17 +1,8 @@
 package cilium
 
 import (
+	apiv1_annotations "github.com/canonical/k8s-snap-api/api/v1/annotations/cilium"
 	"github.com/canonical/k8s/pkg/k8sd/types"
-)
-
-const (
-	// List of devices facing cluster/external network (used for BPF NodePort, BPF masquerading and host firewall); supports '+' as wildcard in device name, e.g. 'eth+'
-	// e.g. k8sd/v1alpha1/cilium/devices="eth+ lxdbr+"
-	annotationDevices = "k8sd/v1alpha1/cilium/devices"
-
-	// Device name used to connect nodes in direct routing mode (used by BPF NodePort, BPF host routing; if empty, automatically set to a device with k8s InternalIP/ExternalIP or with a default route
-	// bridge type devices are ingored in automatic selection
-	annotationDirectRoutingDevice = "k8sd/v1alpha1/cilium/direct-routing-device"
 )
 
 type config struct {
@@ -22,11 +13,11 @@ type config struct {
 func internalConfig(annotations types.Annotations) (config, error) {
 	c := config{}
 
-	if v, ok := annotations.Get(annotationDevices); ok {
+	if v, ok := annotations.Get(apiv1_annotations.AnnotationDevices); ok {
 		c.devices = v
 	}
 
-	if v, ok := annotations.Get(annotationDirectRoutingDevice); ok {
+	if v, ok := annotations.Get(apiv1_annotations.AnnotationDirectRoutingDevice); ok {
 		c.directRoutingDevice = v
 	}
 
--- a/src/k8s/pkg/k8sd/features/cilium/internal.go
+++ b/src/k8s/pkg/k8sd/features/cilium/internal.go
@@ -1,0 +1,34 @@
+package cilium
+
+import (
+	"github.com/canonical/k8s/pkg/k8sd/types"
+)
+
+const (
+	// List of devices facing cluster/external network (used for BPF NodePort, BPF masquerading and host firewall); supports '+' as wildcard in device name, e.g. 'eth+'
+	// e.g. k8sd/v1alpha1/cilium/devices="eth+ lxdbr+"
+	annotationDevices = "k8sd/v1alpha1/cilium/devices"
+
+	// Device name used to connect nodes in direct routing mode (used by BPF NodePort, BPF host routing; if empty, automatically set to a device with k8s InternalIP/ExternalIP or with a default route
+	// bridge type devices are ingored in automatic selection
+	annotationDirectRoutingDevice = "k8sd/v1alpha1/cilium/direct-routing-device"
+)
+
+type config struct {
+	devices             string
+	directRoutingDevice string
+}
+
+func internalConfig(annotations types.Annotations) (config, error) {
+	c := config{}
+
+	if v, ok := annotations.Get(annotationDevices); ok {
+		c.devices = v
+	}
+
+	if v, ok := annotations.Get(annotationDirectRoutingDevice); ok {
+		c.directRoutingDevice = v
+	}
+
+	return c, nil
+}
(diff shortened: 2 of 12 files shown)


src/k8s/pkg/k8sd/features/cilium/internal.go
```diff
@@ -0,0 +1,34 @@
+package cilium
+
+import (
+	"github.com/canonical/k8s/pkg/k8sd/types"
+)
+
+const (
+	// List of devices facing cluster/external network (used for BPF NodePort, BPF masquerading and host firewall); supports '+' as wildcard in device name, e.g. 'eth+'
```
bschimke95, line 8: Should we just directly support regular expressions? IMHO just supporting `+` is a bit weird.

src/k8s/pkg/k8sd/features/cilium/internal.go
```diff
@@ -0,0 +1,34 @@
...
+	// e.g. k8sd/v1alpha1/cilium/devices="eth+ lxdbr+"
+	annotationDevices = "k8sd/v1alpha1/cilium/devices"
```
bschimke95, line 10: Those should be part of the API https://github.com/canonical/k8s-snap-api/blob/main/api/v1/const.go

bschimke95: Please directly add them also to the docs page (IMHO this two lines do not justify a new card).

src/k8s/pkg/k8sd/features/cilium/internal.go
bschimke95, line 1: nit: `internal` sounds like this is something a user would not be able to configure, which is not true.
Maybe something like `config.go` or something else (I'm bad at naming things) would fit better?

bschimke95 (commented):
Great stuff, did a first pass

src/k8s/pkg/k8sd/features/cilium/internal.go
berkayoz: This is not something we control, it is [upstream](https://docs.cilium.io/en/latest/network/kubernetes/kubeproxy-free/#nodeport-devices-port-and-bind-settings) implementation. They are only mentioning the `+` option and not explicitly saying it's regex.

src/k8s/pkg/k8sd/features/cilium/internal.go
berkayoz: I'm not sure about this, we don't have calico related bits in the api as well. Should we put feature dependent annotations on the api directly as these can be interchanged?

src/k8s/pkg/k8sd/features/cilium/internal.go
berkayoz: I followed what we have for calico on the naming, we can adjust it to your suggestion if you'd like. Although for me `internal` is fitting since these are internal config to the actual implementation.

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -60,6 +70,16 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, apiserver types.APIServer
 		}, err
 	}
 
+	ciliumNodePortValues := map[string]any{
+		"enabled": true,
+		// kube-proxy also binds to the same port for health checks so we need to disable it
+		"enableHealthCheck": false,
+	}
+
+	if config.directRoutingDevice != "" {
```
HomayoonAlimohammadi, line 79: nit: Maybe we can also check that these config values are present in the final applied helm `values` in `network_test.go`.

HomayoonAlimohammadi (commented):
Great job! Thanks a lot @berkayoz! Left a minor comment.

src/k8s/pkg/k8sd/features/cilium/internal.go
HomayoonAlimohammadi: I think it's a good idea to move all these annotations to the API repo. Maybe in the future we can also move Calico's (or any other feature) annotations there as well.

HomayoonAlimohammadi: If the annotations are moved there, we might also consider re-organizing them from being only in the `consts.go` file to a more structured format that better conveys the annotations intention (something for the future ofc).

src/k8s/pkg/k8sd/features/cilium/internal.go
bschimke95: Right, I noticed this later in the review but forgot to remove my comment. Thanks

src/k8s/cmd/k8s/k8s_bootstrap_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	apiv1 "github.com/canonical/k8s-snap-api/api/v1"
+	apiv1_annotations "github.com/canonical/k8s-snap-api/api/v1/annotations"
```
HomayoonAlimohammadi, line 11: nit: Maybe `annotationsv1`?

src/k8s/pkg/k8sd/features/cilium/network_test.go
```diff
@@ -390,4 +397,14 @@ func validateNetworkValues(g Gomega, values map[string]any, network types.Networ
 	g.Expect(values["ipam"].(map[string]any)["operator"].(map[string]any)["clusterPoolIPv6PodCIDRList"]).To(Equal(ipv6CIDR))
 	g.Expect(values["ipv4"].(map[string]any)["enabled"]).To(Equal((ipv4CIDR != "")))
 	g.Expect(values["ipv6"].(map[string]any)["enabled"]).To(Equal((ipv6CIDR != "")))
+
+	devices, exists := annotations.Get(apiv1_annotations.AnnotationDevices)
```
HomayoonAlimohammadi, line 401: nit: Maybe inline?

HomayoonAlimohammadi (approved):
Thanks a lot @berkayoz! LGTM. Just some minor nit comments.

src/k8s/pkg/k8sd/features/cilium/network_test.go
berkayoz: I don't think we should inline since the returned value is used.

HomayoonAlimohammadi: oopsies! My bad!

src/k8s/cmd/k8s/k8s_bootstrap_test.go
HomayoonAlimohammadi: I don't think this is a good idea. The `corev1` and `metav1` conventions are there because we have `/core/v1` and `/meta/v1`. I guess `apiv1_annotations` is a better choice. I'm OK with both TBH but that's completely up to you. Sorry for the misleading comment.

mateoflorido (approved):

PR #730, opened 2024-10-11
Enable "Canonical Headers" linter

bschimke95:
Rename the worker headers for this to pass.


mateoflorido (approved):
LGTM! Just to confirm, there's no need to ensure backwards compatibility, as this is only used during bootstrap, right?

bschimke95:
yes, this is only used in the worker code.

PR #733, opened 2024-10-11
Enable error linting

bschimke95:
The err checkers require some more refactoring, They are useful but I will consider them in a second iteration of the linting. 

The `errorlint` is super useful!


src/k8s/pkg/k8sd/setup/auth-token-webhook.conf
eaudetcobello, line 1: Should this file be part of the PR?

src/k8s/pkg/k8sd/setup/k8s-apiserver-proxy.json
eaudetcobello, line 1: This file too? Also, missing newline at EOF

eaudetcobello (changes requested):

src/k8s/pkg/k8sd/setup/auth-token-webhook.conf
bschimke95: Thanks - they were created by the unittests and slipped through

eaudetcobello (approved):

PR #735, opened 2024-10-11
Enable type assertions check

bschimke95:
There were a bunch of type assertions that were based on trust and prayers.
Let's validate the type before dereferencing some random pointers.


eaudetcobello (changes requested):
Wrap errors

eaudetcobello (approved):

PR #737, opened 2024-10-11
add ginko linter

bschimke95:
Pretty cool linter that makes nits around Ginko ("better use this assertion, buddy") obsolete.


src/k8s/pkg/utils/errors/errors.go
bschimke95, line 1: obsolete files that where not used anywhere anymore.

eaudetcobello:
This is really good stuff!

PR #741, opened 2024-10-14
IPv6-only address CIDR and Scope

bschimke95:
# Summary

This PR allows the user create IPv6-only clusters on a Dualstack infrastructure.

# Rationale
Right now, there is no way of setting up an IPv6-only cluster on a dualstack host. This is a rare case so setting up a bootstrap config would not be useful and error-prone. Skipping Loopback addresses allows the user to set the preference directly by configuring the address with an IPv6 CIDR of `--address ::/0` 
Also, this PR adds support for scope in link-local IPv6 (`[ip%eth0]:port`). `net.SplitHostPort` supports this out-of-the-box but with the CIDR notation this becomes more tricky.

Note that this is also useful in a CAPI context where we don't know the ip address before and just know that we want "ip6"

# Details
* Skips Loopback IPs in IP discovery for CIDR in `--address`
* Refactor address resolving to accept IPv6 with scopes.
* remove `dualstack` marker and replace it with `network_type` which can be IPv4, Dualstack & IPv6 (Ipv4 is default)
* Modifies dualstack IPv6-only test to set IPv6 CIDR instead of specific address
* Add ipv6-only test for IPv6-only LXD containers to verify that discover works out of the box in this case.


src/k8s/pkg/utils/cidr.go
```diff
@@ -101,6 +103,59 @@ func ParseAddressString(address string, port int64) (string, error) {
 	return util.CanonicalNetworkAddress(address, port), nil
 }
 
+// GetDefaultAddress returns the default IPv4 and IPv6 addresses of the host.
+func GetDefaultAddress() (ipv4, ipv6 string, err error) {
```
HomayoonAlimohammadi, line 107: Is this intended to only be used in the tests?

HomayoonAlimohammadi (commented):
Thanks a lot @bschimke95! Did a first pass, left some minor comments and questions.

tests/integration/tests/conftest.py
```diff
@@ -142,8 +146,8 @@ def instances(
     disable_k8s_bootstrapping: bool,
     no_setup: bool,
     bootstrap_config: Union[str, None],
-    dualstack: bool,
     request,
+    network_type: bool,
```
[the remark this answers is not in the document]
bschimke95: Thanks, Oh Python my love...

src/k8s/pkg/utils/cidr.go
bschimke95: Yeah, I initially also used in the normal implementation but did not use it anymore. However, it simplifies the tests so I decided to keep it. (In utils as it might be useful in other context as well.)

src/k8s/pkg/utils/cidr.go
```diff
@@ -75,6 +78,19 @@ func GetKubernetesServiceIPsFromServiceCIDRs(serviceCIDR string) ([]net.IP, erro
 
 // ParseAddressString parses an address string and returns a canonical network address.
 func ParseAddressString(address string, port int64) (string, error) {
+	re := regexp.MustCompile(`^\[?([a-z0-9:.]+\/[0-9]+)\]?`)
```
berkayoz, line 81: Should we add some example matching strings/comments to make it easier to understand this regex?

PR #743, opened 2024-10-16
fix: ensure nf_conntrack module loaded for kube-proxy.

aznashwan:
This patch ensures that the `nf_conntrack` kernel module is loaded before the `kube-proxy` service is started so it can read some necessary conntrack module-related params from procfs.

Previously, although the `kube-proxy` service always crashed if the module wasn't loaded, this wasn't that common of an occurrence in practice as there are quite a few ways `nf_conntrack` gets loaded transparently:
* Cilium [automatically loads `iptable_nat`](https://github.com/cilium/cilium/blob/63cd391f93b4e2c865268241d384504348672042/pkg/datapath/iptables/iptables.go#L367-L368) after a small startup delay, whose dependency tree includes `nf_conntrack`
* starting firewalld/ufw/most other firewall services
* setting iptables/nftables rules which imply session tracking

By explicitly loading `nf_conntrack` from the `kube-proxy` service wrapper directly, it should ensure the procfs values kube-proxy reads are always present on startup.

### DIFF
--- a/tests/integration/lxd/setup-image.sh
+++ b/tests/integration/lxd/setup-image.sh
@@ -69,6 +69,8 @@ case "${BASE_DISTRO}" in
     # snapd is preinstalled on Ubuntu OSes
     lxc shell tmp-builder -- bash -c 'snap wait core seed.loaded'
     lxc shell tmp-builder -- bash -c 'snap install '"${BASE_SNAP}"
+    # NOTE(aznashwan): 'nf_conntrack' required by kube-proxy:
+    lxc shell tmp-builder -- bash -c 'apt update && apt install -y "linux-modules-$(uname -r)"}'
     ;;
   almalinux)
     # install snapd and ensure /snap/bin is in the environment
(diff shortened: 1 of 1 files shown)


build-scripts/patches/strict/0001-Strict-patch.patch
```diff
@@ -257,6 +257,15 @@ index 54b5fc0b..01631684 100644
 +      load: "on-boot"
 +    - name: xt_socket
 +      load: "on-boot"
++  # Kube-proxy reads some values related to the nf_conntrack module
++  # from procfs on startup, so we must ensure it's loaded.
++  # Note that the 'iptable_nat' module which is automatically loaded
++  # by Cilium would load 'nf_conntrack' as a dependency anyway,
++  # but we add a separate explicit entry for it for safety.
++  conntrack-module-load:
```
bschimke95, line 265: This causes the build to fail:
```
yaml.parser.ParserError: while parsing a block mapping
  in "/home/runner/work/k8s-snap/k8s-snap/snap/snapcraft.yaml", line 4[33](https://github.com/canonical/k8s-snap/actions/runs/11369169743/job/31625959021?pr=743#step:8:34), column 5
expected <block end>, but found '-'
  in "/home/runner/work/k8s-snap/k8s-snap/snap/snapcraft.yaml", line 4[34](https://github.com/canonical/k8s-snap/actions/runs/11369169743/job/31625959021?pr=743#step:8:35), column 5
  ```

bschimke95 (commented):
Generally LGTM but there seems an issue with the patch

bschimke95:
looks like this breaks the integration tests.

aznashwan:
> looks like this breaks the integration tests.

Yep, currently working through the failure on 20.04 in strict mode and will get to the main clustering tests afterwards!

aznashwan:
@bschimke95 the mixed-version cluster test seems to have behaved nicely this time, and the only [Go failure is an unrelated spellcheck](https://github.com/canonical/k8s-snap/actions/runs/11891398483/job/33132064162?pr=743#step:9:22), so PTAL!

bschimke95:
Go unit test failure is unrelated and fixed here: https://github.com/canonical/k8s-snap/pull/809

bschimke95 (approved):
LGTM

PR #747, opened 2024-10-18
Add Annotations Support for Cilium

mateoflorido:
## Overview
Add annotation support for Cilium network features

## Rationale
To enable the Sunbeam team to configure specific functionalities in Cilium, we are adding one of the requested features (VLANBPFBypass) to the annotations support. This configuration can then be passed during the bootstrap  via the configuration file.

## Related Issues
https://github.com/canonical/k8s-operator/issues/112

### DIFF
--- a/src/k8s/pkg/k8sd/features/cilium/internal.go
+++ b/src/k8s/pkg/k8sd/features/cilium/internal.go
@@ -11,10 +11,8 @@ import (
 )
 
 const (
-	// maxVLANTags is the maximum number of VLAN tags that can be configured
-	maxVLANTags = 5
 	// minVLANIDValue is the minimum valid 802.1Q VLAN ID value
-	minVLANIDValue = 1
+	minVLANIDValue = 0
 	// maxVLANIDValue is the maximum valid 802.1Q VLAN ID value
 	maxVLANIDValue = 4094
 )
@@ -31,15 +29,6 @@ func validateVLANBPFBypass(vlanList string) ([]int, error) {
 	vlanList = strings.Trim(vlanList, "{}")
 	vlans := strings.Split(vlanList, ",")
 
-	// Special case: wildcard "0" allows all VLANs
-	if len(vlans) == 1 && strings.TrimSpace(vlans[0]) == "0" {
-		return []int{0}, nil
-	}
-
-	if len(vlans) > maxVLANTags {
-		return []int{}, fmt.Errorf("the VLAN tag list cannot contain more than %d entries unless '0' is used to allow all VLANs", maxVLANTags)
-	}
-
 	vlanTags := make([]int, 0, len(vlans))
 	seenTags := make(map[int]struct{})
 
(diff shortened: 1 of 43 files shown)


src/k8s/pkg/k8sd/features/cilium/internal.go
```diff
@@ -1,18 +1,78 @@
...
+}
+
+func validateVLANBPFBypass(vlanList string) ([]int, error) {
+	vlanList = strings.TrimSpace(vlanList)
+	// Maintain compatibility with the Cilium chart definition
+	vlanList = strings.Trim(vlanList, "{}")
+	vlans := strings.Split(vlanList, ",")
+
+	// Special case: wildcard "0" allows all VLANs
+	if len(vlans) == 1 && strings.TrimSpace(vlans[0]) == "0" {
```
bschimke95, line 35: What happens if a use sets something like this:
```
10,0,11
```
Will cilium handle this ID or will it fail?

src/k8s/pkg/k8sd/features/cilium/internal.go
```diff
@@ -1,18 +1,78 @@
...
+		return []int{0}, nil
+	}
+
+	if len(vlans) > maxVLANTags {
```
bschimke95, line 39: Why do we have this limitation? Is this imposed by cilium?

src/k8s/pkg/k8sd/features/cilium/internal.go
```diff
@@ -1,18 +1,78 @@
...
+		vlanID, err := strconv.Atoi(strings.TrimSpace(vlan))
+		if err != nil {
+			return []int{}, fmt.Errorf("failed to parse VLAN tag: %w", err)
+		}
+		if vlanID < minVLANIDValue || vlanID > maxVLANIDValue {
+			return []int{}, fmt.Errorf("VLAN tag must be between 0 and %d", maxVLANIDValue)
+		}
+
+		if _, ok := seenTags[vlanID]; ok {
+			return []int{}, fmt.Errorf("VLAN tag %d is duplicated", vlanID)
```
bschimke95, line 56: Up for discussion: Is this really an issue that we should fail on?

bschimke95 (commented):
Great work, couple of questions but generally LGTM

src/k8s/pkg/k8sd/features/cilium/internal.go
mateoflorido: Yes. Unfortunately, this a limitation imposed by the `cilium-agent`: https://docs.cilium.io/en/stable/configuration/vlan-802.1q/

src/k8s/pkg/k8sd/features/cilium/internal.go
mateoflorido: I believe we can relax this check and not fail when we are encountering a duplicate. We could simply avoid including the same value multiple times in the chart configuration value. Wdyt?

src/k8s/pkg/k8sd/features/cilium/internal.go
mateoflorido: Cilium will handle that and allow all the VLAN IDs on the interface: https://github.com/cilium/cilium/blob/main/pkg/datapath/linux/config/config.go#L851-L861
@bschimke95 Should I refactor the code to allow the user to provide the wildcard along with other VLAN IDs?

src/k8s/pkg/k8sd/features/cilium/internal.go
bschimke95: sounds good

bschimke95 (approved):
LGTM

PR #748, opened 2024-10-18
Strict interfaces test

louiseschmidtgen:
# Strict interfaces test
This test verifies that the snap store has auto-connected all necessary interfaces for our strict snap in a nightly test on our strict/edge snaps.


.github/workflows/nightly-test.yaml
```diff
@@ -63,3 +63,24 @@ jobs:
...
         timeout-minutes: 10
+
+  test-strict-interfaces:
+    name: Strict Interfaces Test ${{ matrix.release }} ${{ matrix.arch }}
+    strategy:
+      matrix:
+        os: ["ubuntu:20.04"]
+        arch: ["amd64", "arm64"]
+        release: ["1.31/edge", "1.30/edge"]
+      fail-fast: false # TODO: remove once arm64 works
```
bschimke95, line 74: this should work now, right @addyess ?

.github/workflows/nightly-test.yaml
```diff
@@ -63,3 +63,24 @@ jobs:
...
         uses: mxschmitt/action-tmate@v3
         timeout-minutes: 10
+
+  test-strict-interfaces:
+    name: Strict Interfaces Test ${{ matrix.release }} ${{ matrix.arch }}
+    strategy:
+      matrix:
+        os: ["ubuntu:20.04"]
+        arch: ["amd64", "arm64"]
+        release: ["1.31/edge", "1.30/edge"]
```
bschimke95, line 73: This will not be maintainable.
Please do something like this:
https://github.com/canonical/k8s-snap/blob/main/.github/workflows/nightly-test.yaml#L47

so that the channels are dynamically obtained:
https://github.com/canonical/k8s-snap/blob/main/tests/integration/tests/test_version_upgrades.py#L23

Bonus points if you refactor this logic into a reusable function.

bschimke95 (commented):
Generally great work but a comment on maintainability.
We cannot hardcode the channels.

.github/workflows/nightly-test.yaml
louiseschmidtgen: Baking this into Python, then we can integrate this nicely in the integration test like your version upgrades.

tests/integration/tests/test_strict_interfaces.py
```diff
@@ -0,0 +1,76 @@
...
+        cp.exec(
+            [
+                "snap",
+                "run",
+                "--shell",
+                snap_name,
+                "-c",
+                f"snapctl is-connected {interface}",
+            ],
+            check=True,
```
bschimke95, line 75: nit: check should be True by default.

bschimke95 (approved):
LGTM, minor nit

PR #751, opened 2024-10-18
Allow tests to override how long wait for slower clouds providers to pull images

addyess:
### Overview
* GH runners may be quick, but others may be slower
* how long to wait for k8s_ready can differ on these cloud substrates, the tests should be adjustable

### Detail
* using OS env vars to set the RETRIES and DELAY_S for various clouds
* `wait_until_k8s_ready` will default to be `150s`, but can be extended to wait longer or shorter

### DIFF
--- a/tests/integration/tests/test_networking.py
+++ b/tests/integration/tests/test_networking.py
@@ -123,7 +123,9 @@ def test_ipv6_only_on_dualstack_infra(instances: List[harness.Instance]):
         )
 
     # This might take a while
-    util.stubbornly(retries=30, delay_s=20).until(util.ready_nodes(main) == 3)
+    util.stubbornly(retries=config.DEFAULT_WAIT_RETRIES, delay_s=20).until(
+        util.ready_nodes(main) == 3
+    )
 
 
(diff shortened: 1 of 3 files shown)


tests/integration/tests/test_networking.py
```diff
@@ -190,4 +192,6 @@ def test_ipv6_only_on_ipv6_infra(instances: List[harness.Instance]):
         )
 
     # This might take a while
-    util.stubbornly(retries=30, delay_s=20).until(util.ready_nodes(main) == 3)
+    util.stubbornly(retries=config.DEFAULT_WAIT_RETRIES, delay_s=20).until(
```
mateoflorido, line 195: Should we use `DEFAULT_WAIT_DELAY_S` instead of `20`?

mateoflorido (approved):
LGTM! Quick question about the `delay_s` value

tests/integration/tests/test_networking.py
addyess: i don't actually know if this needs 600s to setup or if 150 will do.  Do you know

PR #752, opened 2024-10-18
Fix conflict using manual trivy and the trivy action

addyess:
### Overview
Resolve a conflict trying to run trivy manually and using the trivy-action

### Details
The `trivy-action` clones itself into a directory called `trivy` overwriting the binary just downloaded in that same spot. 
By moving the downloaded file to another directory, they don't' conflict


eaudetcobello (approved):

bschimke95 (approved):
LGTM, thanks for fixing this annoying thing

PR #757, opened 2024-10-22
Add timeout to integration tests LXD container removal

bschimke95:
Sometimes the removing an LXD container is stuck forever. Adding a timeout prevents the CI from hanging indefinitely in this case.
It is very likely that this happens because of how we configure the LXD profile. Since this will undergo significant changes in the future, I did not dig too deeply into this issue.
https://discourse.ubuntu.com/t/lxd-containers-not-removed-in-ci-even-with-force-flag/49039

### DIFF
--- a/tests/integration/tests/test_util/harness/lxd.py
+++ b/tests/integration/tests/test_util/harness/lxd.py
@@ -230,10 +230,10 @@ class LXDHarness(Harness):
         try:
             # There are cases where the instance is not deleted properly and this command is stuck.
             # A timeout prevents this.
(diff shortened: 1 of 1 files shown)


louiseschmidtgen (commented):
Can we please add a TODO here with a note to make a proper fix. 
IMHO, we should rather work on a fix before we keep this timeout in forever.
If we add a timeout I would much rather put a timeout on the entire test to avoid it getting stuck for 6 hours.

HomayoonAlimohammadi (approved):
Thanks a lot @bschimke95! LGTM.

PR #768, opened 2024-11-05
Remove IPv6-only on IPv6 only infra test

bschimke95:
Initially, those tests were skipped on main because of Cilium LB limitations. We now moved to Metallb for this and therefore can enable it everywhere.

Note that we also remove the `test_ipv6_on_ipv6_only_infra` because of limitations on the github runners: https://github.com/actions/runner-images/issues/668


bschimke95:
seems like snap has some issues with IPv6-only LXD containers on 24.04. I'll investigate.

bschimke95:
Turned out that the issue was not 24.04 but rather the ipv6 connectivity, which we cannot support on the runners: https://github.com/actions/runner-images/issues/668

This only popped up on 24.04 because on the previous versions `core20` is already shipped with the image. This is a requirement for the k8s snap and it failed to fetch it due to missing connectivity.

PR #769, opened 2024-11-07
Add minimum release for upgrade tests

bschimke95:
Upgrading from 1.30 to 1.31 is not supported due to incompatible changes. The upgrade tests serve as a promotion gate, so we need to have a way to exclude this release from the tests.

### DIFF
--- a/tests/integration/tests/test_util/snap.py
+++ b/tests/integration/tests/test_util/snap.py
@@ -84,7 +84,7 @@ def get_most_stable_channels(
 
         if min_release is not None:
             _min_release = major_minor(min_release)
-            if version_key < _min_release:
+            if _min_release is not None and version_key < _min_release:
                 continue
 
         if version_key not in channel_map or RISK_LEVELS.index(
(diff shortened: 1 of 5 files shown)


tests/integration/tests/test_util/snap.py
```diff
@@ -76,6 +82,11 @@ def get_most_stable_channels(
     for channel, major, minor, risk in arch_flavor_channels:
         version_key = (int(major), int(minor))
 
+        if min_release is not None:
+            _min_release = major_minor(min_release)
+            if version_key < _min_release:
```
HomayoonAlimohammadi, line 87: nit: `_min_release` might be `None` which will cause `TypeError: '<' not supported between instances of 'tuple' and 'NoneType'`
```suggestion
            if _min_release is not None and version_key < _min_release:
```

HomayoonAlimohammadi (approved):
Great work @bschimke95! LGTM. Just a minor nit, feel free to ignore.

tests/integration/tests/test_util/snap.py
bschimke95: Good catch!

PR #772, opened 2024-11-07
Updates the containerd-related paths

claudiubelu:
We're updating the default paths for containerd to align with the default paths used by upstream. This makes sure any application that requires interaction with containerd, such as the GPU operator can be deployed with upstream instructions.

Adds ContainerdSocketPath, making it easier to refer to this path and simplifying tests.

Adds a check in PreInitChecks, which ensures that there isn't a containerd.sock file already existent in this location, meaning that there is already a containerd instance running. This method is called in the bootstrap and join hooks.

Added unit tests for PreInitChecks.

[UpgradeImpact]: The default locations for the containerd-related files has been changed to align with the default paths used by upstream, instead of the snap's common directory.


src/k8s/pkg/snap/snap.go
```diff
@@ -323,6 +340,17 @@ func (s *snap) PreInitChecks(ctx context.Context, config types.ClusterConfig) er
 		}
 	}
 
+	// check if the containerd.sock file already exists, signaling the fact that another containerd instance
+	// is already running on this node, which will conflict with the snap.
+	socketPath := s.ContainerdSocketPath()
+	if _, err := os.Stat(socketPath); err == nil {
+		return fmt.Errorf("The path '%s' to be used for the containerd socket already exists. "+
```
eaudetcobello, line 347:
```suggestion
		return fmt.Errorf("The path '%s' required for the containerd socket already exists. "+
```

eaudetcobello (approved):
Nice work! Thank you! I found no issues other than a wording nitpick.

src/k8s/pkg/snap/interface.go
```diff
@@ -39,11 +39,12 @@ type Snap interface {
...
-	ContainerdConfigDir() string         // /var/snap/k8s/common/etc/containerd
-	ContainerdExtraConfigDir() string    // /var/snap/k8s/common/etc/containerd/conf.d
-	ContainerdRegistryConfigDir() string // /var/snap/k8s/common/etc/containerd/hosts.d
+	ContainerdConfigDir() string         // /etc/containerd
+	ContainerdExtraConfigDir() string    // /etc/containerd/conf.d
+	ContainerdRegistryConfigDir() string // /etc/containerd/hosts.d
 	ContainerdRootDir() string           // /var/snap/k8s/common/var/lib/containerd
-	ContainerdSocketDir() string         // /var/snap/k8s/common/run
+	ContainerdSocketDir() string         // /run/containerd
+	ContainerdSocketPath() string        // /run/containerd/containerd.sock
```
eaudetcobello, line 47: Actually, can you document both paths here? Like `classic confinement: ..., strict confinement: ...` ?

eaudetcobello (changes requested):

eaudetcobello (approved):

src/k8s/pkg/snap/snap.go
claudiubelu: Done

src/k8s/pkg/snap/interface.go
claudiubelu: Done

PR #773, opened 2024-11-07
Doc build fix

evilnick:
Fixes the doc build system so it builds locally and on RTD


docs/src/charm/explanation/security.md
```diff
@@ -1,2 +1,4 @@
-```{include} /snap/explanation/security.md
+# Security
```
nhennigan, line 1: This creates a double header and double entry on side navigation
image

docs/src/charm/reference/proxy.md
```diff
@@ -1,2 +1,4 @@
-```{include} /snap/reference/proxy.md
+# Proxies
+
```
nhennigan, line 2: Same here again. Do you want to rename it?
image

nhennigan (commented):
I still see the same issues with importing kroki on my machine which is weird because the python versions are the same. The rest of the PR looks fine. I left 2 comments on titles. If this builds fine on rtd I would say go ahead and merge as it is probably a local issue.

nhennigan (approved):

PR #775, opened 2024-11-08
Pass localhost address from the start hook instead of embedding into cluster config

berkayoz:
Pass localhost address from the start hook instead of embedding into cluster config.

### DIFF
--- a/src/k8s/pkg/k8sd/app/cluster_util.go
+++ b/src/k8s/pkg/k8sd/app/cluster_util.go
@@ -3,10 +3,12 @@ package app
 import (
 	"context"
 	"fmt"
+	"net"
 
 	"github.com/canonical/k8s/pkg/k8sd/setup"
 	"github.com/canonical/k8s/pkg/snap"
 	snaputil "github.com/canonical/k8s/pkg/snap/util"
+	mctypes "github.com/canonical/microcluster/v3/rest/types"
 )
 
 func startControlPlaneServices(ctx context.Context, snap snap.Snap, datastore string) error {
@@ -57,4 +59,23 @@ func waitApiServerReady(ctx context.Context, snap snap.Snap) error {
 	}
 
 	return nil
+}
+
+func DetermineLocalhostAddress(clusterMembers []mctypes.ClusterMember) (string, error) {
+	// Check if any of the cluster members have an IPv6 address, if so return "::1"
+	// if one member has an IPv6 address, other members should also have IPv6 interfaces
+	for _, clusterMember := range clusterMembers {
+		memberAddress := clusterMember.Address.Addr().String()
+		nodeIP := net.ParseIP(memberAddress)
+		if nodeIP == nil {
+			return "", fmt.Errorf("failed to parse node IP address %q", memberAddress)
+		}
+
+		if nodeIP.To4() == nil {
+			return "::1", nil
+		}
+	}
+
+	// If no IPv6 addresses are found this means the cluster is IPv4 only
+	return "127.0.0.1", nil
 }
--- a/src/k8s/pkg/k8sd/app/cluster_util.go
+++ b/src/k8s/pkg/k8sd/app/cluster_util.go
@@ -72,7 +72,7 @@ func DetermineLocalhostAddress(clusterMembers []mctypes.ClusterMember) (string, 
 		}
 
 		if nodeIP.To4() == nil {
-			return "::1", nil
+			return "[::1]", nil
 		}
 	}
 
--- a/src/k8s/pkg/k8sd/app/hooks_start.go
+++ b/src/k8s/pkg/k8sd/app/hooks_start.go
@@ -5,7 +5,6 @@ import (
 	"crypto/rsa"
 	"database/sql"
 	"fmt"
-	"net"
 	"time"
 
 	"github.com/canonical/k8s/pkg/k8sd/database"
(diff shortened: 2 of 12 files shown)


src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -61,6 +62,33 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
...
+					return "", fmt.Errorf("failed to get leader client: %w", err)
+				}
+
+				clusterMembers, err := c.GetClusterMembers(ctx)
+				if err != nil {
+					return "", fmt.Errorf("failed to get cluster members: %w", err)
+				}
+
+				// Check if any of the cluster members have an IPv6 address, if so return "::1"
+				// if one member has an IPv6 address, other members should also have IPv6 interfaces
```
HomayoonAlimohammadi, line 77: Quick question:

> other members should also have IPv6 interfaces

Are we assuming this "should"? Maybe we can also force it?
E.g. by making sure every member has a non-nil `nodeIP.To6()`?

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -72,14 +72,19 @@ func NewFeatureController(opts FeatureControllerOpts) *FeatureController {
...
 	setFeatureStatus func(ctx context.Context, name types.FeatureName, featureStatus types.FeatureStatus) error,
 ) {
 	c.waitReady()
 	ctx = log.NewContext(ctx, log.FromContext(ctx).WithValues("controller", "feature"))
 
 	go c.reconcileLoop(ctx, getClusterConfig, setFeatureStatus, features.Network, c.triggerNetworkCh, c.reconciledNetworkCh, func(cfg types.ClusterConfig) (types.FeatureStatus, error) {
-		return features.Implementation.ApplyNetwork(ctx, c.snap, cfg.APIServer, cfg.Network, cfg.Annotations)
+		localhostAddress, err := getLocalhostAddress()
+		if err != nil {
+			return types.FeatureStatus{Enabled: cfg.Network.GetEnabled(), Message: "failed to determine the localhost address"}, fmt.Errorf("failed to get localhost address: %w", err)
```
HomayoonAlimohammadi, line 85: I think we should return `Enabled: false` in this case. According to previous (very old) discussions in the team, mostly with Angelos.

HomayoonAlimohammadi: Now that I'm thinking about it, the feature is "enabled", but not "working" or "responsive" or "ready". 
Maybe we need to change how we show this in the future, but for now, I think `Enabled: false` is probably the way to go.

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -61,6 +62,33 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
...
+					}
+
+					if nodeIP.To4() == nil {
+						return "::1", nil
+					}
+				}
+
+				// If no IPv6 addresses are found this means the cluster is IPv4 only
+				return "127.0.0.1", nil
+			},
```
HomayoonAlimohammadi, line 91: Is it possible to have any sort of test for this?

HomayoonAlimohammadi (commented):
Great work! Thanks a lot @berkayoz! Almost LGTM, just minor comments and questions.

src/k8s/pkg/k8sd/app/hooks_start.go
berkayoz: This is implicit, if a member is able to join with an IPv6 address this means the other members have ipv6 interfaces. If rest of the members did not have ipv6 routing available, the join operation / clustering would never work. I'm also not exactly sure on if IPv4 / IPv6 mixed clustering is supported by microcluster and dqlite. I believe this is not something for us to enforce.

src/k8s/pkg/k8sd/controllers/feature.go
berkayoz: I've went with `Enabled: false` initially however what happens in the case that 
1. we already had a successful deployment of the helm chart
2. reconcile loop kicked in
3. localhost address determination failed
4. retry
AFAIK the deployment is still there, the feature is "enabled" technically right? 

I'm open to setting it to `false` just feels confusing tbh

src/k8s/pkg/k8sd/app/cluster_util.go
```diff
@@ -58,3 +60,22 @@ func waitApiServerReady(ctx context.Context, snap snap.Snap) error {
...
+	// if one member has an IPv6 address, other members should also have IPv6 interfaces
+	for _, clusterMember := range clusterMembers {
+		memberAddress := clusterMember.Address.Addr().String()
+		nodeIP := net.ParseIP(memberAddress)
+		if nodeIP == nil {
+			return "", fmt.Errorf("failed to parse node IP address %q", memberAddress)
+		}
+
+		if nodeIP.To4() == nil {
+			return "::1", nil
```
HomayoonAlimohammadi, line 75: Shouldn't this be `[::1]`?

HomayoonAlimohammadi (approved):
Thanks a lot! LGTM!

src/k8s/pkg/k8sd/features/calico/network.go
```diff
@@ -23,7 +23,7 @@ const (
 // deployment.
 // ApplyNetwork returns an error if anything fails. The error is also wrapped in the .Message field of the
 // returned FeatureStatus.
-func ApplyNetwork(ctx context.Context, snap snap.Snap, apiserver types.APIServer, network types.Network, annotations types.Annotations) (types.FeatureStatus, error) {
+func ApplyNetwork(ctx context.Context, snap snap.Snap, _ string, apiserver types.APIServer, network types.Network, annotations types.Annotations) (types.FeatureStatus, error) {
```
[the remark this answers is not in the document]
berkayoz: Localhost address is not used in Calico currently, compiler would complain about the unused variable.

src/k8s/pkg/k8sd/app/cluster_util.go
berkayoz: Cilium seems to add the square brackets on it's own, so having the brackets here is turned into `[[::1]]` which breaks. Updated to return `[::1]` here and trim these on cilium apply instead.

PR #776, opened 2024-11-08
Add min-tls-version flag

bschimke95:
DISA requires a TLS version of 1.2 as a minimum.

### DIFF
--- a/src/k8s/pkg/k8sd/setup/kube_apiserver.go
+++ b/src/k8s/pkg/k8sd/setup/kube_apiserver.go
@@ -85,6 +85,7 @@ func KubeAPIServer(snap snap.Snap, securePort int, nodeIP net.IP, serviceCIDR st
 		"--service-cluster-ip-range":                 serviceCIDR,
 		"--tls-cert-file":                            filepath.Join(snap.KubernetesPKIDir(), "apiserver.crt"),
 		"--tls-cipher-suites":                        strings.Join(apiserverTLSCipherSuites, ","),
+		"--tls-min-version":                          "VersionTLS12",
 		"--tls-private-key-file":                     filepath.Join(snap.KubernetesPKIDir(), "apiserver.key"),
 	}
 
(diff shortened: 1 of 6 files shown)


src/k8s/pkg/k8sd/setup/kube_apiserver.go
```diff
@@ -85,6 +85,7 @@ func KubeAPIServer(snap snap.Snap, securePort int, nodeIP net.IP, serviceCIDR st
 		"--service-cluster-ip-range":                 serviceCIDR,
 		"--tls-cert-file":                            filepath.Join(snap.KubernetesPKIDir(), "apiserver.crt"),
 		"--tls-cipher-suites":                        strings.Join(apiserverTLSCipherSuites, ","),
+		"--tls-min-version":                          "VersionTLS12",
```
HomayoonAlimohammadi, line 88: Maybe we can extract this `VersionTLS12`? 
```go
const TLSVersion12 string = "VersionTLS12"
```
I was also thinking that we can also go with 1.3 just to be "even more secure". WDYT?

HomayoonAlimohammadi (approved):
Thanks a lot @bschimke95. LGTM, just a minor comment.

src/k8s/pkg/k8sd/setup/kube_apiserver.go
bschimke95: tbh, I don't think that makes much sense and will just cause another indirection. The VersionString does not change. 
I *could* see something like
```
const minTLSVersion string = VersionTLS12
```
but I'd still say that the indirection is not necessary.

As for the 1.3 version: We need to have 1.2 for the DISA but I don't see a reason to further restrict the version (and with that potential users) without a good reason.

HomayoonAlimohammadi: Makes sense :+1:

bschimke95:
SBOM etc fail because of other issues, not related to this PR - merging.

PR #778, opened 2024-11-08
Refactor file writing logic

HomayoonAlimohammadi:
### Overview
Calling `os.WriteFile` might cause partial writes as we've seen with `microk8s-cluster-agent`: https://github.com/canonical/microk8s-cluster-agent/pull/62
This PR adds a `utils.WriteFile` that should be used instead of `os.WriteFile` to prevent partial writes and empty files.


bschimke95 (approved):
LGTM, thanks!

PR #779, opened 2024-11-08
Enable upgrade tests on PRs

bschimke95:
This can only be merged after 1.31 is populated with the current main, otherwise the upgrade test fails.


bschimke95:
Upgrades should run on nightly tests only

PR #781, opened 2024-11-08
remove kroki and replace with SVGS for now

evilnick:
Kroki causes issues with:
 - rate limiting as we are too lazy to host this ourselves
 - build problems on different versions of Python
 
 This PR removes the kroki extension, the directives in the text and replaces with static SVG images until we determine a better replacement.


bschimke95 (approved):
LGTM, thanks!

PR #784, opened 2024-11-08
add cloud-provider to k8s_get

eaudetcobello:
Depends on: https://github.com/canonical/k8s-snap-api/pull/17

cloud-provider config was not available through `k8s get cloud-provider`

### DIFF
--- a/src/k8s/cmd/k8s/k8s_get.go
+++ b/src/k8s/cmd/k8s/k8s_get.go
@@ -48,7 +48,6 @@ func newGetCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 			config := response.Config
 
 			config.MetricsServer = apiv1.MetricsServerConfig{}
-			config.CloudProvider = nil
 			config.Annotations = nil
 
 			var key string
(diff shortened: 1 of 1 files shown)


src/k8s/cmd/k8s/k8s_get.go
```diff
@@ -116,6 +115,8 @@ func newGetCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				output = config.LoadBalancer.GetBGPPeerPort()
 			case fmt.Sprintf("%s.bgp-peer-asn", features.LoadBalancer):
 				output = config.LoadBalancer.GetBGPPeerASN()
+			case "cloud-provider":
+				output = config.GetCloudProvider()
```
addyess, line 119: what happens if it's unset?  I guess that's okay -- it doesn't panic by de referencing nil or something?

eaudetcobello: util.Deref(c.CloudProvider) in the k8s-snap-api dereferences in a safe way. If it's not set, it won't print anything.

PR #785, opened 2024-11-08
Check if Microk8s is installed

eaudetcobello:
Adds a check that uses the snapd API to let the user know microk8s must be uninstalled.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_bootstrap.go
+++ b/src/k8s/cmd/k8s/k8s_bootstrap.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"net"
 	"os"
 	"slices"
 	"strings"
@@ -98,15 +97,6 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				env.Exit(1)
 				return
 			}
-
-			a := fmt.Sprintf("%s:%d", opts.address, config.DefaultPort)
-			ln, err := net.Listen("tcp", a)
-			if err != nil {
-				cmd.PrintErrf("Error: Failed to bind to required port %d. Please make sure the port is not in use.\n", config.DefaultPort)
-				env.Exit(1)
-				return
-			}
-			_ = ln.Close()
 
 			client, err := env.Snap.K8sdClient("")
 			if err != nil {
--- a/src/k8s/cmd/k8s/k8s_bootstrap.go
+++ b/src/k8s/cmd/k8s/k8s_bootstrap.go
@@ -60,7 +60,7 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				env.Exit(1)
 				return
 			}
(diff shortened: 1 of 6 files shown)


src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -47,6 +48,24 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
+				return
+			}
+			microk8sInfo, err := snapdClient.GetSnapInfo("microk8s")
+			if err != nil {
+				cmd.PrintErrln("Error: failed to check if microk8s is installed: %w", err)
+				env.Exit(1)
+				return
+			}
+			if microk8sInfo.StatusCode == 200 {
+				cmd.PrintErrln("Error: microk8s snap is installed, please remove it and try again.")
```
HomayoonAlimohammadi, line 64: Just as a suggestion, maybe we can also guide the user on how to do that? e.g. "... By running `snap remove microk8s`"

src/k8s/pkg/client/snapd/snap_info.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func (c *Client) GetSnapInfo(snap string) (*snapdSnapInfoResponse, error) {
+	resp, err := c.client.Get(fmt.Sprintf("http://localhost/v2/snaps/%s", snap))
+	if err != nil {
+		return nil, fmt.Errorf("failed to get snapd snap info: %w", err)
+	}
+	defer resp.Body.Close()
+
+	resBody, err := io.ReadAll(resp.Body)
```
HomayoonAlimohammadi, line 20: Maybe we can used buffered read here?
p.s. The response is probably never that big so feel free to ignore this. Just as a precaution.

HomayoonAlimohammadi (approved):
Great work! Thanks @eaudetcobello! LGTM, just some minor comments and questions.

berkayoz (approved):
LGTM, just to make sure we only get response code `200` if snap is installed right? 

We should also mention this in our docs, we can do that in a separate PR/card.

HomayoonAlimohammadi:
Quick question: Do we also care if people have any other thing installed that binds to the ports we care about? E.g. maybe `k3s` or...?

eaudetcobello:
> Quick question: Do we also care if people have any other thing installed that binds to the ports we care about? E.g. maybe `k3s` or...? 

Definitely, we have a card for this.

addyess (approved):
LGTM.  The queue of suggestions is full but merge when ready

src/k8s/pkg/client/snapd/snap_info.go
eaudetcobello: I don't think that's necessary for a few kilobytes (if even that).

src/k8s/cmd/k8s/k8s_bootstrap.go
eaudetcobello: Sure

PR #786, opened 2024-11-08
KU-2052 CIS hardening explanation page

nhennigan:


eaudetcobello (approved):
LGTM

evilnick:
I think when we have more of these we will probably want a more general 'compliance' section to put these in, but this is fine for now

evilnick (approved):
:cake:

PR #787, opened 2024-11-11
Temporary fix for sbom generation for metrics-server

berkayoz:
Temporary fix for sbom generation for metrics-server


louiseschmidtgen (approved):
LGTM, consider making `0.7.0/rockcraft.yaml` a variable. This way, we only have it hard-coded in one place until updated.

PR #788, opened 2024-11-12
Cleanup left-over iptables rules from kubeproxy and cilium

berkayoz:
`cilium-dbg` seems to clean up interfaces, bpf programs, network namespaces properly. Seems like we are leaking some iptables rules so added a cleanup step for left-over iptables rules from kubeproxy and cilium.

### DIFF
--- a/src/k8s/pkg/k8sd/features/cilium/cleanup.go
+++ b/src/k8s/pkg/k8sd/features/cilium/cleanup.go
@@ -27,11 +27,8 @@ func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
 
 		lines := strings.Split(string(out), "\n")
 		for i, line := range lines {
-			for _, word := range []string{"cilium", "kube", "CILIUM", "KUBE"} {
-				if strings.Contains(line, word) {
-					lines[i] = ""
-					break
-				}
+			if strings.Contains(strings.ToLower(line), "cilium") {
+				lines[i] = ""
 			}
 		}
 
--- a/src/k8s/pkg/k8sd/features/cilium/cleanup.go
+++ b/src/k8s/pkg/k8sd/features/cilium/cleanup.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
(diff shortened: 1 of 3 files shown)


eaudetcobello:
Can we back up the original in /var/snap before overwriting in case we mess up someone's iptables? Also, are you sure about the assumption that every rule created by cilium or kubernetes contains the keywords you defined?

src/k8s/pkg/k8sd/features/cilium/cleanup.go
```diff
@@ -18,5 +19,28 @@ func CleanupNetwork(ctx context.Context, snap snap.Snap) error {
...
+		}
+
+		lines := strings.Split(string(out), "\n")
+		for i, line := range lines {
+			for _, word := range []string{"cilium", "kube", "CILIUM", "KUBE"} {
+				if strings.Contains(line, word) {
+					lines[i] = ""
+					break
+				}
+			}
```
HomayoonAlimohammadi, line 35: Maybe we can go with the following?
(Not sure if it's okay to count words like `Kubernetes` or `Cilium` as well).
```suggestion
			for _, word := range []string{"cilium", "kube"} {
				if strings.Contains(strings.ToLower(line), word) {
					lines[i] = ""
					break
				}
			}
```

src/k8s/pkg/k8sd/features/cilium/cleanup.go
HomayoonAlimohammadi, line 1: Is it possible to have a unit (or even maybe even better) e2e test covering this? 
Maybe something for another card tho.

HomayoonAlimohammadi (commented):
Great work! Thanks a lot Berkay! Overall LGTM, just some minor comments and questions.
+1 on the backup suggestion by Etienne.

src/k8s/pkg/k8sd/features/cilium/cleanup.go
berkayoz: We can look into creating an e2e test as a different card.

HomayoonAlimohammadi (approved):
Thanks a lot @berkayoz! LGTM!

PR #789, opened 2024-11-12
KU-2068 reformatted annotations table

nhennigan:
Due to formatting issues with large amounts of text, the annotations table was quite unclear. Edited to make it more readable

### DIFF
--- a/docs/src/snap/reference/annotations.md
+++ b/docs/src/snap/reference/annotations.md
@@ -1,35 +1,169 @@
 # Annotations
 
 This page outlines the annotations that can be configured during cluster
-[bootstrap]. To do this, set the cluster-config/annotations parameter in
+[bootstrap]. To do this, set the `cluster-config/annotations` parameter in
 the bootstrap configuration.
 
-| Name                                                          | Description                                                                                                                                                                                                                                       | Values          |
-|---------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------|
-| `k8sd/v1alpha/lifecycle/skip-cleanup-kubernetes-node-on-remove` | If set, only microcluster and file cleanup are performed.  This is helpful when an external controller (e.g., CAPI) manages the Kubernetes node lifecycle. By default,  k8sd will remove the Kubernetes node when it is removed from the cluster. | "true"\|"false" |
-| `k8sd/v1alpha/lifecycle/skip-stop-services-on-remove` | If set, the k8s services will not be stopped on the leaving node when removing the node. This is helpful when an external controller (e.g., CAPI) manages the Kubernetes node lifecycle. By default, all services are stopped on leaving nodes. | "true"\|"false" |
-| `k8sd/v1alpha1/csrsigning/auto-approve` | If set, certificate signing requests created by worker nodes are auto approved. | "true"\|"false" |
-| `k8sd/v1alpha1/calico/apiserver-enabled` | Enable the installation of the Calico API server to enable management of Calico APIs using kubectl. | "true"\|"false" |
-| `k8sd/v1alpha1/calico/encapsulation-v4` | The type of encapsulation to use on the IPv4 pool. | "IPIP"\|"VXLAN"\|"IPIPCrossSubnet"\|"VXLANCrossSubnet"\|"None" |
-| `k8sd/v1alpha1/calico/encapsulation-v6` | The type of encapsulation to use on the IPv6 pool. | "IPIP"\|"VXLAN"\|"IPIPCrossSubnet"\|"VXLANCrossSubnet"\|"None" |
-| `k8sd/v1alpha1/calico/autodetection-v4/firstFound` | Use default interface matching parameters to select an interface, performing best-effort filtering based on well-known interface names. | "true"\|"false" |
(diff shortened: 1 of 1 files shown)


docs/src/snap/reference/annotations.md
```diff
@@ -1,35 +1,169 @@
 # Annotations
 
 This page outlines the annotations that can be configured during cluster
-[bootstrap]. To do this, set the cluster-config/annotations parameter in
+[bootstrap]. To do this, set the `cluster-config/annotations` parameter in
 the bootstrap configuration.
```
evilnick, line 5: include an example? I think we should get in the habit of doing this at the top of the file, since that is likely the only hint people will need to be effective.

docs/src/snap/reference/annotations.md
```diff
@@ -1,35 +1,169 @@
...
+|   |   |
+|---|---|
+|**Values**| string |
+|**Description**|Enable IP auto-detection based on which source address on the node is used to reach the specified IP or domain.|
+
+## `k8sd/v1alpha1/calico/autodetection-v6/cidrs`
+
+|   |   |
+|---|---|
+|**Values**| []string (comma separated) |
```
evilnick, line 130: the description here is inconsistent. I get that []string means something to golang, but we aren't expecting all readers to be golang experts. We should probably settle on a single way of describing this. suggestions?

docs/src/snap/reference/annotations.md
```diff
@@ -1,35 +1,169 @@
...
-| `k8sd/v1alpha1/cilium/direct-routing-device` | Device name used to connect nodes in direct routing mode (used by BPF NodePort, BPF host routing); if empty, automatically set to a device with k8s InternalIP/ExternalIP or with a default route. Bridge type devices are ignored in automatic selection | string |
-| `k8sd/v1alpha1/cilium/vlan-bpf-bypass` | Comma separated list of VLAN tags to bypass eBPF filtering on native devices. Cilium enables firewalling on native devices and filters all unknown traffic, including VLAN 802.1q packets, which pass through the main device with the associated tag (e.g., VLAN device eth0.4000 and its main interface eth0). Supports `0` as wildcard for bypassing all VLANs. e.g. `4001,4002` | []string |
-| `k8sd/v1alpha1/metrics-server/image-repo` | Override the default image repository for the metrics-server. | string |
-| `k8sd/v1alpha1/metrics-server/image-tag` | Override the default image tag for the metrics-server. | string |
+## `k8sd/v1alpha/lifecycle/skip-cleanup-kubernetes-node-on-remove`
```
evilnick, line 7: These are unreadable as h2s, however, they need to be h2 to appear in the list on the right, which we definitely want. I wonder if we can do a local override to the h2 style to make these less unpleasant

evilnick:
This is a great improvement! I still think we can do better though

docs/src/snap/reference/annotations.md
```diff
@@ -1,35 +1,179 @@
 # Annotations
 
 This page outlines the annotations that can be configured during cluster
-[bootstrap]. To do this, set the cluster-config/annotations parameter in
+[bootstrap]. To do this, set the `cluster-config/annotations` parameter in
```
eaudetcobello, line 4: People typically expect dots to indicate nesting for json and yaml structures. So, `cluster-config.annotations`

evilnick:
i think we can maybe bump up the size to 1.5em

docs/src/snap/reference/annotations.md
```diff
@@ -1,34 +1,187 @@
...
+
+|   |   |
+|---|---|
+|**Values**| string|
+|**Description**|Override the default image tag for the metrics-server.|
+
+<script>
+const el = document.getElementsByTagName("h2");
+for(var i=0;i<el.length;i++){
+  el[i].style.fontSize = '1.2em';
```
evilnick, line 181:
```suggestion
  el[i].style.fontSize = '1.5em';
```

docs/src/snap/reference/annotations.md
nhennigan: Done

docs/src/snap/reference/annotations.md
```diff
@@ -1,34 +1,187 @@
...
+For example:
+
+```yaml
+cluster-config:
+...
+    annotations:
+        k8sd/v1alpha/lifecycle/skip-cleanup-kubernetes-node-on-remove: true
+        k8sd/v1alpha/lifecycle/skip-stop-services-on-remove: true
+```
+
```
evilnick, line 16: might be nice to have some text here to ease the transition to the actual list of annotations

docs/src/snap/reference/annotations.md
```diff
@@ -1,34 +1,187 @@
 # Annotations
 
 This page outlines the annotations that can be configured during cluster
```
evilnick, line 3: do we want to link to somewhere that explains what annotations are in this context? Like https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/ maybe?

nhennigan: Thats a good idea. I'll add that now.

bschimke95:
Please rebase this PR.

evilnick (approved):
:cake:

PR #792, opened 2024-11-13
Increase timeouts and log test timestamps

petrutlucian94:
We're waiting about 5 minutes for new nodes to become ready, however sometimes this isn't enough. It can take around one minute just to pull the cilium image.

For this reason, we'll double the test timeouts again.

While at it, we're configuring pytest to log timestamps, which allows us to correlate test logs with the k8s logs.


berkayoz (approved):
LGTM, we can adjust the retries later on if this proves to increase total runtime until failure.

PR #803, opened 2024-11-15
Update Cilium to 1.16.3

berkayoz:
Updates Cilium to  `v1.16.3`
Updates Gateway API to version `v1.1.0`
Fixed a bug related to `upgrade.ReuseValues`, chart defaults would not be set on non-existing fields/manifests. Instead switched to `upgrade.ResetThenReuseValues` which first calculates the initial values then applies the user-supplied config on top.
`cilium-dbg cleanup` is renamed to `cilium-dbg post-uninstall-cleanup`


eaudetcobello:
Do we also update the SBOM in generate-sbom.py?

bschimke95:
Good point, updated the version

bschimke95 (approved):
LGTM

PR #806, opened 2024-11-15
Bump coredns to 1.11.3 and coredns chart to 1.36.0

eaudetcobello:
* Bump coredns to 1.11.3
* Bump coredns chart to 1.36.0


build-scripts/hack/sync-images.yaml
```diff
@@ -8,8 +8,8 @@ sync:
   - source: ghcr.io/canonical/cilium:1.15.2-ck2
     target: '{{ env "MIRROR" }}/canonical/cilium:1.15.2-ck2'
     type: image
-  - source: ghcr.io/canonical/coredns:1.11.1-ck4
-    target: '{{ env "MIRROR" }}/canonical/coredns:1.11.1-ck4'
+  - source: ghcr.io/canonical/coredns:1.11.3
+    target: '{{ env "MIRROR" }}/canonical/coredns:1.11.3
```
eaudetcobello, line 12: Missing '

build-scripts/hack/update-coredns-chart.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+VERSION="1.36.0"
+DIR=$(realpath $(dirname "${0}"))
+
+CHARTS_PATH="$DIR/../../k8s/manifests/charts"
+
+cd "$CHARTS_PATH"
```
HomayoonAlimohammadi, line 8: Maybe we can use `-d` flag in `helm pull` instead of `cd`?

src/k8s/pkg/k8sd/features/coredns/chart.go
```diff
@@ -11,12 +11,12 @@ var (
...
-		ManifestPath: filepath.Join("charts", "coredns-1.29.0.tgz"),
+		ManifestPath: filepath.Join("charts", "coredns-1.36.0.tgz"),
 	}
 
 	// imageRepo is the image to use for CoreDNS.
 	imageRepo = "ghcr.io/canonical/coredns"
 
 	// ImageTag is the tag to use for the CoreDNS image.
-	ImageTag = "1.11.1-ck4"
+	ImageTag = "1.11.3"
```
HomayoonAlimohammadi, line 21: Something for later: Maybe we can use resolve these from the actual binaries. Having these versions and numbers duplicated a multiple places scares me...
Created a card for it.

build-scripts/hack/update-coredns-chart.sh
eaudetcobello: We could but that would mean changing the other scripts to do that as well. The user's directory is not changed when he runs the script, so functionally it's the same thing.

build-scripts/hack/sync-images.yaml
bschimke95: fixed

PR #809, opened 2024-11-16
revert spelling change

eaudetcobello:
Revert a spelling change in https://github.com/canonical/k8s-snap/commit/79f3145c038b1b5901264720a1c06daf0969517d that causes the `make go.doc` command to make a change during the workflow, which fails the subsequent `git diff` check.


bschimke95 (approved):
LGTM

PR #811, opened 2024-11-18
Docs refresh 1.31

evilnick:
This refreshes the content for 1.31, which fixes the navigation.

@nhennigan 
The previous docs fixed the tools, but the files in src didn't get updated for some reason - maybe some symlink related nonsense?
I have built this locally and it works, but please test


nhennigan (approved):
Builds on my system locally. LGTM

PR #813, opened 2024-11-19
Add containerd path marker file

bschimke95:
We need to properly clean up the containerd path on snap removal. For that, the path needs to be stored in a file.
This serves two purposes:
1. The existence of the file indicates that the cluster was already bootstrapped and the containerd directory is not created by some other service.
2. The containerd path is configurable, having this information in a file makes it easy to access even after the k8sd service is already stopped.

Also, updates the `PreInitCheck` to check for the existence of the containerd directory, not the unix socket as the service might just be stopped temporarily.

### DIFF
--- a/k8s/lib.sh
+++ b/k8s/lib.sh
@@ -46,7 +46,7 @@ k8s::common::is_strict() {
 # Cleanup configuration left by the network feature
 k8s::remove::network() {
   k8s::common::setup_env
-  
+
   "${SNAP}/bin/kube-proxy" --cleanup || true
 
   k8s::cmd::k8s x-cleanup network || true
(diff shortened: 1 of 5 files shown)


k8s/lib.sh
```diff
@@ -94,6 +94,16 @@ k8s::remove::containers() {
...
 
+k8s::remove::containerd() {
+  k8s::common::setup_env
+
+  # only remove containerd if the snap was already bootstrapped.
+  # this is to prevent removing containerd when it is not installed by the snap.
+  if [ -f "$SNAP_COMMON/lock/containerd-socket-path" ]; then
+     rm -f $(cat "$SNAP_COMMON/lock/containerd-socket-path")
+  fi
+}
```
eaudetcobello, line 105: If this is removing a directory, wouldn't you need `rm -rf`? Also, if this is the containerd directory path, maybe `containerd-socket-path` is not the most accurate name?

eaudetcobello (approved):
Looks good. Please make a note somewhere so we don't forget to cleanup the directory once @claudiubelu's work lands.

PR #815, opened 2024-11-19
Validate ports in PreInitChecks

eaudetcobello:
This function runs before any of the k8s binaries are started, it's the right place to validate that the api-server, kubelet, etc. ports are available.

Note: containerd doesn't seem to use a consistent port, which complicates checking if it's port is available.
Note: When a port is in use, I sometimes get a `The error was: failed to POST /k8sd/cluster: Post "http://control.socket/1.0/k8sd/cluster": EOF` error before the validation happens. I'm investigating this.

### DIFF
--- a/src/k8s/pkg/snap/snap.go
+++ b/src/k8s/pkg/snap/snap.go
@@ -344,8 +344,7 @@ func (s *snap) PreInitChecks(ctx context.Context, config types.ClusterConfig) er
 		for _, address := range []string{"127.0.0.1", "0.0.0.0"} {
 			listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", address, port))
 			if err != nil {
-				listener.Close()
-				return fmt.Errorf("couldn't bind to required address %s. Is it in use?", address)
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/snap/snap.go
```diff
@@ -329,7 +330,25 @@ func (s *snap) SnapctlSet(ctx context.Context, args ...string) error {
 }
 
 func (s *snap) PreInitChecks(ctx context.Context, config types.ClusterConfig) error {
-	// TODO: check for available ports for k8s-dqlite, apiserver, containerd, etc
+	// TODO(eac): Make these configurable or part of a Config struct
+	ports := []int{
+		config.APIServer.GetSecurePort(),    // kubeAPIServerPort
+		config.Datastore.GetK8sDqlitePort(), // k8sDqlitePort
+		10248,                               // kubeletHealthzPort
```
ktsakalozos, line 337: All these ports can be set in the service extra args. We need to check from there for the right values.

PR #817, opened 2024-11-20
Adds configurable containerd base dir during bootstrap / join

claudiubelu:
Adds configurable containerd base dir during bootstrap / join

Currently, for the classic k8s snap, the default locations for the containerd-related files are:

```
- /etc/containerd/
- /run/containerd/
- /var/lib/containerd/
```

These paths can conflict with other containerd installations on the host (e.g. from docker), meaning that the k8s snap cannot be installed. In a developer's usecase, we shouldn't require the other services to be disabled, but we can allow the k8s snap's containerd to be installed in a different location during bootstrap / node join.

Adds the ``--containerd-base-dir`` option for k8s bootstrap and join-cluster commands. If not given, the snap's default paths will be used instead. After the node was initialized, save the containerd base location in a file, so we can properly reference it later / on restart.

Depends-On: https://github.com/canonical/k8s-snap-api/pull/19

### DIFF
--- a/k8s/lib.sh
+++ b/k8s/lib.sh
@@ -99,6 +99,7 @@ k8s::remove::containerd() {
 
   # only remove containerd if the snap was already bootstrapped.
   # this is to prevent removing containerd when it is not installed by the snap.
+  # NOTE: do NOT include .containerd-base-dir! By default, it will contain "/".
   for file in "containerd-socket-path" "containerd-config-dir" "containerd-root-dir" "containerd-cni-bin-dir"; do
     if [ -f "$SNAP_COMMON/lock/$file" ]; then
       rm -rf $(cat "$SNAP_COMMON/lock/$file")
--- a/src/k8s/cmd/util/environ.go
+++ b/src/k8s/cmd/util/environ.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/canonical/k8s/pkg/snap"
@@ -32,10 +33,17 @@ func DefaultExecutionEnvironment() ExecutionEnvironment {
 	var s snap.Snap
 	switch os.Getenv("K8SD_RUNTIME_ENVIRONMENT") {
 	case "", "snap":
+		// If this node is already bootstrapped / joined, we should already know where the
+		// containerd base directory is. If not, leave the defaults.
+		containerdBaseDir := ""
+		if data, err := os.ReadFile(filepath.Join(os.Getenv("SNAP_COMMON"), "lock", snap.ContainerdBaseDir)); err == nil {
+			containerdBaseDir = strings.TrimSpace(string(data))
+		}
 		s = snap.NewSnap(snap.SnapOpts{
-			SnapDir:          os.Getenv("SNAP"),
-			SnapCommonDir:    os.Getenv("SNAP_COMMON"),
-			SnapInstanceName: os.Getenv("SNAP_INSTANCE_NAME"),
+			SnapDir:           os.Getenv("SNAP"),
+			SnapCommonDir:     os.Getenv("SNAP_COMMON"),
+			SnapInstanceName:  os.Getenv("SNAP_INSTANCE_NAME"),
+			ContainerdBaseDir: containerdBaseDir,
 		})
 	case "pebble":
 		s = snap.NewPebble(snap.PebbleOpts{
--- a/src/k8s/pkg/k8sd/api/cluster_bootstrap.go
+++ b/src/k8s/pkg/k8sd/api/cluster_bootstrap.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"path/filepath"
 	"time"
 
 	apiv1 "github.com/canonical/k8s-snap-api/api/v1"
@@ -38,6 +39,12 @@ func (e *Endpoints) postClusterBootstrap(_ state.State, r *http.Request) respons
 
 	if status.Ready {
 		return response.BadRequest(fmt.Errorf("cluster is already bootstrapped"))
+	}
+
+	// If not set, leave the default base dir location.
+	if req.Config.ContainerdBaseDir != "" {
+		// append k8s-containerd to the given base dir, so we don't flood it with our own folders.
+		e.provider.Snap().SetContainerdBaseDir(filepath.Join(req.Config.ContainerdBaseDir, "k8s-containerd"))
 	}
 
 	// NOTE(neoaggelos): microcluster adds an implicit 30 second timeout if no context deadline is set.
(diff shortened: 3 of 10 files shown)


src/k8s/go.mod
```diff
@@ -177,3 +177,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/canonical/k8s-snap-api v1.0.13 => github.com/claudiubelu/k8s-snap-api v0.0.0-20241119131040-dc75989e4d14
```
claudiubelu, line 181: NOTE: this change and the ``go.sum`` one is only for testing purposes.

src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -170,6 +172,7 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 	cmd.Flags().StringVar(&opts.address, "address", "", "microcluster address or CIDR, defaults to the node IP address")
 	cmd.Flags().StringVar(&opts.outputFormat, "output-format", "plain", "set the output format to one of plain, json or yaml")
 	cmd.Flags().DurationVar(&opts.timeout, "timeout", 90*time.Second, "the max time to wait for the command to execute")
+	cmd.Flags().StringVar(&opts.containerdBaseDir, "containerd-base-dir", "", "set the containerd base containerd directory")
```
eaudetcobello, line 175: Has there been agreement from higher-ups that this should be a CLI flag?

src/k8s/pkg/k8sd/api/cluster_bootstrap.go
```diff
@@ -36,6 +36,11 @@ func (e *Endpoints) postClusterBootstrap(_ state.State, r *http.Request) respons
 		return response.BadRequest(fmt.Errorf("cluster is already bootstrapped"))
 	}
 
+	// If not set, leave the default base dir location.
+	if req.ContainerdBaseDir != "" {
+		e.provider.Snap().SetContainerdBaseDir(req.ContainerdBaseDir)
+	}
```
eaudetcobello, line 42: I would rather see this in BootstrapConfig's SetDefaults().

bschimke95 (changes requested):
We don't want this option to be part of the CLI. This should be an optional entry in the bootstrap config file.

src/k8s/pkg/snap/snap.go
```diff
@@ -23,6 +23,10 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
+const (
+	CONTAINERD_BASE_DIR = ".containerd-base-dir"
```
HomayoonAlimohammadi, line 27: I think we should use `ContainerdBaseDir` here: https://go.dev/doc/effective_go#names

src/k8s/cmd/k8s/k8s_bootstrap.go
claudiubelu: Done, went with the bootstrap config file option.

src/k8s/pkg/snap/snap.go
claudiubelu: Done

eaudetcobello:
Instead of creating a new feature, can this be set by the ExtraContainerdArgs bootstrap option we already have?

bschimke95:
please resolve the conflicts

k8s/lib.sh
```diff
@@ -99,6 +99,7 @@ k8s::remove::containerd() {
 
   # only remove containerd if the snap was already bootstrapped.
   # this is to prevent removing containerd when it is not installed by the snap.
+  # NOTE: do NOT include .containerd-base-dir!
```
eaudetcobello, line 102: Please add a short explanation why

src/k8s/go.mod
```diff
@@ -177,3 +177,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/canonical/k8s-snap-api v1.0.13 => github.com/claudiubelu/k8s-snap-api v0.0.0-20241121120954-ac9cc734c153
```
eaudetcobello, line 181: revert before merge

k8s/lib.sh
claudiubelu: Done

bschimke95:
Please fix the go CI

tests/integration/tests/test_cleanup.py
```diff
@@ -27,11 +29,64 @@ def test_node_cleanup(instances: List[harness.Instance], tmp_path):
...
+    for path in all_paths:
         assert f"cannot access '{path}': No such file or directory" in process.stderr
 
     util.setup_k8s_snap(instance, tmp_path)
     instance.exec(["k8s", "bootstrap"])
+
+
+@pytest.mark.node_count(2)
+@pytest.mark.disable_k8s_bootstrapping()
+def test_node_cleanup_new_containerd_path(instances: List[harness.Instance]):
```
bschimke95, line 45: Please add a test level tag. That is the reason why the CI fails.
I think it is sufficient to run this `nightly`.
```
```

bschimke95 (approved):
Generally, LGTM
Please add the test level tag, then we can merge it.

src/k8s/cmd/util/environ.go
```diff
@@ -32,10 +33,17 @@ func DefaultExecutionEnvironment() ExecutionEnvironment {
...
+			containerdBaseDir = strings.TrimSpace(string(data))
+		}
 		s = snap.NewSnap(snap.SnapOpts{
-			SnapDir:          os.Getenv("SNAP"),
-			SnapCommonDir:    os.Getenv("SNAP_COMMON"),
-			SnapInstanceName: os.Getenv("SNAP_INSTANCE_NAME"),
+			SnapDir:           os.Getenv("SNAP"),
+			SnapCommonDir:     os.Getenv("SNAP_COMMON"),
+			SnapInstanceName:  os.Getenv("SNAP_INSTANCE_NAME"),
+			ContainerdBaseDir: containerdBaseDir,
```
bschimke95, line 46: On a side note, I don't think we need the containerd path on the CLI side.

claudiubelu: True, but I'd say the ``DefaultExecutionEnvironment`` comment is not quite true, it's also being used by ``k8sd`` and ``k8s-apiserver-proxy``, and ``DefaultExecutionEnvironment`` is the only caller of ``snap.NewSnap`` anyways.

tests/integration/tests/test_cleanup.py
claudiubelu: Done

PR #818, opened 2024-11-20
Remove obsolete sync-images scripts

bschimke95:
We now use the our own Rocks everywhere, so syncing is not required anymore.


ktsakalozos (approved):
LGTM +1

PR #828, opened 2024-11-21
Add containerd warning to getting-started.md

eaudetcobello:


docs/src/snap/reference/troubleshooting.md
```diff
@@ -82,8 +82,31 @@ EOF
...
+paths. This means that a {{product}} installation will conflict with
+any existing system configuration where containerd is already installed.
+For example, if you have Docker installed, or another Kubernetes distribution
+that uses containerd.
+
+### Solution
+
+We recommend running {{product}} in an isolated environment, for this purpose,
+you can create a LXD container for your installation. See
+[Install {{product}} in LXD][lxd-install] for instructions.
```
nhennigan, line 105: Good idea to link to isolated env instead

docs/src/snap/tutorial/getting-started.md
```diff
@@ -19,6 +19,12 @@ Install the {{product}} snap with:
 sudo snap install k8s --edge --classic
 ```
 
+```{warning}
```
nhennigan, line 22: I think a warning here is a bit strong. I didn't take into account where the tutorial will fail is directly below the requirements section. I think this would be better suited in the "what you will need" section just above - along the lines of: A system without any previous installations of containerd/docker. Installing either with {{product}} may cause conflicts.  and then link to the LXD page as an alternative.

docs/src/snap/reference/troubleshooting.md
eaudetcobello: Wdym?

nhennigan: I just meant this is a better solution than asking the user to uninstall/clean up existing deployments

docs/src/snap/tutorial/getting-started.md
eaudetcobello: Done

nhennigan (approved):
Thanks for implementing those changes

PR #829, opened 2024-11-22
Merge latest changes into release-1.31

bschimke95:
# Notable changes

## Cilium `socketLB` and `apiserver-proxy` change

We've configured Cilium to talk to the localhost address instead of the `kube-proxy` provided service to address the issue we've faced with `socketLB`. We've introduced a new way to determine the localhost address with https://github.com/canonical/k8s-snap/pull/775 which provides a smooth upgrade path. No actions are needed.
LoadBalancer change from Cilium to MetalLB

We've changed the `load-balancer` implementation from Cilium to MetalLB. This will introduce down-time/interruption to these services while the upgrade is happening. It is necessary to follow the specific steps for the upgrade. Not following these steps properly will introduce nasty issues and conflicts!

## Feature components version upgrade

The versions for underlying feature implementations have been upgraded. 

## Containerd Default Path Change

Does not affect existing nodes, new nodes start using the new designated path. Older nodes might need the adjustment by hand to their args files.


bschimke95:
used wrong branch for source - closing

PR #835, opened 2024-11-22
fix: remove `--containerd` Kubelet cAdvisor arg.

aznashwan:
This patch stops passing the `--containerd` argument (value should be path to containerd socket) to `kubelet`.

This argument's backstory is a bit of a doozy but I'll try to lay it out:

- it's [originally a command line argument of cAdvisor](https://github.com/google/cadvisor/blob/3888dda254a8fa12f1c96be24ffdc33288ea9071/container/containerd/factory.go#L34) (container system resource usage monitor by Google)
- there was a time when Kubelet imported cAdvisor directly and thus had the argument transparently defined for it (along with a bunch of other imports transparently defining arguments)
- this was changed in [this Kubelet patch which made all external arguments cAdvisor explicit](https://github.com/kubernetes/kubernetes/pull/57613/files#diff-fecc7baa0b61f11d37dfc453aef25acca98866e1924f7765ba5592ca58fc3045R143), and instantly deprecated them
- [irrelevant] those deprecated flag definitions in kubelet were [moved to facilitate some crossbuilding here](https://github.com/kubernetes/kubernetes/pull/58095/files)

Thus, this setting does not _SEEM_ to be relevant for the k8s-snap.
The actual containerd socket is correctly passed as the `--container-runtime-endpoint` right above it FWIW.

I am unfortunately unsure why its presence is producing the `POST /k8sd/cluster: Post "http://control.socket/1.0/k8sd/cluster": EOF` error on `k8s bootstrap` we've been seeing, but I can vouch from my manual testing that removing it works around that particular crash


addyess:
thanks @aznashwan -- might need a bit more description here.  But i understand why we don't need this argument

eaudetcobello:
Is it *not needed*, or is it in use but deprecated?

bschimke95 (approved):
LGTM, thanks for the investigation.

PR #836, opened 2024-11-24
[Backport] fix: remove `--containerd` Kubelet cAdvisor arg. (#835)

bschimke95:
This patch removes the passing of the `--containerd` argument (value should be path to containerd socket) to `kubelet`.

This argument is actually not a kubelet argument per se, but instead it is a cAdvisor argument (1) which was accidentally transparently declared and shipped in kubelet a very long time ago, but had been explicitly declared within kubelet and marked deprecated since 2017 (2).

[1] https://github.com/google/cadvisor/blob/3888dda254a8fa12f1c96be24ffdc33288ea9071/container/containerd/factory.go#L34
[2] https://github.com/kubernetes/kubernetes/pull/57613/files#diff-fecc7baa0b61f11d37dfc453aef25acca98866e1924f7765ba5592ca58fc3045R143


berkayoz (approved):
LGTM

Issue #796, opened 2024-11-13
Proper Multus support


gruyaume:
#### Summary

At the moment, Multus support is quite awkward. As far as I know, the best approach to enable Multus in Canonical Kubernetes is to: 
1. Disable the Cilium CNI 
2. Install Multus with Helm (or another external tool)
3. Re-deploy Cilium with Helm with the "cni.exclusive" value set to false. 

#### Why is this important?

Multus is commonly used in Telco use cases.

In the context of the Telco engineering team, our User Plane Function (UPF) requires Multus. If we don't add this support, our product documentation will have to adapt from the previous Microk8s approach (microk8s enable community + microk8s enable multus) to this awkward approach. 

#### Are you interested in contributing to this feature?
It depends.

louiseschmidtgen:
Hello @gruyaume,

Unfortunately, Canonical Kubernetes does not support community add-ons similar to MicroK8s. You will need to adjust your product documentation and install Multus via helm or another tool.

There are docs on [how-to set up Multus + SRIOV](https://documentation.ubuntu.com/canonical-kubernetes/latest/src/snap/howto/epa/#multus-and-sriov-setup) as part of our EPA docs.

Any feedback on this set-up and our documentation is welcome!

gruyaume:
> Hello @gruyaume,
> 
> Unfortunately, Canonical Kubernetes does not support community add-ons similar to MicroK8s. You will need to adjust your product documentation and install Multus via helm or another tool.
> 
> There are docs on [how-to set up Multus + SRIOV](https://documentation.ubuntu.com/canonical-kubernetes/latest/src/snap/howto/epa/#multus-and-sriov-setup) as part of our EPA docs.
> 
> Any feedback on this set-up and our documentation is welcome!

Yeah I understand that community add-ons are a thing of the past, I'm not suggesting we take the same route as MicroK8s. I don't really mind about the approach honestly, it could be a new API, or documentation focused, but right now we have neither.

On the topic of the docs. First, they don't quite work. As I mentioned above, Cilium has to be installed with "cni.exclusive" set to false for Multus to work.  Second, they are hidden this EPA. I didn't want to "set up Enhanced Platform Awareness", all I wanted was Multus.

PR #838, opened 2024-11-25
Make cilium cni.exclusive configurable

petrutlucian94:
We'll add an annotation, making the cilium "cni.exclusive" setting configurable. This will allow using additional CNIs such as Multus.

Requires: https://github.com/canonical/k8s-snap-api/pull/20
Fixes: https://github.com/canonical/k8s-snap/issues/796


docs/src/snap/reference/annotations.md
```diff
@@ -24,6 +24,7 @@ the bootstrap configuration.
 | `k8sd/v1alpha1/calico/autodetection-v6/skipInterface` | Enable IP auto-detection based on interfaces that do not match the given regex. | string |
 | `k8sd/v1alpha1/calico/autodetection-v6/canReach` | Enable IP auto-detection based on which source address on the node is used to reach the specified IP or domain. | string |
 | `k8sd/v1alpha1/calico/autodetection-v6/cidrs` | Enable IP auto-detection based on which addresses on the nodes are within one of the provided CIDRs. | []string (comma separated) |
+| `k8sd/v1alpha1/cilium/cni-exclusive` | Make Cilium take ownership over the `/etc/cni/net.d` directory on the node, renaming all non-Cilium CNI configurations to `*.cilium_bak`. This ensures no Pods can be scheduled using other CNI plugins during Cilium agent downtime. Set this to "false" if you wish to use other CNIs such as Multus. | "true"\|"false" |
```
bschimke95, line 27: nit: this sounds like you need to set one of the values.
`false`/not set is the default. Please rephrase to something like
```
Set to "true" if you do not want to allow multiple CNIs.
```

src/k8s/pkg/k8sd/features/cilium/internal_test.go
```diff
@@ -119,7 +140,7 @@ func TestInternalConfig(t *testing.T) {
 		{
 			name:           "Nil annotations",
 			annotations:    nil,
-			expectedConfig: config{},
+			expectedConfig: config{cniExclusive: true},
```
bschimke95, line 143: why do we expect `cniExclusive` here? 
if everything is nil, then we should expect nil and use whatever cilium takes as default.

bschimke95 (commented):
Great stuff, added some questions/comments.

docs/src/snap/reference/annotations.md
petrutlucian94: Yes, the initial implementation expects one of these values. I'll update it.

bschimke95 (approved):
LGTM after CI is green

petrutlucian94:
```Error: pkg/k8sd/features/cilium/internal.go:75:48: undefined: apiv1_annotations.AnnotationCNIExclusive```

We need to merge the other PR first and then bump the requirement.

PR #839, opened 2024-11-25
Fix registry mirror log collection

petrutlucian94:
The registry mirror instance doesn't currently include the k8s snap, which is why the log collection script is missing.

This commit will install the k8s snap on the mirror instance and also capture the "inspect.sh" stderr output.


bschimke95 (approved):
LGTM

PR #840, opened 2024-11-25
Add docs for ports and services

maci3jka:
This PR adds to the documentation list of ports and services Canonical Kubernetes uses.
KU-1041

### DIFF
--- a/docs/src/snap/reference/ports-and-services.md
+++ b/docs/src/snap/reference/ports-and-services.md
@@ -1,0 +1,38 @@
+# Services and ports
+
+Services can be placed in two groups based on the network interface they bind to. Services binding to the localhost interface are only available from within the host. Services binding to the default host interface are available from outside the host and thus are subject to access restrictions.
+
+## Services binding to the default Host interface
+
+| Port  | Service         | Description/Access Restrictions                                                                          |
+|-------|-----------------|----------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


docs/src/snap/reference/ports-and-services.md
```diff
@@ -0,0 +1,38 @@
...
+|-------|-----------------|-------------------------------------------------------------------------|
+| 9234  | cilium-operator | cilium-operator  Address to serve API requests.                         |
+| 9879  | cilium-agent    | TCP port for the Cilium agent health status API.                        |
+| 9890  | cilium-agent    | cilium agent [gops](https://github.com/google/gops) server endpoint.    |     
+| 9891  | cilium-operator | cilium-operator [gops](https://github.com/google/gops) server endpoint. |
+| 10248 | kubelet         | Localhost health check endpoint.                                        |
+| 10249 | kube-proxy      | Port for the metrics server.                                            |
+| 10256 | kube-proxy      | Port for binding the health check server.                               |
+ 
+## Containerd
```
nhennigan, line 32: This feels a bit out of place because there is no mention of sockets up to this point. Can we include it in the opening paragraph? Also maybe change the title of this section to be something along the lines of "Services exposed through unix sockets".

nhennigan (approved):
Thanks for implementing those changes. I think it flows a lot better now.

PR #842, opened 2024-11-26
Add testing level tags

bschimke95:
Introduce tags to mark tests for specific testing levels.

In a follow-up PR, the existing tests will be marked and then we can simply run e.g.
```
tox run -e integration -- --tags up_to_nightly
```
to execute all nightly tests.

### DIFF
--- a/tests/integration/tests/test_util/tags.py
+++ b/tests/integration/tests/test_util/tags.py
@@ -1,0 +1,12 @@
+#
+# Copyright 2024 Canonical, Ltd.
+#
+from pytest_tagging import combine_tags
+
+PULL_REQUEST = "pull_request"
+NIGHTLY = "nightly"
+WEEKLY = "weekly"
+
+# Those tags can be used for a convenient way to run multiple test levels.
+combine_tags("up_to_nightly", PULL_REQUEST, NIGHTLY)
+combine_tags("up_to_weekly", PULL_REQUEST, NIGHTLY, WEEKLY)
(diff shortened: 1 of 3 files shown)


tests/integration/tests/test_util/tags.py
```diff
@@ -0,0 +1,12 @@
...
+# Copyright 2024 Canonical, Ltd.
+#
+from pytest_tagging import combine_tags
+
+PULL_REQUEST = "pull_request"
+NIGHTLY = "nightly"
+WEEKLY = "weekly"
+
+# Those tags can be used for a convenient way to run multiple test levels.
+combine_tags("up_to_nightly", PULL_REQUEST, NIGHTLY)
```
HomayoonAlimohammadi, line 11: Do we also need to have an `UP_TO_NIGHTLY` const? I'm not sure how it's gonna be used, but I imagine it will be _something like_:
```python
@with_tag(SOME_TAG)
def my_test(...):
    ...
```

HomayoonAlimohammadi (approved):
Great work! Thanks a lot @bschimke95! LGTM. Just a minor question.

tests/integration/tests/test_util/tags.py
bschimke95: This is only a convenience for invoking the tests via tox.
`tox run -e integration -- --tags up_to_nightly`
A test only gets assigned one of PULL_REQUEST, NIGHTLY, WEEKLY

PR #844, opened 2024-11-26
Add proxy setting from/etc/environment to inspection report

maci3jka:
Adding proxy settings from `etc/environment` to the inspection report will help solve deployment problems.

KU-2197

### DIFF
--- a/k8s/scripts/inspect.sh
+++ b/k8s/scripts/inspect.sh
@@ -47,11 +47,6 @@ function is_service_active {
   systemctl status "snap.$service" | grep -q "active (running)"
 }
 
-function collect_proxy_from_environment {
-  log_info "Grep proxy form environment file to the final report tarball"
-  grep -i "proxy" /etc/environment > "$INSPECT_DUMP/proxy_in_etc_environment"
-}
-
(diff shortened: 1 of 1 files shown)


k8s/scripts/inspect.sh
```diff
@@ -47,6 +47,11 @@ function is_service_active {
   systemctl status "snap.$service" | grep -q "active (running)"
 }
 
+function collect_environment {
+  log_info "Copy environment file to the final report tarball"
+  cp -r --no-preserve=mode,ownership /etc/environment "$INSPECT_DUMP"
+}
```
eaudetcobello, line 53: I would be careful collecting this. We know that API keys and other sensitive information can be stored in environment variables, and if the user uploads this report to Github, that's no good.

Can you grep for the specific lines we want and only export those instead?

eaudetcobello (changes requested):

k8s/scripts/inspect.sh
maci3jka: makes sense

eaudetcobello (approved):

PR #845, opened 2024-11-26
Change busybox image registry to avoid rate limit errors

petrutlucian94:
Our tests use busybox and nginx images from dockerhub, however the CI jobs often fail due to rate limit errors.

We'll use images from GHCR instead, where the CI jobs are authenticated.

While at it, we'll configure our registry mirror to cache images from rocks.canonical.com.


ktsakalozos (approved):
LGTM +1

PR #847, opened 2024-11-27
Remove harden runner from CI

louiseschmidtgen:
## Description
Removing git harden runner as part of our CI as it is a small startup and not critical.


bschimke95 (approved):
LGTM

PR #848, opened 2024-11-27
Fix cilium ingress, Refactor string literals

HomayoonAlimohammadi:
### Overview
- Fixed Cilium ingress chart options according to https://github.com/cilium/cilium/blob/f221719170636b0e0da2c7b8227c18967a1201c8/install/kubernetes/cilium/values.yaml#L765-L774:
    - `defaultTLSSecret` was replaced by `defaultSecretName`
- Refactored string literals and added constants

### Note to reviewer
The `DefaultTLSSecret` field in the `Ingress` type is not changed because Contour ingress is using that field as well:
- https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/features/contour/ingress.go#L130-L132
- https://github.com/canonical/k8s-snap/blob/main/k8s/manifests/charts/ck-ingress-tls/templates/tlscertificatedelegation.yaml#L8
- https://github.com/canonical/k8s-snap/blob/main/k8s/manifests/charts/ck-ingress-tls/values.yaml#L4

### DIFF
--- a/src/k8s/pkg/k8sd/features/cilium/ingress.go
+++ b/src/k8s/pkg/k8sd/features/cilium/ingress.go
@@ -12,6 +12,14 @@ import (
 const (
 	IngressDeleteFailedMsgTmpl = "Failed to delete Cilium Ingress, the error was: %v"
 	IngressDeployFailedMsgTmpl = "Failed to deploy Cilium Ingress, the error was: %v"
+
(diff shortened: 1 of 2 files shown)


bschimke95 (approved):
LGTM, great stuff @HomayoonAlimohammadi

PR #850, opened 2024-11-27
Use test-level tags in CI

bschimke95:
# Summary

Adds a test level to each integration test and use them in the CI

# Rationale
The integration tests for each PR take way too long and test a lot of things that are very unlikely to break. It is sufficient to run those in the nightly test runs or even weekly (e.g. for conformance tests).

# Changes
* Assigns a test level to each integration test 
* ensures that the integration tests fail if a tag is missing for a test.
* minor refactoring to ensure pull_request level tests and nightly tests are not in the same file
* Use tags in the CI

Note that there is currently no weekly CI but I already added the tag for it. This is work for later.

### DIFF
--- a/.github/workflows/nightly-test.yaml
+++ b/.github/workflows/nightly-test.yaml
@@ -51,7 +51,7 @@ jobs:
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
         run: |
           export PATH="/home/runner/.local/bin:$PATH"
-          cd tests/integration && sg lxd -c 'tox -vve integration'
+          cd tests/integration && sg lxd -c 'tox -vve integration -- --tags up_to_nightly '
       - name: Prepare inspection reports
         if: failure()
         run: |
--- a/tests/integration/tests/test_annotations.py
+++ b/tests/integration/tests/test_annotations.py
@@ -80,14 +80,13 @@ def test_skip_services_stop_on_remove(instances: List[harness.Instance]):
     services = joining_cp.exec(
         ["snap", "services", "k8s"], capture_output=True, text=True
     ).stdout.split("\n")[1:-1]
-    print(services)
     for service in services:
         if "k8s-apiserver-proxy" in service:
             assert (
                 " inactive " in service
             ), "apiserver proxy should be inactive on control-plane"
         else:
-            assert " active " in service, "service should be active"
+            assert " active " in service, f"'{service}' should be active"
 
     cluster_node.exec(["k8s", "remove-node", worker.id])
     nodes = util.ready_nodes(cluster_node)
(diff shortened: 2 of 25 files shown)


.github/workflows/nightly-test.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
         run: |
           export PATH="/home/runner/.local/bin:$PATH"
-          cd tests/integration && sg lxd -c 'tox -vve integration'
+          cd tests/integration && sg lxd -c 'tox -vve integration -- --tags up_to_nightly '
```
eaudetcobello, line 54: I don't see up_to_nightly defined anywhere?

eaudetcobello (approved):
Very useful change, thanks!

berkayoz (approved):
LGTM.

tests/integration/tests/test_annotations.py
```diff
@@ -0,0 +1,110 @@
...
+    # TODO: skip retrying this once the microcluster trust store issue is addressed.
+    util.stubbornly(retries=3, delay_s=5).on(cluster_node).exec(
+        ["k8s", "remove-node", joining_cp.id]
+    )
+    nodes = util.ready_nodes(cluster_node)
+    assert len(nodes) == 2, "cp node should have been removed from the cluster"
+    services = joining_cp.exec(
+        ["snap", "services", "k8s"], capture_output=True, text=True
+    ).stdout.split("\n")[1:-1]
+    print(services)
```
HomayoonAlimohammadi, line 83: Maybe we can remove this?

tests/integration/tests/test_annotations.py
```diff
@@ -0,0 +1,110 @@
...
+    for service in services:
+        if "k8s-apiserver-proxy" in service:
+            assert (
+                " inactive " in service
+            ), "apiserver proxy should be inactive on control-plane"
+        else:
+            assert " active " in service, "service should be active"
```
HomayoonAlimohammadi, line 90: Maybe we can include the name of the service for easier debugging here:
```suggestion
            assert " active " in service, f"'{service}' should be active"
```

HomayoonAlimohammadi (approved):
Great work, thanks a lot @bschimke95! LGTM. Just minor comments and questions.
I'm not sure I fully understand which tag should be applied to which each test.
AFAIU, we mark the core/critical functionalities with `PULL_REQUEST`, less critical ones with `NIGHTLY` and even less critical ones with `WEEKLY`. Overall I think this is definitely the right direction and we can tune those tags as needed in the future.

tests/integration/tests/test_version_upgrades.py
```diff
@@ -15,6 +15,7 @@
 @pytest.mark.skipif(
     not config.VERSION_UPGRADE_CHANNELS, reason="No upgrade channels configured"
 )
+@pytest.mark.tags(tags.NIGHTLY)
```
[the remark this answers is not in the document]
bschimke95: Rather expensive. We test up to 6 releases (once we have them) and install, bootstrap/restart, wait for everything to settle for every release. 
Let's keep it as a nightly run. We have those tests as part of the promotion CI. Hence, there is no way that we break the upgrades in a channel.

PR #853, opened 2024-11-27
Add download links to long files

nhennigan:
Added download links to longer files. I also replaced those files with an {literalinclude} so that they still appear in the docs the same way, but we only need to update one location in the future.


docs/src/assets/configuration.yaml
```diff
@@ -0,0 +1,11 @@
...
+     network:
+       enabled: true
+     dns:
+       enabled: true
+     local-storage:
+       enabled: true
+   extra-node-kubelet-args:
+     --reserved-cpus: "0-31"
+     --cpu-manager-policy: "static"
+     --topology-manager-policy: "best-effort"
```
bschimke95, line 11: nit: please add a newline - otherwise GH complains with this red circle

bschimke95 (approved):
LGTM, only a minor nit

docs/src/assets/configuration.yaml
nhennigan: Fixed thanks for pointing it out

bschimke95:
thanks, feel free to merge if everything is addressed.

PR #855, opened 2024-11-28
How-to Openstack Integration

louiseschmidtgen:
## How-to Openstack Integration

This PR adds docs: how to integrate Canonical K8s with the Openstack cloud it is deployed on, by adding a layer.

### DIFF
--- a/docs/src/charm/howto/index.md
+++ b/docs/src/charm/howto/index.md
@@ -16,6 +16,7 @@ Overview <self>
 
 charm
 install-lxd
+openstack
 Integrate with etcd <etcd>
 Integrate with ceph-csi <ceph-csi>
 proxy
--- a/docs/src/charm/howto/openstack.md
+++ b/docs/src/charm/howto/openstack.md
@@ -67,7 +67,7 @@ A modified overlay template can look like this:
 ```yaml
 applications:
   kubeapi-load-balancer: null
-  kubernetes-control-plane:
+  k8s:
     options:
       allow-privileged: "true"
   openstack-integrator:
@@ -85,8 +85,8 @@ applications:
       - 0 
 relations:
   - [openstack-cloud-controller:certificates,            easyrsa:client]
-  - [openstack-cloud-controller:kube-control,            kubernetes-control-plane:kube-control]
-  - [openstack-cloud-controller:external-cloud-provider, kubernetes-control-plane:external-cloud-provider]
+  - [openstack-cloud-controller:kube-control,            k8s:kube-control]
+  - [openstack-cloud-controller:external-cloud-provider, k8s:external-cloud-provider]
   - [openstack-cloud-controller:openstack,               openstack-integrator:clients]
   - [easyrsa:client,                                     cinder-csi:certificates]
   - [k8s:kube-control,              cinder-csi:kube-control]
--- a/docs/src/charm/howto/openstack.md
+++ b/docs/src/charm/howto/openstack.md
@@ -1,120 +1,79 @@
 # Integrating with OpenStack
 
 This guide explains how to integrate {{product}} with the OpenStack cloud
-platform.
+platform. The `openstack-integrator` charm simplifies working with {{product}}
+on OpenStack. Using the credentials provided to Juju, it acts as a proxy between
+{{product}} and the underlying cloud, granting permissions to dynamically
+create, for example, Cinder volumes.
 
 ## Prerequisites
 
 To follow this guide, you will need:
 
 - An [OpenStack][openstack] cloud environment.
-- Juju set-up as the deployment tool for Openstack. Please refer to
-  [Juju and Openstack][juju-openstack] documentation for more information.
-- A Juju [controller][controller] with [access][credentials] to the OpenStack 
-  cloud environment.
-- A Juju [model][model] for deploying {{product}} on OpenStack.
+- The installation requires Octavia to be available both to support Kubernetes
+  LoadBalancer services and to support the creation of a load balancer for the
+  Kubernetes API.
 - A valid [proxy configuration][proxy] in constrained environments.
 
-## Deploying {{product}} on OpenStack
-
+## Installing {{product}} on OpenStack
 
 To deploy the {{product}} [bundle][bundle] on OpenStack you need an overlay
-bundle which serves as an extension of the core bundle. The overlay bundle
-contains the necessary configuration to deploy {{product}} on OpenStack.
-Applications are deployed through the overlay and relations are established
-between the applications. These include the openstack integrator, cloud
+bundle which serves as an extension of the core bundle. Through the overlay,
+applications are deployed and relations are established between the
+applications. These include the openstack integrator, cloud
 controller, and cinder-csi charm.
+
+### OpenStack Overlay Configurations:
 
 Refer to the base overlay [openstack-overlay.yaml][openstack-overlay] and
 modify it as needed.
 
-### OpenStack Overlay Configurations:
-
-Run `openstack project list` to retrieve your project id and include the
-[project id][project] in the overlay template:
-
 ```yaml
 applications:
-  openstack-integrator:
-    options:
-      project-id: <my-openstack-project-id>
-```
-
-
-Adjust [easyrsa][easyrsa] to avoid creating an LXD machine:
-
-```yaml
-applications:
-  easyrsa:
-    to:
-      - 0
-```
-
-If your set-up includes a load-balancer add the following:
-
-```yaml
-relations:
-  - [k8s:loadbalancer-external, openstack-integrator:lb-consumers]
-```
-
-```yaml
-applications:
-  kubeapi-load-balancer: null 
-```
-
-A modified overlay template can look like this:
-
-```yaml
-applications:
-  kubeapi-load-balancer: null
-  k8s:
-    options:
-      allow-privileged: "true"
   openstack-integrator:
     charm: openstack-integrator
     num_units: 1
     trust: true
-    options:
-      project-id: <my-openstack-project-id>
+    base: ubuntu@22.04
   openstack-cloud-controller:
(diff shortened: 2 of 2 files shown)


docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,104 @@
...
+
+```yaml
+applications:
+  openstack-integrator:
+    options:
+      project-id: <my-openstack-project-id>
+```
+
+
+Adjust [easyrsa][easyrsa] to avoid creating an LXD machine:
```
louiseschmidtgen, line 43: @HomayoonAlimohammadi Is this necessary?

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,104 @@
...
+To follow this guide, you will need:
+
+- An [OpenStack][openstack] cloud environment.
+- Juju set-up as the deployment tool for Openstack. Please refer to
+  [Juju and Openstack][juju-openstack] documentation for more information.
+- An OpenStack [project][project] with the necessary permissions to deploy
+  {{product}}.
+- A juju [controller][controller] with [access][credentials] to the OpenStack 
+  cloud environment.
+- A juju [model][model] for deploying {{product}} on OpenStack.
```
eaudetcobello, line 17:
```suggestion
- A Juju [model][model] for deploying {{product}} on OpenStack.
```

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,104 @@
...
+- Configured [proxy configuration][proxy] in constrained environments.
```
eaudetcobello, line 18:
```suggestion
- A valid [proxy configuration][proxy] in constrained environments.
```

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Deploying {{product}} on OpenStack
+
+
+To deploy the {{product}} [bundle][bundle] on OpenStack you need an overlay
+bundle which serves as an extension of the core bundle. The overlay bundle
+contains the necessary configuration to deploy {{product}} on OpenStack.
```
eaudetcobello, line 25: Can you add a sentence or two about specific things it does?

docs/src/charm/howto/openstack.md
HomayoonAlimohammadi: I don't think so. Maybe @addyess has a better idea but I think it's not a very special thing. I mean the problem in our envs is that we can not create LXD machines, maybe that's not the case for the user. And what if something else is going wrong with the Openstack deployment of the user? e.g. What if - let's say - Keystone is not configured correctly? Also seems like [Openstack on LXD is supported](https://docs.openstack.org/charm-guide/victoria/openstack-on-lxd.html) (tho it's only for testing/development purposes).

HomayoonAlimohammadi: And even if _not using LXD_ is the way to go when deploying on Openstack, I think we should change the [overlay](https://github.com/charmed-kubernetes/bundle/blob/main/overlays/openstack-lb-overlay.yaml) rather than instructing the user to the change the overlay manually.

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,104 @@
...
+platform.
+
+## Prerequisites
+
+To follow this guide, you will need:
+
+- An [OpenStack][openstack] cloud environment.
+- Juju set-up as the deployment tool for Openstack. Please refer to
+  [Juju and Openstack][juju-openstack] documentation for more information.
+- An OpenStack [project][project] with the necessary permissions to deploy
```
HomayoonAlimohammadi, line 13: nit: We already mentioned the we need an Openstack environment in the first point. Maybe we can remove one of these two.

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,104 @@
...
+  openstack-cloud-controller:
+    charm: openstack-cloud-controller
+  cinder-csi:
+    charm: cinder-csi
+  easyrsa: 
+    to:   
+      - 0 
+relations:
+  - [openstack-cloud-controller:certificates,            easyrsa:client]
+  - [openstack-cloud-controller:kube-control,            kubernetes-control-plane:kube-control]
```
HomayoonAlimohammadi, line 75: I think we should change these. For Canonical K8s, it's `k8s` and `k8s-worker`.
Also I'm now thinking that we might need to include an Openstack overlay (or overlay template) somewhere in the `k8s-bundles` repo which is sort of customized for our Canonical K8s charm. 
Overall, I think this overlay needs significant changes to work with Canonical K8s (in the K8s Openstack docs I only tested against `kubernetes-core` bundle).

HomayoonAlimohammadi (commented):
Great work, thanks a lot @louiseschmidtgen! Left some comments.

docs/src/charm/howto/openstack.md
louiseschmidtgen: Agreed we need the k8s template in the k8s-bundles repo and tested before these docs can get published.

louiseschmidtgen: @HomayoonAlimohammadi Are you working on verifying the Canonical k8s overlay template already?

docs/src/charm/howto/index.md
```diff
@@ -16,6 +16,7 @@ Overview <self>
 
 charm
 install-lxd
+openstack
```
addyess, line 19: I imagine at some point, we're going to need to make an "Integrate with Clouds", and detail other clouds like aws, azure, gcp, maas, etc.  Even lxd could be counted as one of these clouds (though there isn't much to integrate with)

addyess, line 19: I imagine at some point, we're going to need to make an "Integrate with Clouds", and detail other clouds like aws, azure, gcp, maas, etc.  Even lxd could be counted as one of these clouds (though there isn't much to integrate with)

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,120 @@
...
+This guide explains how to integrate {{product}} with the OpenStack cloud
+platform.
+
+## Prerequisites
+
+To follow this guide, you will need:
+
+- An [OpenStack][openstack] cloud environment.
+- Juju set-up as the deployment tool for Openstack. Please refer to
+  [Juju and Openstack][juju-openstack] documentation for more information.
```
addyess, line 12: Ehh --- this isn't REALLY necessary.  You just need openstack credentials  -- openstack doesn't have to be deployed BY juju

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,120 @@
...
+- A Juju [controller][controller] with [access][credentials] to the OpenStack 
+  cloud environment.
```
addyess, line 14: again -- this isn't super necessary.  If you have credentials to openstack, you can TOTALLY bootstrap a VM in openstack to use as your juju controller.  

Does any of this from charmed-kubernetes help?
https://ubuntu.com/kubernetes/docs/openstack-integration

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,120 @@
...
+- A Juju [model][model] for deploying {{product}} on OpenStack.
```
addyess, line 15: Once you have bootstrapped a controller on openstack -- it's easy to make a new model for your k8s cluster

docs/src/charm/howto/openstack.md
addyess: Yeah, i don't think this is critical.  I don't know if we ever got to a solution why the openstack instance we're using for testing requires easyrsa to be its own machine.  I think we shouldn't note "avoiding creating this LXD machine"

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,120 @@
...
+
+```yaml
+relations:
+  - [k8s:loadbalancer-external, openstack-integrator:lb-consumers]
+```
+
+```yaml
+applications:
+  kubeapi-load-balancer: null 
+```
```
addyess, line 63: what's this?  I don't think our bundle call out the kubeapi-load-balancer does it?  This is used by charmed-kubernetes V1 -- it shouldn't be referenced here i don't think

addyess (commented):
Alrighty so this has alot of detail about how the bundle works.  I appreciate that.  

Can we use more of the docs from charmed-kubernetes
https://ubuntu.com/kubernetes/docs/openstack-integration

i think it provides a better user experience from the day-2 ops rather than how to get k8s up on openstack

docs/src/charm/howto/openstack.md
HomayoonAlimohammadi: No I will create a card for that, but we don't have that for now.

docs/src/charm/howto/openstack.md
```diff
@@ -0,0 +1,120 @@
...
+
+This guide explains how to integrate {{product}} with the OpenStack cloud
+platform.
+
+## Prerequisites
+
+To follow this guide, you will need:
+
+- An [OpenStack][openstack] cloud environment.
+- Juju set-up as the deployment tool for Openstack. Please refer to
```
evilnick, line 11:
```suggestion
- Juju set-up as the deployment tool for OpenStack. Please refer to
```
:stuck_out_tongue:

addyess: Look at this guy... epic!  Next thing you know, he'll want us spelling tires with a y

docs/src/charm/howto/index.md
louiseschmidtgen: Great point Adam!

HomayoonAlimohammadi (approved):
LGTM, thanks a lot @louiseschmidtgen!

PR #837, opened 2024-11-25
v1.32-rc1

bschimke95:
# Summary

Enable and fix 1.32-rc release

# Rationale

Upstream Kubernetes 1.32 will be released next week. This PR tests if the current snap is compatible with the current release candidate. 

Some changes in the snap are required:
* We noticed some test failures that happen because of an incompatibility of `k8s-dqlite` and the [consistent read from cache](https://kubernetes.io/blog/2024/08/15/consistent-read-from-cache-beta/). By default, this feature is enabled on 1.32. 
For now, we'll disable this feature and do a follow-up investigation what it takes to support that feature in k8s-dqlite.
* updated the `k8s.io` library versions
* update Go version to 1.23.0 to build with the same version as upstream.

PR #857, opened 2024-11-28
Add patches for Kubernetes 1.32

bschimke95:
# Summary

Adapt the patches to the 1.32 release

# Rationale

We slightly modify the upstream Kubernetes for various reasons. Those patches are applied on build and can be found in `build-scripts/components/kubernetes/patches/<release>`.
The patches might - as in this case - need to be adapted to fit the refactored code in new releases hence we need a folder for each release.

# Testing

See https://github.com/canonical/k8s-snap/pull/837 for a run with the latest 1.32-rc


eaudetcobello (approved):
Diffed the old 0001-resolvconf patch and the new patch in an external tool, LGTM

HomayoonAlimohammadi (approved):
Thanks a lot, great work @bschimke95! LGTM. Applied the patches manually to Kubernetes `v1.32.0-rc.0` and everything seems correct.

PR #860, opened 2024-12-02
[Docs] replace install command with literalinclude

nhennigan:
when updating the docs install commands with each release, now we only have to update 5 files rather than 15
- docs/canonicalk8s/src/snap/reference/releases.md
- docs/src/_parts/install.md
- README.md
- docs/canonicalk8s/reuse/substitutions.yaml
- docs/src/assets/how-to-epa-maas-cloud-init

Some linting errors also addressed


eaudetcobello (approved):
LGTM

PR #861, opened 2024-12-03
Fix microk8s snap check

petrutlucian94:
During bootstrap, we're checking if microk8s is installed. If so, we're throwing an error, asking the user to uninstall microk8s first.

The problem is that this check fails when running the CAPI tests under Docker.

```
  k8s bootstrap --name '' --address :2380 --file /capi/etc/config.yaml
  Error: failed to check if microk8s is installed: %w failed to get
  snapd snap info: Get "http://localhost/v2/snaps/microk8s": dial unix
  /run/snapd.socket: connect: no such file or directory
  : exited with status: 1,
```

We'll change the code to log a warning instead of throwing an error when it's unable to check the microk8s snap status.


eaudetcobello (approved):
Nice find!

PR #862, opened 2024-12-03
Testing 1.32 with etcd

bschimke95:


bschimke95:
closing as this was only created to check if we see the same propagation failures in etcd as we do with `k8s-dqlite`. We don't.

PR #863, opened 2024-12-03
fix: ensure containerd-related directories removed on failed `bootstrap/join-cluster`.

aznashwan:
### fix: ensure containerd-related directories removed on failed `bootstrap/join-cluster`

`k8sd` automatically sets up some directories with the appropriate
ownership/permissions to be used by containerd in the early stages
of the `bootstrap` and `join-cluster` commands.

In the classic (non-strict) version of the k8s-snap, these
containerd directories are system-wide (e.g. `/etc/containerd`,
`/run/containerd`, etc).

Should any of the other setup steps fail after the containerd
directories were set up, the directories would still remain on
disk and thus lead to a 'partial installation' of on the host system.

This patch ensures that `k8s` will automatically remove any
containerd-related directories which were created in the event of
the  `bootstrap` / `join-cluster` commands failing.

### fix: ensure containerd Base Dir lockfile is never accidentally deleted.

The containerd Base Dir is the special path all other containerd-related
paths on the snap are derived from.

Under classic confinement and default settings, this path defaults
to the host's root (`/`), and thus extreme care must be taken to
not accidentally include it in k8sd's cleanup routine or the
k8s-snap's remove hook.

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -13,6 +13,7 @@ import (
 	"github.com/canonical/k8s/pkg/k8sd/pki"
 	"github.com/canonical/k8s/pkg/k8sd/setup"
 	"github.com/canonical/k8s/pkg/log"
+	"github.com/canonical/k8s/pkg/snap"
 	snaputil "github.com/canonical/k8s/pkg/snap/util"
 	"github.com/canonical/k8s/pkg/utils/control"
 	"github.com/canonical/microcluster/v2/cluster"
@@ -145,5 +146,48 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 		}
 	}
 
+	tryCleanupContainerdPaths(log, snap)
+
 	return nil
+}
+
+// tryCleanupContainerdPaths attempts to clean up all containerd directories which were
+// created by the k8s-snap based on the existence of their respective lockfiles
+// located in the directory returned by `s.LockFilesDir()`.
+func tryCleanupContainerdPaths(log log.Logger, s snap.Snap) {
+	for lockpath, dirpath := range setup.ContainerdLockPathsForSnap(s) {
+		// Ensure lockfile exists:
+		if _, err := os.Stat(lockpath); os.IsNotExist(err) {
+			log.Info("WARN: failed to find containerd lockfile, no cleanup will be perfomed", "lockfile", lockpath, "directory", dirpath)
+			continue
+		}
+
+		// Ensure lockfile's contents is the one we expect:
+		lockfile_contents := ""
+		if contents, err := os.ReadFile(lockpath); err != nil {
+			log.Info("WARN: failed to read contents of lockfile", "lockfile", lockpath, "error", err)
+			continue
+		} else {
+			lockfile_contents = string(contents)
+		}
+
+		if lockfile_contents != dirpath {
+			log.Info("WARN: lockfile points to different path than expected", "lockfile", lockpath, "expected", dirpath, "actual", lockfile_contents)
+			continue
+		}
+
+		// Check directory exists before attempting to remove:
+		if _, err := os.Stat(dirpath); os.IsNotExist(err) {
+			log.Info("Containerd directory doesn't exist; skipping cleanup", "directory", dirpath)
+		} else {
+			if err := os.RemoveAll(dirpath); err != nil {
+				log.Info("WARN: failed to remove containerd data directory", "directory", dirpath, "error", err)
+				continue // Avoid removing the lockfile path.
+			}
+		}
+
+		if err := os.Remove(lockpath); err != nil {
+			log.Info("WARN: Failed to remove containerd lockfile", "lockfile", lockpath)
+		}
+	}
 }
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -180,8 +180,23 @@ func tryCleanupContainerdPaths(log log.Logger, s snap.Snap) {
 		if _, err := os.Stat(dirpath); os.IsNotExist(err) {
 			log.Info("Containerd directory doesn't exist; skipping cleanup", "directory", dirpath)
 		} else {
+			// NOTE(aznashwan): because of the convoluted interfaces-based way the snap
+			// composes and creates the original lockfiles (see k8sd/setup/containerd.go)
+			// this check is meant to defend against accidental code/configuration errors which
+			// might lead to the root FS being deleted:
+			realPath, err := os.Readlink(dirpath)
+			if err != nil {
+				log.Error(err, fmt.Sprintf("Failed to os.Readlink the directory path for lockfile %q pointing to %q. Skipping cleanup", lockpath, dirpath))
+				continue
+			}
+
+			if realPath == "/" {
+				log.Error(fmt.Errorf("There is some configuration/logic error in the current versions of the k8s-snap related to lockfile %q (meant to lock %q, which points to %q) which could lead to accidental wiping of the root file system.", lockpath, dirpath, realPath), "Please report this issue upstream immediately.")
+				continue
+			}
+
 			if err := os.RemoveAll(dirpath); err != nil {
-				log.Info("WARN: failed to remove containerd data directory", "directory", dirpath, "error", err)
+				log.Info("WARN: failed to remove containerd data directory", "directory", dirpath, "error", err, "realPath", realPath)
 				continue // Avoid removing the lockfile path.
 			}
 		}
--- a/src/k8s/pkg/k8sd/setup/containerd.go
+++ b/src/k8s/pkg/k8sd/setup/containerd.go
@@ -169,13 +169,19 @@ func Containerd(snap snap.Snap, extraContainerdConfig map[string]any, extraArgs 
 // ContainerdLockPathsForSnap returns a mapping between the absolute paths of
 // the lockfiles within the k8s snap and the absolute paths of the containerd
 // directory they lock.
+//
+// WARN: these lockfiles are meant to be used in later cleanup stages.
+// DO NOT include any system paths which are not managed by the k8s-snap!
+//
+// It intentionally does NOT include the containerd base dir lockfile
+// (which most of the rest of the paths are based on), as it is meant
+// to indicate the root of the containerd install ('/' or '/var/snap/k8s/*').
 func ContainerdLockPathsForSnap(s snap.Snap) map[string]string {
 	m := map[string]string{
 		"containerd-socket-path": s.ContainerdSocketDir(),
 		"containerd-config-dir":  s.ContainerdConfigDir(),
 		"containerd-root-dir":    s.ContainerdRootDir(),
 		"containerd-cni-bin-dir": s.CNIBinDir(),
-		snap.ContainerdBaseDir:   s.GetContainerdBaseDir(),
 	}
 
 	prefixed := map[string]string{}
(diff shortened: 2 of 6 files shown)


src/k8s/pkg/k8sd/setup/containerd.go
```diff
@@ -166,23 +167,79 @@ func Containerd(snap snap.Snap, extraContainerdConfig map[string]any, extraArgs
...
+	log := log.L()
+
+	for lockpath, dirpath := range containerdLockPathsForSnap(s) {
+		// Ensure lockfile exists:
+		if _, err := os.Stat(lockpath); os.IsNotExist(err) {
+			log.Info("WARN: failed to find containerd lockfile, no cleanup will be perfomed", "lockfile", lockpath, "directory", dirpath)
+			continue
+		}
+
+		// Ensure lockfile's contents is the one we expect:
```
berkayoz, line 211: What does this check ensure for us? Is it to check someone else did not create the locks?

src/k8s/pkg/k8sd/setup/containerd.go
```diff
@@ -166,23 +167,79 @@ func Containerd(snap snap.Snap, extraContainerdConfig map[string]any, extraArgs
...
+			continue
+		}
+
+		// Check directory exists before attempting to remove:
+		if _, err := os.Stat(dirpath); os.IsNotExist(err) {
+			log.Info("Containerd directory doesn't exist; skipping cleanup", "directory", dirpath)
+		} else {
+			if err := os.RemoveAll(dirpath); err != nil {
+				log.Info("WARN: failed to remove containerd data directory", "directory", dirpath, "error", err)
+				continue // Avoid removing the lockfile path.
```
berkayoz, line 231: Isn't cleanup done on a best effort basis? Do we check the lockfile later on at bootstrap time?

tests/integration/tests/test_cleanup.py
```diff
@@ -17,21 +19,71 @@
 ]
 
 
+def _assert_paths_not_exist(instance: harness.Instance, paths: List[str]):
```
berkayoz, line 22: Feels like it would fit better in `test_util.util`

berkayoz (commented):
LGTM overall, we should probably wait and adjust for #817

src/k8s/pkg/k8sd/setup/containerd.go
aznashwan: I admit I have no concrete usecase where this check would ever be relevant, it just "felt right" to double-check in case someone changed their contents from what the k8s-snap expects.

Note that, as far as I can tell, **the k8s-snap does NOT use the paths written within the lockfiles** (considering they're hardcoded/derived within the k8s-snap itself, it has no need for them).

The only place I've seen the [actual file contents read](https://github.com/canonical/k8s-snap/blob/a00846526c10cca25cfe54d2ad1685e558e7251a/k8s/lib.sh#L104) is in the [`remove` hook of the snap](https://github.com/canonical/k8s-snap/blob/17491b44a5b747e456253254e7d6d89ff350a9c4/snap/hooks/remove#L11-L12).

I could gladly remove it if it's "too redundant" or switch the k8s-snap to always clean up the paths which are written within the file, instead of what the `k8s-snap` expects that file lock to be for...

src/k8s/pkg/k8sd/setup/containerd.go
aznashwan: It is, but IMO it would have been "incorrect" for this code to remove the lockfile path if it had failed to remove the actual path it was pointing too. (IMO either both should be deleted; or none)

> Do we check the lockfile later on at bootstrap time?

AFAIK the `LockFilesDir()` is used only by `k8sd` in this file when setting up the containerd dirs, and [to mark a a node as worker](https://github.com/canonical/k8s-snap/blob/a00846526c10cca25cfe54d2ad1685e558e7251a/src/k8s/pkg/snap/util/node.go#L13) (contents doesn't matter for marking workers, only the file existing does).

These contents of the [lockfiles are only read by the `remove` hook of the k8s-snap](https://github.com/canonical/k8s-snap/blob/a00846526c10cca25cfe54d2ad1685e558e7251a/k8s/lib.sh#L104) (it also neglects to delete the lockfiles after it's done though, unsure whether that's intentional or not...).

tests/integration/tests/test_cleanup.py
aznashwan: Great point!

I've made the function a bit more generic (returns `bool` for each given path) and moved it to `test_util`.

aznashwan:
@berkayoz note that it just hit me that because of the new `pytest.mark.tags(tags.NIGHTLY)` test tagging system the integration test doesn't get executed as part of the PR, sorry 🤦 

While I believe it's better to leave that test for NIGHTLY, I've manually run the test on a 22.04 VM and I have.

I needed to add [this small fix (I had forgotten to actually call `socket.listen()` in the contextmanager)](https://github.com/canonical/k8s-snap/pull/863/commits/83763de8ba79f0ff8fec66fde1df9b2e438596b1#diff-765a6c110620b1ff5a80a7f549d5d2138f096a5fc6cd2e90cc7ef7d4337574c8R545).

Can vouch that the added integration test seems to work when run manually now **BUT it does sometimes exhibit the `EOF` error** on my machine (which I am still positive is unrelated and hope Claudiu will push a fix for soon).

src/k8s/pkg/k8sd/setup/containerd.go
```diff
@@ -166,23 +167,79 @@ func Containerd(snap snap.Snap, extraContainerdConfig map[string]any, extraArgs
...
+			return fmt.Errorf("failed to write %s: %w", lockpath, err)
 		}
 	}
 	return nil
 }
 
+// Attempts to clean up all containerd directories which were created
+// by the k8s-snap based on the existence of their respective lockfiles
+// located in the directory returned by `s.LockFilesDir()`.
+func TryCleanupContainerdPaths(s snap.Snap) error {
```
bschimke95, line 201: A couple of comments for this function:
* The signature says it returns an error but this is never exercised
* This function is very closely tied to our remove logic (best effort base etc.) -> IMHO it fits better somewhere close to the remove hook rather then in `setup`

bschimke95 (commented):
thanks @aznashwan - a couple of comments.

src/k8s/pkg/k8sd/setup/containerd.go
```diff
@@ -166,23 +167,79 @@ func Containerd(snap snap.Snap, extraContainerdConfig map[string]any, extraArgs
...
+	prefixed := map[string]string{}
+	for k, v := range m {
+		prefixed[filepath.Join(s.LockFilesDir(), k)] = v
+	}
+
+	return prefixed
+}
+
+// Creates the lock files for the containerd directory paths to be used for later cleanup.
+func saveSnapContainerdPaths(s snap.Snap) error {
```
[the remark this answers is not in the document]
aznashwan: Done.

src/k8s/pkg/k8sd/setup/containerd.go
```diff
@@ -166,23 +167,79 @@ func Containerd(snap snap.Snap, extraContainerdConfig map[string]any, extraArgs
 	return nil
 }
 
-func saveSnapContainerdPaths(s snap.Snap) error {
-	// Write the containerd-related paths to files to properly clean-up on removal.
+// Returns a mapping between the absolute paths of the lockfiles within the
+// k8s snap and the absolute paths of the containerd directory they lock.
+func containerdLockPathsForSnap(s snap.Snap) map[string]string {
```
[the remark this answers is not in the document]
aznashwan: The link's for this same PR 😆 

I've done a superficial parsing of each of the recently-merged PRs and it's not clear to me which could conflict with this one?

Please advise.

src/k8s/pkg/k8sd/setup/containerd.go
aznashwan: Done.

src/k8s/pkg/k8sd/setup/containerd.go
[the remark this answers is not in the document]
bschimke95: Ah damnit, sorry.
I meant Claudius PR about the containerd base dir changes but after a second look I think we should be fine.

src/k8s/pkg/k8sd/setup/containerd.go
berkayoz: I agree that removing the lockfile on failure, my thought/question is how should we communicate this to the user since these are just logs. Should we have a check in the bootstrap process so we can point out potentially left over paths? @bschimke95

aznashwan:
@bschimke95 @berkayoz related to [this latest defensive commit I had to add to the cleanup](https://github.com/canonical/k8s-snap/pull/863/commits/57f5b04b119d5f79c684ff35fe868c978b20dbd0)

@claudiubelu literally accidentally deleted his root dir using this cleanup code after I re-based my patch over [his recently-merged configurable containerd paths one](https://github.com/canonical/k8s-snap/pull/817) (thank you Claudiu, this was the definition of taking one for the team!)

I think it's really bad that we have our cleanup logic split between the snap's remove hook (on `snap remove`) and in `k8sd` (on failed `bootstrap`/`join-cluster`), and we'll always be exposing ourselves to us/users messing up the lockfiles.

I would strongly suggest we move ALL cleanup logic to a dedicated routine which we re-use internally and expose via a dedicated `k8s cleanup` command which should be best-effort; and directly called in the snap's remove hook.

berkayoz (approved):
LGTM, agreed on turning this into a `cleanup` command we can call on both the snap hook and in bootstrap/join cleanup. We can pick that up as a separate task and backport it to 1.32.

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -145,5 +146,65 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
+	tryCleanupContainerdPaths(snap)
+
 	return nil
 }
+
+// tryCleanupContainerdPaths attempts to clean up all containerd directories which were
+// created by the k8s-snap based on the existence of their respective lockfiles
+// located in the directory returned by `s.LockFilesDir()`.
+func tryCleanupContainerdPaths(s snap.Snap) {
+	log := log.L()
```
berkayoz, line 158: One last thing before merging, can we pass the `onPreRemove` logger through the context and use it here?

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -145,5 +146,65 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
+		if lockfile_contents != dirpath {
+			log.Info("WARN: lockfile points to different path than expected", "lockfile", lockpath, "expected", dirpath, "actual", lockfile_contents)
+			continue
+		}
+
+		// Check directory exists before attempting to remove:
+		if _, err := os.Stat(dirpath); os.IsNotExist(err) {
+			log.Info("Containerd directory doesn't exist; skipping cleanup", "directory", dirpath)
+		} else {
+			// NOTE(aznashwan): because of the convoluted interfaced-based way the snap
```
bschimke95, line 185: nit: typo

bschimke95: interfaced

bschimke95 (approved):
LGTM, minor nit and Berkays comment

bschimke95:
please also add some description to this PR. IMHO the title is not sufficient anymore.

aznashwan:
> please also add some description to this PR. IMHO the title is not sufficient anymore.

Made the commit messages very explicit for posterity; and pasted their descriptions in the PR.

PR #865, opened 2024-12-03
Fix install by revision in integration tests

bschimke95:
The commands in `shlex` need to contain only strings. Passing an `int` fails with
`TypeError: expected string or bytes-like object, got 'int'`

Relates to: https://github.com/canonical/canonical-kubernetes-release-ci/pull/14


addyess (approved):
ok, approved!

PR #867, opened 2024-12-04
Adds DrainConnectionsTimeout microcluster configuration

claudiubelu:
In microcluster, there is now a configuration that allows us to drain any current connections before shutting down the unix / http servers.

Note that the servers will also be shut down when removing a node, or if bootstrapping or joining a node fails. If we're not waiting for the connection to drain, we may close current connections before managing to write a response to the clients, resulting in them getting EOF errors.

### DIFF
--- a/src/k8s/pkg/k8sd/app/app.go
+++ b/src/k8s/pkg/k8sd/app/app.go
@@ -45,6 +45,8 @@ type Config struct {
 	DisableFeatureController bool
 	// DisableCSRSigningController is a bool flag to disable csrsigning controller.
 	DisableCSRSigningController bool
+	// DrainConnectionsTimeout is the amount of time to allow for all connections to drain when shutting down.
+	DrainConnectionsTimeout time.Duration
 }
 
 // App is the k8sd microcluster instance.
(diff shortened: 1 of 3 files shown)


eaudetcobello:
What is the expected impact of this change? What does this improve?

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -232,12 +232,13 @@ func (a *App) Run(ctx context.Context, customHooks *state.Hooks) error {
...
-		Hooks:            hooks,
-		ExtensionServers: api.New(ctx, a),
-		ExtensionsSchema: database.SchemaExtensions,
+		Version:                 string(apiv1.K8sdAPIVersion),
+		Verbose:                 a.config.Verbose,
+		Debug:                   a.config.Debug,
+		Hooks:                   hooks,
+		ExtensionServers:        api.New(ctx, a),
+		ExtensionsSchema:        database.SchemaExtensions,
+		DrainConnectionsTimeout: 10 * time.Second,
```
bschimke95, line 241: let's make this an k8sd option.

bschimke95 (commented):
Do we already run on a microcluster version that contains the fix or do we need to bump microcluster?
Could you link the relevant PR for documentation?

claudiubelu:
> What is the expected impact of this change? What does this improve?

This should finally close the issue with the EOF errors we've been seeing client-side when an error occurs server-side when bootstrapping / joining. This was addressed in microcluster, the PR contains more information on this issue: https://github.com/canonical/microcluster/pull/295

> Do we already run on a microcluster version that contains the fix or do we need to bump microcluster? Could you link the relevant PR for documentation?

We already bumped the microcluster version: https://github.com/canonical/k8s-snap/pull/968 . This contains the backported fix: https://github.com/canonical/microcluster/pull/309 . If the current version of microcluster didn't have the `DrainConnectionsTimeout` field, the CIs should have failed.

src/k8s/pkg/k8sd/app/app.go
claudiubelu: Done

bschimke95 (approved):
LGTM

PR #869, opened 2024-12-05
Add 1.32 fixes and updates

bschimke95:
# Summary
Add fixes and updates for 1.32 release

# Rationale
Upstream Kubernetes 1.32 will be released next week. [This PR](https://github.com/canonical/k8s-snap/pull/837) tests if the current snap is compatible with the current release candidate.

Some changes in the snap were required:

We noticed some test failures that happen because of an incompatibility of k8s-dqlite and the [consistent read from cache](https://kubernetes.io/blog/2024/08/15/consistent-read-from-cache-beta/). By default, this feature is enabled on 1.32.
For now, we'll disable this feature and do a follow-up investigation what it takes to support that feature in k8s-dqlite.
updated the k8s.io library versions
update Go version to 1.23.0 to build with the same version as upstream.

We do not want the release candidate to be on `main` so we separate the fixes from the actual version bump.


HomayoonAlimohammadi (approved):
Thanks a lot @bschimke95! LGTM

berkayoz (approved):
LGTM, I'd like to bring up a point that we'll need to handle cases like this in upgrades. We should create cards/follow up on this as well. See [microk8s PR](https://github.com/canonical/microk8s/pull/4773/files#diff-7f1ea137a27c441e1be1e0ddb37091c53b09335f3124c2b47c4792e82b15290d) for example

bschimke95:
Integration tests succeed: https://github.com/canonical/k8s-snap/actions/runs/12200077471/job/34036197199?pr=837

(only branch management not yet which is expected as there is no 1.32 branch yet)

PR #870, opened 2024-12-05
Loosen pre-configured certificate checks

petrutlucian94:
We recently introduced strict checks for pre-configured certificates. However, this broke cluster-api-k8s.

We'll loosen the checks, ignoring the certificate CN and O fields, checking only:

* NotBefore/NotAfter
* CA
* DNS SANs (fewer checks though)


bschimke95 (approved):
LGTM

PR #873, opened 2024-12-06
Check kubernetes services are running as expected after bootstrapping / joining

claudiubelu:
Currently, we're doing only a few checks that would ensure a successful installation of the k8s-snap (the containerd path is free for us to use, the ports meant for Kubernetes services are free, binaries are sane).

However, due to various reasons, the bootstrap / join process can still succeed, even if some of the services are not running (e.g.: invalid extra service arguments given in the bootstrap config, leading to a service to fail due to an unknown or invalid argument). At the moment, we're only verifying that `kube-apiserver` is accessible.

We're adding a few checks that the Kubernetes services are not stopped / failed after we've started them.

Adds integration test for this scenario.

### DIFF
--- a/k8s/lib.sh
+++ b/k8s/lib.sh
@@ -178,10 +178,10 @@ k8s::common::execute_service() {
 
   k8s::common::setup_env
 
+  set -xe
   # Source arguments and substitute environment variables. Will fail if we cannot read the file.
   declare -a args="($(cat "${SNAP_COMMON}/args/${service_name}"))"
 
-  set -xe
   exec "${SNAP}/bin/${service_name}" "${args[@]}"
 }
 
--- a/src/k8s/pkg/k8sd/app/cluster_util.go
+++ b/src/k8s/pkg/k8sd/app/cluster_util.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/canonical/k8s/pkg/k8sd/setup"
 	"github.com/canonical/k8s/pkg/snap"
 	snaputil "github.com/canonical/k8s/pkg/snap/util"
+	"github.com/canonical/k8s/pkg/utils/control"
 	mctypes "github.com/canonical/microcluster/v2/rest/types"
 )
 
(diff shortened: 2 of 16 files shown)


k8s/lib.sh
```diff
@@ -178,10 +178,10 @@ k8s::common::execute_service() {
 
   k8s::common::setup_env
 
+  set -xe
   # Source arguments and substitute environment variables. Will fail if we cannot read the file.
   declare -a args="($(cat "${SNAP_COMMON}/args/${service_name}"))"
 
-  set -xe
```
claudiubelu, line 184: Note: moved these above. If the args are missing, it should error out as well.

src/k8s/pkg/k8sd/app/cluster_util.go
```diff
@@ -61,6 +63,20 @@ func waitApiServerReady(ctx context.Context, snap snap.Snap) error {
 	return nil
 }
 
+func waitControlPlaneServices(ctx context.Context, snap snap.Snap) error {
+	// The services may be able to start, appearing to be "active", but they might eventually fail due to
+	// various reasons, and they may be restarted. We're checking their activity a few times.
+	if err := control.Consistently(ctx, 3, 5*time.Second, func() error {
+		if err := snaputil.CheckControlPlaneServicesStates(ctx, snap, "active"); err != nil {
+			return fmt.Errorf("failed to ensure all control plane services are active: %w", err)
+		}
```
eaudetcobello, line 72: I'm not the biggest fan of adding 15 seconds to every bootstrap...

src/k8s/pkg/snap/pebble.go
```diff
@@ -66,6 +73,31 @@ func (s *pebble) RestartService(ctx context.Context, name string) error {
 	return s.runCommand(ctx, []string{filepath.Join(s.snapDir, "bin", "pebble"), "restart", name})
 }
 
+// GetServiceState returns a k8s service state. The name can be either prefixed or not.
+func (s *pebble) GetServiceState(ctx context.Context, name string) (string, error) {
```
eaudetcobello, line 77: Please add a unit test for this function

src/k8s/pkg/k8sd/app/cluster_util.go
eaudetcobello: I am not a fan of any kind of hard-coded time. What about this?

> Is it not possible to poll the state frequently and continue as soon as all services are active?

claudiubelu: Not really, systemd reports a unit as `active` immediately after starting it, even if it'll fail almost immediately afterwards. Ofcourse, it'll eventually mark it as failed afterwards, and it'll try again to start it due to the restart policy.

If you are to check the unit state immediately after starting it, you'll most likely see it as `active`. We can easily test this. Let's say we have a ``--foo=lish`` argument in ``/var/snap/k8s/common/args/kubelet``, which will cause ``kubelet`` to fail. If we then run the following command:

```
svc_name="snap.k8s.kubelet.service" && sudo systemctl stop $svc_name && sudo systemctl start $svc_name && systemctl is-active $svc_name && sleep 5 && systemctl is-active $svc_name
```

We'd see the following:

```
active
failed
```

If the sleep is shorter (2s), you might even see it in `activating` state.

An idea would be to have `ExecStartPost` in the systemd units, which could help ensure that the service is running. As an interesting tidbit, it seems that setting `ExecStartPost=/usr/bin/sleep 1` would actually protect against immediately-exiting services (though I'm not 100% it is an intended behaviour).

src/k8s/pkg/snap/pebble.go
claudiubelu: Done

eaudetcobello:
This has been inactive for a while. Feel free to re-open if still relevant.

PR #874, opened 2024-12-06
Fix pebble Containerd path

HomayoonAlimohammadi:
### Overview
This PR fixes the Pebble Containerd path. Without the leading slash, the path is considered relative.


petrutlucian94 (approved):
Thank for addressing this! I can confirm that this fixes the cluster-api-k8s tests.

PR #876, opened 2024-12-06
Fix nightly tests

petrutlucian94:
Fix nightly job

* fix yaml definition
* ensure that the containerd base path setting is properly parsed when joining nodes
* fix containerd test assertions
* run all tests whenever there are changes under the ``tests`` dir and in case of backports

### DIFF
--- a/.github/workflows/nightly-test.yaml
+++ b/.github/workflows/nightly-test.yaml
@@ -30,6 +30,8 @@ jobs:
           sudo lxd init --auto
           sudo usermod --append --groups lxd $USER
           sg lxd -c 'lxc version'
+          sudo iptables -I DOCKER-USER -i lxdbr0 -j ACCEPT
+          sudo iptables -I DOCKER-USER -o lxdbr0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
       - name: Create build directory
         run: mkdir -p build
       - name: Install ${{ matrix.release }} k8s snap
--- a/.github/workflows/nightly-test.yaml
+++ b/.github/workflows/nightly-test.yaml
@@ -17,7 +17,7 @@ jobs:
         release: ["latest/edge"]
       fail-fast: false # TODO: remove once arm64 works
 
-    runs-on: ${{ matrix.arch == 'arm64' && ["self-hosted", "Linux", "ARM64", "jammy", "large"] || ["self-hosted", "Linux", "AMD64", "jammy", "large"] }}
+    runs-on: ${{ matrix.arch == 'arm64' && 'self-hosted-linux-arm64-jammy-large' || 'self-hosted-linux-amd64-jammy-large' }}
 
     steps:
       - name: Checking out repo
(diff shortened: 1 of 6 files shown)


.github/workflows/nightly-test.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
         release: ["latest/edge"]
       fail-fast: false # TODO: remove once arm64 works
 
-    runs-on: ${{ matrix.arch == 'arm64' && ["self-hosted", "Linux", "ARM64", "jammy", "large"] || ["self-hosted", "Linux", "AMD64", "jammy", "large"] }}
+    runs-on: ${{ matrix.arch == 'arm64' && 'self-hosted-linux-arm64-jammy-large' || 'self-hosted-linux-amd64-jammy-large' }}
```
louiseschmidtgen, line 20: Should we make this adjustment in our daily test as well?

tests/integration/tests/test_cleanup.py
```diff
@@ -64,15 +64,28 @@ def test_node_cleanup_new_containerd_path(instances: List[harness.Instance]):
...
     new_containerd_paths = [
-        os.path.join(boostrap_config["containerd-base-dir"], "k8s-containerd", p)
+        os.path.join(
+            boostrap_config["containerd-base-dir"], "k8s-containerd", p.lstrip("/")
+        )
         for p in CONTAINERD_PATHS
     ]
+
+    # We expect containerd to use our custom paths instead of the default ones.
+    # However, the test fixture places registry configuration in /etc/containerd
```
louiseschmidtgen, line 74: Can we please fix this in the fixture instead of cleaning up the paths after the fact?

louiseschmidtgen (commented):
Thanks for looking into the containerd issue @petrutlucian94!
Left some comments here:

.github/workflows/nightly-test.yaml
petrutlucian94: Not necessarily, we need it here because of the ``matrix.arch == 'arm64' &&`` check, which doesn't work when followed by a list.

tests/integration/tests/test_cleanup.py
petrutlucian94: Thanks for the suggestion. I think we can add a marker for custom containerd paths to accommodate this test.

.github/workflows/nightly-test.yaml
```diff
@@ -3,6 +3,7 @@ name: Nightly Latest/Edge Tests
 on:
   schedule:
     - cron: "0 0 * * *" # Runs every midnight
+  pull_request:
```
bschimke95, line 6: reminder: remove before merge

bschimke95 (approved):
LGTM

src/k8s/pkg/k8sd/api/cluster_join.go
```diff
@@ -37,16 +37,16 @@ func (e *Endpoints) postClusterJoin(s state.State, r *http.Request) response.Res
 
 	joinConfig := struct {
 		// We only care about this field from the entire join config.
-		containerdBaseDir string `yaml:"containerd-base-dir,omitempty"`
+		ContainerdBaseDir string `yaml:"containerd-base-dir,omitempty"`
```
berkayoz, line 40: Can we drop a comment on why the uppercase is needed? (Fixing the marshall bug, just as a reminder to ourselves)

berkayoz (approved):
LGTM, one small thing

src/k8s/pkg/k8sd/api/cluster_join.go
petrutlucian94: Sure, done.

PR #880, opened 2024-12-09
Add unit tests for token validator endpoints

eaudetcobello:
Add unit tests for the ValidateNodeTokenAccessHandler.
Add mock for Provider interface.

PR #885, opened 2024-12-11
implement TestValidateCAPIAuthTokenAccessHandler

eaudetcobello:
* TestValidateCAPIAuthTokenAccessHandler
* rename WithDB to WithState and move to a new location with package name microcluster_testenv
* WithState gives access to the whole state.State type instead of state.State.Database() only

Depends on https://github.com/canonical/k8s-snap/pull/880

### DIFF
--- a/src/k8s/pkg/k8sd/api/capi_access_handler_test.go
+++ b/src/k8s/pkg/k8sd/api/capi_access_handler_test.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/canonical/k8s/pkg/k8sd/api"
 	"github.com/canonical/k8s/pkg/k8sd/database"
-	microcluster_testenv "github.com/canonical/k8s/pkg/utils/microcluster"
+	testenv "github.com/canonical/k8s/pkg/utils/microcluster"
 	"github.com/canonical/microcluster/v2/state"
 	. "github.com/onsi/gomega"
 )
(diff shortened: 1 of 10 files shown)


src/k8s/pkg/k8sd/api/capi_access_handler_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"context"
+	"database/sql"
+	"net/http"
+	"testing"
+
+	"github.com/canonical/k8s/pkg/k8sd/api"
+	"github.com/canonical/k8s/pkg/k8sd/database"
+	microcluster_testenv "github.com/canonical/k8s/pkg/utils/microcluster"
```
bschimke95, line 11: nit: not sure if we need to be that verbose but fine for me

bschimke95 (approved):
LGTM

src/k8s/pkg/k8sd/api/capi_access_handler_test.go
```diff
@@ -0,0 +1,90 @@
...
+					err = s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
+						return database.SetClusterAPIToken(ctx, tx, tc.tokenDBContent)
+					})
+					g.Expect(err).To(Not(HaveOccurred()))
+				}
+
+				req := &http.Request{
+					Header: make(http.Header),
+				}
+				req.Header.Set(api.CAPITokenHeaderName, tc.tokenHeaderContent)
```
HomayoonAlimohammadi, line 75: Ah I really wish we could have the `node-token` in the k8s API as well...
Would you mind adding it or making a card to add it?

HomayoonAlimohammadi (approved):
Thanks a lot @eaudetcobello, LGTM!

src/k8s/pkg/k8sd/api/capi_access_handler_test.go
eaudetcobello: Changed to testenv

PR #892, opened 2024-12-11
Deduplicate github actions

petrutlucian94:
We have multiple github actions that run e2e tests and share a significant amount of logic.

We'll add reusable actions, making the workflows much easier to maintain.


petrutlucian94:
There may be a problem with the TICS job, it fails even without these changes:

```
2024-12-12 10:28:46.636: [INFO 3005] License expires in 139 days.
2024-12-12 10:28:46.739: [INFO 3000] Running TICS.bin -branchdir=/home/runner/work/k8s-snap/k8s-snap/ -calc=ALL -overviews -project=k8s-snap -results -showsynopsis -timeinfo -tmpdir=/tmp/tics -totaloverviews from /home/runner/work/k8s-snap/k8s-snap/src/k8s
2024-12-12 10:28:51.709: [INFO 3000] Creating database backup for project 'k8s-snap'.
Wrote backup to /opt/tiobe/TICS/WebServer/TICS/backups/qdb/k8s-snap-140-2024-12-12_10_28_52.fbk
[ERROR 5333] Project database 'k8s-snap' already is connected to a started TICSQServer process that has not registered a stop time. To avoid data corruption, analysis is aborted. Details of the other TICSQServer process: started at 2024-12-12 10:24:16.000 by user runner.
```

.github/workflows/nightly-test.yaml
```diff
@@ -14,58 +18,11 @@ jobs:
...
-      - name: Tmate debugging session
-        if: ${{ failure() && github.event_name == 'pull_request' }}
-        uses: mxschmitt/action-tmate@v3
-        timeout-minutes: 10
+    uses: ./.github/workflows/e2e-tests.yaml
+    with:
+      arch: ${{ matrix.arch }}
+      os: ${{ matrix.os }}
+      channel: ${{ matrix.channel }}
+      test-tags: 'up_to_weekly'
```
petrutlucian94, line 28: TODO: this should be ``up_to_nightly``

petrutlucian94:
``tests/test_ingress.py::test_ingress`` is constantly failing on Moonray, need to figure out why.

tests/integration/tests/test_ingress.py
```diff
@@ -41,7 +41,7 @@ def get_ingress_service_node_port(p):
 def get_external_service_ip(instance: harness.Instance, service_namespace) -> str:
     try_count = 0
     ingress_ip = None
-    while ingress_ip is None and try_count < 5:
+    while ingress_ip is None and try_count < 20:
```
berkayoz, line 44: The changes here are not leftovers from testing right? Just making sure these are intentional.

k8s/scripts/inspect.sh
```diff
@@ -54,7 +54,7 @@ function collect_args {
 
 function collect_cluster_info {
   log_info "Copy k8s cluster-info dump to the final report tarball"
-  k8s kubectl cluster-info dump --output-directory "$INSPECT_DUMP/cluster-info" &>/dev/null
+  k8s kubectl cluster-info dump --all-namespaces --output-directory "$INSPECT_DUMP/cluster-info" &>/dev/null
```
berkayoz, line 57: Is this the part where we switch to collect logs from all namespaces? I'm not too sure about this script is used by actual users as well. We shouldn't get logs to their workloads which might contain secrets, sensitive info etc.

.github/workflows/nightly-test.yaml
```diff
@@ -14,58 +19,86 @@ jobs:
...
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
+    strategy:
+      matrix:
+        include:
+          # Latest branches
+          - { branch: main, channel: latest/edge }
+          # Stable branches
+          # Add branches to test here
+          - { branch: release-1.30, channel: 1.30-classic/edge }
+          - { branch: release-1.31, channel: 1.31-classic/edge }
```
berkayoz, line 43: Are we missing 1.32 here? @bschimke95

.github/workflows/nightly-test.yaml
```diff
@@ -14,58 +19,86 @@ jobs:
...
-        uses: mxschmitt/action-tmate@v3
-        timeout-minutes: 10
+          ref: ${{matrix.branch}}
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.22"
```
berkayoz, line 69: We change the go version often, should we get this from a file like https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

berkayoz (commented):
Did a first pass, great work

.github/workflows/nightly-test.yaml
```diff
@@ -14,58 +19,86 @@ jobs:
       matrix:
         os: ["ubuntu:20.04", "ubuntu:22.04", "ubuntu:24.04"]
         arch: ["amd64", "arm64"]
-        release: ["latest/edge"]
+        channel: ["latest/edge"]
```
louiseschmidtgen, line 22: We should probably test our release channels as well for the nightly test and add `1.32-classic/stable`.

.github/workflows/nightly-test.yaml
louiseschmidtgen, line 43: Can we automate this please? We don't want to add each release branch manually for every release.

tests/integration/tests/test_cleanup.py
```diff
@@ -114,6 +114,7 @@ def test_node_cleanup_new_containerd_path(instances: List[harness.Instance]):
 @pytest.mark.node_count(1)
 @pytest.mark.no_setup()
 @pytest.mark.tags(tags.NIGHTLY)
+@pytest.mark.skip(reason="broken, the open port won't affect the bootstrap under lxd")
```
louiseschmidtgen, line 117: Please rephrase the error reason.

louiseschmidtgen (commented):
Nice work @petrutlucian94, I love the move of shared bits into actions!
Did a first pass and left some suggestions.

.github/workflows/nightly-test.yaml
louiseschmidtgen: Yes

louiseschmidtgen:
Could you please also update the integration test in k8s-dqlite to use your new actions? https://github.com/canonical/k8s-dqlite/blob/master/.github/workflows/k8s-snap-integration.yaml

.github/actions/download-k8s-snap/action.yaml
```diff
@@ -0,0 +1,61 @@
...
+      id: retrieve-path
+      run: |
+        if [[ -n "${{ inputs.artifact }}" ]]; then
+          snap_path="${{ github.workspace }}/${{ inputs.artifact }}"
+        else
+          snap_path="${{ github.workspace }}/k8s.snap"
+        fi
+        echo "snap-path=$snap_path" >> "$GITHUB_OUTPUT"
+
+        echo "Downloaded snap: $snap_path"
```
bschimke95, line 60: nit: we don't download anything in this step.
If `input.channel` is set you'd print "Downloaded snap: ..." twice

.github/actions/install-lxd/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: Install lxd
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install lxd snap
+      shell: bash
+      run: |
+        sudo snap refresh lxd --channel 5.21/stable
```
bschimke95, line 9: if you make this a separate action, I'd make this configurable

.github/actions/install-lxd/action.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - name: Initialize lxd
+      shell: bash
+      run: |
+        sudo lxd init --auto
+        sudo usermod --append --groups lxd $USER
+        sg lxd -c 'lxc version'
+    - name: Apply Docker iptables workaround
+      shell: bash
+      run: |
```
bschimke95, line 18: please add a command why this is required. 
You can link to https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker

.github/workflows/build-snap.yaml
```diff
@@ -0,0 +1,51 @@
...
+        type: string
+    outputs:
+      snap-artifact:
+        description: Name of the uploaded snap artifact
+        value: ${{ jobs.build-snap.outputs.snap-artifact }}
+
+jobs:
+  build-snap:
+    name: Build snap
+    runs-on: ubuntu-20.04
```
bschimke95, line 17: let's bump the versions here and elsewhere as we are here. 20.04 is EoL in 4 months.

.github/workflows/lint_and_integration.yaml
```diff
@@ -0,0 +1,140 @@
...
+  push:
+    paths-ignore:
+      - 'docs/**'
+    branches:
+      - main
+      - autoupdate/strict
+      - autoupdate/moonray
+      - 'release-[0-9]+.[0-9]+'
+      - 'autoupdate/release-[0-9]+.[0-9]+-strict'
+      - 'autoupdate/sync/**'
```
bschimke95, line 13: I wonder why we are so explicit about the autoupdate branches, can you remember @berkayoz?
Why don't we just use
```
  - main
  - autoupdate/**                
  - 'release-[0-9]+\.[0-9]+'  
  ```

File: tests/integration/tests/test_ingress.py (hunk 2 above)
Comment (petrutlucian94) [reply tests/integration/tests/test_ingress.py:44]: These changes are intentional, aiming to reduce test flakiness.

File: .github/workflows/lint_and_integration.yaml [hunk 13]
```diff
@@ -0,0 +1,140 @@
...
+        os: ["ubuntu:20.04", "ubuntu:22.04", "ubuntu:24.04"]
+    needs: [build-snap, get-e2e-tags, go-lint-and-unit, python-lint]
+    uses: ./.github/workflows/e2e-tests.yaml
+    with:
+      arch: amd64
+      os: ${{ matrix.os }}
+      test-tags: ${{ needs.get-e2e-tags.outputs.test-tags}}
+      artifact: k8s.snap
+
+  test-integration-informing:
```
bschimke95, line 115: @berkayoz shoulkd we just merge this informing into `test-integration`?
IIRC we did this separation historically because there were some flaky tests (or something like this?) but I don't think this is necessary anymore.

.github/workflows/lint_and_integration.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+  go-lint-and-unit:
+    name: Go lint and unit tests
+    uses: ./.github/workflows/go.yaml
+    permissions:
+      contents: read  # for actions/checkout to fetch code
+      pull-requests: write  # for marocchino/sticky-pull-request-comment to create or update PR comment
+      checks: write # for golangci/golangci-lint-action to checks to allow the action to annotate code in the PR.
+
+  get-e2e-tags:
```
bschimke95, line 73: nit: we don't need to run this if the linters fail.

k8s/scripts/inspect.sh
petrutlucian94: Good point, I assumed that it's only used by the CI. We could add an option called ``--verbose`` or ``--detailed``. It can be a follow-up to avoid delaying this PR.

.github/workflows/nightly-test.yaml
petrutlucian94: Missed that, thanks! 1.32 wasn't released when I prepared the first commit.

.github/workflows/nightly-test.yaml
```diff
@@ -14,58 +19,86 @@ jobs:
...
       fail-fast: false # TODO: remove once arm64 works
```
bschimke95, line 23: nit: arm64 should work now - we can remove that.

.github/workflows/nightly-test.yaml
petrutlucian94: Good point.

.github/workflows/nightly-test.yaml
bschimke95: yes, this eventually needs to be automated. 
@petrutlucian94 let's add a todo for follow up work. This PR is already large enough.

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Run e2e tests
+        env:
+          TEST_SNAP: ${{ steps.download-snap.outputs.snap-path }}
+          TEST_SUBSTRATE: lxd
+          TEST_LXD_IMAGE: ${{ inputs.os }}
+          TEST_FLAVOR: ${{ inputs.flavor }}
+          TEST_INSPECTION_REPORTS_DIR: ${{ github.workspace }}/inspection-reports
+          # Test the latest (up to) 6 releases for the flavour
+          # TODO(ben): upgrade nightly to run all flavours
+          TEST_VERSION_UPGRADE_CHANNELS: "recent 6 classic"
```
bschimke95, line 64: we need "recent 6 \<flavor\>" here and use "classic" if flavor is empty.

.github/actions/download-k8s-snap/action.yaml
```diff
@@ -0,0 +1,61 @@
+name: Download k8s-snap
```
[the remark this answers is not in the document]
petrutlucian94: Yeah, unfortunately it has to be called action.yaml.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow

https://stackoverflow.com/questions/74350826/github-composite-actions-cant-find-action-yml (I'm not a big fan of pasting stackoverflow answers as reference but the docs are a bit vague).

.github/workflows/nightly-test.yaml
bschimke95: We don't need that.
We run the full test suite before populating/backporting to `release-1.XX` branches. This would just run the same tests again.
So, we only need `latest/edge`

.github/actions/download-k8s-snap/action.yaml
```diff
@@ -0,0 +1,61 @@
...
+      run: |
+        echo "No k8s-snap artifact or channel specified..."
+        exit 1
+    - name: Exit if multiple inputs provided
+      if: inputs.artifact != '' && inputs.channel != ''
+      shell: bash
+      run: |
+        echo "Received snap artifact AND snap channel."
+        exit 1
+
```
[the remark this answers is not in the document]
petrutlucian94: Indeed, that would be cleaner.

.github/workflows/nightly-test.yaml
bschimke95: I think it makes sense to have a separate "get-latest-releases" action that can be reused across the workflows as we likely need this in other places as well.

.github/workflows/nightly-test.yaml
bschimke95: +1

.github/workflows/nightly-test.yaml
petrutlucian94: I'd do that in a separate PR if that's ok with you. To be honest, I'm not entirely sure if that would work with ``matrix`` input.

.github/workflows/nightly-test.yaml
```diff
@@ -14,58 +19,86 @@ jobs:
...
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.22"
+      - name: go mod download
+        working-directory: src/k8s
+        run: go mod download
+      - name: TICS scan
+        run: |
+          export TICSAUTHTOKEN=${{ secrets.TICSAUTHTOKEN }}
```
bschimke95, line 75: We eventually need to move those steps into a script that can be executed in one line. Feels free to just add a todo for now and we can return to this.

.github/workflows/security-scan.yaml
```diff
@@ -0,0 +1,80 @@
...
+        type: string
+      channel:
+        description: k8s snap channel.
+        type: string
+      checkout-ref:
+        description: k8s-snap git checkout ref, optional.
+        type: string
+
+jobs:
+  get-tags:
```
bschimke95, line 22: I guess this should be called something like `security-scan`?

petrutlucian94:
> 

Sure, we can handle the k8s-dqlite actions after this PR merges. Let's add a separate card though.

.github/workflows/security-scan.yaml
```diff
@@ -0,0 +1,80 @@
...
+    name: Security scan
+    runs-on: ubuntu-20.04
```
bschimke95, line 24: just use `ubuntu-latest` if this breaks our CI it is good to know and we need to investigate anyway.

.github/workflows/security-scan.yaml
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          channel: ${{ inputs.channel }}
+          artifact: ${{ inputs.artifact }}
+      - name: Checking out tested repo branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.checkout-ref }}
+          # Persist downloaded artifacts
+          clean: 'false'
+      - name: Setup Trivy vulnerability scanner
```
bschimke95, line 48: those trivy steps also need to move into a script. (I know that you simply moved this workflow over - just a comment).

k8s/scripts/inspect.sh
bschimke95: let's make this a configurable flag of the script. 
For the inspection report, we can then enable this flag (or specify all of our namespace).
I'd tackle this in a separate PR.

bschimke95 (commented):
Great work @petrutlucian94 
Did a first pass

.github/workflows/lint_and_integration.yaml
berkayoz: I'm not sure about the why but personally being more explicit/expecting a pattern feels less prone to errors. Up to you folks

.github/workflows/lint_and_integration.yaml
berkayoz: Sure we can merge it, this will probably be fleshed out as part of the flavors spec. Eventually I'd like "flavors" as a separate job/step, to have a visual separation between the default track.

.github/actions/download-k8s-snap/action.yaml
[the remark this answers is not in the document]
petrutlucian94: > If you combine both steps you won't know which error happened. I'd keep it like this.

I was thinking about having a single action but two separate checks inside the "run" block but I guess it's ok either way. I can leave it like this for now.

.github/workflows/nightly-test.yaml
petrutlucian94: @bschimke95 thanks for the clarification.

.github/workflows/nightly-test.yaml
louiseschmidtgen: Let's use the automation in Microk8s workflows as well. We have the same issues there.

.github/actions/download-k8s-snap/action.yaml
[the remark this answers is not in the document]
louiseschmidtgen: I see, in that case there's nothing we can do.

.github/actions/download-k8s-snap/action.yaml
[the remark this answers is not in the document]
louiseschmidtgen: Cool, in that case I would suggest the names: `Fail if neither artifact nor channel are specified` and `Fail if both artifact and channel are specified`.

.github/workflows/security-scan.yaml
petrutlucian94: Whops, indeed.

.github/workflows/lint_and_integration.yaml
bschimke95: Discussed offline `autoupdate/sync` is obsolete.
@petrutlucian94 feel free to either use `autoupdate/**` or `autoupdate/moonray`, ... but remove the `autoupdate/sync` in any case

.github/workflows/nightly-test.yaml
petrutlucian94: The integration tests are still a bit flakey, we could keep ``fail-fast: false`` for now to see if the other jobs pass.

.github/workflows/lint_and_integration.yaml
petrutlucian94: One thing to note is that ``test-integration`` runs against ubuntu 20.04, 22.04 and 24.04 while the informing/moonray job only covers one version. If we merge them, we'll get additional moonray jobs, which implies additional pressure on the worker pool.

petrutlucian94:
Follow-up items that will be handled in subsequent PRs:

* move Trivy and TICS checks to separate scripts
* add a job to automatically detect snap releases
    * should also be used by microk8s
* add "verbose" inspect.sh flag
* update k8s-dqlite job: https://github.com/canonical/k8s-dqlite/blob/master/.github/workflows/k8s-snap-integration.yaml

PR #898, opened 2024-12-12
integration: fix all linter warnings in test harness.

aznashwan:


tests/integration/tests/test_util/util.py
```diff
@@ -43,6 +43,65 @@ def run(command: list, **kwargs) -> subprocess.CompletedProcess:
...
 
+class Retriable:
+    def __init__(self, retry_kwargs: Optional[Mapping[str, Any]]) -> None:
+        self._condition = None
+        self._run = partial(run, capture_output=True)
+        if retry_kwargs is None:
+            retry_kwargs = {}
+        self._retry_kwargs = retry_kwargs
+
+    def exec(
```
berkayoz, line 54: Why the switch from `@retry(**_retry_args)`?

berkayoz (approved):
LGTM

tests/integration/tests/test_util/util.py
aznashwan: The `_retry_args` were composed in `stubbornly` and applied as a decorator on `Retriable.exec()`; now they're simply passed as instance attrbutes (`self._retry_kwargs = retry_kwargs` above)

There was no practical reason to (re)declare the `Retriable` class every time someone called `stubbornly` just to force the `@decorator` syntactic sugar, so it's been made an explicit decorator call below (`return retry(**self._retry_kwargs)(self._exec)(...)`)

PR #899, opened 2024-12-12
1.32 release docs

nhennigan:
Release docs for 1.32

### DIFF
--- a/docs/canonicalk8s/reuse/substitutions.yaml
+++ b/docs/canonicalk8s/reuse/substitutions.yaml
@@ -1,6 +1,6 @@
 product: 'Canonical Kubernetes'
-version: '1.31'
-channel: '1.31/candidate'
+version: '1.32'
+channel: '1.32/edge'
 multi_line_example: |-
   *Multi-line* text
    that uses basic **markup**.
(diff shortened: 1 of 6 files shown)


docs/canonicalk8s/reuse/substitutions.yaml
```diff
@@ -1,6 +1,6 @@
 product: 'Canonical Kubernetes'
-version: '1.31'
-channel: '1.31/candidate'
+version: '1.32'
+channel: '1.32/stable'
```
bschimke95, line 3: Why do we define those channels here but still need to change the channels in a bunch of places?

bschimke95 (approved):
LGTM, minor comment 

Btw, please refrain from putting the Jira ticket number in the title or description of the PR. It is an internal reference that external contributors/users cannot access.
It's fine to have it in the branch name so that Jira picks it up, though.

docs/canonicalk8s/reuse/substitutions.yaml
nhennigan: These {{channels}}/{{version}} cannot be included in code snippets so that is why we have the install file instead. I'm not sure if they will be used in the future.

HomayoonAlimohammadi (approved):
Great work, thanks a lot @nhennigan! LGTM

PR #908, opened 2024-12-16
Cleans up k8s-dqlite state dir and stops it on remove hook

claudiubelu:
``cfg.Datastore.GetType()`` may return an empty string if the bootstrap action failed before ``database.SetClusterConfig`` has been called. Because of this, we're not removing the state dir for k8s-dqlite, which will be wrongfully removed by ``setup.K8sDqlite`` on the next bootstrap attempt.

We're now opportunistically cleaning up the ``k8s-dqlite`` related state directory.

If a bootstrap attempt fails, the ``k8s-dqlite`` service will still be running, which will cause the next bootstrap attempt to fail, as the ``k8s-dqlite`` port will be currently in use.

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -92,10 +92,6 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 				log.Error(err, "Failed to create k8s-dqlite client: %w")
 			}
 
-			log.Info("Cleaning up k8s-dqlite directory")
-			if err := os.RemoveAll(snap.K8sDqliteStateDir()); err != nil {
-				return fmt.Errorf("failed to cleanup k8s-dqlite state directory: %w", err)
-			}
 		case "external":
 			log.Info("Cleaning up external datastore certificates")
 			if _, err := setup.EnsureExtDatastorePKI(snap, &pki.ExternalDatastorePKI{}); err != nil {
@@ -107,6 +103,10 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 		log.Error(err, "Failed to retrieve cluster config")
 	}
 
+	log.Info("Cleaning up k8s-dqlite directory")
+	if err := os.RemoveAll(snap.K8sDqliteStateDir()); err != nil {
+		log.Error(err, "failed to cleanup k8s-dqlite state directory")
+	}
 	for _, dir := range []string{snap.ServiceArgumentsDir()} {
 		log.WithValues("directory", dir).Info("Cleaning up config files", dir)
 		if err := os.RemoveAll(dir); err != nil {
@@ -143,6 +143,11 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 		log.Info("Stopping control plane services")
 		if err := snaputil.StopControlPlaneServices(ctx, snap); err != nil {
 			log.Error(err, "Failed to stop control-plane services")
+		}
+
+		log.Info("Stopping k8s-dqlite")
+		if err := snaputil.StopK8sDqliteServices(ctx, snap); err != nil {
+			log.Error(err, "Failed to stop k8s-dqlite service")
 		}
 	}
 
--- a/src/k8s/pkg/k8sd/app/cluster_util.go
+++ b/src/k8s/pkg/k8sd/app/cluster_util.go
@@ -29,24 +29,6 @@ func startControlPlaneServices(ctx context.Context, snap snap.Snap, datastore st
 	return nil
 }
 
-func stopControlPlaneServices(ctx context.Context, snap snap.Snap, datastore string) error {
-	// Stop services
-	switch datastore {
-	case "k8s-dqlite":
-		if err := snaputil.StopK8sDqliteServices(ctx, snap); err != nil {
-			return fmt.Errorf("failed to stop k8s-dqlite service: %w", err)
-		}
-	case "external":
-	default:
-		return fmt.Errorf("unsupported datastore %s, must be one of %v", datastore, setup.SupportedDatastores)
-	}
-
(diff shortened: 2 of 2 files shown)


src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -92,13 +91,9 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
 			}
 
-			log.Info("Cleaning up k8s-dqlite directory")
-			if err := os.RemoveAll(snap.K8sDqliteStateDir()); err != nil {
-				return fmt.Errorf("failed to cleanup k8s-dqlite state directory: %w", err)
-			}
 		case "external":
 			log.Info("Cleaning up external datastore certificates")
-			if _, err := setup.EnsureExtDatastorePKI(snap, &pki.ExternalDatastorePKI{}); err != nil {
+			if err := setup.RemoveExtDatastorePKIFiles(snap); err != nil {
```
bschimke95, line 96: This is not required. The files are deleted if the respective PKI file is not set.
See https://github.com/canonical/k8s-snap/blob/28919a246052d403b865cada03b25f35cc3f6a74/src/k8s/pkg/k8sd/setup/certificates.go#L15

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -118,7 +117,7 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 	// Trying to detect the node type is not reliable as the node might have been marked as worker
 	// or not, depending on which step it failed.
 	log.Info("Cleaning up worker certificates")
-	if _, err := setup.EnsureWorkerPKI(snap, &pki.WorkerNodePKI{}); err != nil {
+	if err := setup.RemoveWorkerPKIFiles(snap); err != nil {
```
bschimke95, line 120: This is also not required.

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -130,7 +129,7 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 	}
 
 	log.Info("Cleaning up control plane certificates")
-	if _, err := setup.EnsureControlPlanePKI(snap, &pki.ControlPlanePKI{}); err != nil {
+	if err := setup.RemoveControlNodePKIFiles(snap); err != nil {
```
bschimke95, line 132: same here.

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -144,6 +143,11 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 		if err := snaputil.StopControlPlaneServices(ctx, snap); err != nil {
 			log.Error(err, "Failed to stop control-plane services")
 		}
+
+		log.Info("Stopping k8s-dqlite")
+		if err := snaputil.StopK8sDqliteServices(ctx, snap); err != nil {
```
bschimke95, line 148: We have a similar implementation here: https://github.com/canonical/k8s-snap/blob/28919a246052d403b865cada03b25f35cc3f6a74/src/k8s/pkg/k8sd/app/cluster_util.go#L35

But it seems unused now. Consider either using the references implementation or remove it as part of this PR please.

src/k8s/pkg/k8sd/app/hooks_remove.go
claudiubelu: Ack. The name `EnsureExtDatastorePKI` is a bit confusing, the first thing that would come to mind is "ensure the files exists". Reverted this change.

src/k8s/pkg/k8sd/app/hooks_remove.go
claudiubelu: Done

src/k8s/pkg/k8sd/app/hooks_remove.go
claudiubelu: Done

src/k8s/pkg/k8sd/app/hooks_remove.go
claudiubelu: Right. I've chosen not to use it in this case, because it would return an error if it couldn't stop the k8s-dqlite services (e.g.: if doesn't exist), and leave the control plane services running, and for the remove hook, we should stop any started service regardless.

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -107,6 +103,10 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
 		log.Error(err, "Failed to retrieve cluster config")
 	}
 
+	log.Info("Cleaning up k8s-dqlite directory")
+	if err := os.RemoveAll(snap.K8sDqliteStateDir()); err != nil {
+		return fmt.Errorf("failed to cleanup k8s-dqlite state directory: %w", err)
```
berkayoz, line 108: Is this a best effort cleanup, if so I guess we should just `log.Error` instead of failing the whole hook right?

berkayoz (commented):
LGTM overall, one minor comment

src/k8s/pkg/k8sd/app/hooks_remove.go
claudiubelu: Indeed, you are right. We were originally removing the error as well, so this was moved as is. Corrected.

Though, `onPreRemove` returns an error. If we're only doing best effort here, what error should we eventually return, considering the fact that there are many error cases?

berkayoz: That is a good question. What could be a better way is maybe to use an error group, run through all steps and return error(s) through there? @bschimke95

berkayoz (approved):
LGTM

PR #912, opened 2024-12-17
integration: completely remove LocalHarness.

aznashwan:
The LocalHarness is a harness used for running integration tests
on the local machine where the tests are directly invoked
(be it via `tox`, `pytest`, etc).

It presents numerous limitations (can't run any multi-node tests)
and poses a lot of potential risks (cleanup failing in case of
fatal errors in the test fixtures) which outweighs most of its
convenience benefits (especially when compared to the LXD
substrate).

This patch completely removes the LocalHarness and all references to it
from the documentation, making LXD the new default.


bschimke95 (commented):
LGTM, please add a description to the PR on why we remove this.

berkayoz (approved):
LGTM

PR #915, opened 2024-12-18
Disable TICS runs

ktsakalozos:
Disable the TICS job


bschimke95 (approved):
For context: Tiobe performs updates on this workflow which will break this job. We were asked to disable the Tics jobs until further notice.

ktsakalozos:
Latest info we got is that we would ask TiCS to handle the upgrade in a more transparent way and/or we will live with the contaminated metrics.

PR #916, opened 2024-12-18
Automatically manage pre-release branches

petrutlucian94:
We need to automatically create k8s-snap branches for upstream pre-releases such as v1.33.0-alpha.1.
For v1.33.0-alpha.1, we'll generate a branch called ``autoupdate/v1.33.0-alpha``. This pre-release branch will be automatically deleted once the upstream stable release is published (e.g. 1.33.0).

Note that these branches are used by the Launchpad recipes to produce and publish k8s-snap builds.

The new workflow will be executed nightly and whenever a patch merges on main, keeping the pre-release branches in sync with the main branch.


petrutlucian94:
Instead of continuously/removing branches such as ``autoupdate/v1.33.0-alpha.0`` -> autoupdate/v1.33.0-beta.0, we may also consider having a single branch called ``autoupdate/v1.33-prerelease``.

.github/workflows/update-pre-release-branches.yaml
```diff
@@ -0,0 +1,77 @@
...
+          git fetch origin
+
+          # Log the latest release for reference.
+          latestRelease=`python3 ./build-scripts/k8s_releases.py get_latest_release`
+          echo "Latest k8s release: $latestRelease"
+
+          for outstandingPreRelease in `python3 ./build-scripts/k8s_releases.py get_obsolete_prereleases`; do
+            branch="autoupdate/${outstandingPreRelease}"
+            if git branch -r | grep "origin/$branch"; then
+              echo "Cleaning up obsolete pre-release branch: $branch"
```
bschimke95, line 72: If you are doing the cleanup here, then what is this for: https://github.com/canonical/canonical-kubernetes-release-ci/pull/18/files#diff-ce5e51b2f5af159f9e18385b27ee73318dad2d035680225b48298aba83cd60e2R90

I'd leave the branch creation and deletion to the ci repo and only perform the version update here.

.github/workflows/update-pre-release-branches.yaml
```diff
@@ -0,0 +1,77 @@
...
+        run: |
+          # Needed to create commits.
+          git config --global user.name "Github Actions"
+          git config --global user.email "worker@org.com"
+      - name: 'Update k8s component version: ${{ steps.determine.outputs.preRelease }}'
+        if: ${{ steps.determine.outputs.preRelease }} != ''
+        run: |
+          echo ${{ steps.determine.outputs.preRelease }} > ./build-scripts/components/kubernetes/version
+          git add ./build-scripts/components/kubernetes/version
+          git commit -m "Update k8s version to ${{ steps.determine.outputs.preRelease }}"
```
bschimke95, line 55: please move all of those steps into a script, similar to https://github.com/canonical/k8s-snap/blob/30d3f3b8429fb84e3df94a32928fcb1605403671/build-scripts/hack/update-component-versions.py#L1

build-scripts/k8s_releases.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
bschimke95, line 1: I don't think we need to add all of this here (see my comment above).

bschimke95 (commented):
I think we are doing two separate things here, branch management and version updates.

Let's reuse the existing `update-components` scripts and CI. I believe we don't need to create a separate workflow for this.

We only need to update the logic in the [upgrade scripts](https://github.com/canonical/k8s-snap/blob/30d3f3b8429fb84e3df94a32928fcb1605403671/build-scripts/hack/update-component-versions.py#L57) to correctly determine the version if this is a pre-release version. The rest should be similar to the upgrade-components workflow, no?

bschimke95:
@petrutlucian94 regarding your comment
```

Instead of continuously adding/removing branches such as autoupdate/v1.33.0-alpha.0 -> autoupdate/v1.33.0-beta.0, we may also consider having a single branch called autoupdate/v1.33-prerelease.
```

We don't need `autoupdate/v1.33.0-alpha.0` but `autoupdate/v1.33.0-alpha`.
`utoupdate/v1.33.0-prerelease` does not work as we need to populate alpha to edge, beta to beta and release candidate to candiate branches so we need to maintain all 3 branches.

.github/workflows/update-pre-release-branches.yaml
petrutlucian94: This workflow handles git branches, the other one manages launchpad recipes. IMO it would be redundant to have two workflows that manage git branches.

build-scripts/k8s_releases.py
petrutlucian94: I'm not sure I understand what you mean.

petrutlucian94:
>We don't need autoupdate/v1.33.0-alpha.0 but autoupdate/v1.33.0-alpha.
utoupdate/v1.33.0-prerelease does not work as we need to populate alpha to edge, beta to beta and release candidate to candiate branches so we need to maintain all 3 branches.

I see. Do we keep the alpha and beta branches once the rc comes out?

petrutlucian94:
> We only need to update the logic in the [upgrade scripts](https://github.com/canonical/k8s-snap/blob/30d3f3b8429fb84e3df94a32928fcb1605403671/build-scripts/hack/update-component-versions.py#L57) to correctly determine the version if this is a pre-release version. The rest should be similar to the upgrade-components workflow, no?

That script fetches the latest stable k8s version, we could add a parameter to get pre-releases instead and maybe have a separate step in ``update-components.yaml`` for pre-releases.

build-scripts/k8s_releases.py
petrutlucian94: Oh, I see. You mean moving the pre-release checks to the "update-components.yaml" script and the branch management to the other project.

petrutlucian94: One problem is that ``update-components.yaml`` will simply create a PR that updaets the ``component`` versions. However, we also need to sync the changes from the ``main`` branch of k8s-snap, which is why the new job currently does a force push.

.github/workflows/update-pre-release-branches.yaml
petrutlucian94: I've moved this logic to ``k8s_release.py``.

bschimke95 (approved):
LGTM

build-scripts/k8s_release.py
```diff
@@ -0,0 +1,199 @@
...
+        project_basedir, "build-scripts", "components", "kubernetes", "version"
+    )
+    with open(k8s_component_path, "w") as f:
+        f.write(k8s_version)
+
+
+def prepare_prerelease_git_branch(project_basedir: str, remote: str = "origin"):
+    prerelease = get_outstanding_prerelease()
+    if not prerelease:
+        LOG.info("No outstanding k8s pre-release.")
```
petrutlucian94, line 127: Whops, need to return here.

Issue #537, opened 2024-07-09
Docs: Edit icon doesn't go to the right location


YanisaHS:
The **Edit this page** icon gives a 404 error in GitHub and doesn't navigate correctly. I think your links are missing `src` in the path.

I only clicked it on a few pages, but this was the case for all of them. For example, I'm guessing your home page should navigate to `docs/src/index.md` instead of `docs/index.md`.

![Screenshot from 2024-07-09 15-41-40](https://github.com/canonical/k8s-snap/assets/53576988/b6e159e1-a2d0-40fb-a994-3c35d661363d)

![Screenshot from 2024-07-09 15-36-45](https://github.com/canonical/k8s-snap/assets/53576988/dfceeb08-4e37-4684-8d6b-3a3f4aec1620)

---
*Reported from: https://documentation.ubuntu.com/canonical-kubernetes/latest/*

YanisaHS:
@evilnick

evilnick:
So, the edit this page at the bottom works, the one at the top doesn't, because they both interpret the path in a different way. I think I will try and remove the top link, though as it doesn't show up when rendering locally I'm not entirely sure where it gets generated

evilnick:
I think this is something added automatically by RTD, which is likely to go away soon, but now I have more info on where it is generated (RTD manipulates the sphinx config and adds a load of things) I may be able to work around it

nhennigan:
The top edit button was removed and the bottom edit button path was updated.

PR #917, opened 2024-12-19
Move files under src so edit button works

nhennigan:
It was reported in issue #537 that the edit this page button did not work. Previous work was done to fix most pages but the about and community page at the top level were still affected. This PR fixes that functionality


louiseschmidtgen (approved):
LGTM, thanks!

PR #921, opened 2024-12-19
Include system information in the inspect.sh script

claudiubelu:
When investigating the state of a Kubernetes node, it may be useful to check a few host resources and their availability, to root cause potential memory / disk pressure issues.

### DIFF
--- a/k8s/scripts/inspect.sh
+++ b/k8s/scripts/inspect.sh
@@ -62,6 +62,33 @@ function collect_sbom {
   cp --no-preserve=mode,ownership /snap/k8s/current/bom.json "$INSPECT_DUMP"/sbom.json
 }
 
+function collect_system_info {
+  mkdir -p $INSPECT_DUMP/sys
+  log_info "Copy processes list to the final report tarball"
+  ps -ef > $INSPECT_DUMP/sys/ps
+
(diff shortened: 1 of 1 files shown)


k8s/scripts/inspect.sh
```diff
@@ -62,6 +62,24 @@ function collect_sbom {
...
+
+  log_info "Copy memory usage information to the final report tarball"
+  free -m > $INSPECT_DUMP/sys/memory_usage
+
+  log_info "Copy node uptime to the final report tarball"
+  uptime > $INSPECT_DUMP/sys/uptime
+
+  log_info "Copy /etc/os-release to the final report tarball"
+  cp /etc/os-release $INSPECT_DUMP/sys/etc-os-release
+}
```
bschimke95, line 81: Let's also add the loaded kernel modules(smod), CPU info(lscpu), network connections (`ss -tuln`), mounts (`cat /proc/mounts`) and swap (`swapon`) - you never know.

bschimke95 (commented):
LGTM, minor addition

k8s/scripts/inspect.sh
claudiubelu: Done

bschimke95 (approved):
LGTM

PR #924, opened 2024-12-20
[Docs] Fix broken links

nhennigan:
Fix broken links and add list of links to ignore to custom_conf due to known sphinx issue with anchors causing false positives.

### DIFF
--- a/docs/canonicalk8s/custom_conf.py
+++ b/docs/canonicalk8s/custom_conf.py
@@ -137,7 +137,10 @@ redirects = {}
 
 # Links to ignore when checking links
 linkcheck_ignore = [
-    'http://127.0.0.1:8000'
+    'http://127.0.0.1:8000',
+    'http://rocks.canonical.com',
+    'about',
+    'https://ceph.io/'
     ]
 
 # Pages on which to ignore anchors
@@ -166,7 +169,7 @@ custom_extensions = [
     'canonical.custom-rst-roles',
     'canonical.terminal-output',
     'notfound.extension',
-    'sphinxcontrib.cairosvgconverter', 
+    'sphinxcontrib.cairosvgconverter',
     ]
 
 # Add custom required Python modules that must be added to the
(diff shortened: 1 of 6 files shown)


louiseschmidtgen (approved):
LGTM, I'm no expert on Sphinx but the changes look sound to me.

docs/canonicalk8s/custom_conf.py
```diff
@@ -137,7 +137,10 @@
 
 # Links to ignore when checking links
 linkcheck_ignore = [
-    'http://127.0.0.1:8000'
+    'http://127.0.0.1:8000',
+    'http://rocks.canonical.com',
+    'about',
+    'https://ceph.io/'
```
bschimke95, line 143: why do we ignore the ceph link?

docs/canonicalk8s/custom_conf.py
```diff
@@ -231,3 +234,11 @@
 .. role:: h2
     :class: hclass2
 '''
+
+custom_linkcheck_anchors_ignore_for_url = [
```
bschimke95, line 238: what does this list do? Could you add a brief comment, please?

bschimke95 (commented):
Generally LGTM, only two minor questions.

docs/canonicalk8s/custom_conf.py
nhennigan: The ceph.io and ceph.com website seems to be returning an error on the linkcheck when I run it. However, when I go from the docs it resolves fine. I am not sure why it is raising a false positive with the ceph site.

docs/canonicalk8s/custom_conf.py
nhennigan: There is a [known issue](https://github.com/sphinx-doc/sphinx/issues/9016) with sphinx that it is unable to resolve some anchors (link to specific part of the page) in urls, even if they exist. With this custom_linkcheck_anchors_ignore_for_url it will ignore the anchors listed in those urls. When that issue is resolved upstream, I will hopefully be able to remove these urls.

PR #927, opened 2024-12-31
CI improvements

petrutlucian94:
* inspect.sh "--detailed" flag that collects logs from all namespaces
* ~~moved~~ copied TICS and Trivy tests to separate scripts
  * these workflows run against other branches (``main``, ``release/*``). We need to backport the scripts there before being able to clean up the workflows.

### DIFF
--- a/k8s/scripts/inspect.sh
+++ b/k8s/scripts/inspect.sh
@@ -7,22 +7,24 @@
 # elevated permissions (sudo).
 #
 # Usage:
-#   ./inspect.sh [output_file] [--detailed]
+#   ./inspect.sh [output_file] [--all-namespaces]
 #
 # Arguments:
-#   output_file  (Optional) The full path and filename for the generated tarball.
-#                If not provided, a default filename based on the current date
-#                and time will be used.
-#   --detailed   (Optional) Acquire detailed debugging information, including logs
-#                from all Kubernetes namespaces.
+#   output_file        (Optional) The full path and filename for the generated tarball.
+#                      If not provided, a default filename based on the current date
+#                      and time will be used.
+#   --all-namespaces   (Optional) Acquire detailed debugging information, including logs
+#                      from all Kubernetes namespaces.
 #
 # Example:
 #   ./inspect.sh /path/to/output.tar.gz
 #   ./inspect.sh  # This will generate a tarball with a default name.
-#   ./inspect.sh --detailed  # Obtain logs from all k8s namespaces.
+#   ./inspect.sh --all-namespaces  # Obtain logs from all k8s namespaces.
 
 INSPECT_DUMP=$(pwd)/inspection-report
-DETAILED=0
+# We won't fetch all namespaces by default to avoid logging potentially sensitive
+# user data.
+ALL_NAMESPACES=0
 
 function log_success {
   printf -- '\033[32m SUCCESS: \033[0m %s\n' "$1"
@@ -59,7 +61,7 @@ function collect_args {
 function collect_cluster_info {
   log_info "Copy k8s cluster-info dump to the final report tarball"
   local FLAGS=""
-  if [[ "$DETAILED" == "1" ]]; then
+  if [[ "$ALL_NAMESPACES" == "1" ]]; then
     FLAGS="--all-namespaces"
   fi
   k8s kubectl cluster-info dump $FLAGS --output-directory "$INSPECT_DUMP/cluster-info" &>/dev/null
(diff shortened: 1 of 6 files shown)


eaudetcobello:
Thanks @petrutlucian94! Adding a flag is OK with me, but is it possible to always collect info. from all namespaces? Does this pollute the inspection report too much for the majority of cases?

tests/tics-scan.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$(dirname "$BASH_SOURCE")
```
eaudetcobello, line 3: I tend to favor this, because it resolves to the script's absolute path:

```suggestion
SCRIPT_DIR=$(realpath $(dirname "$BASH_SOURCE"))
```

tests/tics-scan.sh
```diff
@@ -0,0 +1,34 @@
...
+
+set -ex
+cd "${SCRIPT_DIR}/.."
+
+# Install python dependencies
+pip install -r tests/integration/requirements-test.txt
+pip install -r tests/integration/requirements-dev.txt
```
eaudetcobello, line 10: This will fail on Ubuntu 24(?) because of PEP 668. Use a venv instead.

tests/tics-scan.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Install the TICS and staticcheck
+go install honnef.co/go/tools/cmd/staticcheck@v0.5.1
+. <(curl --silent --show-error 'https://canonical.tiobe.com/tiobeweb/TICS/api/public/v1/fapi/installtics/Script?cfg=default&platform=linux&url=https://canonical.tiobe.com/tiobeweb/TICS/')
+
+# We need to have our project built
+# We load the dqlite libs here instead of doing through make because TICS
+# will try to build parts of the project itself
+sudo add-apt-repository -y ppa:dqlite/dev
+sudo apt install dqlite-tools-v2 libdqlite1.17-dev
```
eaudetcobello, line 30: Missing -y flag

tests/tics-scan.sh
```diff
@@ -0,0 +1,34 @@
...
+sudo make clean
+go build -a ./...
+
+TICSQServer -project k8s-snap -tmpdir /tmp/tics -branchdir $HOME/work/k8s-snap/k8s-snap/
```
eaudetcobello, line 34: Change hardcoded directory `$HOME/work`.

k8s/scripts/inspect.sh
```diff
@@ -173,6 +180,25 @@ if [ "$EUID" -ne 0 ]; then
   exit 1
 fi
 
+POSITIONAL_ARGS=()
+while [[ $# -gt 0 ]]; do
```
bschimke95, line 184: Bash is such a beautiful and simple language.

k8s/scripts/inspect.sh
```diff
@@ -7,18 +7,22 @@
...
 #
 # Usage:
-#   ./inspect.sh [output_file]
+#   ./inspect.sh [output_file] [--detailed]
 #
 # Arguments:
 #   output_file  (Optional) The full path and filename for the generated tarball.
 #                If not provided, a default filename based on the current date
 #                and time will be used.
+#   --detailed   (Optional) Acquire detailed debugging information, including logs
```
bschimke95, line 16: Could you remind me, why we need a separate flag for that and not just use `--all-namespaces` everywhere? I recall there was a reason but lost it over the Christmas break. 

It feels a bit weird that we have a `--detailed` flag for an inspection report that is already supposed to collect all information.

bschimke95 (commented):
Generally LGTM, +1 on @eaudetcobello comments

k8s/scripts/inspect.sh
berkayoz: IIRC we wanted to collect all logs on CI runs however we don't want to collect all logs from the users reporting issues since they have workloads and logs might contain private information.

bschimke95: Right, I remember - could you add the reason to the comment please?

berkayoz: Also why not also rename this to `--all-namespaces` directly? Feels like the intention is clearer.

petrutlucian94:
> Thanks @petrutlucian94! Adding a flag is OK with me, but is it possible to always collect info. from all namespaces? Does this pollute the inspection report too much for the majority of cases?

There were concerns about collecting sensitive user data: https://github.com/canonical/k8s-snap/pull/892#discussion_r1888033671

tests/tics-scan.sh
petrutlucian94: whops, that's a leftover from the github workflow.

k8s/scripts/inspect.sh
petrutlucian94: My take was that a generic ``--detailed`` flag could be reused for other logs as well but I can rename it to ``--all-namespaces``.

eaudetcobello (approved):

PR #930, opened 2025-01-07
Ensure lxd is installed before attempting snap refresh

bschimke95:
- Add a check to determine if the `lxd` snap is already installed before attempting to refresh or install it. This is required because the LXD snap is not shipped by default in 24.04 anymore.

- Replace the use of `sg lxd -c` with `sudo --user "$USER"`. This eliminates the need for password entry during the `lxc` command execution.
See https://github.com/actions/runner-images/issues/9932#issuecomment-2573170305


louiseschmidtgen (approved):
Thanks Ben!

PR #932, opened 2025-01-08
Add a systemd override to add containerd defaults

### LINKED ISSUE #718: Need to manually set LimitNOFILE in the containerd service file
### Summary

I think it's related to this issue from 2019: https://github.com/containerd/containerd/issues/3201

I had some issues with both Kasten and Cilium pods reporting Too Many Open Files, when checking in a running pod the limit was showing as 1024, in order to increase this I needed to manually add an override for the snap.k8s.containerd service file and add the line `LimitNOFILE=1048576` to the `[Service]` section.

I feel like maybe this should be included in the default service file, Cilium seems to need this and that seems to be the chosen CNI.  Apologies if I've misunderstood something though :)

### Reproduction Steps

1. Bootstrap a cluster
2. Add some workloads
3. Monitor Cilium pods for errors

### System information

n/a

### Can you suggest a fix?

_No response_

_No response_

berkayoz:
Closes #718

Adds a systemd `.conf`  override file to match upstream recommendations, see https://github.com/containerd/containerd/blob/release/1.6/containerd.service for example

File is only copied over if it does not exist to make sure changes are not over-written

This approach might not work in strict as we are editing `/etc/systemd/...`

### DIFF
--- a/k8s/lib.sh
+++ b/k8s/lib.sh
@@ -219,3 +219,16 @@ k8s::util::load_kernel_modules() {
 
   modprobe $@
 }
+
+k8s::containerd::ensure_systemd_defaults() {
+  k8s::common::setup_env
+
+  local override_dir="/etc/systemd/system/snap.k8s.containerd.service.d"
+  local override_file="$SNAP/k8s/systemd/containerd-defaults.conf"
+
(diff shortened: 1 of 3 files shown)


bschimke95 (approved):
LGTM

PR #935, opened 2025-01-08
Custom bootstrap configuration Documentation

louiseschmidtgen:
## Custom bootstrap configuration Documentation

This PR adds the custom bootstrap configuration documentation which includes bootstrapping with the interactive mode and through a configuration file.

### DIFF
--- a/docs/src/snap/howto/custom-bootstrap-config.md
+++ b/docs/src/snap/howto/custom-bootstrap-config.md
@@ -1,0 +1,73 @@
+# Custom Bootstrap Configuration
+
+When creating a {{ product }} cluster that differs from the default
+configuration you can choose to use a custom bootstrap configuration.
+The CLI's interactive mode or a custom bootstrap config file allow you to 
+modify the configuration of the first node that will join your cluster. 
+
+## Configuration Options
+
+Please consult the [bootstrap-configuration-reference page] for all of the
+available configuration options and their defaults. These configuration options
+may only be adjusted on bootstrap and not after the cluster is bootstrapped.
+
+## Interactive mode
+
+The interactive mode allows for the selection of the built-in features, the pod
+CIDR and the Service CIDR.
+
+To bootstrap interactively, run:
+
+```
+sudo k8s bootstrap --timeout 10m --interactive
+```
+
+Here is an example custom configuration:
+
+```
+Which features would you like to enable? (network, dns, gateway, ingress, local-storage, load-balancer) [network, dns, gateway, local-storage]: network,ingress,dns
+Please set the Pod CIDR: [10.1.0.0/16]: 10.1.0.0/16,fd01::/108
+Please set the Service CIDR: [10.152.183.0/24]: 10.152.183.0/24,fd98::/108
+```
+
+The output for this example would be:
+
+```
+Bootstrapping the cluster. This may take a few seconds, please wait.
+Bootstrapped a new Kubernetes cluster with node address "192.168.3.117:6400".
+The node will be 'Ready' to host workloads after the CNI is deployed successfully.
+```
+
+## Bootstrap Configuration File
+
+If your deployment requires a more fine tuned configuration, use the bootstrap
+configuration file. A good starting point can be the default
+[bootstrap-config-full.yaml].
+
+For this example, create a custom bootstrap configuration file:
+
+```yaml
+cat <<EOF > bootstrap.yaml
+cluster-config:
+  network:
+    enabled: false
+EOF
+```
+
+
+Then, apply the bootstrap configuration file:
+
+```
+sudo k8s bootstrap --timeout 10m --file /path/to/bootstrap.yaml
+```
+
+To verify any changes to the built-in features run:
+
+```
+sudo k8s status
+```
+
+<!-- LINKS -->
+
+[bootstrap-configuration-reference page]: /src/snap/reference/bootstrap-config-reference.md
+[bootstrap-config-full.yaml]: https://raw.githubusercontent.com/canonical/k8s-snap/refs/heads/main/src/k8s/cmd/k8s/testdata/bootstrap-config-full.yaml
--- a/docs/src/snap/howto/custom-bootstrap-config.md
+++ b/docs/src/snap/howto/custom-bootstrap-config.md
@@ -2,8 +2,8 @@
 
 When creating a {{ product }} cluster that differs from the default
 configuration you can choose to use a custom bootstrap configuration.
-The CLI's interactive mode or a custom bootstrap config file allow you to 
-modify the configuration of the first node that will join your cluster. 
+The CLI's interactive mode or a custom bootstrap configuration file allow you
+to modify the configuration of the first node that will join your cluster. 
 
 ## Configuration Options
 
--- a/docs/src/snap/howto/index.md
+++ b/docs/src/snap/howto/index.md
@@ -18,7 +18,6 @@ install/index
 networking/index
 storage/index
 security/index
-custom-bootstrap-config
 Use an external datastore <external-datastore>
 backup-restore
 refresh-certs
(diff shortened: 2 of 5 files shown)


docs/src/snap/howto/custom-bootstrap-config.md
```diff
@@ -0,0 +1,73 @@
...
+The node will be 'Ready' to host workloads after the CNI is deployed successfully.
+```
+
+## Bootstrap Configuration File
+
+If your deployment requires a more fine tuned configuration, use the bootstrap
+configuration file. A good starting point can be the default
+[bootstrap-config-full.yaml].
+
+For this example, create a custom bootstrap configuration file:
```
nhennigan, line 47: maybe add what this custom bootstrap file is doing - file that sets network enable to false/disables the network

docs/src/snap/howto/index.md
```diff
@@ -18,6 +18,7 @@ install/index
 networking/index
 storage/index
 security/index
+custom-bootstrap-config
```
nhennigan, line 21: As this is done on cluster creation should this be under install dir?

nhennigan (commented):
Great work. :smile_cat:

berkayoz (commented):
LGTM overall, great work! A few points I'd like to bring up
* IIRC All features are disabled by default when a config file is used unless `enabled: true` is set exclusively for desired features. I feel like we should mention this somewhere here. See https://github.com/canonical/k8s-snap/blob/main/tests/integration/templates/bootstrap-all.yaml for example
* Another point is we can pass a "bootstrap file" to the `join-cluster` command as well, which is mostly used to adjust per-node config. Should we make this into another page like join config or mention it here? @nhennigan

docs/src/snap/howto/custom-bootstrap-config.md
```diff
@@ -0,0 +1,73 @@
+# Custom Bootstrap Configuration
+
+When creating a {{ product }} cluster that differs from the default
+configuration you can choose to use a custom bootstrap configuration.
+The CLI's interactive mode or a custom bootstrap configuration file allow you
+to modify the configuration of the first node that will join your cluster. 
+
+## Configuration Options
```
HomayoonAlimohammadi: Maybe?
```suggestion
## Configuration options
```

louiseschmidtgen:
> LGTM overall, great work! A few points I'd like to bring up
> 
>     * IIRC All features are disabled by default when a config file is used unless `enabled: true` is set exclusively for desired features. I feel like we should mention this somewhere here. See https://github.com/canonical/k8s-snap/blob/main/tests/integration/templates/bootstrap-all.yaml for example
> 
>     * Another point is we can pass a "bootstrap file" to the `join-cluster` command as well, which is mostly used to adjust per-node config. Should we make this into another page like join config or mention it here? @nhennigan

Just double checking: according to the bootstrap reference guide the network, dns and gateway are by default enabled and the rest disabled. Is that wrong?

docs/src/snap/howto/custom-bootstrap-config.md
```diff
@@ -0,0 +1,73 @@
...
+
+The output for this example would be:
+
+```
+Bootstrapping the cluster. This may take a few seconds, please wait.
+Bootstrapped a new Kubernetes cluster with node address "192.168.3.117:6400".
+The node will be 'Ready' to host workloads after the CNI is deployed successfully.
+```
+
+## Bootstrap Configuration File
```
HomayoonAlimohammadi: Just to be consistent
```suggestion
## Bootstrap configuration file
```

HomayoonAlimohammadi (commented):
Great work @louiseschmidtgen, thanks a lot! LGTM overall, just some minor nits.

docs/src/snap/howto/custom-bootstrap-config.md
```diff
@@ -0,0 +1,73 @@
...
+
+If your deployment requires a more fine tuned configuration, use the bootstrap
+configuration file. A good starting point can be the default
+[bootstrap-config-full.yaml].
```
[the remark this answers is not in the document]
louiseschmidtgen: I am worried about that too- but seeing that it is part of a test that tests I am quite confident that it will stay updated... Happy to remove it as well if that's the better call

nhennigan:
> * IIRC All features are disabled by default when a config file is used unless `enabled: true` is set exclusively for desired features. I feel like we should mention this somewhere here. See https://github.com/canonical/k8s-snap/blob/main/tests/integration/templates/bootstrap-all.yaml for example

I think it is important to add the line about providing a config file disabling the features by default. The user should be aware of exactly the implications of using a custom config file. 

> * Another point is we can pass a "bootstrap file" to the `join-cluster` command as well, which is mostly used to adjust per-node config. Should we make this into another page like join config or mention it here? @nhennigan
Mateo also did a similar page for the charms [here](https://documentation.ubuntu.com/canonical-kubernetes/latest/src/charm/howto/custom-workers/ ) where the worker nodes custom config join was a separate page. I liked how he had different workers configs for different workloads. It is not a massive deal whether this is two separate pages or one but I would lean towards two.

berkayoz:
Agreed @nhennigan , we should definitely mention it. I recall a user hitting this case and getting confused.

And for @louiseschmidtgen to verify that features are disabled unless specified you can bootstrap a cluster with
```yaml
pod-cidr: 10.100.0.0/16
service-cidr: 10.200.0.0/16
```
as the config file and check `k8s status`  and `k8s kubectl get all -A`.

louiseschmidtgen:
@nhennigan and @berkayoz I've corrected the default feature settings reference.  
+1 on a separate page for the join config.

docs/src/snap/howto/install/custom-bootstrap-config.md
```diff
@@ -47,19 +47,21 @@ The node will be 'Ready' to host workloads after the CNI is deployed successfull
 ## Bootstrap configuration file
 
 If your deployment requires a more fine-tuned configuration, use the bootstrap
-configuration file. By default the network, dns and gateway features
-are enabled, but the ingress, load-balancer and local-storage features are
-disabled.
+configuration file. 
 
+``` {note}
+By default all features including network, dns, gateway, ingress, load-balancer
```
nhennigan, line 53: I would add when using a custom bootstrap configuration file. They are not set disabled by default if you use the generic bootstrap

nhennigan (commented):
Great work. Small nit and then good to merge :smile:

nhennigan (approved):

docs/src/snap/howto/install/custom-bootstrap-config.md
```diff
@@ -0,0 +1,83 @@
+# Custom bootstrap configuration
+
+When creating a {{ product }} cluster that differs from the default
+configuration you can choose to use a custom bootstrap configuration.
+The CLI's interactive mode or a custom bootstrap configuration file allow you
+to modify the configuration of the first node of your cluster.
```
eaudetcobello, line 6: I wouldn't be so specific here. Technically the config. will apply to other nodes as well.

docs/src/snap/howto/install/custom-bootstrap-config.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Configuration options
+
+Please consult the [reference page] for all of the
+available configuration options and their defaults. 
+
+``` {note}
+Most of these configuration options are set during the initial bootstrapping
+and cannot be modified afterward. Runtime changes may be unsupported and
+could require deploying a new cluster. Refer to the reference page to
```
eaudetcobello, line 16: Link to the reference page

louiseschmidtgen: The link is in line 10.

docs/src/snap/howto/install/custom-bootstrap-config.md
louiseschmidtgen: There will be a page for the join-config, this page is about the bootstrap options which concern the first node. Once we have the join config page we can link from this doc there then users won't be confused.

PR #945, opened 2025-01-10
docs: Add ports and services reference for the charm

bschimke95:
The charm exposes the same services and ports on a unit as the snap, hence the reference directly links to the snap docs.


nhennigan (approved):
It's great this was an easy win. Thanks for double checking there were no differences

PR #948, opened 2025-01-10
Simplify the docs for installing via terraform

addyess:
### Overview

Ensure that the user provides a deployment channel for the charms rather than rely on a default in the charm modules. 

### Details
* stop ignoring the channel field, and make the user provide one -- one that matches for both charms
* source the tf modules from github rather than locally checked out modules
* indicates one should run `terraform plan` before running apply

### DIFF
--- a/docs/src/charm/howto/install/install-terraform.md
+++ b/docs/src/charm/howto/install/install-terraform.md
@@ -34,22 +34,37 @@ and k8s-worker child modules:
 Example `main.tf`:
 
 ```hcl
-data "juju_model" "testing" {
-  name = "juju-myk8s"
+provider "juju" {}
+
+resource "juju_model" "this" {
+  name = "juju-myk8s"  # name of the juju model
 }
+
+variable "k8s" {
+  description = "K8s deployment shared configuration"
+  channel = string
+  default = "1.xx/stable"
+}
+
 module "k8s" {
-  source = "path-to/k8s"
-  juju_model = module.juju_model.testing.name
+  source = "git::https://github.com/canonical/k8s-operator//charms/worker/k8s/terraform?ref=main"
+
+  model = juju_model.this.name
+  channel = var.k8s.channel
+  units = 3
 }
 
 module "k8s-worker" {
-  source = "path-to/k8s-worker"
-  juju_model = module.juju_model.testing.name
+  source = "git::https://github.com/canonical/k8s-operator//charms/worker/terraform?ref=main"
+  model = juju_model.this.name
+  channel = var.k8s.channel
+  units = 2
 }
 ```
 
 ```{note} 
 Please ensure that the root module references the correct source path for the `k8s` and `k8s-worker` modules.
+Also, ensure you replace the k8s.channel with {{version}}/stable
 ```
 
 Example `versions.tf`:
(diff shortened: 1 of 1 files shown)


bschimke95 (approved):
LGTM, minor nit

docs/src/charm/howto/install-terraform.md
```diff
@@ -34,22 +34,37 @@ and k8s-worker child modules:
 Example `main.tf`:
 
 ```hcl
-data "juju_model" "testing" {
-  name = "juju-myk8s"
+provider "juju" {}
+
+resource "juju_model" "this" {
+  name = "juju-myk8s"  # name of the juju model
```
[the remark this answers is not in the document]
louiseschmidtgen: Maybe a better name than `this` here? Maybe rather `my-model`?

docs/src/charm/howto/install-terraform.md
```diff
@@ -34,22 +34,37 @@ and k8s-worker child modules:
...
+}
+
+variable "k8s" {
+  description = "K8s deployment shared configuration"
+  channel = string
+  default = "1.xx/stable"
```
louiseschmidtgen, line 46: I don't quite like this variable declaration here. First it should be in an input file and second this should be closer to where it is set >> sub-module.

louiseschmidtgen (commented):
Thanks for the improvement of my docs! Some questions here:

docs/src/charm/howto/install-terraform.md
addyess: so, this `channel` likely should match both `k8s` and `k8s-worker`.  We can eliminate this from the example if you think that's a good idea and have folks go read the individual tf modules to learn what variables are optional.  

i was trying to use this variable as `var.k8s.channel` for both modules.  Did i do this incorrectly?

addyess: another thing, i wish i could use `{{channel}}` here

addyess: @louiseschmidtgen did i close the gap here or are we still an open issue.  I'm unsure how to resolve

nhennigan (commented):
From a docs perspective this looks fine. I'll let Louise give the final approval as the terraform expert

docs/src/charm/howto/install/install-terraform.md
```diff
@@ -36,35 +36,33 @@ Example `main.tf`:
...
+resource "juju_model" "my_model" {
+  name = "juju-myk8s"
 }
 
 variable "k8s" {
   description = "K8s deployment shared configuration"
-  channel = string
-  default = "1.xx/stable"
+  type        = string
 }
```
louiseschmidtgen, line 46: This should be in a separate values.tf file.

docs/src/charm/howto/install/install-terraform.md
```diff
@@ -116,10 +114,18 @@ commands:
 
 ```bash
 terraform init
+export TF_VAR_channel=<charm channel>
```
louiseschmidtgen, line 117: Nice, this works!
We should also be able to do `terraform apply -var k8s=1.xx/stable`
Not sure which is the more terraformy way- you can choose.

louiseschmidtgen (approved):
Some final words, then we're good to go:

docs/src/charm/howto/install/install-terraform.md
addyess: Done

addyess:
@louiseschmidtgen I included the charm relations as well.  I'm glad I caught these before we released. 

Here's a [pb](https://pastebin.ubuntu.com/p/xP7tx2nWWY/) of the plan with these files specifically

PR #949, opened 2025-01-10
[Docs] Clarify node-labels are additive or destructive

addyess:
### Overview
Indicate to users that the charm config `node-labels` is both additive and destructive and can be used for either purpose.

### DIFF
--- a/docs/src/charm/howto/install-custom.md
+++ b/docs/src/charm/howto/install-custom.md
@@ -36,10 +36,16 @@ k8s:
   dns-cluster-domain: "cluster.local"
   dns-upstream-nameservers: "8.8.8.8 8.8.4.4"
 
-  # Add/Remove custom node labels
-  # Using <key>=<value> one can ensure the label is added to all the nodes of this application
-  # Using <key>=-       one can ensure the label is removed from all the nodes of this application
-  node-labels: "environment=production zone=us-east-1 node-role.kubernetes.io/worker=-"
+  # Add & Remove node-labels from the snap's default labels
+  #   The k8s snap applies its default labels, these labels define what
+  #   are added or removed from those defaults
+  # <key>=<value> ensures the label is added to all the nodes of this application
+  # <key>=-       ensures the label is removed from all the nodes of this application
+  # See charm-configuration notes for more information regarding node labelling
+  node-labels: >-
+    environment=production
+    node-role.kubernetes.io/worker=-
+    zone=us-east-1
 
   # Configure local storage
   local-storage-enabled: true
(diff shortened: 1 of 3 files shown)


addyess:
This may need a troubleshooting page on its own.  Right @nhennigan ?

bschimke95 (commented):
Generally LGTM, minor comments and questions.

docs/src/charm/howto/install-custom.md
```diff
@@ -35,13 +36,16 @@ k8s:
   dns-cluster-domain: "cluster.local"
   dns-upstream-nameservers: "8.8.8.8 8.8.4.4"
 
-  # Add custom node labels
-  node-labels: "environment=production zone=us-east-1"
+  # Add/Remove custom node labels
+  # Using <key>=<value> one can ensure the label is added to all the nodes of this application
+  # Using <key>=-       one can ensure the label is removed from all the nodes of this application
+  node-labels: "environment=production zone=us-east-1 node-role.kubernetes.io/worker=-"
```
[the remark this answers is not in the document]
addyess: yes, so there are some "default labels" that the snap is adding -- not defined in this list.  Some of our customers want a way to remove those default labels... this was the only way i could think to do so.

docs/src/charm/reference/troubleshooting.md
```diff
@@ -0,0 +1,78 @@
...
+node-role.kubernetes.io/worker=
+```
+
+### Explanation
+
+Each kubernetes node comes with a set of node labels enabled by default. The k8s
+snap defaults with both control-plane and worker role labels, while the worker
+node only has a role label.
+
+For example, consider the following simply deployment with a worker and a
```
nhennigan, line 28: nit: simple

docs/src/charm/reference/troubleshooting.md
```diff
@@ -0,0 +1,78 @@
...
+control-plane.
+
+```sh
+$ sudo k8s kubectl get nodes
```
nhennigan, line 32: please remove the $

addyess: Done

docs/src/charm/reference/troubleshooting.md
addyess: nice thanks!

docs/src/charm/howto/install-custom.md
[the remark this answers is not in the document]
bschimke95: When reading the code, my initial feeling was that that if the `node-labels` are set, they would completely override the default labels. But I guess this is annoying from a UX perspective since 80% of the time users probably want to keep those labels. 

Could you add a short comment that the default labels are kept if they are not explicitly removed? Thanks!

bschimke95 (approved):
LGTM after my comment is addressed

PR #956, opened 2025-01-14
test_ingress: fix service ip poller

petrutlucian94:
get_external_service_ip repeatedly attempts to retrieve the
specified service ip.

The problem is that it uses "is None" checks to determine if
an ip was retrieved. This returns "False" for empty strings,
so the loop exits prematurely.

Note that "is None" checks are strongly discouraged in Python.

While at it, we're bumpting the timeout as well.

### DIFF
--- a/docs/src/_parts/bootstrap_config.md
+++ b/docs/src/_parts/bootstrap_config.md
@@ -11,7 +11,7 @@ Configuration options for the network feature.
 **Type:** `bool`<br>
 
 Determines if the feature should be enabled.
-If omitted defaults to `false`
+If omitted defaults to `true`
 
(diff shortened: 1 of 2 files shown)


petrutlucian94:
Had a few hundred test runs, I can confirm that this fixes the issue.

bschimke95:
Please fix the diff check, then we can merge

petrutlucian94:
> Please fix the diff check, then we can merge

Right away. Btw, the diff check compares automatically generated docs with the existing files and fails whenever there are changes in the k8s-snap-api repo. IMHO we should remove all auto-generated docs and add them to .gitignore.

PR #964, opened 2025-01-15
Add package management with helm explanation

berkayoz:
Adds the explanation page on package management with helm.

Not sure on the placement of this explanation page, where should this reside in?

### DIFF
--- a/docs/src/snap/explanation/package-management.md
+++ b/docs/src/snap/explanation/package-management.md
@@ -1,0 +1,126 @@
+# Package management with Helm
+
+There are multiple ways of installing and managing packages in Kubernetes.
+This explanation covers the popular package management tool [Helm][] and
+provides details on using Helm with {{product}}.
+
+## Installing Helm
+
+Please check out the upstream documentation for [installing Helm CLI][].
+
+## Charts
+
+[Charts][] are the packaging format used by Helm. A chart is a collection of 
+Kubernetes resource manifests related to an application or to a group of 
+applications that can be packaged into versioned archives. 
+Charts utilize templating and template variables called "values" to enable 
+customization of the resources.
+
+## Chart repository
+
+The [Chart repository][] is a web server that usually consists of an index and 
+packaged charts. Repositories are the preferred way of distributing charts.
+
+A Helm repository can be added to the local client with:
+
+```
+helm repo add <repository-name> <repository-url>
+```
+This command fetches the index from the repository and
+stores it in a cache directory.
+This index is used when doing an installation or performing an upgrade.
+
+Update the cached index regularly to get the latest list of
+available charts and versions.
+Run the following command before performing an upgrade:
+
+```
+helm repo update
+```
+
+## Installing and managing charts
+
+A [Helm installation][] consists of templates with configurable values that are
+rendered into standard Kubernetes resource manifests.
+
+User supplied values are defined in the `values.yaml` file that contains
+defaults, which can be overwritten by users at install time.
+
+View the configuration of a chart's `values.yaml`:
+
+```
+helm show values <repository-name>/<chart-name>
+```
+
+Helm charts can be installed on any Kubernetes cluster, including {{product}}
+because Helm uses a kubeconfig file just like `kubectl` to apply
+the generated manifests to the cluster.
+
+```{note}
+Retrieve a kubeconfig for your {{product}} cluster in accordance to
+your installation method.
+The kubeconfig file can be placed under the default `~/.kube/config` path.
+Alternatively, set the `--kubeconfig <path/to/kubeconfig>` flag to
+point Helm to the kubeconfig file.
+```
+
+A chart installation can be performed with:
+
+```
+helm install <release-name> <repository-name>/<chart-name> --version <version> --values <path/to/values.yaml> --namespace <namespace>
+```
+
+```{note}
+Helm will use the latest available version of a chart if
+the `--version` flag is emitted.
+```
+
+A single chart can be used to install an application multiple times.
+Each Helm installation creates a **Release**,
+which is a group of Kubernetes resources that is managed by Helm.
+The `<release-name>` is used in templates to generate unique
+Kubernetes resources so the same chart can be used for multiple installations.
+
+Existing releases under a namespace can be listed with:
+
+```
+helm list --namespace <namespace>
+```
+
+Upgrading a release with a newer version of a chart is similar
+to the installation process. An [upgrade][] can be performed with:
+
+```
+helm upgrade <release-name> <repository-name>/<chart-name> --version <version> --values <path/to/values.yaml> --namespace <namespace>
+```
+
+```{note}
+The upgrade command can also be used to change the values of an existing release without having to upgrade to a newer version.
+```
+
+Each upgrade operation creates a new release revision.
+This makes rollbacks possible in case things go wrong and
+there is a need to go back to the previous state of a release.
+
+A [rollback][] can be performed with:
+
+```
+helm rollback <release-name> <revision> --namespace <namespace>
+```
+
+A release can be [uninstalled][] from the cluster with:
+
+```
+helm uninstall <release-name> --namespace <namespace>
+```
+
+<!-- LINKS -->
+
+[Helm]: https://helm.sh/
+[installing Helm CLI]: https://helm.sh/docs/intro/install/
+[Charts]: https://helm.sh/docs/topics/charts/
+[Chart repository]: https://helm.sh/docs/topics/chart_repository/
+[helm installation]: https://helm.sh/docs/helm/helm_install/
+[upgrade]: https://helm.sh/docs/helm/helm_upgrade/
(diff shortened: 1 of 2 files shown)


docs/src/snap/explanation/package-management.md
```diff
@@ -0,0 +1,95 @@
+# Package management with Helm
+
+There are multiple ways of installing and managing packages on Kubernetes. This explanation covers the popular package management tool [Helm][] and provides details on using Helm with {{product}}.
```
HomayoonAlimohammadi, line 3: I might be terribly wrong here, but the line length is more than 80. I don't know why the linting passed, but I think it happened for me before as well (because now that I'm reviewing some of my old docs I see the line length issue).

HomayoonAlimohammadi, line 3: Just out of curiosity. Why do we have `[Helm][]`? Isn't `[Helm]` enough?

docs/src/snap/explanation/package-management.md
```diff
@@ -0,0 +1,95 @@
...
+
+```
+helm upgrade <release-name> <repository-name>/<chart-name> --version <version> --values <path/to/values.yaml> --namespace <namespace>
+```
+
+```{note}
+The upgrade command can also be used to change the values of an existing release without having to upgrade to a newer version.
+```
+
+Each upgrade operation will create a new revision for that release. This makes rollbacks possible in case things go wrong and there is a need to go back to the previous state of a release.
```
louiseschmidtgen, line 71:
```suggestion
Each upgrade operation creates a new release revision enabling to the previous state.
```

docs/src/snap/explanation/package-management.md
```diff
@@ -0,0 +1,95 @@
...
+
+A [rollback][] can be performed with:
+
+```
+helm rollback <release-name> <revision> --namespace <namespace>
+```
+
+Releases can be totally removed from the cluster with an [uninstallation][].
+
+A release can be uninstalled with:
```
louiseschmidtgen, line 81:
```suggestion
A release can be [uninstalled][] from the cluster with:
```
Let's embed the link:

docs/src/snap/explanation/package-management.md
```diff
@@ -0,0 +1,95 @@
+# Package management with Helm
```
louiseschmidtgen, line 1:
```suggestion
# Package management with helm
```
Sentence case in our header. Otherwise we need to capitalize Helm in the entire doc

louiseschmidtgen (commented):
Great work Berkay, this provides a great introduction into helm!
I've got some suggestions:

docs/src/snap/explanation/package-management.md
```diff
@@ -0,0 +1,95 @@
...
+
+`values.yaml` of a chart can be viewed with: 
+```
+helm show values <repository-name>/<chart-name>
+```
+
+Helm uses a kubeconfig file just like `kubectl` to apply the generated manifests to the cluster. This means charts can be installed to any Kubernetes cluster, including {{product}}. 
+
+```{note}
+Retrieve a kubeconfig for your {{product}} cluster in accordance to your installation method. The kubeconfig file can be placed under the default `~/.kube/config` path or `--kubeconfig <path/to/kubeconfig>` flag can be used point Helm to the kubeconfig file.
```
HomayoonAlimohammadi, line 40: Do we need to change this part?
`Retrieve a kubeconfig for your {{product}} cluster in accordance to your installation method`. It seems a bit off.

HomayoonAlimohammadi: We may also mention the `KUBECONIG` env

HomayoonAlimohammadi (commented):
I honestly enjoyed (and learned a lot) reading through this one. Very good job!
Left some minor comments and questions.

eaudetcobello:
It looks like you're missing `sudo k8s` before all the commands? `sudo k8s helm`

docs/src/snap/explanation/package-management.md
```diff
@@ -0,0 +1,95 @@
...
+[Charts][] are the packaging format used by Helm. A chart is a collection of Kubernetes resource manifests related to an application or to a group of applications that can be packaged into versioned archives. Charts utilize templating and template variables called "values" to enable customization of the resources. They can be used to deploy simple applications like a web server or something complex such as a web application and all of it's requirements. 
+
+## Chart repository
+
+[Chart repository][] is a web server that usually consists of an index and packaged charts. Repositories are the preferred way of distributing charts.
+
+A helm repository can be added to the local client with:
+
+```
+helm repo add <repository-name> <repository-url>
```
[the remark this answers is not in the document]
berkayoz: In the long term we do not want to ship/support Helm in our snap package.
Additionally this guide is applicable for all installation methods (currently placed under the snap folder but the card mentions it should be general) where `k8s <command>` does not exist/is not applicable . 

The users should install Helm by upstream instructions. We can mention and add a link to the installation page.

docs/src/snap/explanation/package-management.md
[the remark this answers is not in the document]
berkayoz: There isn't an exact "how to get your kubeconfig" page, this is usually the part of installation. Any suggestions on what to link here?

docs/src/snap/explanation/package-management.md
```diff
@@ -0,0 +1,95 @@
...
+
+```
+helm install <release-name> <repository-name>/<chart-name> --version <version> --values <path/to/values.yaml> --namespace <namespace>
+```
+
+```{note}
+Helm will install the latest available version on a repository if the `--version` flag is emitted.
+```
+
+A single chart can be used to install an application multiple times. Each helm installation creates a **Release**, which are the created Kubernetes resources that are managed by helm. `<release-name>` is used in templates to generate unique Kubernetes resources so the same chart can be used for multiple installations. Helm stores certain metadata on the cluster to maintain the state of releases.
```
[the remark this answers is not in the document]
berkayoz: The `is` changes did not make sense to me, I tried to reword it by guessing what you meant

docs/src/snap/explanation/package-management.md
berkayoz: The suggestion here seems to be incomplete? (More specifically the second part about rollbacks)

docs/src/snap/explanation/package-management.md
berkayoz: I'm not sure on this, Helm is a special name like Kubernetes. I've capitalized it in the entire doc.
I'd appreciate thoughts on this from @nhennigan

docs/src/snap/explanation/package-management.md
berkayoz: Good point, seems like I've missed this on my snap troubleshooting docs as well. Did we have a change of thoughts on this @nhennigan

berkayoz: I think they both work, I looked at existing files on how we embed links. Seems like we have both of these in the files and this is inconsistent. @nhennigan Should we decide and stick to one of these?

docs/src/snap/explanation/package-management.md
berkayoz: I'm not sure how to word that better, users would have to get a kubeconfig for their cluster. Also I don't think mentioning `KUBECONFIG` is needed, that feels out of scope for this page. (Maybe even the entire bit about getting a kubeconfig) Not sure where to draw the line between this turning into an how-to vs an explanation

docs/src/snap/explanation/package-management.md
nhennigan: Helm is a product name so it will have to be capitalized throughout the entire doc

docs/src/snap/explanation/package-management.md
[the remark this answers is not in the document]
nhennigan: +1 to keeping it generic as we do not want to ship/support Helm in our snap package. The thought here was that we often tell users to use Helm packages so this explanation page is for users who may not be familiar with package management and they can go to the Helm official page for the install etc

docs/src/snap/explanation/package-management.md
```diff
@@ -0,0 +1,94 @@
...
+
+Update the cached index regularly to get the latest list of available charts and versions. Run the following command before performing an upgrade:
+
+```
+helm repo update
+```
+
+## Installing and managing charts
+
+A [helm installation][] consists of templates with configurable values that are rendered into standard Kubernetes resource manifests.
```
nhennigan, line 28: nit: Helm installation

nhennigan (commented):
Small nit change needed but besides that its good to go :cat:

bschimke95 (approved):
LGTM, minor nits

docs/src/snap/explanation/package-management.md
nhennigan: I think we can fix the lint issues as we come across them. I will also look into the linter. It has brought up other issues so I can't see how it didn't with this page.

bschimke95 (approved):
LGTM

PR #973, opened 2025-01-16
Allow specifying service env variables

petrutlucian94:
We're adding a mechanism to specify custom service env variables:
  ``/var/snap/k8s/common/args/${service_name}-env``

Note that this feature is already supported by microk8s[1] and documented by k8s-dqlite[2].

[1] https://github.com/canonical/microk8s/blob/master/microk8s-resources/wrappers/run-k8s-dqlite-with-args#L38-L43
[2] https://github.com/canonical/k8s-snap/blob/4615c0377b85e6708ffa144b4667b2dd0fb252cf/k8s/lib.sh#L184-L194

### DIFF
--- a/k8s/lib.sh
+++ b/k8s/lib.sh
@@ -191,7 +191,7 @@ k8s::common::execute_service() {
 
   set -xe
   if [[ -f "${SNAP_COMMON}/args/${service_name}-env" ]]; then
-    declare -a env_vars="$(cat "${SNAP_COMMON}/args/${service_name}-env")"
+    mapfile -t env_vars < "${SNAP_COMMON}/args/${service_name}-env"
     exec env -S "${env_vars[@]}" "${SNAP}/bin/${service_name}" "${args[@]}"
   else
     exec "${SNAP}/bin/${service_name}" "${args[@]}"
--- a/k8s/lib.sh
+++ b/k8s/lib.sh
@@ -190,7 +190,12 @@ k8s::common::execute_service() {
   declare -a args="($(cat "${SNAP_COMMON}/args/${service_name}"))"
 
   set -xe
-  exec "${SNAP}/bin/${service_name}" "${args[@]}"
+  if [[ -f "${SNAP_COMMON}/args/${service_name}-env" ]]; then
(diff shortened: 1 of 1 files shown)


k8s/lib.sh
```diff
@@ -189,8 +189,14 @@ k8s::common::execute_service() {
   # Source arguments and substitute environment variables. Will fail if we cannot read the file.
   declare -a args="($(cat "${SNAP_COMMON}/args/${service_name}"))"
 
-  set -xe
-  exec "${SNAP}/bin/${service_name}" "${args[@]}"
+  if [[ -f "${SNAP_COMMON}/args/${service_name}-env" ]]; then
+    set -xe
```
eaudetcobello, line 193: move set -xe to outside the conditional block, since it's repeated in both branches

petrutlucian94: Right away. Tbh, I wanted to avoid having the ``if`` check under ``set -x`` so that the log output would not become too verbose.

k8s/lib.sh
```diff
@@ -190,7 +190,12 @@ k8s::common::execute_service() {
   declare -a args="($(cat "${SNAP_COMMON}/args/${service_name}"))"
 
   set -xe
-  exec "${SNAP}/bin/${service_name}" "${args[@]}"
+  if [[ -f "${SNAP_COMMON}/args/${service_name}-env" ]]; then
+    declare -a env_vars="$(cat "${SNAP_COMMON}/args/${service_name}-env")"
```
eaudetcobello, line 194: See https://ss64.com/bash/mapfile.html for an idiomatic way to do this

```
mapfile -t env_vars < "${SNAP_COMMON}/args/${service_name}-env"
```

k8s/lib.sh
eaudetcobello: No worries, it's a nit

k8s/lib.sh
petrutlucian94: It seems to have exactly the same behavior but sure, I can use that.

eaudetcobello: Yes I tested both, your solution works too.

PR #974, opened 2025-01-16
docs: annotations

eaudetcobello:
* Moves calico annotations to moonray docs
* Removes cilium annotations from moonray docs
* Adds warning about v1alpha annotations

### DIFF
--- a/docs/moonray/reference/annotations.md
+++ b/docs/moonray/reference/annotations.md
@@ -1,0 +1,174 @@
+# Annotations
+
+This page outlines the annotations that can be configured during cluster
+[bootstrap]. To do this, set the `cluster-config.annotations` parameter in
+the bootstrap configuration.
+
+For example:
+
+```yaml
+cluster-config:
+...
+    annotations:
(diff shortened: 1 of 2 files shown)


docs/moonray/reference/annotations.md
```diff
@@ -0,0 +1,174 @@
...
+
+```yaml
+cluster-config:
+...
+    annotations:
+        k8sd/v1alpha/lifecycle/skip-cleanup-kubernetes-node-on-remove: true
+        k8sd/v1alpha/lifecycle/skip-stop-services-on-remove: true
+```
+
+Please refer to the [Kubernetes website] for more information on annnotations.
```
bschimke95, line 17: I would remove that. Those are not annotations in the sense of Kubernetes (e.g. we do not "annotate" a resource).
I'd add a brief explanation in the first paragraph instead.

bschimke95 (approved):
LGTM, minor nit

docs/moonray/reference/annotations.md
addyess: can there be an explanations page for what annotations are and why they are useful/needed

PR #977, opened 2025-01-16
docs: config file reference

eaudetcobello:
* Remove redundant "reference" word from titles in files and therefore navbar
* Move config file references to their own folder, config-files.
* Reference config-files folder in the navbar.
* Use absolute path references to config files

### DIFF
--- a/docs/src/snap/reference/config-files/bootstrap-config.md
+++ b/docs/src/snap/reference/config-files/bootstrap-config.md
@@ -1,4 +1,4 @@
-# Bootstrap configuration file reference
+# Bootstrap configuration file
 
 A YAML file can be supplied to the `k8s join-cluster` command to configure and
 customise the cluster. This reference section provides the format of this file
--- a/docs/src/snap/reference/index.md
+++ b/docs/src/snap/reference/index.md
@@ -12,13 +12,11 @@ Overview <self>
 :titlesonly:
 
 releases
+config-files/index
 commands
 ports-and-services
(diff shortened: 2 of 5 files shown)


docs/src/snap/reference/index.md
```diff
@@ -12,13 +12,11 @@ Overview <self>
 :titlesonly:
 
 releases
+config-files/index
```
nhennigan, line 15: I like this grouping. Makes the navigation easier.

docs/src/snap/reference/config-files/bootstrap-config.md
```diff
@@ -18,6 +18,5 @@ and service CIDRs from the defaults and adds an extra SAN to the generated
 certificates. It is also available to download
  {download}`here </src/assets/reference-bootstrap-config.yaml>`
 
-```{literalinclude} /src/assets/reference-bootstrap-config.yaml 
-:language: yaml
+```{include} /src/assets/reference-bootstrap-config.yaml
```
nhennigan, line 21: Literalinclude was used here so that the output would be formatted as code. You can see in the rendered page that it is displayed as text. Also the earlier include statement in this file needs the path updated :smile_cat:

eaudetcobello: Gotcha, thanks

bschimke95 (approved):
LGTM after Niamh's command was addressed

PR #985, opened 2025-01-17
docs: Fix snap storage

mateoflorido:
## Overview
Include documentation updates for the snap storage guides.

### DIFF
--- a/docs/src/snap/howto/storage/ceph.md
+++ b/docs/src/snap/howto/storage/ceph.md
@@ -90,7 +90,7 @@ data:
   config.json: |-
     [
       {
-        "clusterID": "fsid 6d5c12c9-6dfb-445a-940f-301aa7de0f29",
+        "clusterID": "6d5c12c9-6dfb-445a-940f-301aa7de0f29",
         "monitors": [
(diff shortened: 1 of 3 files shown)


nhennigan (commented):
Just a quick thing, Perk mentioned that storage is now enabled by default in the storage.md file so can we update that with this PR also?

docs/src/snap/howto/storage/ceph.md
```diff
@@ -90,7 +90,7 @@ data:
   config.json: |-
     [
       {
-        "clusterID": "fsid 6d5c12c9-6dfb-445a-940f-301aa7de0f29",
+        "clusterID": "6d5c12c9-6dfb-445a-940f-301aa7de0f29",
```
eaudetcobello, line 93: Out of curiosity, why is this removed?

mateoflorido: This is a copy error, you should place the `fsid` value you got earlier in the `clusterID` value. More info: https://docs.ceph.com/en/latest/rbd/rbd-kubernetes/

nhennigan (approved):
LGTM

PR #989, opened 2025-01-20
Set dqlite failure-domain based on the k8s node AZ label

petrutlucian94:
This commit listens for node label changes and configures the dqlite failure domain based on the availability zone node label.

Dqlite adjusts the cluster node promotion order based on the failure domain, attempting to spread the quorum members over different failure domains.


petrutlucian94:
I can add some tests after the current implementation is approved.

src/k8s/pkg/k8sd/controllers/node_label.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (c *NodeLabelController) reconcileFailureDomain(ctx context.Context, node *v1.Node) error {
+	log := log.FromContext(ctx)
+
+	azLabel, azFound := node.Labels["topology.kubernetes.io/zone"]
+	if azFound {
+		log.Info("Node availability zone found", "label", azLabel)
+	} else {
+		log.Info("The node availability zone label is unset, skipping...")
```
bschimke95, line 64: What happens if a user sets a zone and then wants to remove it again? 
It looks like right now we would keep that zone on the dqlite level even after resetting it on the node label.

src/k8s/pkg/k8sd/controllers/node_label.go
```diff
@@ -0,0 +1,104 @@
...
+
+	log.Info("Waiting for node to be ready")
+	// wait for microcluster node to be ready
+	c.waitReady()
+
+	hostname := c.snap.Hostname()
+	log.Info("Starting node label controller", "hostname", hostname)
+
+	for {
+		client, err := GetNewK8sClientWithRetries(ctx, c.snap)
```
bschimke95, line 38: nit: can you update the other controllers with that new utils function as well?
Is there a reason why this function needs to be public?

src/k8s/pkg/k8sd/controllers/node_label.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return nil
+}
+
+func (c *NodeLabelController) reconcile(ctx context.Context, node *v1.Node) error {
+	log := log.FromContext(ctx)
+	log.Info("reconciling node labels", "name", node.Name)
+
+	err := c.reconcileFailureDomain(ctx, node)
+	if err != nil {
```
bschimke95, line 99: nit: this error check can be inlined

src/k8s/pkg/k8sd/controllers/utils.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/canonical/k8s/pkg/client/kubernetes"
+	"github.com/canonical/k8s/pkg/snap"
+)
+
+func GetNewK8sClientWithRetries(ctx context.Context, snapObj snap.Snap) (*kubernetes.Client, error) {
```
bschimke95, line 11: Please add a comment and make it private or is there a reason to make this function public?

src/k8s/pkg/k8sd/controllers/node_label.go
```diff
@@ -0,0 +1,104 @@
...
+		return fmt.Errorf("failed to update failure-domain, error: %w", err)
+	}
+	log.Info("Updated failure domain", "restart needed", needsRestart)
+
+	if needsRestart {
+		if err := c.snap.RestartService(ctx, "k8s-dqlite"); err != nil {
+			return fmt.Errorf("failed to restart k8s-dqlite to apply failure domain: %w", err)
+		}
+
+		// TODO: k8sd restarts itself, is it safe to do so?
```
bschimke95, line 84: I think it is fine.

src/k8s/pkg/snap/util/dqlite.go
```diff
@@ -0,0 +1,72 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/k8s/pkg/utils"
+)
+
+// Checksum (sha256) node labels to produce uint64 failure domain identifiers.
+func NodeLabelToDqliteFailureDomain(label string) uint64 {
```
bschimke95, line 16: nit: please stick to godoc conventions here and elsewhere
`// NodeLabelToDqliteFailureDomain checks ...`

src/k8s/pkg/k8sd/controllers/utils.go
petrutlucian94, line 11: TODO: get the other controllers to use this helper, removing duplicate code.

src/k8s/pkg/snap/util/dqlite.go
```diff
@@ -0,0 +1,72 @@
...
+	sha256Sum := sha256.Sum256([]byte(label))
+	// Select the first 8 bytes of the sha256 hash.
+	return binary.LittleEndian.Uint64(sha256Sum[:])
+}
+
+func UpdateDqliteFailureDomain(snap snap.Snap, failureDomain uint64) (bool, error) {
```
bschimke95, line 22: I think we should move this logic on the controller level. 
We basically just hide implementation without reducing the complexity.

For example, a bool is returned to indicate that *any* dqlite domain is modified. We don't know which one so the controller will restart both services.

bschimke95 (commented):
Great work @petrutlucian94 - did a first round, mainly focused on overall structure.

src/k8s/pkg/k8sd/controllers/node_label.go
petrutlucian94: Sure, I'll update the other controllers in the following commit.

Good point, the helper doesn't have to be public.

src/k8s/pkg/k8sd/controllers/utils.go
petrutlucian94: No, we can make it private.

src/k8s/pkg/k8sd/controllers/node_label.go
petrutlucian94: I've updated it to clear the failure domain if the label is removed.

src/k8s/pkg/k8sd/controllers/node_label.go
```diff
@@ -0,0 +1,130 @@
...
+		// need to be updated (k8sd and k8s-dqlite).
+		failureDomain = snaputil.NodeLabelToDqliteFailureDomain(azLabel)
+	} else {
+		log.Info("The node availability zone label is unset, clearing failure domain")
+		failureDomain = 0
+	}
+
+	log.Info("Setting failure domain", "failure domain", failureDomain, "availability zone", azLabel)
+	err := c.updateDqliteFailureDomain(ctx, c.snap, failureDomain)
+	if err != nil {
```
berkayoz, line 76: nit: Inline err check

src/k8s/pkg/snap/util/dqlite.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return false, err
+	}
+	if existingFailureDomain == failureDomain {
+		// Failure domain already set.
+		return false, nil
+	} else {
+		// Updating failure domain.
+		err := os.WriteFile(failureDomainFile, []byte(failureDomainStr), 0o644)
+		if err != nil {
```
berkayoz, line 79: nit: Inline err check

berkayoz (approved):
Great work, LGTM. Left small nits

PR #990, opened 2025-01-20
implement timeout in inspect.sh for k8s commands

eaudetcobello:

PR #991, opened 2025-01-20
misc inspection script cleanup

eaudetcobello:
* Declare + assign to a variable when there is no risk doing so (https://www.shellcheck.net/wiki/SC2155)
* Run shell-format
* Replace leftover magic number with NUM_SNAP_LOG_ENTRIES
* Add report directory location message to stdout
* Use only one log_warning_red when a private key is detected

Depends on: https://github.com/canonical/k8s-snap/pull/990


bschimke95 (approved):
LGTM

PR #992, opened 2025-01-20
Retry Temporary API Failures in Microcluster

bschimke95:
When bootstrapping or starting a cluster, we wait for the k8sd server to be fully ready before interacting with it. However, there are edge cases—such as during a snap refresh—where the snap attempts to interact with the CLI (e.g., to configure snap settings) while the database is still initializing.

In these scenarios, immediate failure is unnecessary. The k8sd client now retries such requests, ensuring smoother operation.

This behavior applies only to specific edge cases where it is known that the microcluster database will eventually become available.

### DIFF
--- a/src/k8s/pkg/client/k8sd/query.go
+++ b/src/k8s/pkg/client/k8sd/query.go
@@ -6,14 +6,40 @@ import (
 	"strings"
 
 	apiv1 "github.com/canonical/k8s-snap-api/api/v1"
+	"github.com/canonical/k8s/pkg/log"
+	"github.com/canonical/k8s/pkg/utils/control"
 	"github.com/canonical/lxd/shared/api"
 )
 
-// query is a helper method to wrap common error checking and response handling.
-func query[T any](ctx context.Context, c *k8sd, method string, path string, in any, out *T) (T, error) {
-	if err := c.client.Query(ctx, method, apiv1.K8sdAPIVersion, api.NewURL().Path(strings.Split(path, "/")...), in, out); err != nil {
-		var zero T
-		return zero, fmt.Errorf("failed to %s /%s: %w", method, path, err)
+// query is a helper method for sending requests to the k8sd client with common error checking and automatic retries.
+// It retries on temporary microcluster errors and returns the deserialized response.
+func query[T any](ctx context.Context, c *k8sd, method, path string, in any, out *T) (T, error) {
+	var result T
+
+	retryErr := control.WaitUntilReady(ctx, func() (bool, error) {
+		err := c.client.Query(ctx, method, apiv1.K8sdAPIVersion, api.NewURL().Path(strings.Split(path, "/")...), in, out)
+		if err != nil {
+			if isTemporary(err) {
+				log.FromContext(ctx).Info("Temporary error from k8sd: %v", err)
+				return false, nil
+			}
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/client/k8sd/query.go
```diff
@@ -6,14 +6,37 @@ import (
...
-// query is a helper method to wrap common error checking and response handling.
-func query[T any](ctx context.Context, c *k8sd, method string, path string, in any, out *T) (T, error) {
-	if err := c.client.Query(ctx, method, apiv1.K8sdAPIVersion, api.NewURL().Path(strings.Split(path, "/")...), in, out); err != nil {
-		var zero T
-		return zero, fmt.Errorf("failed to %s /%s: %w", method, path, err)
+// query is a helper method for sending requests to the k8sd client with common error checking and automatic retries.
+// It retries on temporary microcluster errors and returns the deserialized response.
+func query[T any](ctx context.Context, c *k8sd, method, path string, in any, out *T) (T, error) {
+	var result T
+
```
mateoflorido, line 18: From what I've seen, we usually provide the structs where we unmarshal the response in `out`. I think this might be a minor nit, but it could be worthwhile to add a check to validate if the generic `out` pointer is nil and return early, instead of panicking down the road:
```suggestion
func query[T any](ctx context.Context, c *k8sd, method, path string, in any, out *T) (T, error) {
	var result T
	if out == nil {
		return result, fmt.Errorf("out must be a non-nil pointer")
	}
```

src/k8s/pkg/client/k8sd/query.go
```diff
@@ -6,14 +6,37 @@ import (
...
+	retryErr := control.WaitUntilReady(ctx, func() (bool, error) {
```
eaudetcobello, line 19: Should a timeout/max number of retries be introduced here?

src/k8s/pkg/client/k8sd/query.go
```diff
@@ -6,14 +6,37 @@ import (
...
+				log.FromContext(ctx).Info("Temporary error from k8sd: %v", err)
+				return false, nil
+			}
+			return false, fmt.Errorf("failed to %s /%s: %w", method, path, err)
+		}
+		return true, nil
+	})
+
+	if retryErr != nil {
+		return result, fmt.Errorf("failed after potential retry: %w", retryErr)
```
eaudetcobello, line 32: The error message could be improved. I'm not sure what you mean by "potential retry"

src/k8s/pkg/client/k8sd/query.go
bschimke95: the timeout/max number of retries is handled through the ctx. If the context is cancelled, the loop will exit.
In general, you don't want to have arbitrary timeouts and retries in the code. You may remember the hassle we had because microcluster introduces random 30s timeouts in their codebase.

src/k8s/pkg/client/k8sd/query.go
bschimke95: I couldn't really come up with a better way of wrapping that error. The actual issue is already good described on line 29. 
In this line, we basically just return that error we got from the retry function. However, we don't know if we saw a retry error before or just directly exited.
Do you have a good suggestion on what error message you'd like to see here?

berkayoz (approved):
LGTM, I wonder if we can introduce a mock test for this later on

PR #996, opened 2025-01-23
docs: Bump GitPython to 3.1.41 to fix CVEs

HomayoonAlimohammadi:
### Overview
Fixes: https://github.com/canonical/k8s-snap/security/code-scanning/128
This PR puts a >=3.1.41 constraint on the `GitPython` requirement which is suggested by the scanner.
The mentioned scan contains multiple CVEs, all pointing to the `GitPython` dependency and each one points to a different issue that is fixed in a different version of `GitPython`. The highest version (that should contain all the fixes) is `3.1.41`.


bschimke95 (approved):
LGTM

PR #1002, opened 2025-01-23
Don't remove Kubernetes certificates and containerd if skipped

claudiubelu:
Currently, if the `k8sd/v1alpha/lifecycle/skip-stop-services-on-remove` annotation is set, we're not stopping the Kubernetes-related services, but we're still removing its certificates and containerd-related paths. This will end up paralyzing services like kubelet, which might have to do Pod evictions, blocking it from finishing its job, and resulting CAPI not being able to complete its downscaling or upgrade operations.

We should remove those certificates only if we're also stopping the services.

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -115,14 +115,6 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 		}
 	}
 
-	// Perform all cleanup steps regardless of if this is a worker node or control plane.
-	// Trying to detect the node type is not reliable as the node might have been marked as worker
-	// or not, depending on which step it failed.
-	log.Info("Cleaning up worker certificates")
-	if _, err := setup.EnsureWorkerPKI(snap, &pki.WorkerNodePKI{}); err != nil {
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -115,27 +115,30 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
 		}
 	}
 
-	log.Info("Cleaning up control plane certificates")
-	if _, err := setup.EnsureControlPlanePKI(snap, &pki.ControlPlanePKI{}); err != nil {
-		log.Error(err, "failed to cleanup control plane certificates")
-	}
-
+	// NOTE(claudiub): If we don't stop the Kubernetes services, we shouldn't remove the certificates
+	// either. Removing them will result in the Kubernetes services being paralyzed and unable to
```
eaudetcobello, line 126: Can you please change the double negative here? If we stop the kubernetes services, we should...

eaudetcobello (approved):
Good catch. Please address the comment before merge.

src/k8s/pkg/k8sd/app/hooks_remove.go
claudiubelu: Done

PR #1006, opened 2025-01-24
[Docs] Move two-node HA page to /moonray

nhennigan:
The two node high availability is more of a proof of concept rather than a fully supported feature. This was done at request for moonray so moving there.


bschimke95 (approved):
LGTM

PR #1007, opened 2025-01-27
Switch to cilium native routing mode for ipv6 only setup

berkayoz:
Tunneling/encapsulation is not supported for ipv6 only clusters https://github.com/cilium/cilium/issues/17240 .
This PR configures Cilium to run in native routing mode on ipv6 only setups.


bschimke95 (approved):
LGTM

PR #1012, opened 2025-01-27
The command "kubectl config show" does not exist

WizardBit:
The command "kubectl config show" does not exist and should be "kubectl config view".

```
kubectl config show
error: unknown command "show"
See 'kubectl config -h' for help and examples
```


eaudetcobello (approved):
Thanks!

PR #1014, opened 2025-01-28
Generate and collect core dumps

petrutlucian94:
We need core dumps in order to effectively debug crashes, especially the ones caused by external C libraries such as dqlite.

The service start wrapper will now set ``GOTRACEBACK=crash`` and adjust the core dump limit.

We'll also need debug symbols in order to make use of the collected core dumps, which is why we'll include debug symbols even for release builds. This increases the snap size from 107MB to 135MB, which seems reasonable.

``inspect.sh`` as well as the e2e tests are updated to collect the core dumps. The core dump location defaults to /var/crash.


louiseschmidtgen (approved):
LGTM

PR #1015, opened 2025-01-28
Log a message if the cluster is unitialized

petrutlucian94:
Users coming from microk8s may not be used to having to bootstrap the cluster.

We'll check k8sd errors and if the message contains "Database is not yet initialized", we'll ask the users to either bootstrap a new cluster or join an existing one.

We're adding this check to the query function of the k8sd client so that all the k8s commands may benefit from it.

Implements: KU-2481


bschimke95 (approved):
LGTM

PR #1016, opened 2025-01-28
Add k8s inspect command

petrutlucian94:
We're adding a "k8s inspect" command that will invoke the "inspect.sh" script, aiming to improve the user experience.

Note that we'll avoid parsing the arguments twice since that's unnecessary and would complicate the process of adding new parameters.


eaudetcobello (approved):
Thanks!

PR #1028, opened 2025-01-31
inspect.sh: avoid logging an error if there are no core dumps

petrutlucian94:
The inspect.sh script will log an error if the core dump dir is empty. We'll add a check to improve the user experience.

   INFO:  Copy dmesg entries
   INFO:  Collecting core dumps from /var/crash. Size: 4.0K    /var/crash
   cp: cannot stat '/var/crash/*': No such file or directory
   Collecting snap and related information


eaudetcobello (approved):
Replace `` with $() and then lgtm

PR #1032, opened 2025-02-03
Enrich node configuration controller tests with signed configmaps

rapour:
This PR suggests some more tests cases to include scenarios where configmaps are signed

### DIFF
--- a/src/k8s/pkg/k8sd/controllers/node_configuration_test.go
+++ b/src/k8s/pkg/k8sd/controllers/node_configuration_test.go
@@ -60,62 +60,6 @@ func TestConfigPropagation(t *testing.T) {
 			expectRestart: true,
 		},
 		{
-			name: "InitialWithSignature",
-			configmap: &corev1.ConfigMap{
-				ObjectMeta: metav1.ObjectMeta{Name: "k8sd-config", Namespace: "kube-system"},
-				Data: map[string]string{
-					"cluster-dns":    "10.152.1.1",
-					"cluster-domain": "test-cluster.local",
-					"cloud-provider": "provider",
-				},
-			},
-			expectArgs: map[string]string{
-				"--cluster-dns":    "10.152.1.1",
-				"--cluster-domain": "test-cluster.local",
-				"--cloud-provider": "provider",
-			},
-			privKey:       privKey,
-			pubKey:        &privKey.PublicKey,
-			expectRestart: true,
-		},
-		{
-			name: "InitialExpectedSignature",
-			configmap: &corev1.ConfigMap{
-				ObjectMeta: metav1.ObjectMeta{Name: "k8sd-config", Namespace: "kube-system"},
-				Data: map[string]string{
-					"cluster-dns":    "10.152.1.1",
-					"cluster-domain": "test-cluster.local",
-					"cloud-provider": "provider",
-				},
-			},
-			expectArgs: map[string]string{
-				"--cluster-dns":    "10.152.1.1",
-				"--cluster-domain": "test-cluster.local",
-				"--cloud-provider": "provider",
-			},
-			pubKey:        &privKey.PublicKey,
-			expectRestart: false,
-		},
-		{
-			name: "InitialInvalidSignature",
-			configmap: &corev1.ConfigMap{
-				ObjectMeta: metav1.ObjectMeta{Name: "k8sd-config", Namespace: "kube-system"},
-				Data: map[string]string{
-					"cluster-dns":    "10.152.1.1",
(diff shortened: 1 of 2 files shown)


src/k8s/pkg/k8sd/controllers/node_configuration_test.go
```diff
@@ -110,6 +120,62 @@ func TestConfigPropagation(t *testing.T) {
...
+			expectArgs: map[string]string{
+				"--cluster-dns":    "10.152.1.1",
+				"--cluster-domain": "test-cluster.local",
+				"--cloud-provider": "provider",
+			},
+			pubKey:        &privKey.PublicKey,
+			expectRestart: false,
+		},
+		{
+			name: "InvalidSignature",
```
bschimke95, line 161: An error is implicitly indicated by `expectedRestart` == false, right?

src/k8s/pkg/k8sd/controllers/node_configuration_test.go
```diff
@@ -110,6 +120,62 @@ func TestConfigPropagation(t *testing.T) {
...
+				"--cluster-dns":    "10.152.1.1",
+				"--cluster-domain": "test-cluster.local",
+				"--cloud-provider": "provider",
+			},
+			privKey:       privKey,
+			pubKey:        &privKey.PublicKey,
+			expectRestart: true,
+		},
+		{
+			name: "ExpectedSignature",
```
bschimke95, line 143: nit: I was wondering at first where the difference to `WithSignature` is. Maybe `MissingPrivKey` better expresses the test intent.

src/k8s/pkg/k8sd/controllers/node_configuration_test.go
```diff
@@ -139,8 +206,18 @@ func TestConfigPropagation(t *testing.T) {
...
+				kubelet, err := types.KubeletFromConfigMap(tc.configmap.Data, nil)
+				g.Expect(err).To(Not(HaveOccurred()))
+
+				tc.configmap.Data, err = kubelet.ToConfigMap(tc.privKey)
+				g.Expect(err).To(Not(HaveOccurred()))
+			}
+
 			watcher.Add(tc.configmap)
 
+			keyCh <- tc.pubKey
```
bschimke95, line 219: What happens if tc.pubKey is nil? Do we need to guard this call?

src/k8s/pkg/k8sd/controllers/node_configuration_test.go
```diff
@@ -139,8 +206,18 @@ func TestConfigPropagation(t *testing.T) {
...
+
 			// TODO: this is to ensure that the controller has handled the event. This should ideally
 			// be replaced with something like a "<-sentCh" instead
 			time.Sleep(100 * time.Millisecond)
```
bschimke95, line 223: I don't have much context for this test. Do we still need this and how much work would it be to implement this?

bschimke95 (commented):
Great work @rapour 
I added a couple of comments but it already looks very good.

src/k8s/pkg/k8sd/controllers/node_configuration_test.go
rapour: The previous tests don't include signatures but the new ones does. Since we only provide the `getRSAKey` argument once, this channel is used to return `nil` when the test case doesn't have any signatures (and corresponding keys) and return the `tc.pubkey` when the case does have one. So, I think it should not cause any troubles as `tc.pubkey` is already `nil` for the old test cases.

src/k8s/pkg/k8sd/controllers/node_configuration_test.go
rapour: I think we are stuck with `time.Sleep` because we don't have any handles to capture `NodeConfigurationController` errors. We can implement a callback-ish mechanism to see the actual results of each control loop. Something like: 
```
func (c *NodeConfigurationController) Run(ctx context.Context, getRSAKey func(context.Context) (*rsa.PublicKey, error), callback(err error)) {
...
if err := client.WatchConfigMap(...); err != nil {
			// This also can fail during bootstrapping/start up when api-server is not ready
			// So the watch requests get connection refused replies
   			callback(err)
                        log.WithValues(...)
		}
...
}
```
And then capture the error in the tests.

bschimke95: ok, thanks for the context. Sounds like a separate problem to address in another PR then.

bschimke95 (approved):
LGTM, great work @rapour!

PR #1034, opened 2025-02-03
tests: collect all inspection reports before cleaning up nodes

petrutlucian94:
This change will collect all the inspection reports before initiating the node cleanup process. Otherwise we interfere with the observed cluster, potentially breaking it, which can impede the debugging process.


bschimke95 (approved):
LGTM

PR #1037, opened 2025-02-04
Remove obsolete loadbalancer feature check

bschimke95:
Historically, we used the Cilium loadbalancer for the loadbalancer feature. With the move to Metallb, the dependency to the network feature is no longer required.


eaudetcobello (approved):
lgtm

PR #1044, opened 2025-02-05
Downgrade k8s-dqlite (1.4.0 -> 1.3.0 and microcluster)

petrutlucian94:
A recent k8s-dqlite PR (https://github.com/canonical/k8s-dqlite/pull/212) seems to have introduced a regression. The kubelet service is now missing pod events and cilium hangs in "PodScheduled" state.

We'll switch back to k8s-dqlite 1.3.0 for now.

Note that the microcluster dependency has to be downgraded as well, otherwise we're hitting panics.


petrutlucian94:
We decided not to downgrade k8s-dqlite but instead wait for a fix.

PR #1046, opened 2025-02-06
Fix custom containerd paths

petrutlucian94:
For some reason, there are two almost identical kubelet containerd flags and we have to set both:

```
$ snap logs k8s.kubelet -n 30000 | grep FLAG | grep containerd
      FLAG: --container-runtime-endpoint="/home/ubuntu/containerd/k8s-containerd/run/containerd/containerd.sock"
      FLAG: --containerd="/run/containerd/containerd.sock"
      FLAG: --containerd-namespace="k8s.io"
```

```
$ kubelet -h | grep containerd
      --container-runtime-endpoint string    The endpoint of container runtime service.
      --containerd string                    containerd endpoint
```

This change will:

* pass the missing containerd flag
* update the e2e test for custom containerd paths to check if the cluster actually becomes available after bootstrap
* update the dev doc to enable the net, dns and local storage features
  * do the same for the e2e test


HomayoonAlimohammadi (approved):
Great work, thanks a lot @petrutlucian94! LGTM

PR #1053, opened 2025-02-07
test: Remove --maxfail 1 from integration tests

bschimke95:
The CI takes some time to run. We would like to catch all failing tests in one run so that they can be fixed all individually before doing another CI run.

### DIFF
--- a/tests/integration/tox.ini
+++ b/tests/integration/tox.ini
@@ -34,7 +34,6 @@ deps =
     -r {toxinidir}/requirements-test.txt
 commands =
     pytest -vv \
-        --maxfail 1 \
         --tb native \
         --log-cli-level DEBUG \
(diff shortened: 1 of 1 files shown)


tests/integration/tox.ini
```diff
@@ -34,7 +34,6 @@ deps =
     -r {toxinidir}/requirements-test.txt
 commands =
     pytest -vv \
-        --maxfail 1 \
         --tb native \
         --log-cli-level DEBUG \
         --log-format "%(asctime)s %(levelname)s %(message)s" \
```
addyess, line 39: ok, so this will direct pytest to just run all the tests.  This will be nice to see all the failures created by this pr rather than just the first one.  It should help us change more of the code on the next commit.

addyess (approved):
LGTM

PR #1058, opened 2025-02-11
actions: Adds backport github action

claudiubelu:
Sets default action parameters explicitly.

In order to backport a PR, the PR should have the an appropriate label: ``"backport branch-name"`` (e.g.: ``"backport release-1.32"``). When the PR merges, the github action will automatically backport it. The label can also be added after the PR merges, after which a user should comment ``/backport`` in order to trigger the action.

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -1,0 +1,66 @@
+name: Backport merged pull request
+on:
+  pull_request_target:
+    types: [closed]
+  issue_comment:
+    types: [created]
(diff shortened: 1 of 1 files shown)


bschimke95 (commented):
Generally LGTM. It would be great if we can automatically label the backport PRs with "automerge" so that our automerger CI will merge them automatically.

I don't see an option for that in the action but we can probably just add a separate step for that.

bschimke95 (approved):
LGTM!

PR #1064, opened 2025-02-12
Docs: add to charm troubleshooting 

nhennigan:
A common issue faced by users is the fact bridge interfaces are not automatically picked up by Cilium for direct routing mode. Adding to the troubleshooting page how to add the bridge interfaces to the device list.


addyess (approved):
I guess it's a "work-around".  This isn't a good "charmed" way of doing things.  It'd be nicer if the charm knew it was running on maas and could do this for you.

PR #1074, opened 2025-02-14
Fix auto-merge script

bschimke95:
There where two issues:
1. "neutral" test runs (e.g. Trivy) do not "pass" but are marked as "skipped"
2. All PRs must be approved


addyess (approved):
Why are skipped tests okay?  What causes a test to be skipped?

We might want to adapt the callable workflows [here](https://github.com/canonical/k8s-workflows/blob/main/.github/workflows/auto-merge-successful-prs.yaml) 
but make adjustments to it if the caller wants to allow for skipping tests

addyess:
well, disregard that "called" workflow -- it ignores "skipping" tests. 

the callable workflow looks for unready test if they are failed, pending, or cancelled.  That means pass and skipping are okay
```python
        any_unready = (
            len(pr.checks.failed) or len(pr.checks.pending) or len(pr.checks.cancel)
        )
```

PR #1080, opened 2025-02-17
Fix test_gateway retries

petrutlucian94:
test_gateway performs a few retries, waiting for the service IPs to be set.

The problem is that the checks use "is None" syntax, which won't properly handle empty strings. For this reason, the retry loop stops prematurely, causing test flakiness.

```
Execute command ['curl', '', '-H', 'Host: foo.bar.com'] in instance k8s-integration-a6f840-37
```

In general "is None" / "is not None" checks should be avoided. We'll update the k8s-snap tests accordingly.

### DIFF
--- a/tests/branch_management/tests/test_branches.py
+++ b/tests/branch_management/tests/test_branches.py
@@ -25,7 +25,7 @@ def _sh(*args, **kwargs):
 
 def _branch_flavours(branch: str = None):
     patch_dir = Path("build-scripts/patches")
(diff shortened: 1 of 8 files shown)


petrutlucian94:
Another dqlite failure, retrying.

```
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed to allocate a serviceIP: rpc error: code = Unknown desc = exec (try: 500): database is locked
```

PR #1085, opened 2025-02-19
bug: reject unknown bootstrap arguments

petrutlucian94:
We're updating the "k8s bootstrap" command to throw an error when receiving unknown positional arguments.

For example, when passing a bootstrap config as a positional argument instead of a named argument, the command will now error out instead of ignoring the parameter, which can be confusing to end users.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_bootstrap.go
+++ b/src/k8s/cmd/k8s/k8s_bootstrap.go
@@ -47,6 +47,7 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 		Short:  "Bootstrap a new Kubernetes cluster",
 		Long:   "Generate certificates, configure service arguments and start the Kubernetes services.",
 		PreRun: chainPreRunHooks(hookRequireRoot(env), hookInitializeFormatter(env, &opts.outputFormat), hookCheckLXD()),
+		Args:   cobra.NoArgs,
 		Run: func(cmd *cobra.Command, args []string) {
 			snapdClient, err := snapd.NewClient()
 			if err != nil {
--- a/src/k8s/cmd/k8s/k8s_config.go
+++ b/src/k8s/cmd/k8s/k8s_config.go
@@ -18,6 +18,7 @@ func newKubeConfigCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 		Use:    "config",
 		Hidden: true,
 		Short:  "Generate an admin kubeconfig that can be used to access the Kubernetes cluster",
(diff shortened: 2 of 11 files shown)


src/k8s/cmd/k8s/k8s_bootstrap.go
```diff
@@ -47,6 +47,7 @@ func newBootstrapCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 		Short:  "Bootstrap a new Kubernetes cluster",
 		Long:   "Generate certificates, configure service arguments and start the Kubernetes services.",
 		PreRun: chainPreRunHooks(hookRequireRoot(env), hookInitializeFormatter(env, &opts.outputFormat), hookCheckLXD()),
+		Args:   cobra.NoArgs,
```
HomayoonAlimohammadi, line 50: Maybe we can set this for other k8s commands that don't require any argument as well?

HomayoonAlimohammadi (commented):
Nice work, thanks a lot @petrutlucian94! Maybe we can also have this (or similar ones) for other commands?
Maybe even some `cobra.OnlyValidArgs` if necessary?

src/k8s/cmd/k8s/k8s_bootstrap.go
petrutlucian94: sure, I'll add that

petrutlucian94:
I've added ``NoArgs`` to all the commands that don't take positional arguments. ``cobra.OnlyValidArgs`` is not applicable since we aren't using ``ValidArgs``, the commands that do use positional arguments already perform the necessary validation.

petrutlucian94:
no further output in 20 minutes:

```
2025-02-19 09:05:42 INFO Waiting for model:
  k8s/0 [executing] waiting: Unready Pods: kube-system/cilium-4dx6j, kube-system/ck-storage-rawfile-csi-node-5m5x9, kube-system/ck-storage-rawfile-csi-node-cvjmj
  k8s/1 [executing] waiting: Node not Ready
  k8s/2 [idle] waiting: Unready Pods: kube-system/cilium-4dx6j, kube-system/ck-storage-rawfile-csi-node-5m5x9, kube-system/ck-storage-rawfile-csi-node-cvjmj
```

src/k8s/cmd/k8s/k8s_bootstrap.go
bschimke95: I think it is rather weird that we explicitly need to set this in Cobra.
Anyway, thanks for the fix!

bschimke95 (approved):
LGTM

PR #1086, opened 2025-02-19
WIP: Adds cilium 1.17.1 rocks

claudiubelu:


src/k8s/pkg/k8sd/app/util.go
```diff
@@ -0,0 +1,89 @@
...
+	"net"
+	"slices"
+)
+
+// getIPv46Addresses returns an IPv4 and IPv6 pair if possible, from the local network interface which has the
+// given IP address.
+// getIPv46Addresses will return an array containing only the given IP address if the pair could not be found.
+// If the given IP address cannot be found locally, an error will be returned.
+func getIPv46Addresses(ip net.IP) ([]net.IP, error) {
+	ifaces, err := net.Interfaces()
```
eaudetcobello, line 14: Please add testing for this function.

src/k8s/pkg/k8sd/app/util.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get local interfaces: %w", err)
+	}
+
+	isIP := func(ifaceIP net.IP) bool {
+		return ip.Equal(ifaceIP)
+	}
```
eaudetcobello, line 21: I can't figure out why this variable is called isIP? Did you mean isIPEqualTo or something?

src/k8s/pkg/k8sd/app/util.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, iface := range ifaces {
+		addrs, err := iface.Addrs()
+		if err != nil {
+			// Continue, we may still find the right interface.
+			// Return lastError if we don't find it.
+			lastError = fmt.Errorf("Failed to get local interface addresses: %w", err)
+			continue
+		}
+
+		ips, err := parseIPAddresses(addrs)
```
eaudetcobello, line 33: Maybe rename this to ifaceIPs

src/k8s/pkg/k8sd/app/util.go
```diff
@@ -0,0 +1,89 @@
...
+			// Continue, we may still find the right interface.
+			// Return lastError if we don't find it.
+			lastError = err
+			continue
+		}
+
+		// Check if the given IP is in the list.
+		if !slices.ContainsFunc(ips, isIP) {
+			continue
+		}
```
eaudetcobello, line 44: Declare isIP (renamed) above this block, instead of at the beginning of the function. Prefer declaring close to use.

src/k8s/pkg/k8sd/app/util.go
eaudetcobello, line 14: cyclomatic complexity of this function is quite high. is it possible to reduce nesting?

src/k8s/pkg/k8sd/app/util.go
```diff
@@ -0,0 +1,89 @@
...
+
+		// Return a [IPv4, IPv6] pair if possible.
+		givenIPv6 := ip.To4() == nil
+		for _, ifIP := range ips {
+			isIPv6 := ifIP.To4() == nil
+			if isIPv6 == givenIPv6 {
+				// same type, ignore.
+				continue
```
eaudetcobello, line 52: givenIPv6 is not a clear name. Consider this:
	1.	“The IPv6 we were given” (suggesting givenIPv6 holds an IPv6 address or some related data).
	2.	“Were we given an IPv6?” (suggesting givenIPv6 is a boolean indicating whether we received an IPv6 address).

claudiubelu: Right. It is a bit confusing,  but it's basically option 2, as the documentation states that ``ip.To4()`` returns `nil` if the IP is IPv6.

Simplified a bit, it should make things a bit clearer.

src/k8s/pkg/k8sd/app/util.go
claudiubelu: Done

src/k8s/pkg/k8sd/app/util.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"fmt"
+	"net"
+	"slices"
+)
+
+// getIPv46Addresses returns an IPv4 and IPv6 pair if possible, from the local network interface which has the
+// given IP address.
+// getIPv46Addresses will return an array containing only the given IP address if the pair could not be found.
+// If the given IP address cannot be found locally, an error will be returned.
```
eaudetcobello, line 12: Could you add to the documentation (or comments) that this looks through all IPs in the interface's range and finds the first global unicast (routable?) IPv6 IP?

eaudetcobello (approved):
LGTM. Please address documentation comment

src/k8s/pkg/k8sd/app/util.go
claudiubelu: Done

src/k8s/pkg/k8sd/app/util.go
```diff
@@ -0,0 +1,93 @@
+package app
```
berkayoz, line 1: We already have some networking utils in https://github.com/canonical/k8s-snap/tree/main/src/k8s/pkg/utils , wouldn't it be better to put these  there?

src/k8s/pkg/k8sd/setup/kubelet.go
```diff
@@ -80,8 +80,15 @@ func kubelet(snap snap.Snap, hostname string, nodeIP net.IP, clusterDNS string,
 	if clusterDomain != "" {
 		args["--cluster-domain"] = clusterDomain
 	}
-	if nodeIP != nil && !nodeIP.IsLoopback() {
-		args["--node-ip"] = nodeIP.String()
+
+	ips := []string{}
+	for _, nodeIP := range nodeIPs {
+		if !nodeIP.IsLoopback() {
```
berkayoz, line 86: Did we check if this is okay with using `127.0.0.1` as the node address? https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/install/dev-env/#changing-ip-addresses

berkayoz (commented):
LGTM, great work! Left some comments and a question,

src/k8s/pkg/k8sd/app/util.go
claudiubelu: Done

src/k8s/pkg/k8sd/setup/kubelet.go
claudiubelu: Right. So, functionally, I'm not changing anything here with this patch: even before, if the given node IP was a loopback address, the ``--node-ip`` would not be passed, allowing it to take its default value (according to the documentation, if not provided, "the kubelet uses the node's default IPv4 address, if any; if the node has no IPv4 addresses then the kubelet uses the node's default IPv6 address.". This remains the same with this patch.

Now, regarding the documentation link. I think it mostly applies to ``kube-apiserver``, allowing it to still be accessible / usable even if the local IP address changed (e.g.: moved to another WIFI network). From what I've experimented, we don't need to touch anything here for kubelet however. In my env, I have the following ``kubeconfig`` (ran: ``sudo k8s bootstrap --address "127.0.0.1"``):

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0tLS1...
    server: https://127.0.0.1:6443
  name: k8s
```

Now let's check the node IP address:

```
sudo k8s kubectl get nodes -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: Node
  ...
  status:
    addresses:
    - address: 172.26.39.15
      type: InternalIP
    - address: ubuntu
      type: Hostname
```

As you can see, the actual IP address of the node is used. I've then updated my netplan to have a different IP:

```
network:
    ethernets:
        eth0:
            dhcp4: false
            addresses:
            - 172.26.39.150/20
            routes:
              - to: default
                via: 172.26.32.1
            nameservers:
              addresses: [8.8.8.8, 8.8.4.4]
```

After applying the netplan, if I check the node again:

```
sudo k8s kubectl get nodes -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: Node
  ...
  status:
    addresses:
    - address: 172.26.39.150
      type: InternalIP
    - address: ubuntu
      type: Hostname
```

As you can see, the node address also updated. I haven't ran anything other than ``sudo netplan apply``.

Anyways. This PR doesn't affect this behaviour, as we're ignoring loopback addresses anyways, as we did before.

PR #1089, opened 2025-02-19
bug: fix cluster config merge checks

petrutlucian94:
MergeClusterConfig shouldn't allow changing the local storage reclaim policy if the local storage feature is enabled.

The problem is that the checks aren't correctly implemented:

```
allowChange: !existing.LocalStorage.GetEnabled() || !new.LocalStorage.GetEnabled()
```

Let's take the following example:

```
sudo k8s set local-storage.reclaim-policy=Retain
```

We're changing the reclaim policy, however the LocalStorage field is unset. ``new.LocalStorage.GetEnabled()`` would return ``false``, despite the fact that the local storage feature may still be enabled, it's just that the new config didn't contain this value.

To address this, we'll add a new helper that properly checks if a feature was enabled and remained enabled.

Implements: KU-2701

### DIFF
--- a/docs/canonicalk8s/snap/howto/storage/storage.md
+++ b/docs/canonicalk8s/snap/howto/storage/storage.md
@@ -53,6 +53,12 @@ The local-storage feature provides the following configuration options:
   is not changed. If this flag is not set and the cluster does not have a
   default class set then the class from the local-storage becomes the default.
 
+
+Please set the desired reclaim policy before enabling the ``local-storage``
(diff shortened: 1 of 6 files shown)


eaudetcobello:
Does DNS.GetEnabled() at line 52 suffer from the safe problem?

petrutlucian94:
> Does DNS.GetEnabled() at line 52 suffer from the same problem?

Indeed, so I covered that as well.

eaudetcobello (approved):

PR #1090, opened 2025-02-19
CI: fix artifact name e2e test

louiseschmidtgen:
## Description
Resolves error: `Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run` by adding `arch` in the artifact name.


bschimke95 (approved):
LGTM

PR #1093, opened 2025-02-20
feat(k8sd)!: introduce support for storing helm charts on microcluster db

berkayoz:
- Added a table to the microcluster database for storing helm charts.
- Added queries and helper functions for inserting and selecting helm charts.
- Added the ChartLoader interface and updated the helm client to enable loading helm charts in different ways.
- Added the database based ChartLoader to fetch charts from microcluser.
- Added the embed based ChartLoader to fetch charts from the embedded charts, used in testing.
- Added a helm chart controller that synchronizes available charts in the snap to the microcluster database.
- Moved helm charts from `k8s/manifests` into respective feature directories and embedded the charts using go embed package.
- Packed helm charts that are in directory form to tarballs.
- Updated feature controller to create the helm char with the database loader and updated features to consume a helm client.
- Fixed the bug of ports still being in use, in the testing util `WithState` by adding a retry loop for microcluster bootstrapping.
- Updated InstallableChart type to include the name and version of the chart.


src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,133 @@
...
+func NewHelmChartController(snap snap.Snap, waitReady func()) *HelmChartController {
+	return &HelmChartController{
+		snap:      snap,
+		waitReady: waitReady,
+	}
+}
+
+// Run runs the helm chart controller.
+// The controller reconciles helm charts available in the snap with the microcluster database until all valid charts are inserted.
+func (c *HelmChartController) Run(ctx context.Context, s state.State) {
```
bschimke95, line 41: So basically this controller only runs once and then returns after all charts are synced? 
This means it only runs:
a) when bootstrapping the cluster
b) when the snap was refreshed/restarted
right?
I feel like that this could simply run in the microcluster OnStart hook or am I missing something? Why does this need to be a separate controller?

src/k8s/pkg/client/helm/chart.go
```diff
@@ -2,13 +2,15 @@ package helm
...
-	// Namespace is the namespace to install the chart.
-	Namespace string
+	// Version is the version of the chart.
+	Version string
 
-	// ManifestPath is the path to the chart's manifest, typically relative to "$SNAP/k8s/manifests".
-	// TODO(neoaggelos): this should be a *chart.Chart, and we should use the "embed" package to load it when building k8sd.
-	ManifestPath string
+	// InstallName is the install name of the chart.
+	InstallName string
```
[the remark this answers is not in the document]
bschimke95: Nevermind, answered below. Maybe you can extend the comment a bit.

bschimke95 (commented):
Did a first pass, great work!

src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,133 @@
...
+	if exists {
+		return ErrChartExists
+	}
+
+	if err := s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
+		err := database.InsertHelmChart(ctx, tx, chart.Metadata.Name, chart.Metadata.Version, chartBytes)
+		return err
+	}); err != nil {
+		return fmt.Errorf("failed to insert helm chart: %w", err)
+	}
```
louiseschmidtgen, line 128: We can do this all in a query:
INSERT INTO table SELECT ...; 
See https://www.sqlite.org/lang_insert.html for reference.

louiseschmidtgen (commented):
First review of the Microcluster queries:

src/k8s/pkg/k8sd/database/sql/queries/helm-charts/exists.sql
```diff
@@ -0,0 +1,6 @@
+SELECT
+    EXISTS (
```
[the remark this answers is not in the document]
berkayoz: I did not want to fetch the contents column(the chart blob) to check for this. Also handling for this with the get query felt weird since I believe we'd need to check for the error to be `ErrNoRows` https://pkg.go.dev/database/sql#Stmt.QueryRowContext

No strong feelings, I can change it if we really don't want to have this query.

src/k8s/pkg/k8sd/database/sql/queries/helm-charts/select.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+    c.name, c.version, c.contents
+FROM
+    helm_charts AS c
+WHERE
+    ( c.name = ? ) AND ( c.version = ? )
```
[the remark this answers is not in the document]
berkayoz: I'm not sure if I understand correctly, this query selects a helm chart for the given name and version. There can only ever be 1 row due to the primary key, how does most recent come into play here?

src/k8s/pkg/k8sd/database/sql/migrations/helm-charts/002-create.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE helm_charts (
+    name TEXT NOT NULL,
+    version TEXT NOT NULL,
+    contents BLOB NOT NULL,
+    PRIMARY KEY (name, version)
```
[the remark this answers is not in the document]
berkayoz: If I understand correctly the `PRIMARY KEY` already creates an internal index, https://www.sqlite.org/lang_createtable.html#unique_constraints

src/k8s/pkg/k8sd/controllers/helm_chart.go
berkayoz: I'm not sure if I understand this.
> A new entry is inserted into the table for each row of data returned by executing the SELECT statement. 

Do we want `INSERT OR IGNORE` to do something like insert if not exists?

src/k8s/pkg/k8sd/database/helm_chart.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/canonical/k8s/pkg/k8sd/types"
+	"github.com/canonical/microcluster/v2/cluster"
+)
+
+var chartStmts = map[string]int{
```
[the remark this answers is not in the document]
berkayoz: Agreed, this also depends on other features that do not exist yet(knowing chart's in use etc.). I'll create a card to add this later on.

src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,133 @@
...
+
+		if !retryRequired {
+			log.Info("Reconcilation of helm charts complete")
+			return
+		}
+
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(3 * time.Second):
```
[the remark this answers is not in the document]
berkayoz: I believe there is no need, this is just to retry after 3 seconds. We do have the same hard-coded interval in other controllers too.

src/k8s/pkg/k8sd/controllers/helm_chart.go
berkayoz: Correct, this is a fire once operation. The main reason this was made into a "controller" is
1. The logic is quite large, so this would still need to be a function of sorts. 
2. We'd still need the same knobs for disabling etc. if necessary
3. Sync might never fully succeed and retry periodically

Basically the requirements / approach is the same as a controller. I feel like we'd end up something very similar to a controller already and I consider this approach it as a controller listening for the `start` event. We could actually create a full loop, add channels and send a signal there on start to make it feel more like a "controller". Returning after success felt simpler.

src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,133 @@
...
+			entries, err := chartFS.ReadDir("charts")
+			if err != nil {
+				log.Error(err, "failed to read helm charts directory")
+				retryRequired = true
+				continue
+			}
+
+			for _, e := range entries {
+				if e.IsDir() {
+					// TODO(berkayoz): Add support for directories by creating a tarball(.tgz) automatically
```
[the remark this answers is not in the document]
berkayoz: It does not block upgrades per say. It would be just a convenience factor.

Since we need to "serialize" and save the charts into the db, we can't really do this for a directory(in the directory form). So the current approach is just doing `helm package` to create the archive manually. To get rid of this manual step we could look into doing the "helm package" here on the fly, but we already keep most of our charts as tarballs.

src/k8s/pkg/client/helm/chart.go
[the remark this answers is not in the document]
berkayoz: Good point, extended the comment.

src/k8s/pkg/client/helm/loader/db.go
```diff
@@ -0,0 +1,50 @@
...
+		var err error
+		chartEntry, err = database.GetHelmChart(ctx, tx, f.Name, f.Version)
+		return err
+	}); err != nil {
+		return nil, fmt.Errorf("failed to check if helm chart exists: %w", err)
+	}
+
+	if chartEntry == nil {
+		return nil, fmt.Errorf("helm chart %s-%s not found", f.Name, f.Version)
+	}
```
HomayoonAlimohammadi, line 42: Are we going to hit this? I feel like if `database.GetHelmChart` doesn't return any errors we expect the chartEntry to also not be nil. I'm all in for safer and more defensive approaches, but maybe this one is not necessary.

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -126,16 +126,22 @@ func (c *FeatureController) Run(
...
 		return fmt.Errorf("failed to retrieve cluster configuration: %w", err)
 	}
 
-	status, applyErr := apply(cfg)
+	s := getState()
+
+	// helm client with database loader
+	m := c.snap.HelmClient(loader.NewDatabaseLoader(s))
+
+	status, applyErr := apply(m, cfg)
```
HomayoonAlimohammadi, line 144: Optional nit: Maybe we can inline these?

src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,134 @@
...
+				chartBytes, err := chartFS.ReadFile(filepath.Join("charts", e.Name()))
+				if err != nil {
+					log.WithValues("entry", e.Name()).Error(err, "failed to read helm chart")
+					retryRequired = true
+					continue
+				}
+
+				if err := c.reconcile(ctx, s, chartBytes); err != nil {
+					if errors.Is(err, ErrChartExists) || errors.Is(err, ErrChartParseFailed) {
+						log.WithValues("entry", e.Name()).Info("skipping reconciliation of entry")
```
HomayoonAlimohammadi, line 81: nit: Maybe we can have a log.Error for the `ErrChartParseFailed` case?

src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,134 @@
...
+		if !retryRequired {
+			log.Info("Reconcilation of helm charts complete")
+			return
+		}
+
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(3 * time.Second):
+		}
```
HomayoonAlimohammadi, line 99: Dumb question: Why do we need this block? We either return if the context is done or we wait 3 seconds? Do we have to wait for something? Why can't we return immediately?

src/k8s/pkg/k8sd/controllers/helm_chart.go
HomayoonAlimohammadi: Doing this in a single query is a good idea. I think there are multiple ways to do it (IGNORE, ON CONFLICT DO NOTHING, INSERT INTO ... SELECT ... WHERE NOT EXISTS), I'm ok with either of these.

src/k8s/pkg/k8sd/database/helm_chart.go
```diff
@@ -0,0 +1,68 @@
...
+	"insert-chart": MustPrepareStatement("helm-charts", "insert.sql"),
+	"select":       MustPrepareStatement("helm-charts", "select.sql"),
+	"exists":       MustPrepareStatement("helm-charts", "exists.sql"),
+}
+
+// HelmChartExists checks if a helm chart exists in the database by name and version.
+func HelmChartExists(ctx context.Context, tx *sql.Tx, name string, version string) (bool, error) {
```
HomayoonAlimohammadi, line 19: Maybe remove this if the insert and ignore logic was implemented.

src/k8s/pkg/k8sd/database/sql/migrations/helm-charts/002-create.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE helm_charts (
+    name TEXT NOT NULL,
+    version TEXT NOT NULL,
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: We might see the benefits of this if we have millions of charts in the db or more. Good idea overall, maybe not necessary for now.

src/k8s/pkg/k8sd/database/sql/migrations/helm-charts/002-create.sql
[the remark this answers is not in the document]
HomayoonAlimohammadi: Maybe we can add indexes on name, version or the name/version suggested change. I'm also ok with leaving it like this for now tho.

src/k8s/pkg/k8sd/database/sql/migrations/helm-charts/002-create.sql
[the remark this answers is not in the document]
HomayoonAlimohammadi: Should the name/version combination be unique? If so, we can also have a `UNIQUE` constraint on that combined column as well. Now we can't have UNIQUE constraints for neither of name or version.

src/k8s/pkg/k8sd/database/sql/queries/helm-charts/select.sql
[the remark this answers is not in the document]
HomayoonAlimohammadi: I _think_ we can have multiple rows here since `name` or `version` do not have UNIQUE constraints. I don't think that the primary key play any significant role here. It's just auto-incremented. WDYT?

src/k8s/pkg/k8sd/features/calico/network_test.go
```diff
@@ -46,7 +47,8 @@ func TestDisabled(t *testing.T) {
 			SecurePort: ptr.To(6443),
 		}
 
-		status, err := calico.ApplyNetwork(context.Background(), snapM, nil, apiserver, network, nil)
+		mc := snapM.HelmClient(loader.NewFileLoader("./charts"))
```
HomayoonAlimohammadi, line 50: nit: Maybe we can centralize the `"./charts"` string?

HomayoonAlimohammadi (commented):
Great work, thanks a lot @berkayoz! Did a first pass, LGTM overall, left some minor comments and questions.

HomayoonAlimohammadi:
And a quick question. I noticed the `!` in the PR title, How are we gonna backport this if it's a breaking change?

src/k8s/pkg/k8sd/controllers/helm_chart.go
berkayoz: The block here is to wait 3 seconds before retrying the loop. This block is a pretty standard pattern in go to wait for some period while making sure we don't block the exit/shutdown of the program unnecessarily. (Whichever channel fires first will get to execution)

src/k8s/pkg/k8sd/controllers/feature.go
berkayoz: The line felt a bit large when I initially inlined these, feels like compromising from readability

src/k8s/pkg/k8sd/features/calico/network_test.go
berkayoz: Good point, switched to use the `embed.FS` directly instead.

src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,112 @@
...
+
+	chartBytes, err := chartFS.ReadFile(filepath.Join("charts", e.Name()))
+	if err != nil {
+		return fmt.Errorf("failed to read entry: %w", err)
+	}
+
+	chart, err := loader.LoadArchive(bytes.NewReader(chartBytes))
+	if err != nil {
+		log.Error(err, "failed to parse entry, skipping reconciliation")
+		return nil
```
eaudetcobello, line 99: Shouldn't this return an error?

src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (c *HelmChartController) reconcile(ctx context.Context, s state.State, chartFS *embed.FS, e fs.DirEntry) error {
+	log := log.FromContext(ctx).WithValues("entry", e.Name())
+
+	if e.IsDir() {
+		// TODO(berkayoz): Add support for directories by creating a tarball(.tgz) automatically
+		// Directories should be converted manually via `helm package <directory>` command until then
+		log.Info("entry is a directory, skipping reconciliation")
+		return nil
+	}
```
eaudetcobello, line 89: return an error?

src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,112 @@
...
+
+	chartBytes, err := chartFS.ReadFile(filepath.Join("charts", e.Name()))
```
eaudetcobello, line 91: Could this be a path traversal security risk? Say a user sets the chart name to "../../../other-archive.tgz" ?

src/k8s/pkg/client/helm/loader/embed.go
```diff
@@ -0,0 +1,40 @@
...
+// NewEmbedLoader creates a new embed loader.
+func NewEmbedLoader(chartFS *embed.FS) *embedLoader {
+	return &embedLoader{
+		chartFS: chartFS,
+	}
+}
+
+// Load loads a helm chart from the filesystem by name and version.
+func (l *embedLoader) Load(ctx context.Context, f helm.InstallableChart) (*chart.Chart, error) {
+	chartBytes, err := l.chartFS.ReadFile(filepath.Join("charts", fmt.Sprintf("%s-%s.tgz", f.Name, f.Version)))
```
eaudetcobello, line 29: Could this lead to a path traversal vulnerability where someone passes ../../../chart.tgz to load a file they don't have access to normally?

eaudetcobello, line 29: ctx is not used

src/k8s/pkg/k8sd/controllers/helm_chart.go
berkayoz: This is to make sure we ignore any files that is not an actual helm chart. Returning an error here would require us to make sure nothing but helm charts are placed in the appropriate `charts` directory of features.

src/k8s/pkg/k8sd/controllers/helm_chart.go
berkayoz: The same reason mentioned above applies here, we might like to keep the source directory for the helm chart and do `helm package` as necessary.

src/k8s/pkg/client/helm/loader/embed.go
berkayoz: We read from the embedded charts not from the actual filesystem, so the only accessible files would be other charts for that feature. Additionally this specific loader is used in testing only.

src/k8s/pkg/k8sd/controllers/helm_chart.go
```diff
@@ -0,0 +1,112 @@
...
+	for {
+		log.Info("Reconciling helm charts")
+
+		var retryRequired bool
+
+		for _, chartFS := range charts.Charts() {
+			entries, err := chartFS.ReadDir("charts")
+			if err != nil {
+				log.Error(err, "failed to read helm charts directory")
+				retryRequired = true
```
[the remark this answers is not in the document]
berkayoz: It is checked later on with `if !retryRequired` to return if we synced everything successfully.

src/k8s/pkg/k8sd/controllers/helm_chart.go
HomayoonAlimohammadi: Oh I didn't notice the for loop. All good.

src/k8s/pkg/k8sd/database/sql/queries/helm-charts/select.sql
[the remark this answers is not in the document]
HomayoonAlimohammadi: Aaah I didn't notice the `PRIMARY KEY (name, version)`, thought it's a regular primary key. Awesome.

bschimke95 (approved):
LGTM for an initial version. 
We probably need to iterate on that

src/k8s/pkg/client/helm/loader/embed.go
```diff
@@ -0,0 +1,40 @@
...
+// NewEmbedLoader creates a new embed loader.
+func NewEmbedLoader(chartFS *embed.FS) *embedLoader {
+	return &embedLoader{
+		chartFS: chartFS,
+	}
+}
+
+// Load loads a helm chart from the filesystem by name and version.
+func (l *embedLoader) Load(_ context.Context, f helm.InstallableChart) (*chart.Chart, error) {
+	chartBytes, err := l.chartFS.ReadFile(filepath.Join("charts", fmt.Sprintf("%s-%s.tgz", f.Name, f.Version)))
```
louiseschmidtgen, line 29: Have we considered users wanting to use this option for air-gap deployments?

src/k8s/pkg/k8sd/database/sql/queries/helm-charts/insert.sql
```diff
@@ -0,0 +1,4 @@
+INSERT OR IGNORE INTO
```
louiseschmidtgen, line 1: This might be a tricky one to debug when sqlite deals with a conflict.
Have you considered doing a SELECT + INSERT pattern instead of the IGNORE statement?
Then it won't fail either and you can check if the transaction returns an id if the row was inserted?

src/k8s/pkg/k8sd/database/helm_chart.go
```diff
@@ -0,0 +1,50 @@
...
+	err = selectTxStmt.QueryRowContext(ctx, name, version).Scan(&chart.Name, &chart.Version, &chart.Contents)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query helm chart: %w", err)
+	}
+
+	return &chart, nil
+}
+
+// InsertHelmChart inserts a helm chart into the database.
+func InsertHelmChart(ctx context.Context, tx *sql.Tx, name string, version string, contents []byte) error {
```
louiseschmidtgen, line 35: Why don't we use the HelmChartEntry struct in this file?

louiseschmidtgen (commented):
Nice work Berkay, a few more coments:

src/k8s/pkg/client/helm/loader/embed.go
berkayoz: The database loader should work on air-gap deployments as well, this is only for testing purposes.

src/k8s/pkg/k8sd/database/sql/queries/helm-charts/insert.sql
berkayoz: I'm not really familiar with the SELECT + INSERT pattern, could you elaborate on how it differs and provide an example please?

PR #1097, opened 2025-02-21
docs: CAPI landing pages rework 

nhennigan:
Similar to my last PR, this is to change the CAPI landing pages to conform to the standard and provide more info.


addyess (approved):
LGTM

PR #1098, opened 2025-02-21
ci: add top level permissions to the stale job

petrutlucian94:
We're getting code scanning alerts as the stale job doesn't
contain explicit top-level permissions.


louiseschmidtgen (approved):
lgtm

PR #1101, opened 2025-02-21
test: Add airgap test scenarios

bschimke95:
Adds integration tests for:
* CK8s setup with internet access through proxy
* CK8s setup with private image registry

### DIFF
--- a/tests/integration/tests/conftest.py
+++ b/tests/integration/tests/conftest.py
@@ -110,12 +110,7 @@ def h() -> harness.Harness:
 
 @pytest.fixture(scope="session")
 def registry(h: harness.Harness) -> Optional[Registry]:
-    if config.USE_LOCAL_MIRROR:
-        yield Registry(h)
-    else:
-        # local image mirror disabled, avoid initializing the
-        # registry mirror instance.
-        yield None
+    yield Registry(h)
 
 
 @pytest.fixture(scope="session", autouse=True)
--- a/tests/integration/tests/test_airgapped.py
+++ b/tests/integration/tests/test_airgapped.py
@@ -8,10 +8,11 @@ from typing import List
 import pytest
 from test_util import harness, registry, tags, util
 
+REGISTRY_PORT = 5000
 
 def setup_proxy(proxy: harness.Instance):
     """Installs and configures Squid proxy on the given instance."""
-    proxy.exec("sudo apt install squid --yes".split())
+    proxy.exec("apt install squid --yes".split())
     proxy.exec("echo 'http_access allow all' >> /etc/squid/conf.d/allow.conf".split())
     time.sleep(1)
     proxy.exec("systemctl restart squid.service".split())
@@ -75,7 +76,7 @@ def test_airgapped_with_proxy(instances: List[harness.Instance]):
     configure_proxy_env(instance, proxy_ip, instance_ip)
     restrict_network(instance, allow_ports=[3128])
 
-    # Verify connectivity
+    # Verify connectivity without the proxy is blocked.
     assert (
         instance.exec(
             "curl -I -4 --noproxy '*' https://www.google.com".split(),
@@ -84,6 +85,9 @@ def test_airgapped_with_proxy(instances: List[harness.Instance]):
         ).returncode
         == 7
     )
+
+    # Export the proxy settings and verify connectivity through proxy.
+    # This is required because the proxy settings are not available to the Python subprocess shell that runs the connectivity test.
     instance.exec(
         "export $(grep -v '^#' /etc/environment | xargs) && curl -I -4 https://www.google.com".split()
     )
@@ -109,7 +113,7 @@ def test_airgapped_with_proxy_setup_and_image_mirror(
     configure_proxy_env(registry.instance, proxy_ip, registry_ip)
     restrict_network(registry.instance, allow_ports=[3128])
 
-    # Verify connectivity
+    # Verify connectivity without the proxy is blocked.
     assert (
         registry.instance.exec(
             "curl -I -4 --noproxy '*' https://www.google.com".split(),
@@ -118,6 +122,7 @@ def test_airgapped_with_proxy_setup_and_image_mirror(
         ).returncode
         == 7
     )
+    # Verify connectivity through the proxy.
     registry.instance.exec(
         "export $(grep -v '^#' /etc/environment | xargs) && curl -I -4 https://www.google.com".split()
     )
@@ -130,7 +135,7 @@ def test_airgapped_with_proxy_setup_and_image_mirror(
     images = out.stdout.splitlines()
     for image in images:
         link = "/".join(image.split("/")[1:])
-        tag = f"{registry_ip}:5000/{link}"
+        tag = f"{registry_ip}:{REGISTRY_PORT}/{link}"
         registry.instance.exec(
             "export $(grep -v '^#' /etc/environment | xargs) && "
             + f"/snap/k8s/current/bin/ctr images pull --all-platforms {image}".split()
@@ -172,7 +177,7 @@ def test_airgapped_with_proxy_setup_and_image_mirror(
         == 7
     )
 
-    restrict_network(instance, allow_ports=[5000])
+    restrict_network(instance, allow_ports=[REGISTRY_PORT])
     util.setup_k8s_snap(instance, Path("/"))
     registry.apply_configuration(instance)
     instance.exec("sudo k8s bootstrap".split())
--- a/tests/integration/tests/test_airgapped.py
+++ b/tests/integration/tests/test_airgapped.py
@@ -1,9 +1,13 @@
-from pathlib import Path
+#
+# Copyright 2025 Canonical, Ltd.
+#
 import time
+from pathlib import Path
 from typing import List
 
 import pytest
-from test_util import harness, tags, util, registry
+from test_util import harness, registry, tags, util
+
 
 def setup_proxy(proxy: harness.Instance):
     """Installs and configures Squid proxy on the given instance."""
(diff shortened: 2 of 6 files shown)


tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,179 @@
...
+from pathlib import Path
+from typing import List
+
+import pytest
+from test_util import harness, registry, tags, util
+
+
+def setup_proxy(proxy: harness.Instance):
+    """Installs and configures Squid proxy on the given instance."""
+    proxy.exec("sudo apt install squid --yes".split())
```
eaudetcobello, line 14: nit: sudo not needed

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,179 @@
...
+        instance.exec(
+            "curl -I -4 --noproxy '*' https://www.google.com".split(),
+            check=False,
+            capture_output=True,
+        ).returncode
+        == 7
+    )
+    instance.exec(
+        "export $(grep -v '^#' /etc/environment | xargs) && curl -I -4 https://www.google.com".split()
+    )
```
eaudetcobello, line 89: Please add a comment explaining what this does. "Export uncommented env. variables from /etc/environment"

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,179 @@
...
+        registry.instance.exec(
+            "curl -I -4 --noproxy '*' https://www.google.com".split(),
+            check=False,
+            capture_output=True,
+        ).returncode
+        == 7
+    )
+    registry.instance.exec(
+        "export $(grep -v '^#' /etc/environment | xargs) && curl -I -4 https://www.google.com".split()
+    )
```
eaudetcobello, line 123: This is duplicated between tests, might be worth putting in a function

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,179 @@
...
+
+    setup_containerd_proxy(registry.instance, proxy_ip)
+    registry.instance.exec("sudo k8s bootstrap".split())
+
+    # Mirror images
+    out = registry.instance.exec(["k8s", "list-images"], capture_output=True, text=True)
+    images = out.stdout.splitlines()
```
eaudetcobello, line 130:
```suggestion
    out = registry.instance.exec(["k8s", "list-images"], capture_output=True, text=True)
    images = out.stdout.strip().splitlines()
```

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,179 @@
...
+    for image in images:
+        link = "/".join(image.split("/")[1:])
+        tag = f"{registry_ip}:5000/{link}"
+        registry.instance.exec(
+            "export $(grep -v '^#' /etc/environment | xargs) && "
+            + f"/snap/k8s/current/bin/ctr images pull --all-platforms {image}".split()
+        )
```
eaudetcobello, line 137: This gets a bit convoluted. How about this?
```py
proxy_env_file = "/etc/proxy-environment"
registry.instance.exec(f"grep -v '^#' /etc/environment > {proxy_env_file}".split())

...

registry.instance.exec(
            "source /etc/proxy-environment && "
            + f"/snap/k8s/current/bin/ctr images pull --all-platforms {image}".split()
        )
```

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,179 @@
...
+        registry.instance.exec(
+            "export $(grep -v '^#' /etc/environment | xargs) && "
+            + f"/snap/k8s/current/bin/ctr images tag {image} {tag}".split()
+        )
+
+        # The 443 port is required to upload to the local registry. So, we need to temporarily allow it.
+        registry.instance.exec(
```
eaudetcobello, line 144:
```suggestion
		# Delete the reject rule
        registry.instance.exec(
```

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,179 @@
...
+            "iptables -D OUTPUT -p tcp --dport 443 -j REJECT".split()
```
eaudetcobello, line 145:
```suggestion
			# Add an allow rule
            "iptables -D OUTPUT -p tcp --dport 443 -j REJECT".split()
```

tests/integration/tests/test_util/harness/lxd.py
```diff
@@ -225,7 +225,7 @@ def exec(self, instance_id: str, command: list, **kwargs):
 
         LOG.debug("Execute command %s in instance %s", command, instance_id)
 
-        if ">" in " ".join(command):
+        if ">" in " ".join(command) or "&&" in " ".join(command):
```
eaudetcobello, line 228: Would it be cleaner to use something like

```
registry.instance.exec([
    "bash", "-c", 
    f"source /etc/proxy-environment && /snap/k8s/current/bin/ctr images pull --all-platforms {image}"
])
```

through the tests? Right now you are splitting the string when calling .exec("xyz".split()) and then in exec you are re-joining the string.

tests/integration/tests/test_util/registry.py
```diff
@@ -165,6 +166,16 @@ def ip(self) -> str:
...
 
+    def exec(self, command: List[str], **kwargs) -> subprocess.CompletedProcess:
+        """
+        Execute a command on the registry instance.
+
+        Args:
+            command (List[str]): The command to execute
+            **kwargs: Additional keyword arguments to pass to the exec
+        """
+        return self.instance.exec(command, **kwargs)
```
eaudetcobello, line 177: Why is this needed instead of instance.exec directly?

tests/integration/tests/test_airgapped.py
bschimke95: Duplication is (often) fine for test cases to be transparent on what and why things are tested.

I don't think moving that into a separate function makes the test intention clearer. 
Also, it would probably be something like "def verifyProxyAndNoProxyConnectivity" which is a bit ugly

bschimke95: That being said, I added some more comments to better explain this test scenario.

tests/integration/tests/test_airgapped.py
bschimke95: Hm I see your point but I think explicitly putting `/etc/environment` in the line is more transparent. Otherwise, the reader would need to check where "proxy-environment" comes from

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,179 @@
...
+    assert (
+        registry.instance.exec(
+            "curl -I -4 https://www.google.com".split(),
+            check=False,
+            capture_output=True,
+        ).returncode
+        == 7
+    )
+
+    restrict_network(instance, allow_ports=[5000])
```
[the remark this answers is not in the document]
bschimke95: Good point, done

tests/integration/tests/test_util/registry.py
bschimke95: Convenience and abstraction.
I got tired of typing `registry.instance.exec`

tests/integration/tests/test_util/harness/lxd.py
bschimke95: Hm, I don't think this works. See line 234.
This would end up being `bash -c "bash -c ..."`

tests/integration/tests/conftest.py
```diff
@@ -110,12 +110,7 @@ def h() -> harness.Harness:
 
 @pytest.fixture(scope="session")
 def registry(h: harness.Harness) -> Optional[Registry]:
-    if config.USE_LOCAL_MIRROR:
-        yield Registry(h)
-    else:
-        # local image mirror disabled, avoid initializing the
-        # registry mirror instance.
-        yield None
```
HomayoonAlimohammadi, line 118: I'm a bit confused why we removed this. We do still have the `USE_LOCAL_MIRROR` config...

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,186 @@
...
+
+import pytest
+from test_util import harness, registry, tags, util
+
+REGISTRY_PORT = 5000
+
+
+def setup_proxy(proxy: harness.Instance):
+    """Installs and configures Squid proxy on the given instance."""
+    proxy.exec("apt install squid --yes".split())
```
HomayoonAlimohammadi, line 16: Do we need a `apt update` before? And maybe we can install a fixed version to prevent getting broke.

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,186 @@
...
+    )
+
+
+def setup_containerd_proxy(instance: harness.Instance, proxy_ip: str):
+    """Configures containerd to use the proxy."""
+    config = f"""
+[Service]
+Environment="HTTPS_PROXY=http://{proxy_ip}:3128"
+Environment="HTTP_PROXY=http://{proxy_ip}:3128"
+Environment="NO_PROXY=10.1.0.0/16,10.152.183.0/24,192.168.0.0/16,127.0.0.1,172.16.0.0/12"
```
HomayoonAlimohammadi, line 59: Do we also need `extra_no_proxy`(ies) here?

tests/integration/tests/test_airgapped.py
```diff
@@ -0,0 +1,186 @@
...
+            "export $(grep -v '^#' /etc/environment | xargs) && "
+            + f"/snap/k8s/current/bin/ctr images pull --all-platforms {image}".split()
+        )
+        registry.instance.exec(
+            "export $(grep -v '^#' /etc/environment | xargs) && "
+            + f"/snap/k8s/current/bin/ctr images tag {image} {tag}".split()
+        )
+
+        # The 443 port is required to upload to the local registry. So, we need to temporarily allow it.
+        registry.instance.exec(
```
HomayoonAlimohammadi, line 151: Don't we want to use the `exec` function declared in the `registry.py`? I initially thought that's the main reason we declared that function :D

HomayoonAlimohammadi (approved):
Great work, thanks a lot @bschimke95! LGTM overall, some minor comments and questions.

eaudetcobello (approved):

PR #1111, opened 2025-02-26
CI: use lxd vms by default

petrutlucian94:
We're having issues on ubuntu 22.04 hosts with 24.04 lxd containers, which are no longer able to define cgroup cpusets. As a result, kubelet is no longer able to run.

Since lxd privileged containers are about to become unsupported anyway, we'll use lxd vms instead.

We're adding config options to allow using lxd containers or specify the amount of vm resources.


petrutlucian94:
Closing this in favor of https://github.com/canonical/k8s-snap/pull/1121

PR #1113, opened 2025-02-26
feat(ci): add pr linter

eaudetcobello:
Adds PR linter inspired by MAAS team's configuration. Feel free to suggest new scopes.


.github/workflows/lint_pr.yaml
```diff
@@ -0,0 +1,58 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        id: lint_pr_title
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          ignoreLabels: |
+            automerge
+          scopes: |
```
bschimke95, line 22: I have mixed feelings of those scope and would prefer to just remove them but if the team want these conventional commits, then I'm fine with it.

.github/workflows/lint_pr.yaml
```diff
@@ -0,0 +1,58 @@
...
+            feature
+            k8sd
+            controllers
+            db
```
bschimke95, line 26: What is the difference between db and microcluster?

.github/workflows/lint_pr.yaml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        id: lint_pr_title
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          ignoreLabels: |
+            automerge
+          scopes: |
+            feature
```
bschimke95, line 23: Isn't feature a "type"? https://www.conventionalcommits.org/en/v1.0.0/#summary

eaudetcobello: Well load-balancer, network, dns, are features, so I put feature as a scope.

.github/workflows/lint_pr.yaml
nhennigan: I also have mixed feelings about the inclusion of the scope. I wonder are we making it more difficult to get involved in the project?

nhennigan:
Will WIP be no longer permitted under this?

.github/workflows/lint_pr.yaml
nhennigan: Although it could pave the way for more automated release notes

.github/workflows/lint_pr.yaml
bschimke95: AHH, ignore me - you mean this kind of feature. It was late...

.github/workflows/lint_pr.yaml
berkayoz: I feel like we'll need to think more deeply about the scopes. Can we start without scopes for now, and maybe we can spec this out and take the discussion there? @eaudetcobello

eaudetcobello: Sure thing.

eaudetcobello:
@nhennigan, yes, it's still permitted. Just prefix your pr with [WIP] and the title won't be validated

.github/workflows/lint_pr.yaml
```diff
@@ -0,0 +1,48 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          ignoreLabels: |
+            automerge
+          wip: true
+      - uses: marocchino/sticky-pull-request-comment@v2
+        # When the previous steps fails, the workflow would stop. By adding this
+        # condition you can continue the execution with the populated error message.
+        if: always() && (steps.lint_pr_title.outputs.error_message != null)
```
bschimke95, line 27: why is this `always()` required?

.github/workflows/lint_pr.yaml
```diff
@@ -0,0 +1,48 @@
+name: "Lint PR"
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: write # to comment
+  statuses: write # to set status
```
bschimke95, line 9: The job will set the PR to "requires change" or why is this requiered?

bschimke95 (approved):
LGTM, just some clarification questions

.github/workflows/lint_pr.yaml
eaudetcobello: If the PR title is invalid, the previous step (action-semantic-pull-request) fails (exit code 1). When the previous step fails, we want to add a comment.

.github/workflows/lint_pr.yaml
eaudetcobello: It adds information to a commit status (the red x, the yellow circle, etc.) Example: (Yellow circle) - Waiting for status to be reported — This PR is marked with "[WIP]".

PR #1121, opened 2025-02-27
CI: switch to noble (24.04) runners and k8s-dqlite downgrade

petrutlucian94:
We're hitting an LXD issue with ubuntu 22.04 hosts and 24.04 LXD containers. The containers are not allowed to use cgroup cpu sets.

For now, we're going to use 24.04 self-hosted runners instead.

We'll also revert k8s-dqlite 1.4.1 -> 1.3.1 because of the watcher issue that causes kubelet to miss events.


build-scripts/components/runc/patches/v1.2.5/0001-Disable-static-PIE-on-arm64.patch
HomayoonAlimohammadi, line 1: Maybe we can remove this file from the PR now that https://github.com/canonical/k8s-snap/pull/1123 is there?

bschimke95 (approved):
LGTM if CI is green

petrutlucian94:
The nightly jobs had some failures due to the k8s-dqlite bug but that's expected since these jobs use latest/upstream. I'll go ahead and merge the PR.

PR #1122, opened 2025-02-27
doc: add capi how-to for custom bootstrap config

petrutlucian94:
We're adding a how-to document, showing how to use a custom bootstrap configuration with the Cluster API provider.

Note that this is already outlined by the config reference page: https://documentation.ubuntu.com/canonical-kubernetes/latest/capi/reference/configs/#bootstrapconfig

Implements: KU-2405


docs/canonicalk8s/capi/howto/custom-bootstrap-config.md
```diff
@@ -0,0 +1,106 @@
+# How to use custom bootstrap configuration
+
+The {{product}} bootstrap configuration gets automatically generated based on
+user provided settings described in the [Cluster API configuration reference].
+
+The CAPI provider will also include CA certificates as well as annotations
+and other settings that allow the provider to function properly.
```
eaudetcobello, line 7: I feel this is missing a bit of explanation after "include". Include to what? where? cloud-init? Referencing cloud-init may be too much detail, so maybe there's a better way to phrase this.

petrutlucian94: Good point, those are actually added to the bootstrap config.

I could rephrase this to something like:

```
The configuration generated by the CAPI provider will also include ...
```

bschimke95 (approved):
LGTM

PR #1123, opened 2025-02-27
fix: Add runc 1.2.5 arm patch

petrutlucian94:
The runc version was bumped, we need to update the arm patch, which no longer applies correctly.


petrutlucian94:
24.04 tests are expected to fail, we also need this: https://github.com/canonical/k8s-snap/pull/1121

HomayoonAlimohammadi (approved):
Thanks a lot @petrutlucian94, LGTM

PR #1125, opened 2025-02-27
test: Adds Sonobuoy CNCF e2e nigthly tests

claudiubelu:
Adds the ``test_cncf_conformance`` test. It has the ``conformance_tests`` tag, preventing it from being run, unless specifically included through the tag. This test is only meant to be run on the Nightly job.

### DIFF
--- a/tests/integration/tests/test_util/tags.py
+++ b/tests/integration/tests/test_util/tags.py
@@ -7,8 +7,9 @@ PULL_REQUEST = "pull_request"
 NIGHTLY = "nightly"
 WEEKLY = "weekly"
 GPU = "gpu"
+CONFORMANCE = "conformance_tests"
 
-TEST_LEVELS = [PULL_REQUEST, NIGHTLY, WEEKLY]
+TEST_LEVELS = [PULL_REQUEST, NIGHTLY, WEEKLY, CONFORMANCE]
 
 # Those tags can be used for a convenient way to run multiple test levels.
 combine_tags("up_to_nightly", PULL_REQUEST, NIGHTLY)
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -84,3 +84,9 @@ jobs:
         with:
           name: ${{ env.artifact_name }}
           path: ${{ github.workspace }}/inspection-reports.tar.gz
+      - name: Upload CNCF conformance report artifact
+        if: ${{ failure() && contains(inputs.test-tags, 'conformance_tests') }}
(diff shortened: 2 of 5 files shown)


tests/integration/tests/test_util/tags.py
```diff
@@ -7,8 +7,9 @@
 NIGHTLY = "nightly"
 WEEKLY = "weekly"
 GPU = "gpu"
+CONFORMANCE = "conformance_tests"
 
-TEST_LEVELS = [PULL_REQUEST, NIGHTLY, WEEKLY]
+TEST_LEVELS = [PULL_REQUEST, NIGHTLY, WEEKLY, CONFORMANCE]
```
bschimke95, line 12: I understand why you want to do this as the conformance tests add another 2h to the test time.
However, right now - we would only run those test in nightly tests, not e.g. in backports. 
How about putting the conformance tests in the weekly test level?

claudiubelu: Indeed, the test would run in the Nightly **jobs**, and it is, as it can be seen in the Nightly job logs below, but we run the Nightly tests in the `Lint and Integration` tests will run the **nightly tests** as well (`up_to_weekly`, to be more precise: https://github.com/canonical/k8s-snap/blob/b56cba97dbb6526612c03f51b8e35ccdce822541/.github/workflows/lint_and_integration.yaml#L83-L103) in some cases. Basically, with this we prevent these tests from running in the `Lint and Integration` jobs.

tests/integration/tests/test_util/config.py
```diff
@@ -167,3 +167,11 @@
 
 # Local mirror configuration.
 MIRROR_LIST = json.loads(os.getenv("TEST_MIRROR_LIST", "{}")) or DEFAULT_MIRROR_LIST
+
+# SONOBUOY_VERSION is the version of sonobuoy to use for CNCF conformance tests.
+SONOBUOY_VERSION = os.getenv("TEST_SONOBUOY_VERSION") or "v0.57.3"
+
+
+# sonobuoy_tar_gz returns the download URL of sonobuoy.
+def sonobuoy_tar_gz(architecture: str) -> str:
```
bschimke95, line 176: Let's move this to the sonobuoy test - the `config` file should only contain config

bschimke95 (approved):
Final little comment, then we can merge

tests/integration/tests/test_util/config.py
claudiubelu: Done

bschimke95:
Ignoring failing TICS test as it is unrelated.

PR #1134, opened 2025-02-28
CI: use cdkbot token for automated jobs

petrutlucian94:
The backport job creates the PRs using the cdkbot token but approves them using "GITHUB_TOKEN". For this reason, the approval is ignored and the backports cannot be merged automatically.

At the same time, the job that updates components uses the GH token to create the PRs, which do get merged but the e2e tests are not executed, potentially introducing regressions. Again, we'll have to use the cdkbot token.

Since the same account can't approve its own PRs, we'll try using the GH account to approve/merge PRs.

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -58,7 +58,7 @@ jobs:
 
       - name: Label backports with automerge and approve
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ secrets.BOT_TOKEN }}
         run: |
           for pr_number in ${{ steps.create_backports.outputs.created_pull_numbers }}; do
             gh pr edit "$pr_number" --add-label "automerge"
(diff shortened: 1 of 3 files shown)


.github/workflows/backport.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
 
       - name: Label backports with automerge and approve
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ secrets.BOT_TOKEN }}
```
bschimke95, line 61: I don't think this will work. The CDKbot will basically approve its own PR.

petrutlucian94: hmm, right. I'm trying to figure out what's the cleanest workaround.

berkayoz (approved):
LGTM, have you tried this on your fork?

petrutlucian94:
Couldn't really reproduce this on my fork since I don't have access to the ``k8s-snap`` project settings. As discussed in the daily meeting call, we'll merge this for now and create another bot account if it's unable to merge the PRs.

PR #1140, opened 2025-03-03
CI: remove inclusive language check

nhennigan:
In the documentation workflows where we pull the inclusive language check from, there was some issues  from an update with Vale not pulling the correct dictionaries. Until this is fixed, disable the check. Added small doc change to activate the doc check pipeline

### DIFF
--- a/.github/workflows/docs-checks.yml
+++ b/.github/workflows/docs-checks.yml
@@ -29,12 +29,12 @@ jobs:
         with:
           working-directory: docs/canonicalk8s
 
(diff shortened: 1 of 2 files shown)


addyess (approved):
LGTM, do you wanna just delete it or is the comment there in case it comes back?

nhennigan:
> LGTM, do you wanna just delete it or is the comment there in case it comes back?

I think comment for now. I'll re-enable it when it becomes stable again

PR #1142, opened 2025-03-04
feat(k8sd)!: store available feature manifests on microcluster db

berkayoz:
- Added the FeatureManifestController that inserts available manifests in the snap at k8sd start.
- Added a microcluster table to store feature manifests.
- Added db helpers for inserting and getting feature manifests.

### DIFF
--- a/src/k8s/pkg/k8sd/controllers/feature_manifest.go
+++ b/src/k8s/pkg/k8sd/controllers/feature_manifest.go
@@ -1,0 +1,78 @@
+package controllers
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/canonical/k8s/pkg/k8sd/database"
+	"github.com/canonical/k8s/pkg/k8sd/features/manifests"
+	"github.com/canonical/k8s/pkg/k8sd/types"
+	"github.com/canonical/k8s/pkg/log"
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/microcluster/v2/state"
+)
+
+// FeatureManifestController is a controller that syncs feature manifests available in the snap with the microcluster database.
+type FeatureManifestController struct {
+	snap      snap.Snap
+	waitReady func()
+}
+
+// NewFeatureManifestController creates a new feature manifest controller.
+func NewFeatureManifestController(snap snap.Snap, waitReady func()) *FeatureManifestController {
+	return &FeatureManifestController{
+		snap:      snap,
+		waitReady: waitReady,
+	}
+}
+
+// Run runs the feature manifest controller.
+// The controller reconciles feature manifests available in the snap with the microcluster database until all manifests are inserted.
+func (c *FeatureManifestController) Run(ctx context.Context, s state.State) {
+	ctx = log.NewContext(ctx, log.FromContext(ctx).WithValues("controller", "feature-manifest"))
+	log := log.FromContext(ctx)
+
+	log.Info("Waiting for node to be ready")
+	// wait for microcluster node to be ready
+	c.waitReady()
+
+	log.Info("Starting feature manifest controller")
+
+	for {
+		log.Info("Reconciling feature manifests")
+
+		var retryRequired bool
+
+		for _, featureManifest := range manifests.Manifests() {
+			log := log.WithValues("feature", featureManifest.GetName(), "version", featureManifest.GetVersion())
+
+			if err := c.reconcile(ctx, s, featureManifest); err != nil {
+				log.Error(err, "failed to reconcile feature manifest")
+				retryRequired = true
+			}
+		}
+
+		if !retryRequired {
+			log.Info("Reconcilation of feature manifests complete")
+			return
+		}
+
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(3 * time.Second):
+		}
+	}
+}
+
+func (c *FeatureManifestController) reconcile(ctx context.Context, s state.State, manifest *types.FeatureManifest) error {
+	if err := s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
+		return database.InsertFeatureManifest(ctx, tx, manifest)
+	}); err != nil {
+		return fmt.Errorf("failed to insert feature manifest: %w", err)
+	}
+
+	return nil
+}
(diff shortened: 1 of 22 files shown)


src/k8s/pkg/k8sd/controllers/feature_manifest.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/canonical/k8s/pkg/k8sd/database"
+	"github.com/canonical/k8s/pkg/k8sd/features/manifests"
+	"github.com/canonical/k8s/pkg/k8sd/types"
+	"github.com/canonical/k8s/pkg/log"
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/microcluster/v2/state"
+)
+
+// HelmChartController is a controller that syncs helm charts available in the snap with the microcluster database.
```
bschimke95, line 17: comments are outdated/wrong here and below in the controller

src/k8s/pkg/k8sd/database/feature_manifest.go
```diff
@@ -0,0 +1,62 @@
...
+	selectTxStmt, err := cluster.Stmt(tx, featureStmts["select"])
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare select statement: %w", err)
+	}
+
+	var manifestBytes []byte
+
+	err = selectTxStmt.QueryRowContext(ctx, name, version).Scan(name, version, manifestBytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query helm chart: %w", err)
```
bschimke95, line 29: I think we need to differentiate between NotFound errors and other errors here and in the helm controller. 
e.g. the upgrade controller may need to wait for a specific helm chart or manifest to be available and for that, requires that differentiation.

berkayoz: We can take different approaches for handling `NotFound` errors. Current approach is to retry the operation if there is any error down the chain. Let's say on the case of helm chart not yet insterting a chart feature needs, it is a matter of retry until it becomes available in the database.

We can also opt-in to check for this exclusively, maybe we can discuss if there are any benefits over just retrying.

src/k8s/pkg/k8sd/controllers/feature_manifest.go
```diff
@@ -0,0 +1,79 @@
...
+			return
+		case <-time.After(3 * time.Second):
+		}
+	}
+}
+
+func (c *FeatureManifestController) reconcile(ctx context.Context, s state.State, manifest *types.FeatureManifest) error {
+	if err := s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
+		err := database.InsertFeatureManifest(ctx, tx, manifest)
+		return err
```
HomayoonAlimohammadi, line 73: nit:
```suggestion
		return database.InsertFeatureManifest(ctx, tx, manifest)
```

src/k8s/pkg/k8sd/database/feature_manifest.go
HomayoonAlimohammadi: Dumb question: I can't see the `GetFeatureManifest` being called anywhere. Am I missing something?

Apart from that, here's what I think: It's correct that we might be retrying on the layer above (the layer responsible for calling `GetFeatureManifest`) and whether it's a DB issue or a `NotFound` error, we retry anyways. HOWEVER, I also think that it might be valuable if we provide a way for the layer above to understand whether it's a `NotFound` error or some other DB issue. 
We can argue that we don't need this differentiation _right now_, and this can be added in the future if needed, but as a personal preference I would suggest we implement this differentiation now. That's because I don't think it's a niche requirement, on the contrary, I believe it's very common.

src/k8s/pkg/k8sd/features/manifests/manifests.go
```diff
@@ -0,0 +1,22 @@
...
+var registeredManifests []*types.FeatureManifest
+
+// Manifests returns the list of registered charts.
+func Manifests() []*types.FeatureManifest {
+	if registeredManifests == nil {
+		return nil
+	}
+	manifestList := make([]*types.FeatureManifest, len(registeredManifests))
+	copy(manifestList, registeredManifests)
+	return manifestList
```
HomayoonAlimohammadi, line 15: nit: We can remove the `nil` check if we're okay with returning an empty array instead of a `nil` one. We shouldn't be facing any issue since I think it's recommended to check the `len` of a slice to check if it's empty.
```suggestion
	manifestList := make([]*types.FeatureManifest, len(registeredManifests))
	copy(manifestList, registeredManifests)
	return manifestList
```

src/k8s/pkg/k8sd/features/manifests/manifests.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// Register charts that are used by k8s-snap.
+// Register is used by the `init()` method in individual packages.
+func Register(charts *types.FeatureManifest) {
+	registeredManifests = append(registeredManifests, charts)
```
HomayoonAlimohammadi, line 21: nit: I know that we're doing the same thing in the `images.Register` function, but should we be concerned about thread safety here? Package `init` functions are called sequentially so I assume not. Unless we're planning to call this in other places as well.

HomayoonAlimohammadi (commented):
Great work, thanks a lot @berkayoz! LGTM overall, just some minor comments and questions. I was also wondering if we can have some unit and e2e tests covering this functionality.

src/k8s/pkg/k8sd/database/feature_manifest.go
berkayoz: The `GetFeatureManifest` is being used in later PRs, I've tried to split up the PRs so this is why.

So I might be missing something but I feel like we could just do `if errors.Is(err, sql.ErrNoRows)` since we are wrapping the errors and call it wherever needed. Do we rather have a function that makes this separation and returns `(exists bool, result any, err error)`?

src/k8s/pkg/k8sd/features/manifests/manifests.go
berkayoz: We can keep this as is, this is more appropriate in terms of returning the zero value (`nil`) of a slice.

src/k8s/pkg/k8sd/features/manifests/manifests.go
berkayoz: This is not going to be called at runtime, as is with charts and images only at init time.

PR #1150, opened 2025-03-05
fix: Fix certs refresh panic

HomayoonAlimohammadi:
### Overview
This PR fixes a panic that happens during certs refresh.


mateoflorido:
> Nice catch, LGTM! I wonder why
> 
> https://github.com/canonical/k8s-snap/blob/0e7d9823a43ea78fa2dc006c3b526620bcf75338/tests/integration/tests/test_clustering.py#L164
> 
> ...didn't catch that

Oh, it is marked as nightly!

HomayoonAlimohammadi:
@mateoflorido we live in a sad world :disappointed:
Tests have betrayed us...

PR #1166, opened 2025-03-07
fix: remove airgapped test filter

bschimke95:
## Description

Right now, only the airgapped tests run on the main branch as this was left-over from the airgapped tests PR

## Solution

Remove the accidentally introduced airgapped test filter


addyess (approved):
100%

PR #1167, opened 2025-03-07
CI: fix update components

louiseschmidtgen:
## Description
Update charts path in component upgrade automation.

### DIFF
--- a/build-scripts/hack/update-component-versions.py
+++ b/build-scripts/hack/update-component-versions.py
@@ -27,7 +27,7 @@ LOG = logging.getLogger(__name__)
 DIR = Path(__file__).absolute().parent
 SNAPCRAFT = DIR.parent.parent / "snap/snapcraft.yaml"
 COMPONENTS = DIR.parent / "components"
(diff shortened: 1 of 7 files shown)


bschimke95 (approved):
LGTM, thanks!

build-scripts/hack/update-component-versions.py
```diff
@@ -27,7 +27,7 @@
 DIR = Path(__file__).absolute().parent
 SNAPCRAFT = DIR.parent.parent / "snap/snapcraft.yaml"
 COMPONENTS = DIR.parent / "components"
-CHARTS = DIR.parent.parent / "k8s" / "manifests" / "charts"
+CHARTS = DIR.parent.parent / "src" / "k8s" / "pkg" / "k8sd" / "features" / "coredns" / "charts"
```
berkayoz, line 30: I believe this is not quite right, it seems like we should update the references to this variable with appropriate paths for components, because I see there is a call for metallb, contour and so on.

PR #1170, opened 2025-03-10
feat: Enables cilium protocol differentiation 

claudiubelu:
## Description

A few CNCF Conformance tests would fail on a Canonical Kubernetes cluster. This PR will address those failures by enabling a few Cilium features.

## Solution

Starting with cilium 1.17 [1][2], it now supports protocol differentiation. Setting the ``loadBalancer.protocolDifferentiation.enabled`` will result in ``bpf-lb-proto-diff`` being set in cilium's configmap, which corresponds to this feature. We need this feature in order for the following Conformance test to pass:

```
[sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance] [sig-network, Conformance]
```

However, that is not enough for the test to pass. We also need to support `HostPort`s. According to the documentation [3], because we're the standard kube-proxy and not cilium's replacement (``kubeProxyReplacement=false``), we need to also set the Helm chart option ``cni.chainingMode=portmap``.

[1] https://github.com/cilium/cilium/issues/9207
[2] https://github.com/cilium/cilium/pull/33434
[3] https://github.com/cilium/cilium/blob/v1.17.1/Documentation/installation/cni-chaining-portmap.rst#portmap-hostport

## Issue

## Backport

## Checklist

- [x] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/src/k8s/pkg/k8sd/features/cilium/network.go
+++ b/src/k8s/pkg/k8sd/features/cilium/network.go
@@ -147,9 +147,10 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, m helm.Client, s state.St
 			"enabled": true,
 		},
 		"cni": map[string]any{
-			"confPath":  "/etc/cni/net.d",
-			"binPath":   "/opt/cni/bin",
-			"exclusive": config.cniExclusive,
+			"confPath":     "/etc/cni/net.d",
+			"binPath":      "/opt/cni/bin",
+			"exclusive":    config.cniExclusive,
+			"chainingMode": "portmap",
 		},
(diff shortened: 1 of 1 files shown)


berkayoz (approved):
LGTM, this should be backported to 1.32 with the possible backport of Cilium 1.17

bschimke95 (approved):
LGTM (resolved conflict - can be merged if CI passes)

bschimke95:
Manually backported in https://github.com/canonical/k8s-snap/pull/1360

PR #1185, opened 2025-03-13
test: Adds test snap to test_version_upgrades

claudiubelu:
## Description
Currently, the test_version_upgrades test does not take into account the configured ``TEST_SNAP`` (the built snap for testing the PR) when testing upgrades. This means that if there's a PR that breaks upgrades, we won't notice until the PR merges and gets published to the snap store, which is too late.

## Solution

This commit addresses this issue by adding the ``TEST_SNAP`` to the upgrade test, helping ensure that the PR doesn't break upgrades.

## Issue

## Backport

``release-1.32``

## Checklist

- [x] CLA signed

### DIFF
--- a/tests/integration/tests/test_version_upgrades.py
+++ b/tests/integration/tests/test_version_upgrades.py
@@ -5,6 +5,7 @@ import logging
 from typing import List
 
 import pytest
+import yaml
 from test_util import config, harness, snap, tags, util
 from test_util.registry import Registry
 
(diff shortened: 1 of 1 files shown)


tests/integration/tests/test_version_upgrades.py
```diff
@@ -37,6 +37,8 @@ def test_version_upgrades(instances: List[harness.Instance], tmp_path):
             min_release=config.VERSION_UPGRADE_MIN_RELEASE,
             max_release="1.32",
         )
+        channels.append(config.SNAP)
```
eaudetcobello, line 40: This might be problematic when we get to k8s 1.34 it will try to upgrade from 1.32 to 1.34, which is not supported.

claudiubelu: Indeed. We'd have to bump the `max_version` above to `1.33` when that's going to be the case.

bschimke95:
@claudiubelu the version upgrade test did not actually run in the CI.
Could you force a test run with a change please (which we can revert after verification)?

claudiubelu:
> @claudiubelu the version upgrade test did not actually run in the CI. Could you force a test run with a change please (which we can revert after verification)?

Rebased.

bschimke95 (approved):

bschimke95:
Please rebase once more to verify that the downgrade test is fixed in main (and not a failure on your change). Then you can directly merge.

PR #1205, opened 2025-03-19
fix: Move snapd sync to custom post-refresh hook

bschimke95:
## Description

We disable the snapd/k8sd sync on upgrades because k8sd is not available at that time. Right now, we re-enable the sync in the snap `post-refresh` hook
The snapd/k8sd sync requires access to k8sd to obtain/sync the values. The snap `post-refresh` hook [is executed before the services started](https://snapcraft.io/docs/supported-snap-hooks#heading--post-refresh) and therefore wait until timeout.

## Solution

This commit introduces a custom post-refresh hook that runs after k8sd started and moves the config sync into that hook. In the future, other machinery also depends on k8sd after a refresh (e.g. feature upgrades) so it makes sense to have this kind of common entry point to extend later.

I verified the upgrade from rev 2500 to this version (`latest/edge/ben-refresh-hook`) works for a three node cluster. Once this fix is merged, we can also re-enable the upgrade/downgrade tests until latest/edge.

image

## Issue
-

## Backport

The post-refresh mechanism was [backported to ](https://github.com/canonical/k8s-snap/pull/1164) 1.32, so we also need to backport this fix.

## Checklist

- [ ] Covered by integration tests - already covered by upgrade tests
- [x] CLA signed

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_start.go
+++ b/src/k8s/pkg/k8sd/app/hooks_start.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rsa"
 	"database/sql"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/canonical/k8s/pkg/k8sd/database"
@@ -23,6 +24,20 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
 			log.FromContext(ctx).Error(err, "Failed to mark node as ready")
 		}
 	}()
+
+	// Check if a refresh was performed and if so, run the custom post-refresh hook
+	isPostRefresh, err := utils.FileExists(a.snap.PostRefreshLockPath())
+	if err != nil {
+		return fmt.Errorf("failed to check if snap is post-refresh: %w", err)
+	}
+	if isPostRefresh {
+		if err := a.postRefreshHook(ctx, s); err != nil {
+			return fmt.Errorf("failed to run post-refresh hook: %w", err)
+		}
+		if err := os.Remove(a.snap.PostRefreshLockPath()); err != nil {
+			return fmt.Errorf("failed to remove post-refresh lock file: %w", err)
+		}
+	}
 
 	// start node config controller
 	if a.nodeConfigController != nil {
--- a/snap/hooks/post-refresh
+++ b/snap/hooks/post-refresh
@@ -3,8 +3,9 @@
 
 k8s::common::setup_env
 
-# Enable the snap config reconciler after the snap refresh completes.
-# The reconcile command will synchronize the k8sd configuration with the snap config
-# and then set the meta orb to "snapd," effectively re-enabling the sync process.
-echo "Re-enabling snapd config sync after snap refresh"
-k8s::cmd::k8s x-snapd-config reconcile
+# Indicate that an update has been performed.
+# Many actions require k8sd to be active which is not
+# the case in the snap hook. Hence, we use a lock file
+# to indicate that an update has been performed and
+# run a custom post-refresh hook in the k8sd service.
+touch "$SNAP_COMMON/lock/post-refresh"
--- a/src/k8s/pkg/k8sd/app/hooks_post_refresh.go
+++ b/src/k8s/pkg/k8sd/app/hooks_post_refresh.go
@@ -1,0 +1,28 @@
+package app
+
+import (
+	"context"
+	"fmt"
+
+	databaseutil "github.com/canonical/k8s/pkg/k8sd/database/util"
+	"github.com/canonical/k8s/pkg/log"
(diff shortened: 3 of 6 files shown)


berkayoz (approved):
LGTM. Lock files are not the cleanest approach however I don't see an easier way for now. This should get us going and we can revisit later.

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -24,6 +25,20 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
 		}
 	}()
 
+	// Check if a refresh was performed and if so, run the custom post-refresh hook
+	isPostRefresh, err := utils.FileExists(a.snap.PostRefreshLockPath())
```
addyess, line 29: what kind of errors can `utils.FileExists` encounter? Permissions to look at a path?  Encountering something that isn't path-like?

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -24,6 +25,20 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to check if snap is post-refresh: %w", err)
+	}
+	if isPostRefresh {
+		if err := a.postRefreshHook(ctx, s); err != nil {
+			return fmt.Errorf("failed to run post-refresh hook: %w", err)
+		}
```
addyess, line 36: So we will fail in k8sd -- but because it didn't fail in the post-refresh hook -- what will snapd do?  It can't rollback the snap anymore?  Are we just left with a k8ds that's on the new revision but isn't rolledback?

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -24,6 +25,20 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
...
+		if err := os.Remove(a.snap.PostRefreshLockPath()); err != nil {
+			return fmt.Errorf("failed to remove post-refresh lock file: %w", err)
```
addyess, line 38: This feels like a super ugly hole to land in.  We've upgraded, we've run the the `postRefreshHook`, but k8sd can't start b/c it can't delete the lock file.

addyess (commented):
Mostly questions regarding the now non-snapd-rollback on genuinely failed refreshes that could have been detected in post-refresh but now won't be.

src/k8s/pkg/k8sd/app/hooks_start.go
bschimke95: It uses os.Stat under the hood so basically every I/O syscall issue can happen, e.g.:
os.ErrNotExist – File does not exist
os.ErrPermission – Permission denied
syscall.EINVAL – Invalid file name or argument
syscall.ELOOP – Too many symbolic links encountered
syscall.EIO – I/O error (disk or hardware failure)
syscall.ENOSPC – No space left on device (rare for os.Stat)
syscall.ENAMETOOLONG – File name too long
syscall.EROFS – Read-only file system
syscall.ENOTDIR – A component of the path is not a directory
syscall.EFAULT – Invalid memory address (rare case)

src/k8s/pkg/k8sd/app/hooks_start.go
bschimke95: Well, yes but also no. Removing a file should just work in 99% of the cases or be transient. Worst case, the user has to restart the node. Then, the post-refresh will kick in again but it should be mostly (?!) idempotent. And then the lock removal would be retried.

addyess: Good enough for me

src/k8s/pkg/k8sd/app/hooks_start.go
addyess: ack -- a whole bunch of icky things that neither snapd rollback nor k8ds can manage resolving.

src/k8s/pkg/k8sd/app/hooks_start.go
bschimke95: That is a good point. Yes, we basically loose snapds roll-back capabilities but I feel like this is what we signed up for if we tie k8sd configuration to snap configuration. 
Eventually, we'll track an upgrade in a CRD (#1207). If we have that, then the node would mark the upgrade as failed if the post-refresh hook fails and the orchestration tool would be in charge of the rollback. Which IMHO makes more sense anyway.

addyess: Well lovely.  I think owning our own space rather than relying on snapd to "rollback" is a better solution to.  At the moment it LGTM, i just needed to make sure I was understanding what you're selling here.

PR #1208, opened 2025-03-19
fix: Retry snap install command in tests

bschimke95:
The snapstore occasionally fails to properly install the snap because of transient errors. We should not fail on those cases but instead just retry to install the snap.

See [here](https://github.com/canonical/k8s-snap/actions/runs/13946620511/job/39036184552?pr=1207) for an example run.


bschimke95:
Note: It turned out that the underlying issue is caused by snapd and would not be fixed by this (see #1224) but I guess it does not hurt to retry `snap install` failures as they are basically always transient store issues.

PR #1210, opened 2025-03-19
test: Re-enable upgrade/downgrade tests from/to latest

bschimke95:
The issue that prevented the upgrade/downgrade to `latest/edge` was fixed, so that we can perform this test again now.

### DIFF
--- a/tests/integration/tests/test_version_upgrades.py
+++ b/tests/integration/tests/test_version_upgrades.py
@@ -44,7 +44,6 @@ def test_version_upgrades(
             cp.arch,
             include_latest=False,
             min_release=config.VERSION_UPGRADE_MIN_RELEASE,
-            max_release="1.32",
         )
         if len(channels) < 2:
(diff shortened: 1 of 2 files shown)


tests/integration/tests/test_version_upgrades.py
```diff
@@ -35,7 +35,6 @@ def test_version_upgrades(instances: List[harness.Instance], tmp_path):
             cp.arch,
             include_latest=False,
             min_release=config.VERSION_UPGRADE_MIN_RELEASE,
-            max_release="1.32",
```
claudiubelu, line 38: Since the intention is to also run the upgrade to `latest/edge`, you should remove the line `include_latest=False` above. This line prevents the test from trying to upgrade to 1.33. I'm not sure if the 1.33 snap channel has the upgrade fix.

eaudetcobello (approved):

PR #1211, opened 2025-03-19
tests: test rawfile-localpv dependencies bumps

eaudetcobello:
I bumped the rawfile-localpv dependencies to resolve some CVEs in https://github.com/canonical/rawfile-localpv/pull/26. This PR is just to test that rawfile-localpv still works.


bschimke95:
Not relevant anymore - rawfile localpv is updated already

PR #1213, opened 2025-03-20
fix: Mattermost bot returncode and ping

bschimke95:
* The returncode of `jq` was not properly evaluated causing the statement to always succeed
* Only ping @k8s-engineers to avoid management pings

### DIFF
--- a/.github/workflows/nightly-test.yaml
+++ b/.github/workflows/nightly-test.yaml
@@ -126,7 +126,7 @@ jobs:
         run: |
           RESULTS=$(echo '${{ toJson(needs) }}' | jq -c '[.[] | .result]')
 
-          if echo $RESULTS | jq 'all(. == "success")'; then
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-test.yaml
```diff
@@ -126,7 +126,7 @@ jobs:
         run: |
           RESULTS=$(echo '${{ toJson(needs) }}' | jq -c '[.[] | .result]')
 
-          if echo $RESULTS | jq 'all(. == "success")'; then
+          if echo $RESULTS | jq -e 'all(. == "success")'; then
```
addyess, line 129: ah:

```sh
⇒ echo '["fail"]' | jq 'all(. == "success")' && echo yay || echo boo 
false
yay

⇒ echo '["fail"]' | jq -e 'all(. == "success")' && echo yay || echo boo
false
boo
```

addyess (approved):
LGTM

PR #1216, opened 2025-03-21
feat: Add leadership to feature reconcilation

berkayoz:
# 🚧 DO NOT MERGE 🚧
The PR on https://github.com/canonical/k8s-snap-api/pull/31 has to go in first, a new tag should be created and this PR should be updated to use the new tag. 

- A helper util is added to check if this current node is the microcluster leader.
- Feature controller is upgraded to check for leadership and skip reconcilation on non-leader nodes.
- A k8sd API endpoint is introduced to enable triggering feature reconcilation from other nodes.
- Cluster config endpoint is updated to notify the controller or forward the reconcilation request based on leadership status.

## Backport

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed


src/k8s/pkg/k8sd/api/endpoints.go
```diff
@@ -124,6 +124,12 @@ func (e *Endpoints) Endpoints() []rest.Endpoint {
 			Put:  rest.EndpointAction{Handler: e.putClusterConfig, AccessHandler: e.restrictWorkers},
 			Get:  rest.EndpointAction{Handler: e.getClusterConfig, AccessHandler: e.restrictWorkers},
 		},
+		// Features reconciliation
+		{
+			Name: "ReconcileFeatures",
+			Path: apiv1.ReconcileFeaturesRPC,
+			Post: rest.EndpointAction{Handler: e.postFeaturesReconcile, AccessHandler: e.restrictWorkers},
```
HomayoonAlimohammadi, line 131: Quick note, and not related to this PR. I believe the `restrictWorkers` restriction is not an "actual" barrier as the workers can bypass this by simply deleting the `worker` lock file. This can be generalized to the `IsWorker` function as well. We simply can't blindly trust it. I'm not sure how big of a concern this might be, but I suspect for certain things we really don't want the workers to be able to access certain endpoints. E.g. an endpoint to replace or retrieve the certificates might be abused by a malicious worker.

src/k8s/pkg/k8sd/api/features_reconcile.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	isLeader, err := microclusterutil.IsLeader(s)
+	if err != nil {
+		return response.InternalError(fmt.Errorf("failed to check if node is leader: %w", err))
+	}
+
+	if !isLeader {
+		return response.InternalError(fmt.Errorf("feature reconciliation is not allowed on non-leader nodes"))
+	}
```
HomayoonAlimohammadi, line 27: Do we want to also forward the reconciliation here as well? I'm thinking about when the leader changes after we call this endpoint from another node. So we used to be the leader, but now we're not. Is that a concern or are we happy to fail the request in this case?
If we decided to also forward here we can simply remove this block and instead of `NotifyFeatureController`, put `NotifyOrForwardFeatureReconciliation`.

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -78,12 +78,10 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
 	if a.featureController != nil {
 		go a.featureController.Run(
 			ctx,
+			s,
 			func(ctx context.Context) (types.ClusterConfig, error) {
 				return databaseutil.GetClusterConfig(ctx, s)
 			},
-			func() state.State {
-				return s
-			},
```
HomayoonAlimohammadi, line 86: I suspect we were not benefiting from the lazy evaluation here, right? I can't help but think there was a reason to use lazy evaluation like this to begin with... Because just passing `s` is way more obvious and straight forward that it couldn't be by accident that we were using lazy eval.
Maybe the state was still undergoing change even after we call this function?
Yea it has to be it. Have a look at this example: https://go.dev/play/p/tcgKIhuueg3

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -72,41 +73,39 @@ func NewFeatureController(opts FeatureControllerOpts) *FeatureController {
...
 
-	s := getState()
-
-	go c.reconcileLoop(ctx, getClusterConfig, setFeatureStatus, features.Network, c.triggerNetworkCh, c.reconciledNetworkCh, func(cfg types.ClusterConfig) (types.FeatureStatus, error) {
+	go c.reconcileLoop(ctx, s, getClusterConfig, setFeatureStatus, features.Network, c.triggerNetworkCh, c.reconciledNetworkCh, func(cfg types.ClusterConfig) (types.FeatureStatus, error) {
 		return features.Implementation.ApplyNetwork(ctx, c.snap, s, cfg.APIServer, cfg.Network, cfg.Annotations)
 	})
 
-	go c.reconcileLoop(ctx, getClusterConfig, setFeatureStatus, features.Gateway, c.triggerGatewayCh, c.reconciledGatewayCh, func(cfg types.ClusterConfig) (types.FeatureStatus, error) {
+	go c.reconcileLoop(ctx, s, getClusterConfig, setFeatureStatus, features.Gateway, c.triggerGatewayCh, c.reconciledGatewayCh, func(cfg types.ClusterConfig) (types.FeatureStatus, error) {
```
HomayoonAlimohammadi, line 88: If we decided to go by lazy eval for state, I guess we should change these to lazy evals as well.

HomayoonAlimohammadi (commented):
Great work, thanks a lot @berkayoz! Looks very good overall. Left some suggestions and comments.

src/k8s/pkg/k8sd/api/features_reconcile.go
berkayoz: I've thought about this as well but instead opted in to have a single simple endpoint without any forwarding magic to avoid any looping, requests jumping weirdly. I think the request should fail cleanly and retry should be handled by the caller.

src/k8s/pkg/k8sd/app/hooks_start.go
berkayoz: I don't think this was needed for the state atleast, initially I've passed the state like this to fit the "convention", however it does feel dumb to have a function that just returns a reference(since it is a struct).

src/k8s/pkg/k8sd/controllers/feature.go
berkayoz: Not all of them, we still need to run/evaluate `getClusterConfig` later since it queries the DB. What we could do is though just move the `getClusterConfig` logic in the controller itself since we now have access to the `state.State`.

src/k8s/pkg/k8sd/app/hooks_start.go
HomayoonAlimohammadi: I guess the `state.State` is an interface with the implementation being a pointer to `InternalState` in microcluster.
That being said I also don't see why we need to do lazy loading.

src/k8s/go.mod
```diff
@@ -7,7 +7,7 @@ toolchain go1.23.4
 require (
 	dario.cat/mergo v1.0.0
 	github.com/canonical/go-dqlite/v2 v2.0.0
-	github.com/canonical/k8s-snap-api v1.0.20
+	github.com/canonical/k8s-snap-api v1.0.21-0.20250321082427-cc03cf9c3fca
```
HomayoonAlimohammadi, line 10: Just a placeholder to not forget replacing this with an official release.

HomayoonAlimohammadi (approved):
Really loved this one. Great work @berkayoz! LGTM

berkayoz:
Closed in favor of #1268

PR #1222, opened 2025-03-24
fix: Add quotations to service argument values to care for escape characters

rapour:
## Description

The current `lib.sh`'s `execute_service` has a mechanism for parsing multi-line arguments that fail if complex/escape arguments (e.g. including `%`, `<` and `>`) are provided without quotation marks. This PR suggests an improvement so that user won't be confused as to why their bootstrap process fails. 

The current implementation fails on these configuration files: (Note the quotes on `--eviction-hard` value)
```
cluster-config:
  network:
    enabled: true
  dns:
    enabled: true
  metrics-server:
    enabled: true
extra-node-kubelet-args:
  --authentication-token-webhook-cache-ttl: 3m
  --anonymous-auth: "false"
  --eviction-hard: memory.available<100Mi,nodefs.available<1Gi,imagefs.available<1Gi
```
```
cluster-config:
  network:
    enabled: true
  dns:
    enabled: true
  metrics-server:
    enabled: true
extra-node-kubelet-args:
  --authentication-token-webhook-cache-ttl: 3m
  --anonymous-auth: "false"
  --eviction-hard: "memory.available<100Mi,nodefs.available<1Gi,imagefs.available<1Gi"
```
```
cluster-config:
  network:
    enabled: true
  dns:
    enabled: true
  metrics-server:
    enabled: true
extra-node-kubelet-args:
  --authentication-token-webhook-cache-ttl: 3m
  --anonymous-auth: "false"
  --eviction-hard: 'memory.available<100Mi,nodefs.available<1Gi,imagefs.available<1Gi'
```
It passes the following configuration:
```
cluster-config:
  network:
    enabled: true
  dns:
    enabled: true
  metrics-server:
    enabled: true
extra-node-kubelet-args:
  --authentication-token-webhook-cache-ttl: 3m
  --anonymous-auth: "false"
  --eviction-hard: "'memory.available<100Mi,nodefs.available<1Gi,imagefs.available<1Gi'"

```

This PR suggests a change that wraps all parameters in double quotation before adding it to argument files


k8s/lib.sh
```diff
@@ -179,24 +179,30 @@ k8s::util::wait_kube_apiserver() {
   done
 }
 
+# Source arguments and substitute environment variables. Will fail if we cannot read the file.
+k8s::util::parse_arguments_file() {
+  arguments_file_path="$1"
+
+  declare -ag parsed_args="($(cat $arguments_file_path))"  
+}
```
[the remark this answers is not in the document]
rapour: It doesn't. It's a WIP. I just want to have a nice test suite before changing this function. Probably the solution would be to switch from `declare` to something like `read -r -d '' -a args`

src/k8s/pkg/snap/util/arguments.go
```diff
@@ -79,7 +79,7 @@ func UpdateServiceArguments(snap snap.Snap, serviceName string, updateMap map[st
 		existingArguments[key] = struct{}{}
 		if newValue, ok := updateMap[key]; ok {
 			// update argument with new value
-			newArguments = append(newArguments, fmt.Sprintf("%s=%s", key, newValue))
+			newArguments = append(newArguments, fmt.Sprintf(`%s="%s"`, key, newValue))
```
berkayoz, line 82: Why do we change the logic here instead of quoting when storing these in the file in `SerializeArgumentFile`? Especially if we do `value = strings.Trim(value, "\"")` while reading?

rapour: `SerializeArgumentFile` doesn't seem to be used anywhere such as the `UpdateSerivceArguments`. Am I missing something? Should we have been using it?

ader1990:
Hello, I would remove the chore notation, as this has the potential of a fix.

src/k8s/pkg/snap/util/arguments.go
berkayoz: Seems like this function is not being used now, we can address it in another PR.

berkayoz (approved):
LGTM

PR #1224, opened 2025-03-24
fix: Pre-install xdelta3 to mitigate LXD rate-limit issues

bschimke95:
## Description

Recently, installing LXD has triggered rate-limit problems, resulting in context cancelled errors during snap install. 

## Solution

Based on feedback from the Snap Store team, pre-installing xdelta3 helps alleviate these issues.
This is a temporary solution until the snapd team resolved the issue.

### DIFF
--- a/.github/actions/install-lxd/action.yaml
+++ b/.github/actions/install-lxd/action.yaml
@@ -12,6 +12,10 @@ runs:
     - name: Install lxd snap
       shell: bash
       run: |
+        # note(ben): Temporary workaround until snapd/snapcraft issue is resolved.
+        sudo apt-get update
+        sudo apt-get install xdelta3 --yes
+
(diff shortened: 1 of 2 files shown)


.github/actions/install-lxd/action.yaml
```diff
@@ -12,6 +12,10 @@ runs:
     - name: Install lxd snap
       shell: bash
       run: |
+        # note(ben): Temporary workaround until snapd/snapcraft issue is resolved.
+        sudo apt-get update
+        sudo apt-get install xdelta3 --yes
+
```
addyess, line 18: the upstream package doesn't seem to have any dangers, so LGTM

bschimke95:
The airgap tests fail but the fix is delivered here: https://github.com/canonical/k8s-snap/pull/1228

Merging this

PR #1207, opened 2025-03-19
feat: Add Update CRD

bschimke95:
## Description

Add a custom resource to track upgrades of the cluster. Also, a mechanism for loading all CRDs in the k8s/crds directory is created. This is preparation work for the feature upgrades.

## Issue

-

## Backport

The feature upgrade work will not be backported.

## Checklist

- [x] PR title formatted as `type: title`
- [ ] Covered by unit tests - created follow-up card
- [x] Covered by integration tests
- [ ] Documentation updated - N/A
- [x] CLA signed
- [x] Backport label added if necessary 

If any item on the checklist is not complete, please provide justification why.

PR #1228, opened 2025-03-25
fix: Execute post-refresh hook after node is ready

bschimke95:
## Description

Initially, the custom post-refresh hook was executed in the microcluster `onStart` hook. This hook is executed right after the microcluster daemon started which could lead to "Database not initialized" error if the node was not bootstrapped yet.

## Solution

By introducing a nodeReadyHook and execute the refresh from there, we ensure that the refresh is only executed after a `snap refresh` but not after a `snap install`

Note that this hook is also used in https://github.com/canonical/k8s-snap/pull/1207 but since we don't want to backport this PR, we need to introduce the hook here.

## Issue

Fixes the CI tests together with https://github.com/canonical/k8s-snap/pull/1224
Follow-up fix for https://github.com/canonical/k8s-snap/pull/1205

## Backport

We backported the initial [PR ](https://github.com/canonical/k8s-snap/pull/1205), so this fix also needs to be backported.

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -71,6 +71,7 @@ jobs:
           TEST_VERSION_UPGRADE_MIN_RELEASE: "1.31"
           TEST_STRICT_INTERFACE_CHANNELS: "recent 6 strict"
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
+          TEST_SKIP_CLEANUP: 1
         run: |
           cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }}
       - name: Prepare inspection reports
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -33,7 +33,7 @@ on:
 jobs:
   test-integration:
     name: Tests
-    runs-on: ${{ inputs.arch == 'arm64' && 'self-hosted-linux-arm64-noble-large' || 'self-hosted-linux-amd64-noble-large' }}
+    runs-on: ${{ inputs.arch == 'arm64' && 'self-hosted-linux-arm64-noble-large' || 'self-hosted-linux-amd64-jammy-large' }}
     steps:
       - name: Check out code
         uses: actions/checkout@v4
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -72,7 +72,7 @@ jobs:
           TEST_STRICT_INTERFACE_CHANNELS: "recent 6 strict"
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
         run: |
-          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }}
+          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }} -k test_skip_services_stop_on_remove
       - name: Prepare inspection reports
         if: failure()
         run: |
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -71,7 +71,6 @@ jobs:
           TEST_VERSION_UPGRADE_MIN_RELEASE: "1.31"
           TEST_STRICT_INTERFACE_CHANNELS: "recent 6 strict"
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
-          TEST_SKIP_CLEANUP: 1
         run: |
           cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }}
       - name: Prepare inspection reports
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -72,7 +72,7 @@ jobs:
           TEST_STRICT_INTERFACE_CHANNELS: "recent 6 strict"
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
         run: |
-          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }} -k test_skip_services_stop_on_remove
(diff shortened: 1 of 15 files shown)


src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/canonical/microcluster/v2/state"
 )
 
+// postRefreshHook is executed after the node is ready after a `snap refresh` operation
+// See nodeReadyHook for details on when a node is considered ready.
+// Note that the postRefreshHook is NOT executed after a `snap install` operation which is
+// different to the underlying snap hook.
```
berkayoz, line 16: What do you mean by this exactly? Do you mean `post-refresh` of snapd is executed on `snap install` command as well?

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -278,6 +278,10 @@ func (a *App) markNodeReady(ctx context.Context, s state.State) error {
 		return fmt.Errorf("failed to wait for kubernetes endpoint: %w", err)
 	}
 
+	if err := a.onNodeReady(ctx, s); err != nil {
+		return fmt.Errorf("failed to execute onNodeReady hook: %w", err)
+	}
+
```
berkayoz, line 284: Failing here means the node will not be marked as ready if any of the hooks in `onNodeReady` fails right? Should we update the comment for the `markNodeReady` func?

And I assume this is the reason why we are not doing something similar to our controllers (waitReady)

berkayoz (commented):
LGTM overall, some quick comments

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
bschimke95: Yes, If I see this correctly then the post-refresh hook is executed when snap installing
https://snapcraft.io/docs/supported-snap-hooks#heading--post-refresh

bschimke95: tbh, I think separate hooks for install and refresh would make sense in the snap but I guess we need to work with what we have.

src/k8s/pkg/k8sd/app/app.go
bschimke95: Yes, exactly the node should not be marked as ready.

I will add the command after the CI run so that it not needs to spin up everything again.

tests/integration/tests/test_airgapped.py
```diff
@@ -142,10 +142,12 @@ def test_airgapped_with_proxy_setup_and_image_mirror(
     for image in images:
         link = "/".join(image.split("/")[1:])
         tag = f"{registry_ip}:{REGISTRY_PORT}/{link}"
+        # Pull the image from the upstream registry and push it to the local registry.
+        # Pipe the pull and push output to /dev/null as ctr is very verbose.
         registry.exec(
             (
                 "export $(grep -v '^#' /etc/environment | xargs) && "
-                + f"/snap/k8s/current/bin/ctr images pull --all-platforms {image}"
+                + f"/snap/k8s/current/bin/ctr images pull --all-platforms {image} > /dev/null"
```
bschimke95, line 150: ctr is very verbose (progress bar that cannot be turned off). Piping this into the void.

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -252,6 +252,7 @@ func (a *App) Run(ctx context.Context, customHooks *state.Hooks) error {
 // The node is ready if:
 // - the microcluster database is accessible
 // - the kubernetes endpoint is reachable.
+// - the onNodeReady hook succeeds.
 func (a *App) markNodeReady(ctx context.Context, s state.State) error {
```
louiseschmidtgen, line 256: Where is the function invocation for `markNodeReady`? Is this function retried on errors?

bschimke95: https://github.com/canonical/k8s-snap/blob/a85fe3973565f01efb26231dc252ede7f47f57e4/src/k8s/pkg/k8sd/app/hooks_start.go#L23

louiseschmidtgen: Should we have a retry upon failure to mark the node as ready?

bschimke95: I'm not sure if I understand - the `markNodeReady` function already retries until the readiness status is reached, no?

louiseschmidtgen: Do we retry on when receiving an error from `err := a.onNodeReady`?

bschimke95: IMHO we should not retry onNodeReady
The steps inside this hook should be stable or perform retries as required.

.github/workflows/e2e-tests.yaml
```diff
@@ -72,7 +72,7 @@ jobs:
           TEST_STRICT_INTERFACE_CHANNELS: "recent 6 strict"
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
         run: |
-          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }}
+          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }} -k test_skip_services_stop_on_remove
```
bschimke95, line 75: revert after investigation

.github/workflows/e2e-tests.yaml
```diff
@@ -84,3 +84,6 @@ jobs:
         with:
           name: ${{ env.artifact_name }}
           path: ${{ github.workspace }}/inspection-reports.tar.gz
+      - name: Setup tmate session
```
bschimke95, line 87: same here

.github/workflows/lint_and_integration.yaml
```diff
@@ -107,7 +107,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu:24.04"]
+        os: ["ubuntu:22.04"]
```
bschimke95, line 110: same

tests/integration/tests/test_clustering.py
```diff
@@ -90,6 +91,12 @@ def test_worker_nodes(instances: List[harness.Instance]):
     assert "worker" in util.get_local_node_status(joining_node)
     assert "worker" in util.get_local_node_status(other_joining_node)
 
+    # Expect all snap services to be up and running. Add a little sleep to
+    # ensure that services not just started but are also stable.
+    time.sleep(5)
+    for instance in instances:
+        util.check_snap_services(instance)
```
petrutlucian94, line 98: Should we just move ``check_snap_services`` here? https://github.com/canonical/k8s-snap/blob/c51fe1347c9747f1ca6cdf771a9ac51d7b83303a/tests/integration/tests/test_util/util.py#L297

tests/integration/tests/test_util/util.py
```diff
@@ -339,6 +339,11 @@ def is_node_ready(
         LOG.info(f"Node not ready yet: {node_name}, failed to retrieve node info: {ex}")
         return False
 
+    # Expect all snap services to be up and running. Add a little sleep to
+    # ensure that services not just started but are also stable.
+    time.sleep(5)
+    check_snap_services(node_name)
```
petrutlucian94, line 345: I'd catch exceptions raised by ``check_snap_services`` and just return False since that's what the callers expect.

You could move it in the try/except block above and update the exception message to ``LOG.info(f"Node not ready yet: {node_name}, exception: {ex}")

tests/integration/tests/test_clustering_race.py
```diff
@@ -7,16 +7,23 @@
...
-@pytest.mark.node_count(3)
+# Note(ben): Commented out as otherwise the setup would still happen for xfail tests.
+# @pytest.mark.node_count(3)
 @pytest.mark.tags(tags.NIGHTLY)
 def test_wrong_token_race(instances: List[harness.Instance]):
+    # Note(ben): k8s-dqlite sometimes takes very long to shutdown (to be investigated) and
+    # since microcluster has a 30s fixed timeout for the remove hooks this test sometimes fails.
+    # The timeout will be configurable in https://github.com/canonical/microcluster/pull/365)
+    # The k8s-dqlite issue will be investigated separately.
+    pytest.xfail("This test is currently flaky because of a k8s-dqlite shutdown issue.")
```
bschimke95, line 18: This is fixed here: https://github.com/canonical/k8s-snap/pull/1262
I wanted those to be separate PRs.

src/k8s/pkg/k8sd/app/hooks_node_ready.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/canonical/k8s/pkg/utils"
+	"github.com/canonical/microcluster/v2/state"
+)
+
+// onNodeReady is called when the node is ready, right after the wait group is released.
+// The node is ready if:
+// - the microcluster database is accessible
+// - the kubernetes endpoint is reachable.
+// Note that this is not a microcluster hook, but a custom k8sd hook.
+func (a *App) onNodeReady(ctx context.Context, s state.State) error {
```
louiseschmidtgen, line 18:
```suggestion
func (a *App) onK8sdNodeReady(ctx context.Context, s state.State) error {
```
It might be handy to clarify it is a k8sd hook in the function name.

louiseschmidtgen (approved):
LGTM, tiny suggestion (you can use it or not)

PR #1231, opened 2025-03-25
ci: retry launchpad clone

eaudetcobello:
Launchpad occasionally returns 500 error during clone which fails the whole CI. This adds a simple retry to try and prevent the fail if possible.


bschimke95 (commented):
I guess something went wrong here - the diff does not fit the PR description.

bschimke95 (approved):

PR #1236, opened 2025-03-26
fix: Cilium possible conflict with fan networking

rapour:
Cilium feature creates a VXLAN interface on destination port `8472` which sometimes conflicts with other tools using the same port for their own VXLAN interface e.g., fan-networking. In such scenario, the user should be notified of the conflict so they can change the Cilium's destination port through annotations. This PR suggests changes to check for possible conflicts and also introduces a temporary workaround for an [issue in the upstream](https://github.com/cilium/cilium/issues/38581).


src/k8s/go.mod
mateoflorido, line 1: I'm not sure if adding a whole new dependency is worth it, given that we are using two analogous commands (i.e., `ip -j link show type vxlan` and `ip link delete`. I'm not sure if the `/sys/class/net` has something useful to infer the type of the link, probably not as VXLAN requires Netlink right? You've probably dealt with this before, so let me know if my suggestion to use `exec` calls is completely crazy.

rapour: Thank you for bringing it up. Actually I was going to address it in the description but I totally forgot. I had researched a little bit and apparently Go doesn't have a standard library to get detailed interface information like destination ports. The alternative seems to be opening up a socket to the kernel's `netlink` and then encode/decode the binary streams, which introduces a lot of maintenance. The `vishvananda/netlink` is used in big projects like Moby and Cilium and I thought it may not be that bad to use it. If we decide to go with the netlink socket, I'm up for it. I just couldn't do the trade-off weighing correctly.

rapour: Agreed. Let's rethink this a bit more

rapour: Based on the conversations we had in the team, I have changed the functions to use `iproute2`'s `ip` command.

src/k8s/pkg/k8sd/features/cilium/chart.go
```diff
@@ -46,4 +46,8 @@ var (
 
 	// ciliumOperatorImageTag is the tag to use for the cilium-operator image.
 	ciliumOperatorImageTag = "1.17.1-ck0"
+
+	ciliumDefaultVXLANPort int = 8472
```
berkayoz, line 50:
```suggestion
	ciliumDefaultVXLANPort = 8472
```

Do we need to put `int` here explicitly?

src/k8s/pkg/k8sd/features/cilium/internal.go
```diff
@@ -23,6 +24,18 @@ type config struct {
...
+
+func validatePort(portStr string) (int, error) {
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		return 0, errors.New("invalid port: not a number")
+	}
+	if port < 1 || port > 65535 {
+		return 0, errors.New("invalid port: out of range")
+	}
+	return int(port), nil
```
berkayoz, line 38: Do we really need `int(port)` here, doesn't `strconv.Atoi` already return an `int`?

src/k8s/pkg/utils/netlink.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	var ipCmdLinks []struct {
+		IfName   string `json:"ifname"`
+		LinkInfo struct {
+			InfoData struct {
+				Port int `json:"port"`
+			} `json:"info_data"`
+		} `json:"linkinfo"`
+	}
```
berkayoz, line 31: Let's move the custom struct types outside the function body and declare them above.

berkayoz (commented):
Great work, left some minor comments

src/k8s/pkg/utils/netlink.go
rapour: Done

src/k8s/pkg/utils/netlink.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if err := json.Unmarshal(out, &ipCmdLinks); err != nil {
+		return vxlanDevices, fmt.Errorf("unmarshaling ip command output failed: %w", err)
+	}
+
+	for _, link := range ipCmdLinks {
+		ifi, err := net.InterfaceByName(link.IfName)
```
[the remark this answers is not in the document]
rapour: Done

src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -133,6 +166,19 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, s state.State, apiserver
 		ciliumNodePortValues["directRoutingDevice"] = config.directRoutingDevice
 	}
 
+	vxlanPort := ciliumDefaultVXLANPort
+	if config.tunnelPort != 0 {
+		vxlanPort = int(config.tunnelPort)
+	}
```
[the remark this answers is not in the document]
rapour: we have to check for the cilium port irrespective of whether we set it as a helm value or not. I think it might be better to explicitly set the port when we are checking for it. This way, we will be sure if Cilium change their default port anyday in the future, we don't check for a wrong port without knowing it. Because of that I'm inclined to the first option if It's OK with you

nhennigan (commented):
Just a few minor things from the docs side. Great work!

docs/canonicalk8s/snap/reference/annotations.md
```diff
@@ -1,10 +1,11 @@
 # Annotations
 
 This page outlines the annotations that can be configured during cluster
-[bootstrap]. To do this, set the `cluster-config.annotations` parameter in
-the bootstrap configuration.
+[bootstrap] or when setting cluster configuration later on.
 
-For example:
+For example you can specify annotations during the bootstrap process.
```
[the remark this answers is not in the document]
rapour: Fixed, thanks a lot

docs/canonicalk8s/snap/reference/annotations.md
```diff
@@ -1,10 +1,11 @@
 # Annotations
 
 This page outlines the annotations that can be configured during cluster
-[bootstrap]. To do this, set the `cluster-config.annotations` parameter in
-the bootstrap configuration.
+[bootstrap] or when setting cluster configuration later on.
```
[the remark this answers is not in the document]
rapour: Done

src/k8s/pkg/utils/netlink.go
mateoflorido, line 1: Most of the comments bellow are regarding the naming convention for acronyms: https://google.github.io/styleguide/go/decisions#initialisms

src/k8s/pkg/utils/netlink.go
```diff
@@ -0,0 +1,59 @@
+package utils
+
+import (
+	"encoding/json"
+	"fmt"
+	"net"
+	"os/exec"
+)
+
+type Vxlan struct {
```
mateoflorido, line 10: I think this is very generic, we can use something like:
```suggestion
type VXLANInterface struct {
```

src/k8s/pkg/utils/netlink.go
```diff
@@ -0,0 +1,59 @@
...
+	net.Interface
+	Port int
+}
+
+var ipCmdLinks []struct {
```
mateoflorido, line 15:
```suggestion
var ipLinks []struct {
```

src/k8s/pkg/utils/netlink.go
```diff
@@ -0,0 +1,59 @@
...
+	IfName   string `json:"ifname"`
+	LinkInfo struct {
+		InfoData struct {
+			Port int `json:"port"`
+		} `json:"info_data"`
+	} `json:"linkinfo"`
+}
+
+func VxlanDevices() ([]Vxlan, error) {
```
mateoflorido, line 24:
```suggestion
func ListVXLANInterfaces() ([]VXLANInterfaces, error) {
```

mateoflorido (commented):
Great work, I've left a few comments targeting the code style

src/k8s/pkg/utils/netlink.go
rapour: Done

src/k8s/pkg/utils/netlink.go
rapour: Done

src/k8s/pkg/utils/netlink.go
rapour: Done

src/k8s/pkg/k8sd/features/cilium/internal.go
```diff
@@ -81,5 +94,18 @@ func internalConfig(annotations types.Annotations) (config, error) {
...
 
+	if v, ok := annotations.Get(apiv1_annotations.AnnotationTunnelPort); ok {
+		tunnelPort, err := validatePort(v)
+		if err != nil {
+			return config{}, fmt.Errorf("failed to parse Tunnel encapsulation port: %w", err)
+		}
+
+		if tunnelPort == 0 {
+			tunnelPort = ciliumDefaultVXLANPort
+		}
```
HomayoonAlimohammadi, line 105: QQ (nit): Don't we want to set the default only when the user is not passing the value? It shouldn't really matter in this case since a 0 port is invalid but I'm thinking about something like this:
```go
if v, ok := annotations.Get(apiv1_annotations.AnnotationTunnelPort); ok {
    tunnelPort, err := validatePort(v)
    // handle error
    c.tunnelPort  = tunnelPort
} else {
    c.tunnelPort = ciliumDefaultVXLANPort
}
```

HomayoonAlimohammadi (commented):
Great work, thanks a lot @rapour! Looks very good. Left some nits and questions.

src/k8s/pkg/k8sd/features/cilium/internal.go
rapour: Actually, leaving it as is would have created a bug! Thank you Hue

src/k8s/pkg/utils/netlink.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	return vxlanDevices, nil
+}
+
+func RemoveLink(name string) error {
+	cmd := exec.Command("ip", "link", "delete", name)
+
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("running ip command failed: %s", string(out))
```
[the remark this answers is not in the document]
rapour: The error includes only the exit code and the `out` contains the actual `stderr` that the executed command returns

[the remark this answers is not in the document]
HomayoonAlimohammadi: Got it. That's odd though. Maybe we can refactor this later. IMO it's a better practice to not rely on other returned values when there is an error, at least in Go. Something for later tho.

HomayoonAlimohammadi (approved):
Looks great, thanks a lot @rapour!

PR #1246, opened 2025-03-28
ci: add sbom retry

eaudetcobello:
The SBOM step clones launchpad repositories. These clones sometimes result in a failed SBOM step, even though a retry could have prevented the failure. This implements a retry mechanism by adding retry_n and retry_delay with 5 as the default values.

### DIFF
--- a/build-scripts/hack/util.py
+++ b/build-scripts/hack/util.py
@@ -1,17 +1,18 @@
 import contextlib
-from typing import Any, Generator
-import tempfile
-import subprocess
 import logging
-from urllib.request import urlopen
+import subprocess
+import tempfile
+import time
 from pathlib import Path
+from typing import Any, Generator
+from urllib.request import urlopen
 
 LOG = logging.getLogger(__name__)
 
 
 @contextlib.contextmanager
 def git_repo(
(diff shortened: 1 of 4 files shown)


build-scripts/hack/util.py
```diff
@@ -30,7 +35,17 @@ def git_repo(
...
+            try:
+                parse_output(cmd)
+                break
+            except subprocess.CalledProcessError as e:
+                if attempt == retry_n - 1:
+                    raise e
+                LOG.warning(
+                    f"Failed to clone {repo_url} @ {repo_tag}, retrying in {retry_delay}s"
+                )
+                time.sleep(retry_delay)
```
HomayoonAlimohammadi, line 48: Maybe we can use `stubbornly`? And also maybe log something in case we failed even after all those retries?

HomayoonAlimohammadi (approved):
Thanks a lot @eaudetcobello! LGTM overall, just a minor comment.

build-scripts/hack/util.py
eaudetcobello: Stubbornly isn't available here :/ I would have to duplicate it

PR #1247, opened 2025-03-28
fix: ensure worker token base64 decoding handles padding correctly

bschimke95:
## Description

In some cases, a worker token was wrongly interpreted as an control-plane/microcluster token because the decoding mechanism used `base64.RawStdEncoding` which caused a decoding error if the worker node contained [padding](https://medium.com/@partha.pratimnayak/understanding-base64-encoding-7764b4ecce3c).
   
## Solution

Ensure that Base64 encoding of the worker token uses standard encoding with padding to avoid decoding errors. 

## Issue

## Backport

This is a bugfix which we need to backport to 1.32

## Checklist

- [x] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/src/k8s/pkg/k8sd/types/worker.go
+++ b/src/k8s/pkg/k8sd/types/worker.go
@@ -21,7 +21,7 @@ type InternalWorkerNodeToken struct {
 // internalWorkerNodeTokenSerializeMagicString is used to validate serialized worker node tokens.
 const internalWorkerNodeTokenSerializeMagicString = "m!!"
 
-var encoding = base64.RawStdEncoding
+var encoding = base64.StdEncoding
 
 type serializableWorkerNodeInfo struct {
 	InternalWorkerNodeToken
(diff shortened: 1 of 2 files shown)


src/k8s/pkg/k8sd/types/worker.go
```diff
@@ -21,7 +21,7 @@ type InternalWorkerNodeToken struct {
 // internalWorkerNodeTokenSerializeMagicString is used to validate serialized worker node tokens.
 const internalWorkerNodeTokenSerializeMagicString = "m!!"
 
-var encoding = base64.RawStdEncoding
+var encoding = base64.StdEncoding
```
mateoflorido, line 24: This is interesting. We were using the Raw encoding which does not add padding to the encoded string, but decoding it with the same engine. I'm curious how padding ended up in that string.

addyess: very curious indeed.  Does it need a deep dive to determine why?

bschimke95: As discussed, the user used added the padding manually at the end. So this was not the issue. 
However, I want to merge this to have the same encoding for CP and worker nodes

bschimke95: We don't need to backport this though

mateoflorido (approved):
Cool, let's make it consistent. LGTM!

addyess (approved):
LGTM too...

so if the token is padded or not, we will still be able to decode the token as a worker token right?

bschimke95:
yes, @addyess we define the encoding so this should be fine

PR #1250, opened 2025-03-30
feat: Register nodes in Upgrade CRD after upgrade

bschimke95:
## **Description**  

During an upgrade, feature updates should only occur once all nodes have been upgraded to the latest version.  
This PR introduces logic to register upgraded nodes in the Upgrade CRD and verify when all nodes have been updated.  

## **Solution**  

At a high level, this PR:  
- During a node upgrade (triggered by a snap refresh), checks if an Upgrade CR already exists (indicating an ongoing upgrade) and creates one if not.  
- After upgrading, registers the node in the `upgradedNodes` list of the CR and updates the upgrade status to `FeatureUpgrade` once all nodes in the microcluster are upgraded.  

Additionally, this PR:  
- Adds `jq`
- Fixes a bug in `lib.sh` where an incorrect `$PATH` caused helper binaries to be missing.  
- Enhances the feature upgrade test to upload the snap to two different branches and perform an upgrade between them.  

## **Limitations**  

- The last upgraded node will transition the upgrade phase to **"FeatureUpgrade"**, but the actual feature upgrade process is not yet implemented (to be addressed in a follow-up PR).  
- The feature controller is not yet blocked during the upgrade (to be handled in a future PR).
- The handling of the Upgrade CR is done in snap hooks. This will be moved to a separate controller in the future.

## Backport

The feature upgrade work will not be backported.

## Checklist

- [ ] Covered by unit tests - not yet
- [x] CLA signed

c


k8s/lib.sh
```diff
@@ -239,3 +237,37 @@ k8s::containerd::ensure_systemd_defaults() {
...
+
+  echo "Preparing for upgrade." >> "$SNAP_COMMON/logs/upgrade.log"
+  echo "Current upgrade CRDs:" >> "$SNAP_COMMON/logs/upgrade.log"
+  k8s::cmd::k8s kubectl get upgrade --no-headers -o json >> "$SNAP_COMMON/logs/upgrade.log"
+
+  # Create a new Upgrade CRD to track the upgrade across the cluster if there is none in progress.
+  upgrade_name=$(k8s::upgrade::in_progress_upgrade)
+  echo "Upgrade name: $upgrade_name" >> "$SNAP_COMMON/logs/upgrade.log"
+
+  if [[ -n "$upgrade_name" ]]; then
```
eaudetcobello, line 260: in_progress_upgrade can return one name per line, and this will break. Example

```
upgrade_name="upgrade-1
upgrade-2"

kubectl patch upgrade "upgrade-1
upgrade-2"
```

k8s/lib.sh
```diff
@@ -239,3 +237,37 @@ k8s::containerd::ensure_systemd_defaults() {
...
+  mkdir -p "$SNAP_COMMON/logs"
+
+  echo "Preparing for upgrade." >> "$SNAP_COMMON/logs/upgrade.log"
+  echo "Current upgrade CRDs:" >> "$SNAP_COMMON/logs/upgrade.log"
+  k8s::cmd::k8s kubectl get upgrade --no-headers -o json >> "$SNAP_COMMON/logs/upgrade.log"
+
+  # Create a new Upgrade CRD to track the upgrade across the cluster if there is none in progress.
+  upgrade_name=$(k8s::upgrade::in_progress_upgrade)
+  echo "Upgrade name: $upgrade_name" >> "$SNAP_COMMON/logs/upgrade.log"
+
```
eaudetcobello, line 259: Potential pitfall: there may be more than one upgrade in progress and this assumes upgrade_name will always contain one string.

in_progress_upgrade might return:

"cluster-1
cluster-2"

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (c *Client) GetInProgressUpgrade(ctx context.Context) (*Upgrade, error) {
+	log := log.FromContext(ctx).WithValues("upgrades", "GetInProgressUpgrade")
+
+	restClient, err := c.K8sdIoRestClient()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create REST client for k8sd.io group: %w", err)
+	}
+
+	upgrades, err := restClient.Get().AbsPath("/apis/k8sd.io/v1alpha/upgrades").DoRaw(ctx)
```
eaudetcobello, line 48: hardcoded path

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	if err := json.Unmarshal(upgrades, &result); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal upgrades: %w", err)
+	}
+
+	for _, upgrade := range result.Items {
+		if upgrade.Status.Phase != "Failed" && upgrade.Status.Phase != "Completed" {
+			return &upgrade, nil
+		}
+	}
```
eaudetcobello, line 64: if there are multiple upgrades, this picks the first one and silently ignores the rest. is this intended?

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,169 @@
...
+
+	log.Info("Marking node as upgraded", "patch", patch)
+
+	restClient, err := c.K8sdIoRestClient()
+	if err != nil {
+		return fmt.Errorf("failed to create REST client for k8sd.io group: %w", err)
+	}
+	result := restClient.Patch(types.MergePatchType).
+		AbsPath("/apis/k8sd.io/v1alpha/upgrades/" + upgrade.Metadata.Name + "/status").
```
eaudetcobello, line 123: hardcoded path

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -24,5 +25,79 @@ func (a *App) postRefreshHook(ctx context.Context, s state.State) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get Kubernetes client: %w", err)
+	}
+
+	upgrade, err := k8sClient.GetInProgressUpgrade(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get in progress upgrade: %w", err)
+	}
+
+	if upgrade != nil {
```
eaudetcobello, line 38: You can remove a level of indentation with a guard clause here
```suggestion

	if upgrade == nil {
	   log.Info()
	   return
    }
    ...
```

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	restClient, err := c.K8sdIoRestClient()
+	if err != nil {
+		return fmt.Errorf("failed to create REST client for k8sd.io group: %w", err)
+	}
+	result := restClient.Patch(types.MergePatchType).
+		AbsPath("/apis/k8sd.io/v1alpha/upgrades/" + upgrade.Metadata.Name + "/status").
+		Body(updatedUpgrade).
+		Do(ctx)
```
eaudetcobello, line 161: Hmm. This seems particularly brittle. Why do we need to use a custom rest path instead of using a kubernetes client to get objects of a certain type?

bschimke95: The current implementation uses direct API calls to handle Upgrades. A better way would be to use tools like kubebuilder to generate the CRD definitions in Go and use well-defined Go structs instead of hard-coded API path. We have a card to investigate options as future work. This is out-of-scope for the initial PoC.

(I updated the "Limitations" sections in the PR description)

src/k8s/pkg/client/kubernetes/upgrades.go
bschimke95: We don't generate Go structs for the CRD yet in the initial proof of concept, so you don't have much choice (AFAIK let me know if you know a better way).
See also comment below.

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
bschimke95: I also thought about this but if feels a bit dangerous. If people later add upgrade-unrelated steps below this logic then this logic is silently not executed if no update is in progress.

k8s/lib.sh
bschimke95: obsolete now

.github/workflows/e2e-tests.yaml
```diff
@@ -71,8 +75,9 @@ jobs:
           TEST_VERSION_UPGRADE_MIN_RELEASE: "1.31"
           TEST_STRICT_INTERFACE_CHANNELS: "recent 6 strict"
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
+          TEST_SNAPCRAFT_STORE_CREDENTIALS: '{{ secrets.SNAPCRAFT_STORE_CREDENTIALS }}'
         run: |
-          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }}
+          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }} -k "test_feature_upgrades"
```
bschimke95, line 80: TODO: remove before merge

k8s/lib.sh
```diff
@@ -9,11 +9,9 @@ k8s::common::setup_env() {
     return 0
   fi
 
-  local SNAP_CURRENT="${SNAP%"${SNAP_REVISION}"}current"
-
   # Configure PATH, LD_LIBRARY_PATH
-  export PATH="$SNAP_CURRENT/usr/bin:$SNAP_CURRENT/bin:$SNAP_CURRENT/usr/sbin:$SNAP_CURRENT/sbin:$REAL_PATH"
-  export LD_LIBRARY_PATH="$SNAP_LIBRARY_PATH:$SNAP_CURRENT/lib:$SNAP_CURRENT/usr/lib:$SNAP_CURRENT/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP_CURRENT/usr/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP_CURRENT/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/ceph:${REAL_LD_LIBRARY_PATH:-}"
+  export PATH="$SNAP/usr/bin:$SNAP/bin:$SNAP/usr/sbin:$SNAP/sbin:$REAL_PATH"
+  export LD_LIBRARY_PATH="$SNAP_LIBRARY_PATH:$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/ceph:${REAL_LD_LIBRARY_PATH:-}"
```
bschimke95, line 14: This is actually a bug that we dragged around since the very initial commit but didn't pop up until one explicitly used a binary from the snap. 
The `SNAP_CURRENT=.../current` path does not exist - it should just be $SNAP (points to the current revision already

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,179 @@
...
+	for _, upgrade := range result.Items {
+		if upgrade.Status.Phase != UpgradePhaseFailed && upgrade.Status.Phase != UpgradePhaseCompleted {
+			matches = append(matches, upgrade)
+		}
+	}
+	if len(matches) == 0 {
+		return nil, nil
+	}
+	if len(matches) > 1 {
+		log.Info("Warning: Found multiple in-progress upgrades", "inprogress upgrades", len(matches))
```
bschimke95, line 106: @eaudetcobello normally we should only have one update but you raised a good point. 
I think it is the best solution to:
1. log the issue and upgrades
2. sort by name (naming is `cluster-upgrade-to-rev-<rev>`)
3. return the last entry.
This will always return the latest upgrade even if there is a stale upgrade left for some reason.

src/k8s/pkg/client/kubernetes/upgrades.go
bschimke95: @eaudetcobello normally we should only have one update but you raised a good point.
I think it is the best solution to:

log the issue and upgrades
sort by name (naming is cluster-upgrade-to-rev-<rev>)
return the last entry.
This will always return the latest upgrade even if there is a stale upgrade left for some reason.

bschimke95:
Note: nightly tests are expected to fail since this changes are not on `latest/edge` yet.

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,179 @@
...
+		Body(body).
+		Do(ctx)
+	if result.Error() != nil {
+		responseBody, _ := result.Raw()
+		log.Error(result.Error(), "failed to create upgrade", "response", string(responseBody))
+		return fmt.Errorf("failed to create upgrade: %w", result.Error())
+	}
+
+	// The status field needs to be patches separatly since it is a subresource.
+	if err := c.PatchUpgradeStatus(ctx, upgrade.Metadata.Name, upgrade.Status); err != nil {
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: And also maybe the `upgrade` instance?

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -39,5 +40,118 @@ func (a *App) postRefreshHook(ctx context.Context, s state.State) error {
...
+	log.Info("Marking node as upgraded.", "node", s.Name())
+
+	upgradedNodes := upgrade.Status.UpgradedNodes
+	upgradedNodes = append(upgradedNodes, s.Name())
+
+	if err := k8sClient.PatchUpgradeStatus(ctx, upgrade.Metadata.Name, kubernetes.Status{UpgradedNodes: upgradedNodes}); err != nil {
+		return fmt.Errorf("failed to mark node as upgraded: %w", err)
+	}
+
+	clusterUpgradeDone, err := allNodesUpgraded(ctx, s, k8sClient)
```
HomayoonAlimohammadi, line 103: Maybe we can also pass the upgrade instance here to prevent re-fetching?

HomayoonAlimohammadi (commented):
Great work, thanks a lot @bschimke95! Did a first pass. Looks good overall, left some comments and questions.

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,179 @@
...
+	var result struct {
+		Items []Upgrade `json:"items"`
+	}
+	if err := json.Unmarshal(upgrades, &result); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal upgrades: %w", err)
+	}
+
+	var matches []Upgrade
+	for _, upgrade := range result.Items {
+		if upgrade.Status.Phase != UpgradePhaseFailed && upgrade.Status.Phase != UpgradePhaseCompleted {
```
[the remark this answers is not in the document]
bschimke95: Woudn't the same argument apply in that case then? If I add a new Phase that is considered "in-progress" I would also need to update this check. 
I feel like it is less likely that a new "final" state appears (a Upgrade is either Complete or failed) than a new in-progress step. WDYT?

src/k8s/pkg/client/kubernetes/upgrades.go
bschimke95: We don't want to have multiple upgrades - regardless if they have the same name or not, right? e.g. if we have currently two upgrades running, on wants to upgrade to revision X, the other to Y then this would be bad as well.

A minor note: we don't know the `from` for the current implementation because everything runs in the `post-refresh` routine to be backward-compatible. For 1.33 we can do such nice things though.

src/k8s/pkg/client/kubernetes/upgrades.go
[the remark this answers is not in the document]
HomayoonAlimohammadi: Makes sense. Let's not do this.

.github/workflows/e2e-tests.yaml
```diff
@@ -112,8 +116,10 @@ jobs:
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
           TEST_GH_REF: ${{ github.ref_name }}
           TEST_GH_BASE_REF: ${{ github.base_ref }}
+          TEST_SNAPCRAFT_STORE_CREDENTIALS: '${{ secrets.SNAPCRAFT_STORE_CREDENTIALS }}'
+          TEST_SKIP_CLEANUP: "1"
         run: |
-          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }} ${{ inputs.extra-test-args }}
+          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }} ${{ inputs.extra-test-args }} -k "test_feature_upgrades"
```
bschimke95, line 122: remove before merge

.github/workflows/e2e-tests.yaml
```diff
@@ -112,8 +116,10 @@ jobs:
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
           TEST_GH_REF: ${{ github.ref_name }}
           TEST_GH_BASE_REF: ${{ github.base_ref }}
+          TEST_SNAPCRAFT_STORE_CREDENTIALS: '${{ secrets.SNAPCRAFT_STORE_CREDENTIALS }}'
+          TEST_SKIP_CLEANUP: "1"
```
bschimke95, line 120: remove before merge

.github/workflows/e2e-tests.yaml
```diff
@@ -112,6 +116,7 @@ jobs:
           TEST_MIRROR_LIST: '[{"name": "ghcr.io", "port": 5000, "remote": "https://ghcr.io", "username": "${{ github.actor }}", "password": "${{ secrets.GITHUB_TOKEN }}"}, {"name": "docker.io", "port": 5001, "remote": "https://registry-1.docker.io", "username": "", "password": ""}, {"name": "rocks.canonical.com", "port": 5002, "remote": "https://rocks.canonical.com/cdk"}]'
           TEST_GH_REF: ${{ github.ref_name }}
           TEST_GH_BASE_REF: ${{ github.base_ref }}
+          TEST_SNAPCRAFT_STORE_CREDENTIALS: '${{ secrets.SNAPCRAFT_STORE_CREDENTIALS }}'
```
berkayoz, line 119: This is a risky secret, we should be careful on which steps/actions can access this.

k8s/lib.sh
berkayoz: The use of `$SNAP_CURRENT` is correct although the way we calculate it might be broken.

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,178 @@
...
+type Upgrade struct {
+	APIVersion string          `json:"apiVersion,omitempty"`
+	Kind       string          `json:"kind,omitempty"`
+	Metadata   UpgradeMetadata `json:"metadata,omitempty"`
+	Status     UpgradeStatus   `json:"status,omitempty"`
+}
+
+func NewUpgrade(name string) Upgrade {
+	return Upgrade{
+		APIVersion: group + "/" + version,
```
berkayoz, line 50: nit: Can we use `Sprintf` and why not have this as a constant if we're already joining 2 `const`s?

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,178 @@
...
+// TODO(ben): Maybe make this more generic, e.g. GetUpgrade(filterFunc func(Upgrade) bool) (*Upgrade, error)
+func (c *Client) GetInProgressUpgrade(ctx context.Context) (*Upgrade, error) {
+	log := log.FromContext(ctx).WithValues("upgrades", "GetInProgressUpgrade")
+
+	restClient, err := c.k8sdIoRestClient()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create REST client for k8sd.io group: %w", err)
+	}
+
+	upgrades, err := restClient.Get().AbsPath(fmt.Sprintf("/apis/%s/%s/upgrades", group, version)).DoRaw(ctx)
```
berkayoz, line 79: nit: Should we also turn `fmt.Sprintf("/apis/%s/%s/upgrades", group, version)` into a `const`?

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,178 @@
...
+	for _, upgrade := range result.Items {
+		if upgrade.Status.Phase != UpgradePhaseFailed && upgrade.Status.Phase != UpgradePhaseCompleted {
+			matches = append(matches, upgrade)
+		}
+	}
+	if len(matches) == 0 {
+		return nil, nil
+	}
+	if len(matches) > 1 {
+		log.Info("Warning: Found multiple in-progress upgrades", "inprogress upgrades", len(matches))
```
berkayoz, line 105: Communicating this to the user through logs is a bit unfriendly UX wise. Maybe as a follow up we can explore adding an Event, and consider documenting the default behavior.

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	// Sort matches by name
+	sort.Slice(matches, func(i, j int) bool {
+		return matches[i].Metadata.Name < matches[j].Metadata.Name
+	})
+
+	// Return the latest
+	return &matches[len(matches)-1], nil
```
berkayoz, line 113: nit: Can we just assign `len(matches)` to a variable and reuse instead of recalculating every time?

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	body, err := json.Marshal(upgrade)
+	if err != nil {
+		return fmt.Errorf("failed to marshal upgrade: %w", err)
+	}
+
+	log.Info("Creating upgrade", "upgrade", upgrade)
+	result := restClient.Post().
+		AbsPath(fmt.Sprintf("/apis/%s/%s/upgrades", group, version)).
```
berkayoz, line 131: nit: We can reuse the constant path in here/other places that I mentioned above.

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -0,0 +1,178 @@
...
+		Body(body).
+		Do(ctx)
+	if result.Error() != nil {
+		responseBody, _ := result.Raw()
+		log.Error(result.Error(), "failed to create upgrade", "response", string(responseBody))
```
berkayoz, line 136: Should these messages be debug level since I assume we want to print out the whole response?

tests/integration/tests/test_version_upgrades.py
```diff
@@ -234,22 +233,106 @@ def test_version_downgrades_with_rollback(
...
     Once this work is complete, this test will likely be merged with the
-    test_version_upgrades test above and create a single test for all upgrades."""
-
-    cp = instances[0]
+    test_version_upgrades test above and create a single test for all upgrades.
 
-    # Verify that the UpgradeCRD is known to the cluster
-    cp.exec("k8s kubectl get crd upgrades.k8sd.io".split())
+    It is not possible to:
+        * do a snap refresh to a local snap
```
berkayoz, line 247: Can you elaborate on this, do we not get `post-refresh` or other refresh hooks triggered when we do `snap install ./path-to-snap --classic --dangerous` when the snap is already installed?

berkayoz (commented):
Great work! Did a first pass and left some comments

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -39,5 +40,113 @@ func (a *App) postRefreshHook(ctx context.Context, s state.State) error {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to get leader client: %w", err)
+	}
+
+	clusterMembers, err := c.GetClusterMembers(ctx)
+	if err != nil {
+		return false, fmt.Errorf("failed to get cluster members: %w", err)
+	}
+
+	if len(clusterMembers) != len(upgradedNodes) {
```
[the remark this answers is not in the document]
bschimke95: This is only a pre-check. The actual name comparison is done below.
If the lengths are not equal, then I don't see a scenario where this is a complete upgrade. Even if upgradedNodes contains all clusterMembers. If the length is not equal then there is still something going on (or broken).

tests/integration/tests/test_version_upgrades.py
bschimke95: Oh, this actually works:

```
root@k8s-integration-3-f9eb54:~# sudo snap list k8s
Name  Version  Rev  Tracking  Publisher  Notes
k8s   v1.32.3  x1   -         -          classic
root@k8s-integration-3-f9eb54:~# sudo snap install /k8s.snap --dangerous --classic
2025-04-08T10:35:47Z INFO Waiting for "snap.k8s.kube-apiserver.service" to stop.
2025-04-08T10:35:53Z INFO Waiting for "snap.k8s.k8s-dqlite.service" to stop.
k8s v1.32.3 installed
root@k8s-integration-3-f9eb54:~# sudo k8s status
cluster status:           ready
control plane nodes:      10.97.37.40:6400 (voter), 10.97.37.211:6400 (spare)
high availability:        no
datastore:                k8s-dqlite
network:                  enabled
dns:                      enabled at 10.152.183.103
ingress:                  disabled
load-balancer:            disabled
local-storage:            disabled
gateway                   disabled
```
For context:
I started to implement this "load to branch" test because I had pre- and post-hooks. After refactoring this to only post-hooks, this just works without uploading a new branch (and heavily shortens the testing time).
Thanks - I'll refactor this.

.github/workflows/e2e-tests.yaml
bschimke95: We don't need this anymore as per comment below. Thanks!

k8s/lib.sh
bschimke95: Well, according to the docs, $SNAP already points to the current revision:
https://snapcraft.io/docs/environment-variables#heading--snap

So there was no assembling required in the first place, no?

tests/integration/tests/test_util/util.py
```diff
@@ -29,7 +29,7 @@
 
 LOG = logging.getLogger(__name__)
 RISKS = ["stable", "candidate", "beta", "edge"]
-TRACK_RE = re.compile(r"^(\d+)\.(\d+)(\S*)$")
+TRACK_RE = re.compile(r"^v?(\d+)\.(\d+)(.\d+)?(\S*)$")
```
bschimke95, line 32: The regex did not work for version with `v` prefix and patch-level versions

tests/integration/tests/test_util/util.py
```diff
@@ -580,23 +580,29 @@ def _previous_track_from_branch(branch: str) -> Optional[str]:
         # NOTE(Hue): `latest/stable` is not populated at the moment.
         # When it is, we should return `latest` instead.
         LOG.info("Getting current version from upstream k8s")
+        # For the main branch, the previous track is the latest release-branch, e.g.
+        # `1.32/stable` for `main` branch which matches the current upstream version.
```
bschimke95, line 584: @HomayoonAlimohammadi could you check if you agree with this logic? 
IMHO there was a bug in the code how we determine previous tracks.

HomayoonAlimohammadi: That's right I guess the main issue was that we decreased the minor version an additional time on main when reading from upstream.
Nice catch

tests/integration/tests/test_util/util.py
HomayoonAlimohammadi: QQ: Since it's a track, do we need to account for `v` prefix and patch-level version?

bschimke95: This is used in `major_minor` and `previous_track` which is not tied to the track and may contain an upstream Kubernetes version.
Maybe the naming is not the best for this regex.

.github/workflows/e2e-tests.yaml
```diff
@@ -113,7 +113,7 @@ jobs:
           TEST_GH_REF: ${{ github.ref_name }}
           TEST_GH_BASE_REF: ${{ github.base_ref }}
         run: |
-          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }} ${{ inputs.extra-test-args }}
+          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags ${{ inputs.test-tags }} ${{ inputs.extra-test-args }} -k "test_version_upgrades"
```
bschimke95, line 116: remove

PR #1251, opened 2025-03-30
test: Disable failing TICS test

bschimke95:
There seems to be an issue with TICS and the test is constantly failing right now.
We remove the test for now. I added a card to investigate this in the upcoming pulse.


eaudetcobello (approved):
+1

PR #1259, opened 2025-04-01
test: cleanup etcd instances immediately after the test

petrutlucian94:
This change will ensure that etcd instances are cleaned up immediately after the test instead of waiting until the end of the test suite.

By doing so, we avoid the situation in which subsequent tests run out of resources due to the leaked instances.


louiseschmidtgen (approved):
LGTM, let's wait on CI.

PR #1262, opened 2025-04-02
fix: Don't wait for services to terminate on removed node

bschimke95:
## Description

Some services (e.g. k8s-dqlite) take a considerable amount of time to shutdown in some cases. Microcluster has a fixed 30s timeout for the remove hook which caused removing a node to fail in some scenarios.

## Solution

With this commit, the remove hook will not wait anymore on services to terminate. The cleanup on a removed node is on a best-effort basis so in that case we don't care when (and even if) a service is actually stopped.

Code changes:
* Renames/refactored `Start/Stop/RestartService` to `...Services` and accept a slice instead of single string name. If we don't want to wait for snap services to stop, we need to have a way to pass all services at once. If we would do a `--no-wait` and then sequentially call `snap service stop`, this would fail with `snap: change in-progress`. 
* Use `snap` instead of `snapctl`. `snapctl` does not provide a `--no-wait` flag. See discussion [here](https://bugs.launchpad.net/snapd/+bug/2106256)
* Add `extraSnapArgs` to services. We only want to not wait when removing a node. All other service interactions should be synchronous, so we need a way to control that.

You can see in the log timestamps of a removed node, that stopping of the services is now finishing immediately:
```
emove" node="k8s-integration-c673d2-3"
Apr 02 09:53:39 k8s-integration-c673d2-3 k8s.k8sd[2211]: I0402 09:53:39.048725    2211 app/hooks_remove.go:138] "Cleaning up control plane certificates" logger="k8sd" hook="preremove" node="k8s-integration-c673d2-3"
Apr 02 09:53:39 k8s-integration-c673d2-3 k8s.k8sd[2211]: I0402 09:53:39.049043    2211 app/hooks_remove.go:143] "Stopping all services except of k8sd" logger="k8sd" hook="preremove" node="k8s-integration-c673d2-3"
Apr 02 09:53:39 k8s-integration-c673d2-3 k8s.k8sd[5271]: 7
Apr 02 09:53:39 k8s-integration-c673d2-3 k8s.k8sd[2211]: I0402 09:53:39.080249    2211 app/hooks_remove.go:148] "Cleaning up containerd paths" logger="k8sd" hook="preremove" node="k8s-integration-c673d2-3"
```
## Backport

This fixes issues when removing a node, so we should backport this.

## Checklist

- [x] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -140,19 +140,9 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 			log.Error(err, "failed to cleanup control plane certificates")
 		}
 
-		log.Info("Stopping worker services")
-		if err := snaputil.StopWorkerServices(ctx, snap); err != nil {
-			log.Error(err, "Failed to stop worker services")
-		}
-
-		log.Info("Stopping control plane services")
-		if err := snaputil.StopControlPlaneServices(ctx, snap); err != nil {
-			log.Error(err, "Failed to stop control-plane services")
-		}
-
-		log.Info("Stopping k8s-dqlite")
-		if err := snaputil.StopK8sDqliteServices(ctx, snap); err != nil {
-			log.Error(err, "Failed to stop k8s-dqlite service")
+		log.Info("Stopping all services except of k8sd")
+		if err := snaputil.StopK8sServices(ctx, snap, "--no-wait"); err != nil {
+			log.Error(err, "failed to stop k8sd service")
 		}
 
 		log.Info("Cleaning up containerd paths")
--- a/src/k8s/pkg/k8sd/app/hooks_remove.go
+++ b/src/k8s/pkg/k8sd/app/hooks_remove.go
@@ -140,9 +140,9 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr 
 			log.Error(err, "failed to cleanup control plane certificates")
 		}
 
-		log.Info("Stopping all services except of k8sd")
+		log.Info("Stopping all services except k8sd")
 		if err := snaputil.StopK8sServices(ctx, snap, "--no-wait"); err != nil {
-			log.Error(err, "failed to stop k8sd service")
+			log.Error(err, "failed to stop k8s services")
 		}
 
 		log.Info("Cleaning up containerd paths")
--- a/src/k8s/pkg/k8sd/controllers/control_plane_configuration_test.go
+++ b/src/k8s/pkg/k8sd/controllers/control_plane_configuration_test.go
@@ -59,7 +59,7 @@ func TestControlPlaneConfigController(t *testing.T) {
 			expectKubeAPIServerArgs         map[string]string
 			expectKubeControllerManagerArgs map[string]string
 
-			expectServiceRestarts [][]string
+			expectServiceRestarts []string
 			expectFilesToExist    map[string]bool
 		}{
 			{
@@ -78,7 +78,7 @@ func TestControlPlaneConfigController(t *testing.T) {
 					filepath.Join(dir, "etcd-pki", "client.crt"): false,
 					filepath.Join(dir, "etcd-pki", "client.key"): false,
 				},
-				expectServiceRestarts: [][]string{{"kube-apiserver"}},
+				expectServiceRestarts: []string{"kube-apiserver"},
 			},
 			{
 				name: "Certs",
@@ -102,7 +102,7 @@ func TestControlPlaneConfigController(t *testing.T) {
 					filepath.Join(dir, "etcd-pki", "client.crt"): true,
 					filepath.Join(dir, "etcd-pki", "client.key"): true,
 				},
-				expectServiceRestarts: [][]string{{"kube-apiserver"}},
+				expectServiceRestarts: []string{"kube-apiserver"},
 			},
 			{
 				name: "CloudProvider",
@@ -114,7 +114,7 @@ func TestControlPlaneConfigController(t *testing.T) {
 				expectKubeControllerManagerArgs: map[string]string{
 					"--cloud-provider": "external",
 				},
-				expectServiceRestarts: [][]string{{"kube-controller-manager"}},
+				expectServiceRestarts: []string{"kube-controller-manager"},
 			},
 			{
 				name: "NoUpdates",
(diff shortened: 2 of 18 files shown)


src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -140,19 +140,9 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
-
-		log.Info("Stopping control plane services")
-		if err := snaputil.StopControlPlaneServices(ctx, snap); err != nil {
-			log.Error(err, "Failed to stop control-plane services")
-		}
-
-		log.Info("Stopping k8s-dqlite")
-		if err := snaputil.StopK8sDqliteServices(ctx, snap); err != nil {
-			log.Error(err, "Failed to stop k8s-dqlite service")
+		log.Info("Stopping all services except of k8sd")
```
eaudetcobello, line 143:
```suggestion
		log.Info("Stopping all services except k8sd")
```

src/k8s/pkg/snap/pebble.go
```diff
@@ -51,19 +51,22 @@ func NewPebble(opts PebbleOpts) *pebble {
 	return s
 }
 
-// StartService starts a k8s service. The name can be either prefixed or not.
-func (s *pebble) StartService(ctx context.Context, name string) error {
-	return s.runCommand(ctx, []string{filepath.Join(s.snapDir, "bin", "pebble"), "start", name})
+// StartServices starts a k8s service. The name can be either prefixed or not.
+func (s *pebble) StartServices(ctx context.Context, names []string, extraPebbleArgs ...string) error {
+	cmd := append([]string{filepath.Join(s.snapDir, "bin", "pebble"), "start"}, append(names, extraPebbleArgs...)...)
+	return s.runCommand(ctx, cmd)
```
eaudetcobello, line 57: This is a bit hard to read. It's not clear why args are append to names if we already run pebble start.
Could I suggest you do this in two lines?

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -140,19 +140,9 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
+		if err := snaputil.StopK8sServices(ctx, snap, "--no-wait"); err != nil {
+			log.Error(err, "failed to stop k8sd service")
```
HomayoonAlimohammadi, line 145: Isn't this error in contrast with the log above? 
"stopping all services EXCEPT k8sd"
"failed to stop k8sd service"

src/k8s/pkg/k8sd/controllers/control_plane_configuration_test.go
```diff
@@ -59,7 +59,7 @@ func TestControlPlaneConfigController(t *testing.T) {
 			expectKubeAPIServerArgs         map[string]string
 			expectKubeControllerManagerArgs map[string]string
 
-			expectServiceRestarts []string
+			expectServiceRestarts [][]string
```
HomayoonAlimohammadi, line 62: nit: Maybe we can change this to still be a `[]string` (will require some changes but I think it's worth it)?

src/k8s/pkg/k8sd/controllers/control_plane_configuration_test.go
```diff
@@ -189,7 +189,8 @@ func TestControlPlaneConfigController(t *testing.T) {
 				// TODO: this should be changed to call g.Eventually()
 				<-time.After(50 * time.Millisecond)
 
-				g.Expect(s.RestartServiceCalledWith).To(ConsistOf(tc.expectServiceRestarts))
+				g.Expect(s.RestartServicesCalledWith).To(HaveLen(len(tc.expectServiceRestarts)))
+				g.Expect(s.RestartServicesCalledWith).To(Equal(tc.expectServiceRestarts))
```
HomayoonAlimohammadi, line 193: nit: Should we care about ordering here?

src/k8s/pkg/snap/util/services.go
```diff
@@ -24,76 +24,86 @@ var (
 		"kubelet",
 		"kube-apiserver",
 	}
+
+	// k8sServices contains all k8s services except of k8sd.
```
eaudetcobello, line 28:
```suggestion
	// k8sServices contains all k8s services except k8sd.
```

src/k8s/pkg/snap/pebble.go
HomayoonAlimohammadi: I agree that it's a bit hard to read but at the same time I think it's a nice one liner to say `cmd := EVERYTHING ALTOGHER` :D

src/k8s/pkg/snap/util/services.go
```diff
@@ -24,76 +24,86 @@ var (
...
+	k8sServices = []string{
```
HomayoonAlimohammadi, line 29: nit: Do we also want to change `workerServices` to `workerK8sServices` (same for CP) since they don't include k8sd as well?

mateoflorido (commented):
⚠️ Take this comment with a grain of salt!  

Wouldn't it be easier to use a wrapper around the `*Service` functions? That way, we could keep the existing function calls unchanged and only modify the ones that require multiple service restarts.

HomayoonAlimohammadi (approved):
Great work, thanks a lot @bschimke95! LGTM overall, left some minor comments and questions.

src/k8s/pkg/snap/pebble.go
HomayoonAlimohammadi: Oh sorry I commented before the edit. I agree with the duplicated `filepath.Join(...)`, maybe we can extract that.

HomayoonAlimohammadi:
> Wouldn't it be easier to use a wrapper around the *Service functions? That way, we could keep the existing function calls unchanged and only modify the ones that require multiple service restarts.

@mateoflorido I guess we'll then lose the ability to pass `--no-wait` and other snap args, right? But overall this sounds like a very reasonable suggestion.

bschimke95:
@mateoflorido Yes, @HomayoonAlimohammadi is right. While this is a good idea (and my initial approach), we cannot simply wrap the existing `StartService` call. 

There is no queuing in snapd (afaik). If you run 
```
snap start k8s.k8s-dqlite --no-wait
``` 
It will create a snapd change and return its ID.
If you run another change request *immediately* afterwards, e.g.
```
snap start k8s.kube-apiserver --no-wait
```
it will fail with something like 
```
another change is already progress
```

So, we need to be able to pass multiple services to the command at the same time which the current `StartService` API does not support.

Another solution (instead of changing the existing API) would be to add `StartServices` as a new method but I felt like this would be confusing for users and inconsistent since one would support extra args and the other command does not.

src/k8s/pkg/k8sd/app/hooks_remove.go
bschimke95: Good catch!

src/k8s/pkg/k8sd/controllers/control_plane_configuration_test.go
bschimke95: WDYM?

bschimke95:
Manually backported in https://github.com/canonical/k8s-snap/pull/1287

PR #1263, opened 2025-04-02
feat: add partial updates for Control Plane and Worker Nodes

mateoflorido:
> [!WARNING]  
> DO NOT MERGE until we change the version of the `k8s-snap-api` go module ([PR](https://github.com/canonical/k8s-snap-api/pull/33))

## Description

Add support for partial certificate updates on both Control Plane and Worker nodes in the cluster using the `refresh-certs` command.

## Solution

The `refresh-certs` command now includes a `--certificates` argument that allows users to specify which certificates to refresh. This implements the new UX described in the `KU-130` spec.

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [ ] CLA signed


bschimke95 (commented):
I'll do another review tomorrow. Just a small thing for now.
Looks already great!

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -33,7 +34,142 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type NodeRole = string
+
+const (
+	Worker       NodeRole = "worker"
+	ControlPlane NodeRole = "control-plane"
```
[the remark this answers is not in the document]
mateoflorido: Fixed, thanks Ben!

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -45,20 +176,49 @@ func (e *Endpoints) postRefreshCertsPlan(s state.State, r *http.Request) respons
...
+		if len(req.Certificates) > 0 {
+			certsToPlan = req.Certificates
+		} else {
+			certsToPlan = getAllCertsForRole(apiv1.ClusterRoleWorker)
+		}
+	} else {
+		role = apiv1.ClusterRoleControlPlane
+		if len(req.Certificates) > 0 {
+			certsToPlan = req.Certificates
+		}
```
HomayoonAlimohammadi, line 193: Don't we need a `getAllCertsForRole(apiv1.ClusterRoleControlPlane)` here like we have for the workers?

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -367,6 +555,52 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
...
+			CommonName:   fmt.Sprintf("system:node:%s", hostname),
+			Organization: []string{"system:nodes"},
+			Usages:       []certv1.KeyUsage{certv1.UsageDigitalSignature, certv1.UsageKeyEncipherment, certv1.UsageClientAuth},
+			SignerName:   "k8sd.io/kubelet-client",
+			NeedsSANs:    false,
+		},
+		CertProxyClient: {
+			CertName:     CertProxyClient,
+			CSRBaseName:  "worker-kube-proxy-client",
+			CommonName:   "system:kube-proxy",
```
HomayoonAlimohammadi, line 595: Something that I don't feel right about is that we're duplicating these certificate creation logic in a couple of places.
I think it would be nice to have a central package to ask for certificates from and then we can use the returned certs to populate CSRs or do other things.
I guess it's out of scope for this PR but something to focus on in the future.

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -133,6 +298,19 @@ func refreshCertsRunControlPlane(s state.State, r *http.Request, snap snap.Snap)
...
 	certificates.ServiceAccountKey = clusterConfig.Certificates.GetServiceAccountKey()
 
+	manager := newControlPlaneCertificateMarker(certificates)
+	if err := setup.ReadControlPlanePKI(snap, certificates, true); err != nil {
+		return response.InternalError(fmt.Errorf("failed to read managed control plane certificates: %w", err))
+	}
+
+	if err := setup.ReadControlPlanePKI(snap, certificates, false); err != nil {
+		return response.InternalError(fmt.Errorf("failed to read unmanaged control plane certificates: %w", err))
+	}
```
HomayoonAlimohammadi, line 308: QQ: Is it possible to overwrite something in the second read? I guess ideally not since a certificate can't be both managed and unmanaged.

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -367,6 +555,52 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
 	})
 }
 
+// getAllCertsForRole returns a slice of all certificate names for a given role.
+func getAllCertsForRole(role apiv1.ClusterRole) []string {
+	if roleCerts, ok := certsByRole[role]; ok {
+		certs := make([]string, 0, len(roleCerts))
+		for certName := range roleCerts {
+			certs = append(certs, certName)
+		}
```
HomayoonAlimohammadi, line 564: nit: If we're using a pre-occupied slice we can also do
```go
certs := make([]string, len(roleCerts))
for i, certName := range roleCerts {
    certs[i] = certName
}
```

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -133,6 +298,19 @@ func refreshCertsRunControlPlane(s state.State, r *http.Request, snap snap.Snap)
 	certificates.K8sdPublicKey = clusterConfig.Certificates.GetK8sdPublicKey()
 	certificates.ServiceAccountKey = clusterConfig.Certificates.GetServiceAccountKey()
 
+	manager := newControlPlaneCertificateMarker(certificates)
```
HomayoonAlimohammadi, line 301: Shouldn't we create the manager after we've read everything that there is into the `certificates`?

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -333,23 +513,31 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
...
 	// restarting the kube-proxy and kubelet, which would break the
 	// proxy connection and cause missed responses in the proxy side.
 	restartFn := func(ctx context.Context) error {
-		if err := snap.RestartService(ctx, "kubelet"); err != nil {
-			return fmt.Errorf("failed to restart kubelet: %w", err)
+		restartKubelet := slices.Contains(certsToRefresh, CertKubelet) || slices.Contains(certsToRefresh, CertKubeletClient)
+		restartProxy := slices.Contains(certsToRefresh, CertProxyClient)
+
+		if restartKubelet {
+			if err := snap.RestartService(ctx, "kubelet"); err != nil {
```
HomayoonAlimohammadi, line 532: Just a reminder that I think this will need to change if we merge https://github.com/canonical/k8s-snap/pull/1262

HomayoonAlimohammadi (commented):
Great work, thanks a lot @mateoflorido! Looks very good, left a couple of questions and comments. Do we also need to add some unit or e2e tests around this?

docs/canonicalk8s/_parts/commands/k8s_refresh-certs.md
```diff
@@ -9,6 +9,9 @@ k8s refresh-certs [flags]
 ### Options
 
 ```
+      --certificates stringArray       List of certificates to renew in the cluster (only used with --expires-in). Allowed values:
```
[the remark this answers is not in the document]
mateoflorido: Yes, that's the default expiration period when bootstrapping a new cluster. From an user perspective, it's better to make them aware of the new expiration period so they can be mindful of how long they  need the certificates to remain valid.

src/k8s/cmd/k8s/k8s_refresh_certs.go
```diff
@@ -113,6 +120,7 @@ func newRefreshCertsCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 			cmd.Printf("Certificates have been successfully refreshed, and will expire at %v.\n", expiryTimeUNIX)
 		},
 	}
+	cmd.Flags().StringArrayVar(&opts.certificates, "certificates", []string{}, "List of certificates to renew in the cluster (only used with --expires-in). Allowed values:\n  Worker nodes: kubelet, kubelet.conf, proxy.conf\n  Control plane nodes: admin.conf, front-proxy-client, apiserver-kubelet-client, scheduler.conf, controller.conf, apiserver, kubelet.conf, kubelet, proxy.conf")
```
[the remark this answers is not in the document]
mateoflorido: Sure, I'll move the definition to `k8s-snap-api` repo

src/k8s/pkg/k8sd/api/certificates_refresh.go
mateoflorido: In this case, we are iterating over a "set" which is of type `map[string]struct{}`. This will unpack the value in `key, value` instead of `index, value` as you suggested.

src/k8s/pkg/k8sd/api/certificates_refresh.go
mateoflorido: It doesn't matter, the constructor of the Marker just links each certificate field to its corresponding string literal. We mark the certificates for refresh after reading them: https://github.com/canonical/k8s-snap/blob/1373e763edac37e40de0f26894c557269b3d72cc/src/k8s/pkg/k8sd/api/certificates_refresh.go#L310-L311

src/k8s/cmd/k8s/k8s_refresh_certs.go
```diff
@@ -26,15 +27,18 @@ func newRefreshCertsCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
+					cmd.PrintErrln("Error: the --expires-in flag is required when not using --external-certificates.")
+					env.Exit(1)
+					return
+				}
+			} else {
+				if opts.expiresIn != "" || len(opts.extraSANs) > 0 || len(opts.certificates) > 0 {
+					cmd.PrintErrln("Error: --external-certificates cannot be used together with --expires-in, --extra-sans, or --certificates.")
+					env.Exit(1)
+					return
+				}
```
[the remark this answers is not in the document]
mateoflorido: Please let me know if I misunderstood your comment:
`externalCerts` and `certificates` are two opposite renewal mechanisms.  When a user provides the `expires-in` argument, it indicates they want to refresh internal (managed) certificates. Even if they do not specify the exact names, the command will default to all certificates (keeping it consistent with the previous UX). This is something that should be included in the command help

src/k8s/pkg/k8sd/api/certificates_refresh.go
mateoflorido: This is not required, as control plane nodes don't need the certificate names to generate CSR (they handle their own refresh process without intervention).

src/k8s/cmd/k8s/k8s_refresh_certs.go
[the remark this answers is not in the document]
mateoflorido: Moved

src/k8s/pkg/k8sd/api/certificates_refresh.go
mateoflorido: Yes, having something like that would be nice. I believe our current implementation is simple, it keeps the strategies within the handlers (`refresh` for internal certs, `update` for external certs, and CSRs for worker nodes), and then we offload the shared process like issuing new certs or filesystem ops to the generic PKI functions.

eaudetcobello (approved):

bschimke95 (approved):

PR #1268, opened 2025-04-03
feat: Trigger feature upgrades after all nodes are in a cluster are refreshed.

bschimke95:
## Description

Trigger feature upgrades after all control-plane nodes in a cluster are upgraded.

## Solution

After every node upgrade, the node checks if the cluster nodes are all upgraded and trigger a feature upgrade if so as part of the post-refresh hook.

This is done by triggering the feature upgrade channels for each feature. The feature controllers take care of the actual helm upgrades.

Right now, there are some limitations to that approach that will be addressed in future work. See also #1250 

## Backport

The feature upgrade work will not be backported as of now.

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed


src/k8s/pkg/client/kubernetes/crds.go
```diff
@@ -86,9 +87,37 @@ func (c *Client) ApplyCRD(ctx context.Context, filePath string) error {
 	}
 
 	log.V(1).Info("Applied CRD", "name", crd.Name, "version", crd.APIVersion, "kind", crd.Kind)
+
+	waitErr := control.WaitUntilReady(ctx, func() (bool, error) {
```
bschimke95, line 91: For some reason, a CRD might not yet be available if it is queried *right after* applying it, hence this check to be on the safe side.

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -111,9 +113,70 @@ func (a *App) performPostUpgrade(ctx context.Context, s state.State) error {
...
-		// TODO: Trigger feature upgrade and unlock feature controllers afterwards.
+		log.Info("Triggering feature upgrades in background")
+
+		// TODO(ben): This is a bit ugly. We cannot wait here for the feature controllers to finish because
+		// the controllers are blocked until the node is marked as ready. For this phase, we can just run a separate
+		// goroutine to trigger the feature controllers and wait for them to finish.
+		// Once we have a separate feature upgrade controller, this should be much cleaner.
+		go func() {
+			log.Info("Triggering feature controllers in separate go routine.")
+			// TODO: Do we need to handle dependencies between features?
```
bschimke95, line 124: I put this up for discussion but I don't think this is required. The feature upgrade controllers already retry themselves if there is any issue. In my tests, the gateway and ingress are just retried after a some seconds because "network upgrade in progress"

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -111,9 +113,70 @@ func (a *App) performPostUpgrade(ctx context.Context, s state.State) error {
...
+					log.Error(ctx.Err(), "Context canceled while waiting for feature controllers to reconcile.")
+					return
+
+				default:
+					for name, ch := range pending {
+						select {
+						case <-ch:
+							log.Info(fmt.Sprintf("%s feature controller reconciled.", name))
+							delete(pending, name)
+						// TODO(ben): Maybe add a timeout here. We don't want to wait forever in this loop.
```
bschimke95, line 161: Handling failed upgrades is out of scope for this PR.

tests/integration/tests/test_version_upgrades.py
```diff
@@ -276,12 +280,48 @@ def test_feature_upgrades(instances: List[harness.Instance], tmp_path: Path):
...
+            current_helm_releases = instance.exec(
+                "k8s helm list -n kube-system -o json".split(),
+                capture_output=True,
+                text=True,
+            ).stdout
+
+            for release in json.loads(current_helm_releases):
+                name = release["name"]
+                assert (
+                    release["updated"] == initial_releases[name]["updated"]
```
bschimke95, line 321: We cannot compare versions here because some features are not updated between 1.32 and latest (e.g. local storage).

src/k8s/pkg/client/kubernetes/crds.go
```diff
@@ -86,9 +87,37 @@ func (c *Client) ApplyCRD(ctx context.Context, filePath string) error {
...
+		return c.CRDExists(ctx, crd.Name)
+	})
+	if waitErr != nil {
```
HomayoonAlimohammadi, line 94: nit: Maybe we can inline the error?

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
HomayoonAlimohammadi, line 161: Isn't the context going to act as the timeout?

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -162,10 +163,31 @@ func (c *FeatureController) reconcileLoop(
...
+				log.Error(err, "failed to check for in-progress upgrade")
+				continue
+			}
+			if upgrade != nil {
+				if upgrade.Status.Phase != kubernetes.UpgradePhaseFeatureUpgrade {
+					log.Info("Upgrade in progress - feature controller blocked", "upgrade", upgrade.Metadata.Name, "phase", upgrade.Status.Phase)
+					continue
+				}
+				log.Info("Upgrade in progress - but in feature upgrade phase - applying configuration", "upgrade", upgrade.Metadata.Name, "phase", upgrade.Status.Phase)
+			}
```
HomayoonAlimohammadi, line 185: nit: Maybe we can refactor this a bit for better readability in this code block:
```go
isUpgrading, phase, err := isClusterUpgrading(...)
if err != nil {
    log(...)
    continue
}
if isUpgrading && phase != FeatureUpgrade {
    log(...)
    continue
}
```
WDYT?

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -111,9 +113,70 @@ func (a *App) performPostUpgrade(ctx context.Context, s state.State) error {
...
+
+			// Trigger all feature controllers
+			// This intentionally blocks until the feature controllers are available.
+			a.triggerFeatureControllerNetworkCh <- struct{}{}
+			a.triggerFeatureControllerGatewayCh <- struct{}{}
+			a.triggerFeatureControllerIngressCh <- struct{}{}
+			a.triggerFeatureControllerDNSCh <- struct{}{}
+			a.triggerFeatureControllerLoadBalancerCh <- struct{}{}
+			a.triggerFeatureControllerLocalStorageCh <- struct{}{}
+			a.triggerFeatureControllerMetricsServerCh <- struct{}{}
```
HomayoonAlimohammadi, line 136: Maybe we can change this part a bit. We can change the current `NotifyFeatureController` to `MaybeNotifyFeatureController` since it's not blocking, and then implement `NotifyFeatureController` that does a blocking trigger of the channels.

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -111,9 +113,70 @@ func (a *App) performPostUpgrade(ctx context.Context, s state.State) error {
...
+
+			log.Info("Waiting for feature controllers to reconcile.")
+			pending := map[string]<-chan struct{}{
+				"Network":       a.featureController.ReconciledNetworkCh,
```
HomayoonAlimohammadi, line 140: I think something's not right here...
We notify the `ReconciledXCh` after EVERY and ANY successful reconciliation. And it's basically a thread safe boolean not a queue of e.g. `ReconciledSuccessfully` events. So we're probably gonna have these channels already filled due to a previous reconciliation (the buffer is 1). 
In other words, we're not waiting for the result of our own reconciliation trigger here. I think the for loop below is going to finish everytime almost instantly because the `ReconciledXCh`(s) are all already full due to previous reconciliations and our own reconciliation trigger has NOTHING to do with that.
Am I talking non-sense?

HomayoonAlimohammadi (commented):
Beautiful work, thanks a lot @bschimke95! Had some questions regarding the reconciliation phase and some minor comments. Overall, looks very promising.

src/k8s/pkg/k8sd/controllers/feature.go
bschimke95: Good idea. I did no use the `isClusterUpgrading` because I don't want to have another wrapper around `GetInProgressUpgrade` but I refactored the logic according to your suggestion. Thanks!

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
bschimke95: Discussed offline, resolved in https://github.com/canonical/k8s-snap/pull/1274

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
bschimke95: Yes, but the context will be the app's context which means this only is cancelled if we stop k8sd.
We might want to indicate a failed upgrade earlier but this is not something we need to solve right now.

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
bschimke95: Good point but I don't think having yet another method that has similar behavior is a good use of the interface.
Instead, I added a `ControllerReadyCh` that will be closed as soon as the controller is able to receive notifications. WDYT?

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -162,13 +205,27 @@ func (c *FeatureController) reconcileLoop(
...
+			blocked, err := c.isBlocked(ctx)
+			if err != nil {
+				log.Error(err, "Failed to check if feature controller is blocked")
+				// notify triggerCh after 5 seconds to retry
+				time.AfterFunc(5*time.Second, func() { utils.MaybeNotify(triggerCh) })
+				continue
+			}
+
+			if blocked {
+				continue
```
berkayoz, line 222: What happens if a user executes `k8s set` while we're undergoing an upgrade? I assume we'll update the cluster configuration but the feature controller will skip here and not perform the operation right?

How do we handle this case? Do we want to retry this, do we want to fail at cmd level e.g. when `k8s set` is run before hitting the API? We should at least log the reconcile was blocked here.

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -179,3 +236,30 @@ func (c *FeatureController) reconcileLoop(
...
+	upgrade, err := k8sClient.GetInProgressUpgrade(ctx)
+	if err != nil {
+		return false, fmt.Errorf("failed to check for in-progress upgrade: %w", err)
+	}
+
+	if upgrade == nil {
+		return false, nil
+	}
+
+	if upgrade.Status.Phase == kubernetes.UpgradePhaseFeatureUpgrade {
```
berkayoz, line 258: This is the check that makes sure we don't block the feature controllers when the upgrades are happening right?

tests/integration/tests/test_version_upgrades.py
```diff
@@ -301,6 +295,18 @@ def test_feature_upgrades(instances: List[harness.Instance], tmp_path: Path):
 
     util.wait_until_k8s_ready(instance, instances)
 
+    # Get initial helm releases to track if they are updated correctly.
+    initial_releases = {
+        release["name"]: release
+        for release in json.loads(
+            main.exec(
+                "k8s helm list -n kube-system -o json".split(),
```
berkayoz, line 303: IIRC we wanted to get rid of the hidden commands such as `k8s helm`, should we discuss this further or shift our test to use `helm` through other means(installing the binary, maybe the snap etc.)

berkayoz (commented):
Great work! LGTM overall, left some questions and comments

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -111,9 +113,65 @@ func (a *App) performPostUpgrade(ctx context.Context, s state.State) error {
...
+
+			for len(pending) > 0 {
+				select {
+				case <-ctx.Done():
+					log.Error(ctx.Err(), "Context canceled while waiting for feature controllers to reconcile.")
+					return
+
+				default:
+					for name, ch := range pending {
+						select {
```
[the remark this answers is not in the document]
bschimke95: Good catch

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -111,9 +113,65 @@ func (a *App) performPostUpgrade(ctx context.Context, s state.State) error {
...
+						case <-ch:
+							log.Info(fmt.Sprintf("%s feature controller reconciled.", name))
+							delete(pending, name)
+						// TODO(ben): Maybe add a timeout here. We don't want to wait forever in this loop.
+						//     		  If we run into the timeout, set the upgrade to failed.
+						// 			  What would be a reasonable timeout here? How could users configure this?
+						default:
+							// Avoid tight looping
+							time.Sleep(100 * time.Millisecond)
```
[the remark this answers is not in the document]
bschimke95: Refactored, good point!

src/k8s/pkg/k8sd/controllers/feature.go
bschimke95: Yes, exactly

src/k8s/pkg/k8sd/controllers/feature.go
bschimke95: I agree. We should block this on API level (as pointed out in the spec under limitations) eventually but I did not want to increase the scope of the PR further.
Right now, the blocked reconcile reason is logged in the isBlocked function

tests/integration/tests/test_version_upgrades.py
bschimke95: That is a good point. I use the helm binary directly

src/k8s/pkg/k8sd/controllers/feature.go
HomayoonAlimohammadi, line 258: QQ: Shouldn't we also consider `kubernetes.UpgradePhaseCompleted` (or even `Failed`, unless we want to handle that separately) as unblocking?

berkayoz (approved):
Great work, LGTM!

bschimke95:
nightly is expected to fail as the changes are not merged yet.

PR #1273, opened 2025-04-04
test: properly wait for k8sd controllers to finish reconciling

petrutlucian94:
The k8sd controller unit tests will emit a test event and wait for the reconcile loop to handle it.

UpdateNodeConfigurationController uses a channel to notify that the reconciliation finished, however the other tests will just wait for about 100ms and expect the reconcile to finish.

The 100ms sleep is flaky, sometimes it takes longer for the events to be processed.

To address this, we'll use the same post-reconciliation notification in all k8sd controllers.

### DIFF
--- a/src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
+++ b/src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
@@ -20,15 +20,18 @@ type ControlPlaneConfigurationController struct {
 	snap      snap.Snap
 	waitReady func()
 	triggerCh <-chan time.Time
+	// reconciledCh is used to notify that the controller has finished its reconciliation loop.
+	reconciledCh chan struct{}
 }
 
 // NewControlPlaneConfigurationController creates a new controller.
 // triggerCh is typically a `time.NewTicker(<duration>).C`.
 func NewControlPlaneConfigurationController(snap snap.Snap, waitReady func(), triggerCh <-chan time.Time) *ControlPlaneConfigurationController {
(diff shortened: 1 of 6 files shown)


src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
```diff
@@ -119,3 +127,8 @@ func (c *ControlPlaneConfigurationController) reconcile(ctx context.Context, con
 
 	return nil
 }
+
+// ReconciledCh returns the channel where the controller pushes when a reconciliation loop is finished.
+func (c *ControlPlaneConfigurationController) ReconciledCh() <-chan struct{} {
```
bschimke95, line 132: Why don't you just make this field public?

bschimke95 (approved):

src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
petrutlucian94: I wanted to stay consistent with the other controllers: https://github.com/canonical/k8s-snap/blob/36930785f0c4759cf4adf5c39071efba79d4b6be/src/k8s/pkg/k8sd/controllers/update_node_configuration.go#L109-L112

bschimke95: Ah, I guess Angelos did this to mark this channel as `<-chan` - read-only and users will not start to push something to it. Makes sense.

PR #1277, opened 2025-04-05
[release-1.31] Update component versions

cdkbot:
[release-1.31] Update component versions


bschimke95:
1.31 is not supported by the k8s-snap

PR #1282, opened 2025-04-08
fix: mark CSR as failed if the CA key is missing

mateoflorido:
## Description

This pull request allows the CSRSigningController to mark the requested CSR as failed if the CA is missing from the ClusterConfiguration on control plane nodes.
## Solution

When the cluster lacks the necessary  context (e.g., when using externally managed certificates), the controller will no longer silently ignore the CSR. Instead, it will explicitly mark it as failed, including a reason and message in the resource to explain the situation.

## Issue

N/A

## Backport

N/A

## Checklist

- [ ] Covered by unit tests
- [ ] Covered by integration tests (existing Nightly tests)
- [ ] Documentation updated
- [x] CLA signed

Documentation will be addressed after all the required features for a partial PKI refresh are in the main branch.


berkayoz (commented):
LGTM, one small comment.

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -586,17 +586,41 @@ func getWorkerCSRDefinitions(hostname string) map[apiv1.CertificateName]*csrDefi
...
 	for _, condition := range csr.Status.Conditions {
-		if condition.Type == certv1.CertificateApproved && condition.Status == corev1.ConditionTrue {
-			return len(csr.Status.Certificate) > 0, nil
-		}
-		if condition.Type == certv1.CertificateDenied && condition.Status == corev1.ConditionTrue {
-			return false, fmt.Errorf("CSR %s was denied: %s", csr.Name, condition.Reason)
-		}
-		if condition.Type == certv1.CertificateFailed && condition.Status == corev1.ConditionTrue {
-			return false, fmt.Errorf("CSR %s failed: %s", csr.Name, condition.Reason)
+		switch condition.Type {
```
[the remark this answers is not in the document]
mateoflorido: Sure, that'd look much better. Thanks @berkayoz!

bschimke95 (approved):

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -586,17 +586,37 @@ func getWorkerCSRDefinitions(hostname string) map[apiv1.CertificateName]*csrDefi
 // approved and issued, false if it is pending, and an error if it is denied
 // or failed.
 func isCertificateSigningRequestApprovedAndIssued(csr *certv1.CertificateSigningRequest) (bool, error) {
+	var approved, failed, denied bool
+	var failReason, failMessage string
+	var denyReason, denyMessage string
```
HomayoonAlimohammadi, line 591: nit: 
```suggestion
	var (
	    approved, failed, denied bool
	    failReason, failMessage string
	    denyReason, denyMessage string
	)
```

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -586,17 +586,37 @@ func getWorkerCSRDefinitions(hostname string) map[apiv1.CertificateName]*csrDefi
...
+				denied = true
+				denyReason = condition.Reason
+				denyMessage = condition.Message
+			case certv1.CertificateFailed:
+				failed = true
+				failReason = condition.Reason
+				failMessage = condition.Message
+			case certv1.CertificateApproved:
+				approved = true
+			}
```
HomayoonAlimohammadi, line 606: Since we're not returning in these cases, in one loop we might see `approved = true` and in another we might see a failed request. We're also relying on the ordering of conditions. Is this intentional? I'm a bit confused.

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -586,17 +586,37 @@ func getWorkerCSRDefinitions(hostname string) map[apiv1.CertificateName]*csrDefi
...
+
+	if denied {
+		return false, fmt.Errorf("CSR %q was denied: %s - %s", csr.Name, denyReason, denyMessage)
+	}
+	if failed {
+		return false, fmt.Errorf("CSR %q failed: %s - %s", csr.Name, failReason, failMessage)
+	}
+	if approved {
+		return len(csr.Status.Certificate) > 0, nil
+	}
```
HomayoonAlimohammadi, line 618: nit: Maybe we could've made the whole function it a bit simpler. Something like:
```go
for _, c := range csr.Status.Conditions {
    if c.Status == corev1.ConditionTrue {
        switch c.Type {
        case certv1.CertificateApproved:
            return len(csr.Status.Certificate) > 0, nil
        case certv1.CertificateDenied, certv1.CertificateFailed:
            return false, fmt.Errorf("failed to approve CSR %s - reason: %s, message: %s", csr.Name, c.Reason, c.Message)
        }
    }
}

return false, nil
```

src/k8s/pkg/k8sd/controllers/csrsigning/reconcile.go
```diff
@@ -116,6 +118,11 @@ func (r *csrSigningReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 		if caKey == nil {
 			log.Error(err, "Cannot sign certificate as CA private key is not available")
+			setFailedCSR(obj, missingKeyFailedReason, missingKeyFailedMessage)
+			if err := r.Client.Status().Update(ctx, obj); err != nil {
+				log.Error(err, "Failed to update failed CSR")
+				return ctrl.Result{}, err
```
HomayoonAlimohammadi, line 124: Maybe we can remove the log and wrap the error instead?

HomayoonAlimohammadi (commented):
Great work, thanks a lot @mateoflorido! LGTM overall. Left some minor nits and questions.

src/k8s/pkg/k8sd/api/certificates_refresh.go
mateoflorido: No, if you look at the diff, it's actually quite similar to the previous implementation. The key issue here is that a certificate can have multiple conditions that can coexist. For example, a user might approve the CSRs, but if a key is missing (which is exactly what this PR is addressing), the CSR could end pu marked as both `Approved,Failed`. It might help to check the CSR [reference](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/certificate-signing-request-v1/#CertificateSigningRequestStatus) to clear up any confusion around this behavior.

src/k8s/pkg/k8sd/api/certificates_refresh.go
mateoflorido: Please check the next comment: https://github.com/canonical/k8s-snap/pull/1282/files#r2035412288

src/k8s/pkg/k8sd/api/certificates_refresh.go
mateoflorido: Also, please note that there's a fourth implicit condition: `Issued`. This is not a condition per se, but it can be inferred from the `len(csr.Status.Certificate) > 0` check.

src/k8s/pkg/k8sd/controllers/csrsigning/reconcile.go
mateoflorido: Sure, but I think we should keep the style consistent with the controller. If you look at the structure, it uses the log/return pattern. Wdyt?

src/k8s/pkg/k8sd/api/certificates_refresh.go
HomayoonAlimohammadi: Oh, ok I there might be multiple conditions available all at the same time, but we do care about them in a certain priority (denied, failed, approved).

HomayoonAlimohammadi (approved):
Amazing, thanks a lot @mateoflorido! LGTM

PR #1302, opened 2025-04-14
test: generate subunit test reports

petrutlucian94:
The test logs are quite verbose (which is good), however it's difficult to find the results of a given test that failed.

To address this, we'll stream the test results in subunit format[1][2][3], which is heavily used by Openstack projects.

The pytest-subunit [4][5] plugin seems unmaintained, however it's lightweight and can be simplified even further. As such, we're cleaning up the plugin and we're including it in the k8s-snap repo.

[1] https://pypi.org/project/python-subunit/
[2] https://launchpad.net/subunit
[3] https://github.com/testing-cabal/subunit

[4] https://pypi.org/project/pytest-subunit/
[5] https://github.com/jelmer/pytest-subunit


tests/integration/requirements-test.txt
```diff
@@ -6,3 +6,5 @@ tenacity==8.2.3
 pylint==3.2.5
 cryptography==44.0.1
 hvac==2.3.0
+python-subunit>=1.2
+os-testr
```
bschimke95, line 10: Do we still need this dependency now?

bschimke95 (commented):
Minor question otherwise LGTM

tests/integration/requirements-test.txt
petrutlucian94: Yes, I dropped pytest-subunit but we still need the following:
* os-testr - includes the subunit2html script
* python-subunit - used by the plugin that's now included in k8s-snap to output subunit results

bschimke95 (approved):

PR #1312, opened 2025-04-16
test: Avoid decoding strings

claudiubelu:
## Description

Currently, we're assuming that a process' ``stdout`` and ``stderr`` are bytes, but that's not always the case. First of all, if the output is not captured, they are None, and if ``text=True``, then the output will be strings.

## Solution

Decode only if the output is in bytes.

## Checklist

- [x] CLA signed


bschimke95 (approved):

claudiubelu:
Both test failed during `setup_k8s_snap`, while running `apt update`.

PR #1317, opened 2025-04-17
feat: Feature upgrade rollout upgrade

bschimke95:
Introduces support for a controlled feature upgrade with a rolling upgrade strategy as done e.g. in CAPI.

See spec KU-137 for details.

### DIFF
--- a/k8s/crds/upgrades.k8sd.io.yaml
+++ b/k8s/crds/upgrades.k8sd.io.yaml
@@ -16,6 +16,10 @@ spec:
       subresources:
         status: {}  # Allows `status` updates separately
       additionalPrinterColumns:
+        - name: Strategy
+          type: string
+          description: "The strategy for this upgrade."
+          jsonPath: ".status.strategy"
         - name: Phase
           type: string
           description: "Current upgrade phase."
@@ -27,6 +31,10 @@ spec:
             status:
               type: object
               properties:
+                strategy:
+                  description: "The strategy for this upgrade."
+                  type: string
+                  enum: ["RollingUpgrade", "RollingDowngrade", "InPlace"]
                 upgradedNodes:
                   description: "List of nodes that have been upgraded."
                   type: array
--- a/src/k8s/pkg/client/kubernetes/upgrades.go
+++ b/src/k8s/pkg/client/kubernetes/upgrades.go
@@ -17,11 +17,21 @@ import (
 )
 
 // TODO: If the upgrade CRD grows, consider using kubebuilder.
+type UpgradePhase string
+
 const (
 	UpgradePhaseNodeUpgrade    = "NodeUpgrade"
 	UpgradePhaseFeatureUpgrade = "FeatureUpgrade"
 	UpgradePhaseFailed         = "Failed"
 	UpgradePhaseCompleted      = "Completed"
+)
+
+type UpgradeStrategy string
+
+const (
+	UpgradeStrategyRollingUpgrade   = "RollingUpgrade"
+	UpgradeStrategyRollingDowngrade = "RollingDowngrade"
+	UpgradeStrategyInPlace          = "InPlace"
 )
 
 const (
@@ -39,8 +49,9 @@ var (
 )
 
 type UpgradeStatus struct {
-	Phase         string   `json:"phase,omitempty"`
-	UpgradedNodes []string `json:"upgradedNodes,omitempty"`
+	Strategy      UpgradeStrategy `json:"strategy,omitempty"`
+	Phase         UpgradePhase    `json:"phase,omitempty"`
+	UpgradedNodes []string        `json:"upgradedNodes,omitempty"`
 }
 
 // TODO(Hue): (KU-3033) Use kubebuilder to generate the CRD .
@@ -101,7 +112,7 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 	return nil
 }
 
-func NewUpgrade(name string) Upgrade {
+func NewUpgrade(name string, strategy UpgradeStrategy) Upgrade {
 	return Upgrade{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: apiVersion,
@@ -110,7 +121,11 @@ func NewUpgrade(name string) Upgrade {
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
 		},
-		Status: UpgradeStatus{Phase: UpgradePhaseNodeUpgrade, UpgradedNodes: []string{}},
+		Status: UpgradeStatus{
+			Strategy:      strategy,
+			Phase:         UpgradePhaseNodeUpgrade,
+			UpgradedNodes: []string{},
+		},
 	}
 }
 
@@ -216,7 +231,7 @@ func (c *Client) PatchUpgradeStatus(ctx context.Context, upgradeName string, sta
 	}
 
 	// Wrap the status in a struct to match the CRD definition.
-	upgrade := NewUpgrade(upgradeName)
+	upgrade := NewUpgrade(upgradeName, status.Strategy)
 	upgrade.Status = status
 
 	body, err := json.Marshal(upgrade)
--- a/src/k8s/pkg/k8sd/app/hooks_join.go
+++ b/src/k8s/pkg/k8sd/app/hooks_join.go
@@ -275,8 +275,6 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
 	return nil
 }
 
-// handleRollOutUpgrade checks this join is part of a rolling upgrade and will create/modify the upgrade CRD
-// accordingly. It will also check if the node version is compatible with the cluster version.
 func handleRollOutUpgrade(ctx context.Context, snap snap.Snap, s state.State, k8sClient *kubernetes.Client) error {
 	log := log.FromContext(ctx).WithValues("step", "rollout-upgrade")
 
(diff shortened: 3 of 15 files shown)


k8s/crds/upgrades.k8sd.io.yaml
```diff
@@ -27,6 +31,10 @@ spec:
             status:
               type: object
               properties:
+                strategy:
+                  description: "The strategy for this upgrade."
+                  type: string
+                  enum: ["RollingUpgrade", "RollingDowngrade", "InPlace"]
```
HomayoonAlimohammadi, line 37: Honsetly I think the point about upgrade or downgrade which I brought up earlier is not valid. The reason I mentioned that was because I was having mixed versions clusters in mind which is not something we're going to support.
I think we should be good with `Rolling` and `InPlace`. WDYT?

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -253,5 +256,141 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
 
+	log.Info("Create Kubernetes client")
+	k8sClient, err := a.snap.KubernetesClient("")
+	if err != nil {
+		return fmt.Errorf("failed to get Kubernetes client: %w", err)
+	}
+
+	// This is required for backwards compatibility.
+	log.Info("Applying custom CRDs")
+	// TODO(ben): This sometimes collides with the CRD application in the node.
```
HomayoonAlimohammadi, line 267: What does the `CRD application` exactly mean?

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -253,5 +256,141 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+					strategy = kubernetes.UpgradeStrategyRollingDowngrade
+				}
+
+				newUpgrade := kubernetes.NewUpgrade(fmt.Sprintf("cluster-upgrade-to-rev-%s", rev), strategy)
+				newUpgrade.Status.UpgradedNodes = []string{s.Name()}
+				return k8sClient.CreateUpgrade(ctx, newUpgrade)
+			}
+		}
+	} else {
+		lowest, highest := lowestHighestK8sVersions(nodeVersions)
```
HomayoonAlimohammadi, line 359: My brain blue screened trying to understand `lowestHighestK8sVersions` :smile: 
Kind of reminds me of financial charts (higher high, lower lower, lower high, higher low)
p.s. the naming is fine, I just needed to also look at the returned values.

k8s/crds/upgrades.k8sd.io.yaml
bschimke95: I think this is required for better UX.
If the user performs an upgrade to 1.33 and tried to join a 1.32 node, we need that info.

HomayoonAlimohammadi: I agree but then shouldn't we also add InPlaceUpgrade/Downgrade?

src/k8s/pkg/k8sd/app/hooks_join.go
bschimke95: Haha, maybe that is not the most straight forward way to implement it but I couldn't really come up with a better way :)

src/k8s/pkg/k8sd/app/hooks_join.go
bschimke95: Whoops - that is a leftover comment - removed

src/k8s/pkg/client/kubernetes/upgrades.go
```diff
@@ -17,13 +17,23 @@ import (
 )
 
 // TODO: If the upgrade CRD grows, consider using kubebuilder.
+type UpgradePhase string
```
berkayoz, line 20: Shouldn't we declare `UpgradePhase` type explicitly here? Wouldn't go infer string for the consts below without this?

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -253,5 +256,141 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+		if clusterK8sVersion == nil {
+			// we should never get here
+			return fmt.Errorf("the cluster has no nodes - cannot determine cluster Kubernetes version")
+		}
+		for node, version := range nodeVersions {
+			log.Info("Checking node version", "this", thisNodeVersion, "other", node, "version", version)
+
+			if thisNodeVersion.EqualTo(version) {
+				continue
+			}
```
HomayoonAlimohammadi, line 331: QQ: Can't we use `clusterK8sVersion` instead of looping through nodes and their versions?

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -253,5 +256,141 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+		return fmt.Errorf("failed to parse node Kubernetes version %q: %w", thisNodeVersionStr, err)
+	}
+
+	nodeVersions, err := k8sClient.NodeVersions(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get node Kubernetes versions: %w", err)
+	}
+
+	nodeName := s.Name()
+	if upgrade == nil {
```
HomayoonAlimohammadi, line 305: nit: Maybe we can break down the logic here? I think it's gonna help with readability.

src/k8s/pkg/snap/snap.go
```diff
@@ -297,6 +297,24 @@ func (s *snap) NodeTokenFile() string {
...
+		return "", fmt.Errorf("failed to get snap info: %w", err)
+	}
+
+	if snap.StatusCode != 200 {
+		return "", fmt.Errorf("failed to get snap info: snapd returned with error code %d", snap.StatusCode)
+	}
+
+	return snap.Result.Version, nil
+}
+
```
berkayoz, line 317: This is great, I wonder if it would be better to use build time variables to inject the version.

k8s/crds/upgrades.k8sd.io.yaml
HomayoonAlimohammadi: Resolved as per the discussion in MM.

src/k8s/pkg/k8sd/app/hooks_join.go
berkayoz: +1  this function feels a bit large.

berkayoz (commented):
Great work, left some comments

src/k8s/pkg/client/kubernetes/upgrades.go
HomayoonAlimohammadi: That's right, but we _can_ ignore this for this PR. I've change this part a bit in `https://github.com/canonical/k8s-snap/pull/1327` PR.

bschimke95: meh, I don't know how to Go...

src/k8s/pkg/snap/snap.go
bschimke95: What advantage do you see in this?

src/k8s/pkg/k8sd/app/hooks_join.go
bschimke95: Fair point - done.

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -253,5 +256,156 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+
+	if thisNodeVersion.Major() == clusterK8sVersion.Major() && thisNodeVersion.Minor() == clusterK8sVersion.Minor() {
+		return fmt.Errorf("the joining node has a different patch version than cluster nodes")
+	}
+
+	if thisNodeVersion.Major() == clusterK8sVersion.Major() && thisNodeVersion.Minor() != clusterK8sVersion.Minor() {
+		return initiateRollingUpgrade(ctx, snap, s, k8sClient, thisNodeVersion, clusterK8sVersion)
+	}
+
+	return nil
```
HomayoonAlimohammadi, line 348: nit: Are we worried about major version changes?
```suggestion
	return initiateRollingUpgrade(ctx, snap, s, k8sClient, thisNodeVersion, clusterK8sVersion)
```

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -253,5 +256,156 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+	var strategy kubernetes.UpgradeStrategy
+	if thisNodeVersion.GreaterThan(clusterK8sVersion) {
+		log.Info("Joining node has a greater version - rolling upgrade")
+		strategy = kubernetes.UpgradeStrategyRollingUpgrade
+	} else {
+		log.Info("Joining node has a lower version - downgrade")
+		strategy = kubernetes.UpgradeStrategyRollingDowngrade
+	}
+
+	newUpgrade := kubernetes.NewUpgrade(fmt.Sprintf("cluster-upgrade-to-rev-%s", rev), strategy)
```
HomayoonAlimohammadi, line 369: I don't know how I feel about naming this `upgrade` when it can be a `RollingDowngrade`, but let's forget about it for now (p.s. the CRD is `Upgrade` as well, so...)

src/k8s/pkg/k8sd/app/hooks_node_ready.go
```diff
@@ -22,7 +22,7 @@ func (a *App) onNodeReady(ctx context.Context, s state.State) error {
 	// Apply all custom CRDs on startup
 	log.Info("Applying custom CRDs")
 	if err := a.applyCustomCRDs(ctx); err != nil {
-		return fmt.Errorf("failed to apply custom CRDs: %w", err)
+		log.Info("failed to apply custom CRDs: %w", err)
```
HomayoonAlimohammadi, line 25: How about `log.Error`?

tests/integration/tests/test_version_upgrades.py
```diff
@@ -340,13 +340,10 @@ def test_feature_upgrades(instances: List[harness.Instance], tmp_path: Path):
         ).stdout
 
         if idx == len(instances) - 1:
-            util.stubbornly(retries=15, delay_s=5).on(instance).until(
```
HomayoonAlimohammadi, line 343: nit: I think the `stubbornly` part is not a bad idea to prevent flakiness (since we're dealing with the upgrade controller). Maybe use it a bit higher when we're getting the `phase`?

HomayoonAlimohammadi (approved):
Great work, thanks a lot @bschimke95! LGTM, just some minor nits that we can also address later. I think it's more important to make sure this one lands in time.

berkayoz (approved):
LGTM overall, rest can be addressed later thanks!

src/k8s/pkg/snap/snap.go
bschimke95: Good point - updated the implementation to use bom.json instead - as discussed offline.

PR #1318, opened 2025-04-17
docs: add charm hardening page

petrutlucian94:
We're mostly reusing the snap how-to guide, just adding a note that the same settings can be applied by either connecting to the Juju units or using settings such as `kube-apiserver-extra-args`.


bschimke95 (approved):
LGTM

PR #1321, opened 2025-04-17
feat: Support multiple pre-releases (#29)

bschimke95:
Right now, only the latest pre-release is supported. However, in transition times there might be multiple non-stable releases (e.g. 1.33.0-rc1 and 1.34.0-alpha) which both need to be supported.

Related PR: https://github.com/canonical/canonical-kubernetes-release-ci/pull/29 (why probably should put those related functions into a util repo or so at some point...)

### DIFF
--- a/build-scripts/k8s_release.py
+++ b/build-scripts/k8s_release.py
@@ -131,7 +131,9 @@ def get_obsolete_prereleases() -> List[str]:
 def _exec(*args, **kwargs) -> tuple[str, str]:
     """Run the specified command and return the stdout/stderr output as a tuple."""
     LOG.debug("Executing: %s, args: %s, kwargs: %s", cmd, args, kwargs)
-    kwargs["text"] = True
+    kwargs.setdefault("text", True)
+    kwargs.setdefault("check", True)
+    kwargs.setdefault("timeout", EXEC_TIMEOUT)
     proc = subprocess.run(*args, **kwargs)
     return proc.stdout, proc.stderr
 
--- a/build-scripts/k8s_release.py
+++ b/build-scripts/k8s_release.py
@@ -189,26 +189,45 @@ def prepare_prerelease_git_branches(project_basedir: str, remote: str = "origin"
     for prerelease in prereleases:
         branch = get_prerelease_git_branch(str(prerelease))
         LOG.info("Preparing pre-release branch: %s", branch)
+
+        # Reset branch to remote main
         _exec(
-            ["git", "checkout", "-B", branch],
+            ["git", "fetch", remote],
+            cwd=project_basedir,
+            capture_output=False,
+        )
+        _exec(
+            ["git", "checkout", "-B", branch, f"{remote}/main"],
             cwd=project_basedir,
             capture_output=False,
         )
 
+        # Update the k8s version and commit
         _update_prerelease_k8s_component(project_basedir, str(prerelease))
         _exec(
             ["git", "add", "./build-scripts/components/kubernetes/version"],
             cwd=project_basedir,
             capture_output=False,
         )
-        _exec(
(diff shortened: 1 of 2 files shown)


build-scripts/k8s_release.py
```diff
@@ -18,72 +18,131 @@
...
+            if (version.major, version.minor) in seen_stable_minors:
+                obsolete.append(tag)
+
+    return obsolete
 
 
-def _exec(cmd: List[str], check=True, timeout=EXEC_TIMEOUT, cwd=None):
+def _exec(
+    cmd: List[str], check=True, capture_output=True, timeout=EXEC_TIMEOUT, cwd=None
+):
```
addyess, line 136: it looks like you just want this to wrap `subprocess.run` to return strings at this point?

```suggestion
def _exec(*args, **kwargs) -> tuple[str, str]:
```

build-scripts/k8s_release.py
```diff
@@ -18,72 +18,131 @@
...
     """Run the specified command and return the stdout/stderr output as a tuple."""
     LOG.debug("Executing: %s, cwd: %s.", cmd, cwd)
     proc = subprocess.run(
```
addyess, line 139:
```suggestion
    kwargs["text"]=True
    proc = subprocess.run(*args, **kwargs)
    return proc.stdout, proc.stderr
```

addyess (approved):
my only fuss here is the param growth of `_exec`.  Pretty nice PR

build-scripts/k8s_release.py
```diff
@@ -18,72 +18,131 @@
...
+    """
+    latest_by_minor: Dict[str, str] = {}
+    version_regex = re.compile(r"^v?(\d+)\.(\d+)\..+")
+
+    for tag in get_k8s_tags():
+        match = version_regex.match(tag)
+        if not match:
+            continue
+        major, minor = match.groups()
+        key = f"{major}.{minor}"
```
petrutlucian94, line 87: Nit: I'd rather parse the version using `packaging.version.Version` instead of regex. It's cleaner / easier to read.

build-scripts/k8s_release.py
```diff
@@ -18,72 +18,131 @@
...
-def _exec(cmd: List[str], check=True, timeout=EXEC_TIMEOUT, cwd=None):
+def _exec(*args, **kwargs) -> tuple[str, str]:
     """Run the specified command and return the stdout/stderr output as a tuple."""
     LOG.debug("Executing: %s, cwd: %s.", cmd, cwd)
-    proc = subprocess.run(
-        cmd, check=check, timeout=timeout, cwd=cwd, capture_output=True, text=True
+    kwargs["text"]=True
+    proc = subprocess.run(*args, **kwargs)
+    return proc.stdout, proc.stderr
+        cmd,
```
petrutlucian94, line 140: Is this a leftover? Also, we're no longer applying the following defaults:
* timeout=EXEC_TIMEOUT
* check=True

build-scripts/k8s_release.py
```diff
@@ -18,73 +18,121 @@
...
+            version = Version(tag.lstrip("v").split("-")[0])
+            if (version.major, version.minor) in seen_stable_minors:
+                obsolete.append(tag)
 
+    return obsolete
 
-def _exec(cmd: List[str], check=True, timeout=EXEC_TIMEOUT, cwd=None):
+
+def _exec(*args, **kwargs) -> tuple[str, str]:
     """Run the specified command and return the stdout/stderr output as a tuple."""
```
petrutlucian94, line 132: I still think we should pass `check` (otherwise it will ignore non-zero exit codes) and `timeout` (so that it doesn't hang indefinitely).

petrutlucian94 (approved):

PR #1337, opened 2025-04-23
test: Remove mixed-version join test

bschimke95:
This test was originally introduced to validate that minor version upgrades work by allowing nodes with different versions to join the same cluster. However, we now have dedicated version upgrade tests that more accurately cover this scenario.

With the introduction of the rollout upgrade feature, this test has started to fail—nodes attempting to join a cluster with a different patch version are now rejected, as expected. The correct process is to upgrade the cluster first, then join the node.

Since the test no longer reflects a valid upgrade path and has become obsolete, it is being removed.


HomayoonAlimohammadi (approved):
This is not required anymore according to a recent spec. Thanks for getting rid of it. LGTM!

PR #1354, opened 2025-04-24

cdkbot:


bschimke95:
The branchmanagement CI is expected to fail since the 1.33 release branch is not cut yet. I will fix this in a follow-up PR.

PR #1363, opened 2025-04-25
fix: update node label controller, allowing custom node names

petrutlucian94:
The k8sd node label controller currently uses the host name when monitoring k8s node events.

We need to obtain the node name from microcluster instead, accommodating custom node names.

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_start.go
+++ b/src/k8s/pkg/k8sd/app/hooks_start.go
@@ -29,7 +29,6 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
 		return fmt.Errorf("failed to retrieve microcluster status: %w", err)
 	}
 	nodeName := serverStatus.Name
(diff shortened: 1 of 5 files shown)


src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -24,6 +24,12 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
 		}
 	}()
 
+	serverStatus, err := a.cluster.Status(ctx)
```
petrutlucian94, line 27: TODO: this returns the host name instead of the node name that was passed to `NewCluster`.

https://chat.canonical.com/canonical/pl/b5c8nzioi3bazbn4k37g9q14xr

We'll need to fix that in microcluster or find another way of retrieving the node name.

bschimke95 (approved):
This should be backported together with the AZ PR @berkayoz

PR #1370, opened 2025-04-29
fix: Properly parse IPv6 endpoints for k8s-apiserver-proxy.json

claudiubelu:
## Description
Currently, in scenarios with clusters with IPv6, we will end up with improper IPv6 addresses in the k8s-apiserver-proxy.json, causing the k8s-apiserver-proxy service to always fail with the following error:

"Failed to start" err="failed to parse endpoint \"fd42:6db9:e49a:486b:216:3eff:fe48:5318:6443\": address fd42:6db9:e49a:486b:216:3eff:fe48:5318:6443: too many colons in address"

According to the documentation [1], the expectation is that the IPv6 address to be enclosed in brackets ([IPv6]:port).

We're writing the k8s-apiserver-proxy.json in 2 cases:
- when joining a worker node.
- when a new control plane node joined the cluster.

In the first scenario, we are properly handling the IPv6 addresses, but not in the second. This PR addresses this issue.

[1] https://pkg.go.dev/net#SplitHostPort

## Solution

The IPv6 addresses are now properly handled.

## Backport

`release-1.33`, `release-1.32`, `release-1.31`.

## Checklist

- [x] Covered by unit tests
- [x] Documentation updated - N/A
- [x] CLA signed

### DIFF
--- a/src/k8s/pkg/client/kubernetes/endpoints.go
+++ b/src/k8s/pkg/client/kubernetes/endpoints.go
@@ -3,7 +3,6 @@ package kubernetes
 import (
 	"context"
 	"fmt"
-	"sort"
 
 	"github.com/canonical/k8s/pkg/utils"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


berkayoz (approved):
Great work LGTM, we should also probably think about the deprecation of `Endpoints` see https://kubernetes.io/blog/2025/04/24/endpoints-deprecation/

PR #1374, opened 2025-04-30
test: Add performance tests machinery

bschimke95:
This PR introduces machinery to create performance tests and
measure them agains a historic baseline.

* Create test structures including tags and `pytest-benchmark` to create
  performance tests
* Introduce structure to store and compare performance against a
  historic baseline that runs on specified hardware (GH runners)
* Create a workflow that creates new performance baselines when requirements
  or performance tests change
* Extend the nightly test with performance tests that validates current
  latests edge with the current baseline and fails if performance
significantly deteriorated

See KU-152 for more details

### DIFF
--- a/.github/workflows/create-perf-baselines.yaml
+++ b/.github/workflows/create-perf-baselines.yaml
@@ -1,0 +1,64 @@
+name: Create Performance Baselines
+description: |
+  This workflow creates performance baselines on the specified runner tags
+  and creates a baseline for future comparisons. It uses pytest-benchmark to run the tests
+  and save the results.
+  It also creates a pull request with the results
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - .github/workflows/create-perf-baselines.yaml
+      - tests/integration/tests/performance/*.py
+
+jobs:
+  baselines:
+    name: Create Baselines for ${{ matrix.os }}-${{ matrix.arch }}
+    strategy:
+      matrix:
+        os: ["ubuntu:22.04", "ubuntu:24.04"]
+        arch: ["amd64", "arm64"]
+    runs-on: ${{ inputs.arch == 'arm64' && 'self-hosted-linux-arm64-noble-medium' || 'self-hosted-linux-amd64-noble-medium' }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Create performance baselines
+        run: |
+          tox -e integration -- --tags performance \
+            --benchmark-storage baselines \
+            --benchmark-name ${ matrix.os }-${{ matrix.arch }}
+      - name: Prepare baseline results
+        run: |
+          echo "artifact_name=${{ inputs.os }}-${{ inputs.arch }}.json" | sed 's/:/-/g' >> $GITHUB_ENV
+      - name: Upload baseline results
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.artifact_name }}
+          path: baselines/**/*.json
+
+  create_pr:
+    name: Create PR with baseline results
+    runs-on: ubuntu-latest
+    needs: baselines
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Download artifact
+        uses: actions/download-artifact@v4
+        with:
+          path: tests/integration/tests/performance/baselines
+      - name: Commit and create PR
+        uses: peter-evans/create-pull-request@v6
+        with:
+          commit-message: "test: Update performance baseline results"
+          title: "test: Automatic Performance Baselines update"
+          body: "Update performance baselines because of changed performance tests"
+          branch: performance/baseline-updates
+          base: ${{ github.head_ref || github.ref_name }}
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/nightly-test.yaml
+++ b/.github/workflows/nightly-test.yaml
@@ -13,78 +13,30 @@ permissions:
   contents: read
 
 jobs:
-  test-integration:
-    name: Integration
+  test-performance:
+    name: Run performance tests for ${{ matrix.os }}-${{ matrix.arch }}
     strategy:
       matrix:
         os: ["ubuntu:22.04", "ubuntu:24.04"]
         arch: ["amd64", "arm64"]
-        channel: ["latest/edge"]
-      fail-fast: false # TODO: remove once we no longer have flaky tests.
-    uses: ./.github/workflows/e2e-tests.yaml
-    with:
-      arch: ${{ matrix.arch }}
-      os: ${{ matrix.os }}
-      channel: ${{ matrix.channel }}
-      test-tags: 'up_to_nightly conformance_tests'
-
-  Trivy:
-    permissions:
-      contents: read # for actions/checkout to fetch code
-      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
-    strategy:
-      matrix:
-        include:
-          # Latest branches
-          - { branch: main, channel: latest/edge }
-          # Stable branches
-          # Add branches to test here
-          # TODO: automatically retrieve the list of channels.
-          - { branch: release-1.30, channel: 1.30-classic/edge }
-          - { branch: release-1.31, channel: 1.31-classic/edge }
-          - { branch: release-1.32, channel: 1.32-classic/edge }
-    uses: ./.github/workflows/security-scan.yaml
-    with:
-      channel: ${{ matrix.channel }}
-      checkout-ref: ${{ matrix.branch }}
-      upload-reports-to-jira: true
-    secrets: inherit
-
-  TICS:
-    if: ${{ always() && github.event_name == 'schedule' }}
-    permissions:
-      contents: read
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        include:
-          - { branch: main }
+    runs-on: ${{ matrix.arch == 'arm64' && 'self-hosted-linux-arm64-noble-medium' || 'self-hosted-linux-amd64-noble-medium' }}
     steps:
-      - name: Checking out repo
+      - name: Checkout repo
         uses: actions/checkout@v4
-      - name: Save tics-scan.sh script
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Run Performance Tests
+        env:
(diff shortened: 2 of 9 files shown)


tests/integration/tests/performance/test_perf_status.py
```diff
@@ -0,0 +1,20 @@
...
+@pytest.mark.tags(tags.PULL_REQUEST, tags.PERFORMANCE)
+def test_perf_status_single_node_cli(
+    instances: List[harness.Instance], tmp_path: str, benchmark
+):
+    node = instances[0]
+
+    def run():
+        node.exec(["k8s", "status"])
+
+    benchmark.pedantic(run, rounds=20)
```
louiseschmidtgen, line 20: Is there a pause between the rounds? Or do they immediately fire after each other?

tests/integration/tests/test_util/config.py
```diff
@@ -188,3 +188,6 @@
 
 # SONOBUOY_VERSION is the version of sonobuoy to use for CNCF conformance tests.
 SONOBUOY_VERSION = os.getenv("TEST_SONOBUOY_VERSION") or "v0.57.3"
+
+# PERF_DEFAULT_ROUNDS is the number of rounds to run for performance tests.
+PERF_DEFAULT_ROUNDS = int(os.getenv("TEST_PERF_DEFAULT_ROUNDS") or 5)
```
louiseschmidtgen, line 193:
```suggestion
PERF_DEFAULT_ROUNDS = int(os.getenv("TEST_PERF_DEFAULT_ROUNDS") or 20)
```
Let's put 20 here, 5 would lead to insufficient results.

.github/workflows/nightly-test.yaml
```diff
@@ -13,78 +13,30 @@ permissions:
...
+        env:
+          TEST_SUBSTRATE: lxd
+          TEST_LXD_IMAGE: ${{ matrix.os }}
         run: |
-          export TICSAUTHTOKEN=${{ secrets.TICSAUTHTOKEN }}
-          /tmp/tics-scan.sh `pwd`
+          cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- --tags performance \
+            --benchmark-storage tests/integration/tests/performance/baselines \
+            --benchmark-compare ${{ matrix.os }}-${{ matrix.arch }}.json
+            --benchmark-compare-fail mean:10%
```
louiseschmidtgen, line 38: We have to be careful with this baseline as our degradation may be less than the statistical error. 
I'm a bit skeptical whether this is going to give us any useful information but let's see what you find out. Maybe I'm wrong :)

louiseschmidtgen (commented):
Nice work Ben, couple of first comments.

tests/integration/tests/test_util/config.py
bschimke95: I think I just remove this altogether and let the performance test define this explicitly.

tests/integration/tests/performance/test_perf_status.py
bschimke95: I couldn't find any specific on that in the documentation. I assume not.
You can have a tiny sleep in the setup/teardown if required, though.

.github/workflows/nightly-test.yaml
bschimke95: Discussed offline. tl;dr:
1. statistical error is reduced through multiple iterations per round which smoothens outliers
2. confidence interval to be considered in future iterations for a more stable metric. This is not supported out-of-the-box by pytest-benchmark but can easily be computated.

bschimke95:
Changing the performance tests will create an automatic PR as in https://github.com/canonical/k8s-snap/pull/1389

.github/workflows/create-perf-baselines.yaml
```diff
@@ -0,0 +1,83 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          path: ../artifacts
+      - name: Flatten results
+        shell: bash
+        run: |
+          mkdir -p tests/integration/tests/performance/baselines
+          # Move the performance baseline files to the correct location and remove the prefix.
+          find ../artifacts -name "performance-baseline-*.json" -exec bash -c 'f="{}"; mv "$f" tests/integration/tests/performance/baselines/${f##*/performance-baseline-}' \;
+      - name: Commit and create PR
```
louiseschmidtgen, line 75: I think these tests should be commited with the PR. What if we open a PR run the test and close the PR again? We'll have the baselines update PR merged already?

.github/workflows/e2e-tests.yaml
```diff
@@ -97,6 +113,17 @@ jobs:
       - name: Setup tmate session
         uses: canonical/action-tmate@main
         if: ${{ github.event_name == 'workflow_dispatch' && inputs.tmate_enabled }}
+      - name: Sanitize test name
```
louiseschmidtgen, line 116: What's sanitizing in this context? Maybe we can come up with a better name? Maybe test_case_name? Wdyt?

louiseschmidtgen (commented):
Nice work Ben!
I'll skip my thoughts on the failure metric- we can deal with it later.

Do we really need to commit the json blob to the repo? It contains a lot of extra info.
The json itself is not very useful when comparing two performance runs in the git commit history. However, the json is nice to do some analysis in python on a run. 
WHy don't you rather generate the json blobs as tmp files and publish a comparison generated with python in an issue comment.

tests/integration/tests/performance/test_perf_clustering.py
HomayoonAlimohammadi, line 1: Since these tests are under `performance` directory, maybe we can remove the `pref` from the file name?

HomayoonAlimohammadi (commented):
Looks great, thanks a lot @bschimke95! Left some comments and suggestions.

.github/workflows/create-perf-baselines.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Download artifact
+        uses: actions/download-artifact@v4
+        with:
+          path: ../artifacts
+      - name: Flatten results
+        shell: bash
+        run: |
+          mkdir -p tests/integration/tests/performance/baselines
+          # Move the performance baseline files to the correct location and remove the prefix.
+          find ../artifacts -name "performance-baseline-*.json" -exec bash -c 'f="{}"; mv "$f" tests/integration/tests/performance/baselines/${f##*/performance-baseline-}' \;
```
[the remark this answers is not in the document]
bschimke95: obsolete now

tests/integration/tests/performance/test_perf_clustering.py
bschimke95: I initially implemented it like this but found it annoying that I always opened the wrong `test_clustering.py` in my VSCode if I typed `test_clus`
I'd rather keep it.

.github/workflows/nightly-test.yaml
```diff
@@ -28,105 +29,118 @@ jobs:
...
+    uses: ./.github/workflows/compare-performance.yaml
     with:
-      channel: ${{ matrix.channel }}
-      checkout-ref: ${{ matrix.branch }}
-      upload-reports-to-jira: true
-    secrets: inherit
+      arch: ${{ matrix.arch }}
+      os: ${{ matrix.os }}
+      channel-1: latest/edge
+      channel-2: 1.32-classic/stable
```
bschimke95, line 43: I'm doing some refactoring [here](https://github.com/canonical/k8s-snap/actions/runs/15209292201/job/42779462615?pr=1468) that makes it easier to select a proper channel. For now, I'll hard-code this.

louiseschmidtgen (changes requested):
Awesome work Ben! I like the comparison much better across snaps. My main concern is the parallelism of the jobs and the test cases. Otherwise a few nit picks:

.github/workflows/compare-performance.yaml
```diff
@@ -0,0 +1,151 @@
...
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install pytest-benchmark
+        run: pip install pytest pytest-benchmark pytest-benchmark[histogram]
```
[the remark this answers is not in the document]
bschimke95: good point - added a new tox environment

.github/workflows/compare-performance.yaml
```diff
@@ -0,0 +1,151 @@
...
+    name: Run performance test with second snap
+    uses: ./.github/workflows/e2e-tests.yaml
+    with:
+      arch: ${{ inputs.arch }}
+      os: ${{ inputs.os }}
+      artifact: ${{ inputs.artifact-2 }}
+      channel: ${{ inputs.channel-2 }}
+      test-tags: performance
+      # All performance tests run in one job to collect results in one report.
+      parallel: false
```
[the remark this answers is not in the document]
bschimke95: Disscussed offline - all jobs run in separate VMs

tests/integration/performance_test_results.py
```diff
@@ -0,0 +1,142 @@
...
+                result_files.append(os.path.join(root, file))
+    return result_files
+
+
+def get_test_name_from_path(file_path):
+    """Extract test name from file path."""
+    return os.path.basename(file_path).split('.')[0]
+
+
+def compare_benchmarks(base_dir):
```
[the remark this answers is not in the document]
bschimke95: This is actually an obsolete file - I discovered later that the CLI already ships with all I need. 
Removed it - thanks!

.github/workflows/e2e-tests.yaml
```diff
@@ -34,6 +34,11 @@ on:
           Additional pytest arguments, use "-k <test_name>" to run a specific test
         default: ""
         type: string
+      parallel:
+        description: Run tests in parallel using GitHub Actions matrix
+        type: boolean
+        default: true
+
```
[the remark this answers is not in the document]
bschimke95: As discussed, noisy neighbor is not a problem here.
Also note that this e2e-tests workflow is not just used for performance tests but for all integration tests which we want to run in parallel.

.github/workflows/compare-performance.yaml
```diff
@@ -0,0 +1,151 @@
...
+        description: Second snap artifact path to compare
+        type: string
+      channel-1:
+        description: Channel to download the first snap from
+        type: string
+        default: ""
+      channel-2:
+        description: Channel to download the second snap from
+        type: string
+        default: ""
```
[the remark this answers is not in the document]
louiseschmidtgen: That's fine

bschimke95:
The failing tests are unrelated and something we also see in the nightly tests.
We'll investigate as part of follow-up work. Merging this.

PR #1381, opened 2025-05-01
docs: Update troubleshooting with cilium tunnel port fix 

nhennigan:
## Description

Users have faced issues when there are other installations of cilium / other networking services also installed on the hose trying to use the default port for VXLAN encapsulation.

## Solution

Now you can alter the default Cilium tunnel port through annotations but this is not in our docs yet

## Backport

1.32, 1.33

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/docs/canonicalk8s/charm/howto/install/install-terraform.md
+++ b/docs/canonicalk8s/charm/howto/install/install-terraform.md
@@ -21,7 +21,7 @@ controller. Choose one of the options outlined in the
 
 The Terraform deployment is done using a root module that specifies the
 Juju model to deploy the submodules to. The root module also references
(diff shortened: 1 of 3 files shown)


rapour:
Great work @nhennigan 
We can simplify the troubleshooting process since Canonical Kubernetes removes the blocked interface and changes the cilium configuration once cilium's tunnel port is set via annotations. The user needs to either disable the fan or change the cilium port and it should be enough

rapour:
The annotations are part of the cluster config and I suppose setting it once in a one of the nodes should be enough

PR #1384, opened 2025-05-02
docs: add notes for new `refresh-certs` ux

mateoflorido:
## Description

Update the `refresh-certs` docs to reflect the new user experience.
## Solution

This pull request updates the `refresh-certs` docs to provide clearer guidance on selective certificate renewal using the `--certificates` flag. The changes include new notes to help users understand how to target specific certificates during the refresh process.

## Issue

N/A

## Backport

N/A

## Checklist

- [ ] PR title formatted as `type: title`
- [ ] Covered by unit tests
- [ ] Documentation updated
- [ ] CLA signed

### DIFF
--- a/docs/canonicalk8s/snap/howto/security/refresh-external-certs.md
+++ b/docs/canonicalk8s/snap/howto/security/refresh-external-certs.md
@@ -21,6 +21,13 @@ argument with a YAML-formatted file. For a complete list of available
 certificate keys, see the
 [certificates refresh configuration file reference page][reference page].
 
+```{note} If your cluster uses a mixed certificate management approach where
+some certificates are managed externally and others internally, you must
+explicitly specify the internally managed certificates to refresh on worker
+nodes using the `--certificates` flag. Externally managed certificates should
+continue to be provided through the `--external-certificates` argument.
+```
+
 Refer to the {{ product }}
 [cluster certificates and configuration directories][certificates]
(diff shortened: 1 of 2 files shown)


addyess (approved):
Thank you @mateoflorido

nhennigan (commented):
Overall I think the info is good but I think we have used too many notes on these 2 files. We should reserve the notes for critical pieces of info otherwise they loose their impact. I am looking at how we can lay out the page better but I think from your side its good. I'll propose some changes shortly

docs/canonicalk8s/snap/howto/security/refresh-external-certs.md
```diff
@@ -21,6 +21,13 @@ argument with a YAML-formatted file. For a complete list of available
 certificate keys, see the
 [certificates refresh configuration file reference page][reference page].
 
+```{note} If your cluster uses a mixed certificate management approach where
```
nhennigan, line 24: Does this mean that if I have external certs and internal certs on my worker node and I just ask it to `sudo k8s refresh-certs --expires-in 10y --timeout 10m` it won't work? I would assume that it would update all the internally managed certs only?

mateoflorido: Yes, it won't work because there's no way to get the necessary information from the worker's end

Issue #1137, opened 2025-03-01
Better support for host systems with existing Docker/containerd


rsyring:
#### Summary

I'd like to see support for install on a developer machine included in the default snap install without the concern of conflicting with an existing Docker/containerd install.

#### Why is this important?

1. Despite being mentioned in numerous places, it's easy to miss this constraint b/c it's not usual for a snap to be able to cause problems outside the snap.
2. While there are other options for running in a dev environment, like LXD, it's more steps.  Easier to install locally means easier to adopt.
3. If it's just a matter of setting a single path in the config, i.e. `containerd-base-dir`, it shouldn't be too hard to detect an existing Docker install and set this path accordingly?

#### Are you interested in contributing to this feature?

No.  But appreciate your consideration.

## Edit

On second thought, why not just use a custom socket path in the first place and never risk the conflict?

rapour:
Hi @rsyring 
Thank you for your interest in using Canonical Kubernetes 

We chose not to support a custom socket path because many upstream projects, like the GPU Operator, expect it to be in the default location. Changing the path would complicate integration with those tools. Additionally, if we automatically adjust the path when another instance of containerd is present, users may lose compatibility with these tools without notification. It’s preferable for users to make that decision themselves to ensure they maintain control over their setup.

Please let me know if I could help with anything

rsyring:
I guess, my point is, can you help the user make the decision?  The warnings in the docs make it seem like it's problematic to use Canonical Kubernetes (CK) on a dev machine.  There is a lot of urging to use a dedicated container/VM/host.  But how important is that really?  If the socket exists, just use it?  Why make the user manually choose the existing path and/or a different path?  I see in the code you are already detecting when the socket exists, so CK is already aware there is a "problem."

I realize I'm showing some ignorance about the internals, but TBH that's kind of the point.  I'm mostly an application developer who wants to try CK.  But I don't want to make it more complicated than it already is by using LXD, etc.  It's already in a snap, that feels like it should be isolated enough.

Can you make the bootstrap "just work" for a development environment?  If not, can you make it more clear about what a developer should do?  For instance, I'm not sure if I should be choosing some random path `/tmp/ck-containerd.sock` or something else.  I.e. maybe when bootstrapping you could offer options?

```
We've detected an existing container.d socket at ($PATH).  Would you like to:

  1. Use it (recommended)
  2. Use a custom path for this Kubernetes install (`/run/canonical-k8s-containerd.sock`)
[...]

PR #1423, opened 2025-05-08
fix: Perform PreInitChecks as early as possible

claudiubelu:
## Description

Currently, when bootstrapping or joining a node, we are doing a few pre-init checks: make sure the binaries work, kubernetes-related ports are not in use, the standard or given containerd path is not currently in use.

However, it takes a while to get to that stage, while we can easily check this client-side much faster, immediately giving feedback to the user that something will go wrong during bootstrap / joining.

Additionally, if the join cluster operation fails because of a pre-init check server-side, the token will have been used and no longer available. The user would have to generate another token and try again. This would not be an issue if we perform these checks client-side.

Doing these checks early will improve the user experience.

## Solution

The pre-init checks are now also being done client-side when running the `k8s bootstrap` or `k8s join-node` commands.

## Backport

`release-1.33`, `release-1.32`

## Checklist

- [x] Covered by unit tests
- [x] Covered by integration tests - existing integration tests
- [ ] Documentation updated
- [x] CLA signed


bschimke95 (changes requested):
I see the value in doing this. However, CLI and API should behave similar. The `validate` function should be the same on the hook and on the CLI for this reason and for maintenance

claudiubelu:
> I see the value in doing this. However, CLI and API should behave similar. The `validate` function should be the same on the hook and on the CLI for this reason and for maintenance

It is calling the same `PreInitChecks` as on the hook. Indeed, the server-side hooks make additional checks (including on the microcluster side), but the checks in `PreInitChecks` are more commonly hit by users, especially new ones (the containerd socket path not being available).

claudiubelu:
Related: https://github.com/canonical/k8s-snap/issues/1137

PR #1425, opened 2025-05-08
fix: patch same upgrade status object

claudiubelu:
## Description

When a second upgraded node joins the cluster, the ``Upgrade.k8sd.io`` object is updated. However, the updated ``upgrade.status`` object does not contain the ``strategy`` field, which is required. This leads to errors like this:

```
err="failed to patch upgrade status: Upgrade.k8sd.io \"cluster-upgrade-to-rev-foo\" is invalid: [status.phase: Required value, status.strategy: Required value]"
```

## Solution

Patch existing `upgrade.status`.

## Issue

## Backport

`release-1.33`, `release-1.32`

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_join.go
+++ b/src/k8s/pkg/k8sd/app/hooks_join.go
@@ -407,9 +407,8 @@ func handleUpgradeInProgress(ctx context.Context, s state.State, k8sClient *kube
 	}
 
 	log.Info("Marking node as upgraded", "node", nodeName)
-	status := upgradesv1alpha.UpgradeStatus{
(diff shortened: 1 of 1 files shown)


HomayoonAlimohammadi (approved):
What dazzles me is that we're "patching" the upgrade resource ([here](https://github.com/canonical/k8s-snap/blob/5d62508ab0a90873656765f9290a52e0469b69d3/src/k8s/pkg/client/kubernetes/upgrades.go#L50)), so I assumed the empty strategy should've been ignored...
Did we notice this in a rollout upgrade test failure? Because IIRC that test was passing.

PR #1434, opened 2025-05-09
docs: update starter pack framework

nhennigan:
## Description

Updating the starter pack version of our docs. The starter pack is Canonical's framework for our documentation. Removing the old unnecessary files should make it easier to contribute to our docs. Keeping up with what the documentation team should also help with maintenance. 

## Solution

Key updates:

- conf.py and custom_conf.py are now just conf.py 
- Old files such as css and templates have been removed as they are available outside the repo
- The spelling check has been moved from our own vale implementation to aspell in the docs checks. We now follow the docs team recommended CI and will get updates automatically so hopefully less maintenance for us.  This also means our wordlist is now at docs/canonicalk8s/.custom_wordlist.txt
- Moving to the docs teams checks also means that our docs must build without errors in order for the spelling, link and inclusive language checks to be run. This will help enforce good habits in our docs.
- Spell check is changed from GB en to US en. (Company decision ) This made this PR look huge but majority of the file changes are one letter.
- Fixed build errors with formatting 
- Vale is now used for inclusive language check 

You shouldn't see a difference in the output of the docs. 

## Issue

N/A

## Backport

I don't think we need to backport. 

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed


.github/workflows/docs-checks.yml
```diff
@@ -9,41 +9,9 @@ permissions:
   contents: read
 
 jobs:
-  doc-checks:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
+  documentation-checks:
+    uses: canonical/documentation-workflows/.github/workflows/documentation-checks.yaml@main
```
addyess, line 13: Really all we have to work with here is `main`
image

docs/canonicalk8s/_parts/common_hardening.md
```diff
@@ -31,7 +31,7 @@ apiserver.
 ```
 sudo sh -c '
 cat >>/var/snap/k8s/common/args/kube-apiserver <<EOL
---encyrption-provider-config=/var/snap/k8s/common/etc/enc.yaml
+--encryption-provider-config=/var/snap/k8s/common/etc/enc.yaml
```
addyess, line 34: NICE!

addyess (approved):
Super huge improvements.  WTG!

.github/workflows/docs-checks.yml
nhennigan: Do you think I should remove main? Main should be stable but I doubt they'll change the default branch

bschimke95: Github will pick the default branch (=main) if you do not set a specific branch, so it doesn't really matter.

Issue #1414, opened 2025-05-07
Docs: CAPI / Provision cluster kubeconfig ignored by sudo k8s kubectl


gustavosr98:
This command is not using the KUBECONFIG pointed at
`KUBECONFIG=./${CLUSTER_NAME}-kubeconfig sudo k8s kubectl get node`

No the actual cluster-x
```bash
# ubuntu@ip-172-31-30-109:~/cluster-api-k8s/templates/aws$ KUBECONFIG=./cluster-x-kubeconfig sudo k8s kubectl get node
NAME               STATUS   ROLES                  AGE   VERSION
ip-172-31-30-109   Ready    control-plane,worker   16m   v1.32.3
```

With normal kubectl it works 
```bash
# ubuntu@ip-172-31-30-109:~/cluster-api-k8s/templates/aws$ KUBECONFIG=./cluster-x-kubeconfig kubectl get node
NAME                          STATUS   ROLES                  AGE     VERSION
ip-10-0-0-179.ec2.internal    Ready    worker                 4m42s   v1.32.3
ip-10-0-12-201.ec2.internal   Ready    control-plane,worker   6m54s   v1.32.3
ip-10-0-26-254.ec2.internal   Ready    control-plane,worker   4m35s   v1.32.3
ip-10-0-36-41.ec2.internal    Ready    control-plane,worker   2m53s   v1.32.3
ip-10-0-8-10.ec2.internal     Ready    worker                 4m47s   v1.32.3
```

---
Document: capi/howto/provision.md

HomayoonAlimohammadi:
Thanks for pointing this out @gustavosr98! I'll raise a PR to address this.

HomayoonAlimohammadi:
I reopened this issue to mark it as resolved, because this is the expected behavior. `sudo` strips the `KUBECONFIG` environment variable when used like that. There are two ways to prevent this from happening:

```bash
sudo -i
KUBECONFIG=... k8s kubectl get pods -A
```

OR

```bash
sudo KUBECONFIG=... k8s kubectl get pods -A
```

PR #1445, opened 2025-05-16
docs(CAPI): Use kubectl instead of k8s kubectl

HomayoonAlimohammadi:
### Overview
In order to communicate with the workload cluster, `kubectl` command should be used instead of `k8s kubectl`, since the latter will get executed with a custom `KUBECONFIG`.

Fixes: https://github.com/canonical/k8s-snap/issues/1414

### DIFF
--- a/docs/canonicalk8s/capi/howto/provision.md
+++ b/docs/canonicalk8s/capi/howto/provision.md
@@ -86,7 +86,13 @@ clusterctl get kubeconfig ${CLUSTER_NAME} > ./${CLUSTER_NAME}-kubeconfig
 You can then see the workload nodes using:
 
 ```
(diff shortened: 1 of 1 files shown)


bschimke95:
Wouldn't it be better if `k8s kubectl` just take into account the `KUBECONFIG` variable here? UX-wise this means:
Connect to the ck8s cluster by default, if KUBECONFIG is set behave as normal kubectl

WDYT?

HomayoonAlimohammadi:
@bschimke95 good point. Lemme try to change it.

nhennigan:
Hi @HomayoonAlimohammadi just confirming there is nothing from a docs perspective to do here? You are working on the implementation of Ben's suggestion?

HomayoonAlimohammadi:
@nhennigan That's right. I have a card to work on this in the backlog. If any changes are required on the docs side as well, I'll let you know.

PR #1449, opened 2025-05-16
fix: update node ip addresses on k8sd post-refresh

### LINKED ISSUE #1448: v1.32.4 Cilium crashes; missing IPv6 address from apiserver on a host with working IPv6
### Summary

Yesterday evening my k8s snap updated from 1.32.3 to 1.32.4. Ever since the update my cilium pod has been in a crash loop.

```
msg="failed to start: daemon creation failed: unable to connect to get node spec from apiserver: node kube does not have an IPv6 address\nfailed to stop: unable to find controller ipcache-inject-labels" subsys=daemon
```

My node most definitely has an IPv6 address :)

```
root@kube:~# ip -6 a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 state UNKNOWN qlen 1000
    inet6 ::1/128 scope host noprefixroute 
       valid_lft forever preferred_lft forever
2: enp5s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
    inet6 2a02:db8:3cc4:3:216:3eff:fed1:7068/64 scope global dynamic mngtmpaddr noprefixroute 
       valid_lft 86145sec preferred_lft 14145sec
    inet6 fe80::216:3eff:fed1:7068/64 scope link 
       valid_lft forever preferred_lft forever
```

**I just restored a snapshot of my LXD VM to yesterday before the update to 1.32.4 and it simply works 👍**. So i'll stick to 1.32.3 for now.

My node has an IPv6 address so it should simply get that data from the apiserver and be able to start cilium.

### Reproduction Steps

No matter how often I restart my apiserver or my whole node, it keeps saying my node does not have a valid IPv6 address.

### System information

[inspection-report-20250516_104214.tar.gz](https://github.com/user-attachments/files/20240917/inspection-report-20250516_104214.tar.gz)

### Can you suggest a fix?

_No response_

_No response_

berkayoz:
## Description

Updates the `--node-ip` flag of kubelet on k8sd post-refresh to include both ipv4 and ipv6 addresses which became a Cilium requirement with 1.17.

## Issue

Fixes #1448

## Backport

* `release-1.33`
* `release-1.32`

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/docs/canonicalk8s/snap/reference/upgrading.md
+++ b/docs/canonicalk8s/snap/reference/upgrading.md
@@ -10,7 +10,6 @@ sudo snap refresh k8s --channel=1.33/stable
 
 All components will be updated automatically.
 
-<!-- TODO(Hue): To be removed after Cilium 1.17 is backported to 1.32 -->
 ### Additional steps for dual-stack environments
 
 If your cluster is configured with dual stack networking (IPv4 and IPv6), 


src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -110,3 +119,48 @@ func (a *App) performPostUpgrade(ctx context.Context, s state.State) error {
 
 	return nil
 }
+
+func (a *App) updateNodeIPAddresses(ctx context.Context, s state.State) error {
```
addyess, line 123: is it REALLY a good idea to expose the kubelet on EVERY nic on the system?

claudiubelu: It probably isn't. In a cloud scenario, where you'd have multiple network planes, Kubernetes internals is the sort of thing you'd limit to only management or internal network planes. Is there a way to specify the NICs to use for this in `k8s-snap`?

Ihis looks similar to the workflow in hooks_join / hooks_bootstrap, so, I guess @addyess question would also apply there. In any case, could we create an util function for this and reusing it in all 3 cases?

berkayoz: I believe we do not expose/use every nic/interface. The cluster member derives the address(and through this the interface) from the `--address` flag provided to bootstrap or join commands.

The IP list here is extracted from the interface that the member IP address belongs to. The steps are basically the same but I'm not sure if we can implement a common util function across all 3 cases. (Most of these changes below are common steps in join/bootstrap hooks and stays at the upper level function, and is passed down the the `setup_kubelet`.)

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -110,3 +119,48 @@ func (a *App) performPostUpgrade(ctx context.Context, s state.State) error {
...
+	snap := a.Snap()
+	nodeIP := net.ParseIP(s.Address().Hostname())
+	if nodeIP == nil {
+		return fmt.Errorf("failed to parse node IP address %s", s.Address().Hostname())
+	}
+
+	// nodeIPs will be passed to kubelet as the --node-ip parameter, allowing it to have multiple node IPs,
+	// including IPv4 and IPv6 addresses for dualstacks.
+	nodeIPs, err := utils.GetIPv46Addresses(nodeIP)
```
addyess, line 132: so @berkayoz are you saying this list of addresses is scoped somehow to microcluster

louiseschmidtgen, line 132:
```suggestion
	nodeIPs, err := utils.GetIPv4IPv6Addresses(nodeIP)
```

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
bschimke95, line 123: I don't really like that this logic lands here as it basically duplicates the logic from the [node_controller](https://github.com/canonical/k8s-snap/blob/5ce8f868784892d259063e103ce551e20429eda8/src/k8s/pkg/k8sd/controllers/node_configuration.go#L71).
I understand why you didn't put it there (this controller operates on the cluster-wide config), though.

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
bschimke95: Looking at the [implementation](https://vscode.dev/github/canonical/k8s-snap/blob/autoupdate/sync/release-1.33/src/k8s/pkg/utils/net.go#L78-L79), we only select the interfaces that contain the IP of the microcluster node which we derive [here](https://github.com/canonical/k8s-snap/pull/1449/files#diff-0b16941c4a34c9e8adfffd9dc32b8a9aa5bfe2308b550dca810255994a5f1f1eR125)

louiseschmidtgen (commented):
Nice work Berkay! How is this tested?

berkayoz:
This should be updated/re-visited after arguments override topic is resolved

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
HomayoonAlimohammadi: It's possible to do this as well. We can pass the internal MC state to the node configuration controller [here](https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/app/hooks_start.go#L36-L47). Do you think that's a better place to put this logic?

HomayoonAlimohammadi: We probably can't put this logic in the node config controller as we want the kubelet to be restarted (with the correct node ips) before the cilium pod is created on that node.
One can argue that the node config controller can also restart the cilium pods if necessary but there are two issues to this I think:
- It makes node config controller a bit too complicated
- We're risking a bried period of Cilium pods being in error state until they get restarted. I'm not sure if it's worth it to take this risk.

bschimke95: That is a fair point. Thanks for investigating it as an option.

PR #1453, opened 2025-05-19
WIP: Centralize controller manager by adding Coordinator

HomayoonAlimohammadi:
### Overview

This PR enables us to use a single `manager.Manager` for both the "upgrade" and "csrSigning" controllers (+ any future controllers).
A `Coordinator` is added which is responsible for creating a single `manager` instance, and registering all the controllers with it.


src/k8s/pkg/k8sd/controllers/coordinator.go
```diff
@@ -0,0 +1,211 @@
...
+	scheme, err := kubernetes.NewScheme()
+	if err != nil {
+		return fmt.Errorf("failed to create scheme: %w", err)
+	}
+
+	mgr, err := manager.New(config, manager.Options{
+		Scheme:                  scheme,
+		Logger:                  logger,
+		LeaderElection:          true,
+		LeaderElectionID:        "oy6981cu.controller-coordinator",
```
HomayoonAlimohammadi, line 89: Something that concerns me is whether or not this is going to cause problems on upgrades. Because for each controller, we're gonna have managers with different leader election IDs. So I assume there would be two running instances of each controller (csrsigning and upgrade) during and actual snap upgrade.

HomayoonAlimohammadi: The feature upgrade test pass. So I assume this shouldn't be a concern. I'm digging deeper tho.

bschimke95 (approved):
Generally LGTM
The feature upgrade tests don't really test if all controllers are functional after a refresh.
Could you do a manual upgrade from e.g. 1.33/stable and then check if the controllers (CSR and upgrade) work as expected?

HomayoonAlimohammadi:
@bschimke95 So I created a 3 CP 1 worker node cluster on 1.32/stable, upgraded the cluster to the snap built from this branch, the upgrade CR was created and successfully transitioned to `Completed` state.
After that, I refreshed cluster certs (on both CP and worker nodes) following [this guide](https://documentation.ubuntu.com/canonical-kubernetes/release-1.32/snap/howto/refresh-certs/). The CSRs get approved as expected and the `csrSigning` controller shows healthy logs.

PR #1454, opened 2025-05-19
feat: bump snap base to core22

berkayoz:
## Description

Added `enable-patchelf` to build attributes to preserve old behavior since it is not enabled by default from core22+.
Updated references of core20 to core22.

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/docs/canonicalk8s/snap/howto/install/offline.md
+++ b/docs/canonicalk8s/snap/howto/install/offline.md
@@ -16,12 +16,12 @@ workloads and {{product}} features.
 ### Download the {{product}} snap
 
 From a machine with access to the internet download the
-`k8s` and `core20` snap with:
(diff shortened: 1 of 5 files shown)


addyess (approved):
thanks so much @berkayoz 

what's `patchelf` 🧝🏻 ?

berkayoz:
> what's patchelf 🧝🏻 ?

Well I'm not exactly sure on the implementation details but `patchelf` is a tool that patches the elf header/section to update dynamically linked libraries so that their paths point to files/libraries from inside the snap (or the base `core20`) instead of the host system.

IIRC this is only necessary for classic snaps since strictly confined snaps will have a "virtual" view of the filesystem with custom mounts.

PR #1462, opened 2025-05-21
test: Run all tests in parallel on Github runners

bschimke95:
This PR does two things:
1. Spin up a job for each test which massively reduces our CI/PR testing time
2. Move to official Github runners instead of self-hosted runners

Notes:
* For normal PRs, we don't run the nightly tests so we *only* spin up ~15 jobs
* We could think about not doing this for our nightly runs since iteration speed is not critical in this scenario.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -67,9 +67,14 @@ on:
         default: false
 
 jobs:
-  test-integration:
-    name: Tests
-    runs-on: ${{ inputs.arch == 'arm64' && 'self-hosted-linux-arm64-noble-large' || 'self-hosted-linux-amd64-noble-large' }}
+  # First job: Prepare the snap and find test files
+  prepare:
+    name: Prepare Environment
+    runs-on: ${{ inputs.arch == 'arm64' && 'ubuntu-latest-arm64' || 'ubuntu-latest' }}
+    outputs:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -67,9 +67,52 @@ on:
...
+          echo "tests=$JSON_ARRAY" >> $GITHUB_OUTPUT
+          echo "Found test files: $TEST_FILES"
+
+  # Second job: Run each test in a separate runner
+  run-tests:
+    name: Run ${{ matrix.test }} - ${{ inputs.os }}-${{ inputs.arch }}
+    needs: prepare
+    runs-on: ${{ inputs.arch == 'arm64' && 'ubuntu-latest-arm64' || 'ubuntu-latest' }}
+    strategy:
+      fail-fast: false
```
addyess, line 113:
```suggestion
    strategy:
      max-parallel: 10    
      fail-fast: false
```

this will limit the number of runners consumed to 10, so the jobs don't swamp and eat all the runners at once

bschimke95: as discussed offline: We use Github runners not self-hosted ones so we should not need to limit this.

berkayoz (approved):
Great work, LGTM overall! As a next step could we have a flag to disable subunit reports completely which we can set.

PR #1492, opened 2025-06-03
chore: Reduce API server calls

HomayoonAlimohammadi:
### Overview
This PR aims to reduce the API server calls that are made by k8sd. Each commit corresponds to a specific improvement.
It needs to be backported to 1.32 and 1.33, but automatic backports will probably fail since the controller implementations have changed.


src/k8s/pkg/client/helm/client.go
```diff
@@ -86,28 +97,46 @@ func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, v
...
 		}
 		return true, nil
 	case isInstalled && desired != StateDeleted:
+		chart, err := loader.Load(filepath.Join(h.manifestsBaseDir, c.ManifestPath))
+		if err != nil {
+			return false, fmt.Errorf("failed to load manifest for %s: %w", c.Name, err)
+		}
+
+		sameValues := jsonEqual(oldConfig, values)
+		sameVersions := release.Chart.Metadata.Version == chart.Metadata.Version
```
berkayoz, line 124: We should probably note that for charts managed by us(for load balancer etc.) , we'll need to make sure to actually bump the version otherwise they'll be not applied.

src/k8s/pkg/client/helm/client.go
```diff
@@ -86,28 +97,46 @@ func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, v
...
 		if err != nil {
-			return false, fmt.Errorf("failed to load manifest for %s: %w", c.Name, err)
+			return false, fmt.Errorf("failed to upgrade %s: %w", c.Name, err)
 		}
 
-		release, err := upgrade.RunWithContext(ctx, c.Name, chart, values)
-		if err != nil {
-			return false, fmt.Errorf("failed to upgrade %s: %w", c.Name, err)
+		if !jsonEqual(values, release.Config) {
+			log.Info("provided values does not match release.Config after upgrade, this is unexpected", "values", values, "newValues", release.Config)
```
berkayoz: Do we really need this? Can we recover / address this at all?

Unless the docs for `release.Config` are wrong, this variable should only contain "extra values" passed by us. I'm not sure if we should verify helm behavior here unless we can address this somehow.

src/k8s/pkg/client/helm/client.go
```diff
@@ -86,28 +97,46 @@ func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, v
...
 		}
 
 		// oldConfig and release.Config are the previous and current values. they are compared by checking their respective JSON, as that is good enough for our needs of comparing unstructured map[string]any data.
-		return !jsonEqual(oldConfig, release.Config), nil
+		return sameValues, nil
```
berkayoz: I believe this would only reflect something has changed  only if values are different, shouldn't we address for different chart versions and check for `sameVersions` as well? 

I think this is just `true` regardless since we eliminate the no changes above?

src/k8s/pkg/k8sd/controllers/upgrade/controller.go
```diff
@@ -58,5 +60,11 @@ func NewController(
 func (c *Controller) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&upgradesv1alpha1.Upgrade{}).
+		WithOptions(controller.Options{
+			RateLimiter: workqueue.NewTypedItemExponentialFailureRateLimiter[ctrl.Request](
+				time.Second,
+				5*time.Minute,
+			),
+		}).
```
berkayoz: Can we add a comment on why this is required/done? Especially since we're diverging from controller-runtime defaults.

src/k8s/pkg/k8sd/controllers/upgrade/reconcile.go
```diff
@@ -170,3 +171,30 @@ func (c *Controller) transitionTo(ctx context.Context, upgrade *upgradesv1alpha.
...
+		c.notifyFeatureController(false, false, false, true, false, false, false)
+	case features.LocalStorage:
+		c.notifyFeatureController(false, false, false, false, true, false, false)
+	case features.MetricsServer:
+		c.notifyFeatureController(false, false, false, false, false, true, false)
+	case features.DNS:
+		c.notifyFeatureController(false, false, false, false, false, false, true)
+	default:
+		return fmt.Errorf("unknown feature %q", name)
+	}
```
berkayoz: Can we add a new function to notify features individually?

src/k8s/pkg/k8sd/controllers/upgrade/reconcile.go
```diff
@@ -146,8 +142,13 @@ func (c *Controller) allNodesUpgraded(ctx context.Context, upgradedNodes []strin
 }
 
 func (c *Controller) waitForFeatureReconciliations(ctx context.Context, log logr.Logger) error {
-	timeout := time.After(c.featureControllerReconcileTimeout)
 	for name, ch := range c.featureToReconciledCh {
+		if err := c.triggerFeature(name); err != nil {
```
berkayoz, line 146: Can we add a comment on how this order is determined ( through the map order [here](https://github.com/canonical/k8s-snap/blob/2914048a216e29b8e826189dc88bea5d873f7053/src/k8s/pkg/k8sd/app/app.go#L192-L200)

src/k8s/pkg/snap/snap.go
```diff
@@ -368,6 +369,7 @@ func (s *snap) HelmClient() helm.Client {
 		func(namespace string) genericclioptions.RESTClientGetter {
 			return s.restClientGetter(filepath.Join(s.KubernetesConfigDir(), "admin.conf"), namespace)
 		},
+		1*time.Minute,
```
berkayoz: I feel like this is a bit low, 5 minutes might be a safer choice wdyt?

src/k8s/pkg/client/helm/client.go
```diff
@@ -86,28 +97,46 @@ func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, v
...
 			return false, fmt.Errorf("failed to install %s: %w", c.Name, err)
 		}
 		return true, nil
 	case isInstalled && desired != StateDeleted:
+		chart, err := loader.Load(filepath.Join(h.manifestsBaseDir, c.ManifestPath))
+		if err != nil {
+			return false, fmt.Errorf("failed to load manifest for %s: %w", c.Name, err)
+		}
+
+		sameValues := jsonEqual(oldConfig, values)
```
berkayoz, line 120: This check will equal to false if a user/admin sets a value that we don't manage through `helm upgrade <release> <chart> --reuse-values --set ...`. Which will fallback to the old behavior of reconciling regardless of changes, can we leave a comment here to indicate this as a start?

I suspect we can also address this case by merging `oldConfig` and our `values`, with `values` taking priority and do a `jsonEqual(oldConfig, mergedValues)` so that we can pickup changes outside our control. (Helm does something similar, maybe there is a helper function)?

src/k8s/pkg/client/helm/client.go
```diff
@@ -46,7 +53,11 @@ func (h *client) newActionConfiguration(ctx context.Context, namespace string) (
 
 // Apply implements the Client interface.
 func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, values map[string]any) (bool, error) {
-	cfg, err := h.newActionConfiguration(ctx, c.Namespace)
+	log := log.FromContext(ctx).WithName("helm").WithValues("chart", c.Name, "desired", desired)
+	applyCtx, cancel := context.WithTimeout(ctx, h.applyTimeout)
```
berkayoz, line 64: I'm not sure how I feel about adding our own timeout here. Do we know if there is a timeout in the kubernetes rest client we use here? I think we can keep this but be more generous with the time.

src/k8s/pkg/k8sd/controllers/upgrade/reconcile.go
HomayoonAlimohammadi: Great suggestion.

src/k8s/pkg/k8sd/controllers/upgrade/reconcile.go
HomayoonAlimohammadi: I think map iteration order is non-deterministic in Go. We can't rely on that.
https://go.dev/play/p/u61QagjK7hf

src/k8s/pkg/client/helm/client.go
HomayoonAlimohammadi: I think we have this issue in our current implementation as well, right? I'll add a note and create a card to address this later.

HomayoonAlimohammadi: `KU-3592`.

src/k8s/pkg/client/helm/client.go
HomayoonAlimohammadi: I increased it to 5 minutes as per your other comment.

HomayoonAlimohammadi: The `ctx` is basically the context that Microcluster is using. It roots back to [here](https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/app/hooks_start.go#L19). Does that answer your question?

src/k8s/pkg/client/helm/client.go
HomayoonAlimohammadi: I added a comment for it. Do you think that is enough for now?

src/k8s/pkg/client/helm/client.go
HomayoonAlimohammadi: Removed.

src/k8s/pkg/client/helm/client.go
HomayoonAlimohammadi: You're right. Returning `true` now.

ktsakalozos (approved):
LGTM +1

berkayoz (approved):
LGTM overall, we can continue iterating later on

PR #1510, opened 2025-06-06
fix: link dqlite against the sqlite we build

ktsakalozos:
## Description

Although we are building sqlite we are not linking against it. This came up with the recent dqlite. Before this patch if you tried to build the snap using dqlite from "master" you would fail with the `configure` step of dqlite saying it the sqlite versions do not match. Even though we are building 3.45 sqlite we were linking against the one from the base which was 3.31. Recent dqlite versions from master required sqlite 3.45+ and so this issue was exposed.

## Solution

Tell dqlite's `confgure` where to search for the sqlite libs

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/src/k8s/hack/dynamic-dqlite.sh
+++ b/src/k8s/hack/dynamic-dqlite.sh
@@ -100,9 +100,9 @@ if [ ! -f "${BUILD_DIR}/dqlite/libdqlite.la" ]; then
       UV_CFLAGS="-I${BUILD_DIR}/libuv/include" \
       UV_LIBS="-L${BUILD_DIR}/libuv/.libs" \
       SQLITE_CFLAGS="-I${BUILD_DIR}/sqlite" \
+      SQLITE_LIBS="-L${BUILD_DIR}/sqlite/.lib" \
       LZ4_CFLAGS="-I${BUILD_DIR}/lz4/lib" \
       LZ4_LIBS="-L${BUILD_DIR}/lz4/lib" \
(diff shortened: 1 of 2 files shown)


src/k8s/hack/dynamic-dqlite.sh
```diff
@@ -102,6 +102,7 @@ if [ ! -f "${BUILD_DIR}/dqlite/libdqlite.la" ]; then
       SQLITE_CFLAGS="-I${BUILD_DIR}/sqlite" \
       LZ4_CFLAGS="-I${BUILD_DIR}/lz4/lib" \
       LZ4_LIBS="-L${BUILD_DIR}/lz4/lib" \
+      SQLITE_LIBS="-L${BUILD_DIR}/sqlite/.lib" \
```
bschimke95: nit: let's keep the lib lines together.
Move this to line 103 (also below).

bschimke95 (commented):
Minor nit, LGTM otherwise.

src/k8s/hack/static-dqlite.sh
```diff
@@ -147,6 +148,7 @@ fi
   cp lz4/lib/*.so* "${INSTALL_DIR}/lib"
   cp sqlite/.libs/*.a "${INSTALL_DIR}/lib"
   cp dqlite/.libs/*.a "${INSTALL_DIR}/lib"
+
```
bschimke95: remove

PR #1496, opened 2025-06-04
DONOTMERGE: feat: FIPS-compliant 

bschimke95:
Note: This PR is **only** meant for reviewing and will not be merged. This PR tests the FIPS-mode and compliance with Kubernetes 1.32 and Go 1.23. 
We will not backport this to 1.32 and only perform this test with 1.32 because the Go 1.24-fips track is not yet available.
This PR does not yet make the Canonical Kubernetes platform fully-FIPS compliant since the features (Cilium etc.) are not yet using the FIPS-compliant Go toolchain. This will be addressed in follow-up work.

See specification KU-111 for more context. The UX improvements that are described there will also be done in a follow-up PR.

This PR introduces several major changes:
1. Use the `go-fips` channel to use the FIPS-enabled [Go toolchain from Microsoft](https://github.com/microsoft/go/blob/microsoft/release-branch.go1.23/eng/doc/fips/README.md) 
2. Modify the build scripts to:
   a) set required environment variables (see spec for more context)
   b) switch to dynamic linking for components that require cryptographic functionality
3. Added `test_fips.py` to verify that all components are built with FIPS support. The test checks that crypto-related components are dynamically linked and fail when FIPS mode is enabled on non-compliant systems. This is an initial test to verify basic functionality. The proper way of testing would be to run the current test suite against a FIPS-enabled snap on a FIPS-enabled machine.
I'm currently discussing with IS ways to get a FIPS-enabled machine on self-hosted runners.
This will be implemented in a **follow-up** PR

<details>
<summary>
For now, I checked manually that the setup generally works on a FIPS machine as expected (click to see steps):
</summary>

1. Installing on a FIPS-machine without core22-fips fails. 
(The error message will be improved as part of KU-3495.) 
1. Verify that this is a FIPS-compliant machine/kernel
```
ubuntu@my-fips-machine:/k8s-snap$ cat /proc/sys/crypto/fips_enabled
1
```
2. Start with a non-FIPS core22
```
ubuntu@my-fips-machine:/k8s-snap$ sudo snap list core22
Name    Version   Rev   Tracking       Publisher   Notes
core22  20250425  1981  latest/stable  canonical✓  base
```
3. Installing the k8s-snap will fail now (since k8sd will fail to start in the postInstall hook)
```
ubuntu@my-fips-machine:/k8s-snap$ sudo snap install ./k8s.snap --dangerous --classic
error: cannot perform the following tasks:
- Run install hook of "k8s" snap if present (run hook "install": 
-----
[...]

PR #1513, opened 2025-06-06
feat: Add FIPS Go toolchain

bschimke95:
* Add necessary configuration and env variables to the build process to support the Go FIPS toolchain
* Build binaries that require crypto dynamically, so that they can reach out to the FIPS-compliant host OpenSSL implementation
* Add a basic test that verifies that the binaries are built correctly

This commit only prepares the build process for FIPS but not actually enables it since Kubernetes 1.33+ requires Go 1.24 which right now does not support a FIPS version.
The `GOEXPERIMENT=opensslcrypto` is only available in the modified FIPS Go toolchain, hence it is commented out for now.

This setup is tested with 1.32 and FIPS Go 1.23 here: https://github.com/canonical/k8s-snap/pull/1496

### DIFF
--- a/tests/integration/tests/test_build.py
+++ b/tests/integration/tests/test_build.py
@@ -13,7 +13,7 @@ LOG = logging.getLogger(__name__)
 @pytest.mark.node_count(1)
 @pytest.mark.disable_k8s_bootstrapping()
 @pytest.mark.tags(tags.NIGHTLY)
-def test_fips(instances: List[harness.Instance]):
+def test_build(instances: List[harness.Instance]):
     """
     Test that all snap components that contain crypto functions are built dynamically.
     """


bschimke95:
The `test_disable_separate_feature_controller` failure is unrelated and also fails in a lot of other places. I investigate/fix this in #1529

build-scripts/components/containerd/build.sh
```diff
@@ -9,8 +9,24 @@ REVISION=$(git rev-parse HEAD)
 sed -i "s,^VERSION.*$,VERSION=${VERSION}," Makefile
 sed -i "s,^REVISION.*$,REVISION=${REVISION}," Makefile
 
-export STATIC=1
-for bin in ctr containerd containerd-shim containerd-shim-runc-v1 containerd-shim-runc-v2; do
+export GOTOOLCHAIN=local
+# # export GOEXPERIMENT=opensslcrypto
```
HomayoonAlimohammadi, line 13: Is this a note?

HomayoonAlimohammadi: I think I got my answer in your other PR.

k8s/wrappers/services/k8sd
```diff
@@ -5,4 +5,16 @@
...
+if k8s::common::is_fips_enabled; then
+  # The Microsoft/go toolchain 1.23 TLS implementation contains a bug
+  # that causes TLS connections to fail when FIPS is enabled.
+  # This is fixed in 1.24 but won't be backported to 1.23.
+  # See https://github.com/microsoft/go/issues/1626
+  # and https://github.com/golang-fips/openssl/pull/272
+  #
+  # As a workaround for 1.23, we set the minTLSVersion to 1.2.
+  # See https://github.com/canonical/lxd/blob/9f30d5799ec0dc45364b95d9e609ab09508d337e/shared/network.go#L65
+  export LXD_INSECURE_TLS=1
```
berkayoz, line 17: We should make sure to not forget this, once we have go 1.24

snap/snapcraft.yaml
```diff
@@ -70,50 +71,59 @@ parts:
...
+
 
   k8s-binaries:
     after: [dqlite]
     source: src/k8s
     build-attributes: [enable-patchelf]
     plugin: nil
     override-build: |
-      INSTALL="${SNAPCRAFT_PART_INSTALL}"
+      INSTALL="${CRAFT_PART_INSTALL}"
```
berkayoz, line 83: Unrelated to the changes here but I feel like we should move this to our `build-components` tooling like the rest in a separate PR/card.

tests/integration/tests/test_fips.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+from test_util import harness, tags
+
+LOG = logging.getLogger(__name__)
+
+
+@pytest.mark.node_count(1)
+@pytest.mark.disable_k8s_bootstrapping()
+@pytest.mark.tags(tags.NIGHTLY)
+def test_fips(instances: List[harness.Instance]):
```
berkayoz: nit: Does this actually test fips, or do we run a sanity check to make sure certain components are dynamically built? Maybe a different name is more fitting?

berkayoz (approved):
LGTM overall, I'm a bit cautious about the `patchelf` from snapcraft so we should make sure we don't have related linting errors.

build-scripts/components/containerd/build.sh
bschimke95: Yes, from the PR description: "The GOEXPERIMENT=opensslcrypto is only available in the modified FIPS Go toolchain, hence it is commented out for now."

This Go experiment is only available in the modified toolchain which is why we cannot set this variable here. I left it here for reference.

k8s/wrappers/services/k8sd
bschimke95: I created KU-3667 as follow-up

snap/snapcraft.yaml
bschimke95: I added KU-3668 to investigate this.
I'm not 100% sure if this fits in that tooling since the current tooling requires a repo and version file but it is worth investigating.

tests/integration/tests/test_fips.py
bschimke95: Fair, it started as a FIPS test but eventually evolved into something else :)

PR #1514, opened 2025-06-06
feat: Introduce snap-env to manage snap-wide environments

bschimke95:
* The user can add a `snap-env` file to $SNAP_COMMON/args to modify the environment for all services in the snap
* `<service>-env` files still take precedence
* We do not fail in the snap postInstall hook anymore as this prevents the installation (and possible config fix) of the snap altogether.

### DIFF
--- a/k8s/lib.sh
+++ b/k8s/lib.sh
@@ -155,21 +155,15 @@ k8s::common::execute() {
 
   # Load global environment variables if the file exists
   if [[ -f "${SNAP_COMMON}/args/snap-env" ]]; then
-    while IFS= read -r line || [[ -n "$line" ]]; do
-      # Skip empty lines and comments
-      [[ -z "${line// }" || "$line" =~ ^# ]] && continue
-      all_env_vars+=("$line")
-    done < "${SNAP_COMMON}/args/snap-env"
+    mapfile -t global_vars < <(grep -vE '^[[:space:]]*($|#)' "${SNAP_COMMON}/args/snap-env")
+    all_env_vars+=("${global_vars[@]}")
   fi
 
   # Load service-specific environment variables if the file exists
   # Service-specific environment variables take precedence over global environment variables.
   if [[ -f "${SNAP_COMMON}/args/${service_name}-env" ]]; then
-    while IFS= read -r line || [[ -n "$line" ]]; do
-      # Skip empty lines and comments
-      [[ -z "${line// }" || "$line" =~ ^# ]] && continue
-      all_env_vars+=("$line")
-    done < "${SNAP_COMMON}/args/${service_name}-env"
+    mapfile -t service_vars < <(grep -vE '^[[:space:]]*($|#)' "${SNAP_COMMON}/args/${service_name}-env")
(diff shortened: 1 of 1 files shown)


k8s/lib.sh
```diff
@@ -127,22 +127,46 @@ k8s::util::wait_kube_apiserver() {
...
+  declare -a all_env_vars=()
+
+  # Load global environment variables if the file exists
+  if [[ -f "${SNAP_COMMON}/args/snap-env" ]]; then
+    while IFS= read -r line || [[ -n "$line" ]]; do
+      # Skip empty lines and comments
+      [[ -z "${line// }" || "$line" =~ ^# ]] && continue
+      all_env_vars+=("$line")
+    done < "${SNAP_COMMON}/args/snap-env"
+  fi
```
HomayoonAlimohammadi: This is some dark bash magic

k8s/lib.sh
```diff
@@ -127,22 +127,46 @@ k8s::util::wait_kube_apiserver() {
...
+
+  # Load service-specific environment variables if the file exists
+  # Service-specific environment variables take precedence over global environment variables.
+  if [[ -f "${SNAP_COMMON}/args/${service_name}-env" ]]; then
+    while IFS= read -r line || [[ -n "$line" ]]; do
+      # Skip empty lines and comments
+      [[ -z "${line// }" || "$line" =~ ^# ]] && continue
```
HomayoonAlimohammadi: This `#` here isn't counted as the beginning of a comment, right?

snap/hooks/install
```diff
@@ -6,9 +6,9 @@ k8s::common::setup_env
...
 # disable snap set/get on startup as k8sd is not up yet.
 # the sync will be enabled in the bootstrap hook.
-k8s::cmd::k8s x-snapd-config disable
+k8s::cmd::k8s x-snapd-config disable || true
 
 # k8s has a REST interface to initialize a cluster.
 # In order to interact with the REST API the k8sd service
 # needs to be started and configured in the installation step.
-k8s::init::k8sd
+k8s::init::k8sd || true
```
berkayoz, line 14: Why do we need ` || true` here and above now? Is it to be on the safe side?

k8s/lib.sh
berkayoz: What's the advantage of this over the previous approach? Couldn't we use that 2 times and merge it somehow?

snap/hooks/install
bschimke95: See PR description: "We do not fail in the snap postInstall hook anymore as this prevents the installation (and possible config fix) of the snap altogether."

k8s/lib.sh
bschimke95: Exactly this is regex that matches a comment (start with #)

k8s/lib.sh
bschimke95: You also end up with a bit of grep magic to filter out comments etc. but I guess it still looks cleaner with `mapfile`. Updated the code.

PR #1515, opened 2025-06-07
fix: address test_loadbalancer flakiness

ktsakalozos:
## Description

The test_loadbalancer tests are flaky. This PR addresses some of the flakiness issues.

## Solution

The following changes/fixes are introduced:

1. A pytest.ini is introduced so we see the LOG messages in the tests
2. The IP v6 test is set to run as part of this test because we have now transitioned main to Cilium 1.17
3. We enable the network and wait for it to some up before we start the loadbalancer (metallb). We wait for metallb to come up and then we configure it. The way we had it before there was a race condition that resulted in an error like this:
  ```
load-balancer:            Failed to deploy MetalLB, the error was: failed to enable LoadBalancer: failed to apply MetalLB LoadBalancer configuration: failed to upgrade metallb-loadbalancer: failed to create resource: Internal error occurred: failed calling webhook "ipaddresspoolvalidationwebhook.metallb.io": failed to call webhook: Post "https://metallb-webhook-service.metallb-system.svc:443/validate-metallb-io-v1beta1-ipaddresspool?timeout=10s": dial tcp 10.200.73.136:443: connect: connection refused
  ```
This error probably indicates that we were trying to configure the LB before it becomes available.
4. If the IP the LB gives to the service is an IPv6 we need to enclose it in `[]` or else curl gets confused about its format.
5. Increased some of the API query intervals and timeouts.

## Backport

A version of this PR needs to be backported to 1.33 and 1.32 to address flakiness there too.

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/tests/integration/pytest.ini
+++ b/tests/integration/pytest.ini
@@ -1,0 +1,5 @@
+[pytest]
+log_cli = 1
+log_cli_level = INFO
+log_cli_format = %(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)
+log_cli_date_format=%Y-%m-%d %H:%M:%S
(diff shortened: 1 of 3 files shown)


HomayoonAlimohammadi (approved):
Thanks a lot @ktsakalozos! LGTM

PR #1528, opened 2025-06-10
feat: Gracefully fail if user tries to run FIPS on non-compliant host.

bschimke95:
Introduce a helpful error message instead of panicking if the user tries to run FIPS on a non-compliant machine.

* Introduced `k8s::common::validate_env()` to validate environment variables, ensuring FIPS mode is only enabled on FIPS-capable hosts. This function also temporarily prefixes variables with `K8SD_` for isolated validation.
* Updated `k8s::common::execute_service()` to validate environment variables using the new `k8s::common::validate_env()` function before running services.
* Added a new integration test `test_user_config()` that verifies that configuration provided by the user is actually used, including:
    * `snap-env` - snap-wide configuration, see #1514 
    * `<service-env> - service specific config (this exists for a while in the code already - now we actually test it.
    * Directly passed environment variables

Manual test:
```
ubuntu@my-k8s-vm:/k8s-snap$ sudo GOFIPS=1 k8s status
FIPS mode is requested (GOFIPS=1 or GODEBUG contains fips140=on) but the host system is not FIPS-enabled.
Please run this service on a FIPS-enabled host to use FIPS mode.
```

Note: The `GOFIPS` variable is used for the integration test because it conveniently works for all services and exits a binary early. This requires #1513 to be merged first.


tests/integration/tests/test_util/util.py
```diff
@@ -945,3 +945,15 @@ def check_snap_services_ready(
...
+
+def host_is_fips_enabled():
+    """
+    Returns True if the host is running with FIPS enabled, False otherwise.
+    """
+    fips_path = "/proc/sys/crypto/fips_enabled"
+    try:
+        with open(fips_path) as f:
+            return f.read().strip() == "1"
+    except (FileNotFoundError, PermissionError):
```
HomayoonAlimohammadi, line 958: I'm a bit curious about the exception of `PermissionError`.

HomayoonAlimohammadi (approved):
I really like the test. Thanks @bschimke95. LGTM! Just a minor question.

tests/integration/tests/test_util/util.py
bschimke95: I guess this is just theoretically/for completeness since this file is readable by everyone:
```
ubuntu@my-fips-machine:~$ ll /proc/sys/crypto/fips_enabled 
-r--r--r-- 1 root root 0 Jun 12 13:24 /proc/sys/crypto/fips_enabled
```
but in theory `f.read` can throw this exception so I added it here.

PR #1529, opened 2025-06-10
test: Fix annotation test

bschimke95:
* We join a node to the cluster and immediately trigger a snap refresh.
At this point, the k8sd cluster reflects both nodes, but the k8s-dqlite cluster appears to be out of sync:
* On [node 1](https://github.com/canonical/k8s-snap/actions/runs/15565098315/job/43839259681? pr=1529#step:10:350), the newly joined node doesn't appear yet.
* On [node 2](https://github.com/canonical/k8s-snap/actions/runs/15565098315/job/43839259681?pr=1529#step:10:355), it is already registered.
* Then, k8s-dqlite is stopped due to the snap refresh, and this inconsistency seems to be persisted (visible in the [inspection reports](https://github.com/canonical/k8s-snap/actions/runs/15492797421/artifacts/3294837888) from the failed runs).
* After the refresh, k8s-dqlite restarts but fails to recover correctly, likely due to the inconsistent cluster state.

As an immediate fix, we query the cluster state on the joining node, which ensures that the node is already registered to the cluster (read your own writes). 
I created KU-3664 to work on a long-term k8s-dqlite fix of this consistency issue.

### DIFF
--- a/.github/workflows/lint_and_integration.yaml
+++ b/.github/workflows/lint_and_integration.yaml
@@ -112,9 +112,10 @@ jobs:
     with:
       arch: amd64
       os: ubuntu:24.04
-      test-tags: ${{ needs.get-e2e-tags.outputs.test-tags}}
+      test-tags: up_to_weekly
       artifact: k8s.snap
       parallel: true
+      extra-test-args: "-k test_disable_separate_feature_upgrades"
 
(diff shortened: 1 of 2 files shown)


ktsakalozos (approved):
Thank you!

PR #1538, opened 2025-06-11
feat(microcluster): Change heartheat interval

HomayoonAlimohammadi:
### Overview

This can be a workaround for the failed heartbeats in Microcluster when a member is cut off.
A broader discussion is ongoing with the Microcluster team in the following PR: https://github.com/canonical/microcluster/pull/412


HomayoonAlimohammadi:
@bschimke95 Exactly. This timeout is a bigger problem honestly, because GET request to the `cluster` endpoint of Microcluster will take at least 30 seconds * num_cut_off_nodes (which in a typical 3 node cluster can be 1 at max).
As an example, `GetClusterMembers` call also calls this GET `cluster` endpoint. We do have a workaround for that tho, and that is the `--timeout` in `k8s` commands.

PR #1546, opened 2025-06-12
fix: Don't print k8s binary outputs in CLI

bschimke95:
In #1423, a pre-init check was introduced on the client-side. In this check, the k8s binaries are started to ensure they do not hang later. However, the output was printed to stdout causing e.g. the `k8s bootstrap` command to be polluted with debug messages.

This commit pipes the output into the void instead.


berkayoz (approved):
LGTM

PR #1556, opened 2025-06-13
test: Add weekly FIPS tests on multipass

bschimke95:
# Summary

This PR introduces a new weekly GitHub Actions workflow and makes several enhancements to the Multipass integration test harness.

# Changes
* Adds a weekly GitHub workflow scheduled to run every Saturday
* Fixes various issues in the Multipass test harness
* Adds IPv6 support for the Multipass harness
* Enables running integration tests on FIPS-enabled Multipass instances as part of the weekly workflow
* Skipped some tests on the multipass harness due to various reasons

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -203,9 +203,6 @@ jobs:
           fi
 
           cd tests/integration && sudo --user "$USER" --preserve-env --preserve-env=PATH -- env -- tox -e integration -- ${{ inputs.parallel && matrix.test || '' }} ${{ env.PERFORMANCE_FLAGS }} --tags ${{ inputs.test-tags }} ${{ inputs.extra-test-args }}
-      - name: Setup tmate session
-        uses: mxschmitt/action-tmate@v3
-        if: failure()
       - name: Prepare inspection reports
         if: failure()
         run: |


.github/workflows/weekly-test.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Test results
+        run: |
+         RESULTS=$(echo '${{ toJson(needs) }}' | jq -c '[.[] | .result]')
+
+         if echo $RESULTS | jq -e 'all(. == "success")'; then
+           echo "RESULT=success" >> $GITHUB_ENV
+         else
+           echo "RESULT=failure" >> $GITHUB_ENV
+         fi
+      - name: Generate Mattermost Message
```
louiseschmidtgen, line 55: Can we turn this MM messages into an action as it is also used by our nightly CI?

tests/integration/tests/test_util/config.py
```diff
@@ -12,6 +12,7 @@
 DEFAULT_WAIT_DELAY_S = int(os.getenv("TEST_DEFAULT_WAIT_DELAY_S") or 5)
 
 MANIFESTS_DIR = DIR / ".." / ".." / "templates"
+CLOUD_INIT_DIR = MANIFESTS_DIR / "cloud-init"
```
louiseschmidtgen, line 15: I think it is a bit confusing to have both `CLOUD_INIT_DIR` and `MULTIPASS_CLOUD_INIT`. Can we not re-use the same env variable?

tests/integration/tests/test_util/harness/multipass.py
```diff
@@ -32,43 +33,103 @@ def __init__(self):
...
+                    r"\$\{([^}]+)\}|\$([A-Za-z_][A-Za-z0-9_]*)",
+                    replace_env_var,
+                    cloud_init_content,
+                )
+
+                LOG.info("Using cloud-init: %s", cloud_init_content)
+                # Note(ben): Multipass does not handle restarts in
+                # cloud-init very well and just times out even if
+                # the underlying machine works just fine.
+                # Hence, we disable the check for this call, continue and hope for the best.
```
louiseschmidtgen: Could we have a log message here and some sort of retry or restart if multipass doesn't start properly? It sounds like this would cause us some flakiness.

.github/workflows/e2e-tests.yaml
```diff
@@ -133,14 +160,35 @@ jobs:
           channel: ${{ inputs.channel }}
           artifact: ${{ inputs.artifact }}
       - name: Setup LXD
+        if: ${{ inputs.substrate == 'lxd' }}
         uses: canonical/setup-lxd@v0.1.3
         with:
           bridges: "lxdbr0,dualstack-br0,ipv6-br0"
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
```
ktsakalozos: I am not sure about this. We should not have in our release pipeline untrusted steps.

.github/workflows/weekly-test.yaml
bschimke95: I added KU-3765 to work on this in a separate PR.

tests/integration/tests/test_util/config.py
bschimke95: CLOUD_INIT_DIR is not an environment variable but a constant that points to the cloud-init files, which means the user cannot directly change it and IMHO cannot get confused by it. I choose a fixed path for CLOUD_INIT_DIR because it can get tricky to set the right files otherwise from another job.

.github/workflows/e2e-tests.yaml
bschimke95: Fair point - removed it.

tests/integration/tests/test_util/harness/multipass.py
bschimke95: The retry is done in the next command.
The problem with the `multipass launch` with `reboot` cloud-init is that the command will fail even if the machine comes up just fine. 
So, instead of retrying this command - which will just return a "machine already exists" error in consecutive calls - we try to interact with the machine in the next command. 
If this fails after retries, we fail the whole test.

I rephrased the comment to reflect this better. Let me know if this works for you now.

louiseschmidtgen (approved):
Great work here Ben! Thanks for all the work to get multipass working properly.

PR #1561, opened 2025-06-16
feat: add managed etcd implementation

berkayoz:
## Description

Based on #500.

Adds a new etcd wrapper/command that mimics k8s-dqlite behavior for joining.

Updates related sections to handle operations in bootstrap, join and remove hooks.

## Backport

## Checklist

- [x] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed


src/k8s/pkg/etcd/register.go
```diff
@@ -0,0 +1,81 @@
...
+			if e.peerURL == url {
+				return "", nil
+			}
+		}
+	}
+
+	members, err := client.Cluster.MemberAdd(ctx, []string{e.peerURL})
+	if err != nil {
+		return "", fmt.Errorf("failed to add cluster member with peerURL %q: %w", e.peerURL, err)
+	}
```
louiseschmidtgen: We should check the error and determine whether it was a context error or a grpc error. Else we should retry the operation. Wdyt?

src/k8s/pkg/snap/snap.go
```diff
@@ -390,6 +395,13 @@ func (s *snap) K8sDqliteClient(ctx context.Context) (*dqlite.Client, error) {
 	return client, nil
 }
 
+func (s *snap) EtcdClient(endpoints []string) (etcd.Client, error) {
```
louiseschmidtgen: If I see correctly, the EtcdClient does not explicitly get closed.
The client documentation warns us: 
> Make sure to close the client after using it. If the client is not closed, the connection will have leaky goroutines.

src/k8s/pkg/client/etcd/client.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	return clientv3.New(clientv3.Config{
+		Endpoints: endpoints,
+		TLS:       tlsConfig,
+	})
+}
+
+func (c *externalClient) RemoveNodeByAddress(ctx context.Context, peerURL string) error {
+	// List all members
```
louiseschmidtgen:
```suggestion
```

src/k8s/pkg/client/etcd/client.go
```diff
@@ -0,0 +1,77 @@
...
+	resp, err := c.etcdClient.MemberList(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to list etcd members: %w", err)
+	}
+
+	// Find the member with the matching peerURL
+	var memberID uint64
```
louiseschmidtgen, line 45: Is the memberID something that we can put in a map with the peerURLs?

src/k8s/cmd/etcd/etcd.go
```diff
@@ -0,0 +1,91 @@
...
+			command.GroupID = group.ID
+		}
+	}
+
+	root.AddCommand(commands...)
+}
+
+func NewRootCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "k8sd",
```
HomayoonAlimohammadi: shouldn't this be `etcd`?

src/k8s/cmd/etcd/etcd.go
```diff
@@ -0,0 +1,91 @@
...
+			log.Configure(log.Options{
+				LogLevel:     rootCmdOpts.logLevel,
+				AddDirHeader: true,
+			})
+
+			instance, err := etcd.New(rootCmdOpts.stateDir)
+			if err != nil {
+				cmd.PrintErrf("Error: Failed to initialize etcd: %v", err)
+				env.Exit(1)
+				return
```
HomayoonAlimohammadi: Why are we using `env.Exit(1)` and `return` here, while in other places we use `logrus.Fatal()`?

src/k8s/cmd/etcd/etcd.go
HomayoonAlimohammadi, line 1: I think this file needs to change a bit to reflect the `etcd` command. Currently I guess it has some `k8sd` related stuff (e.g. options).

src/k8s/pkg/client/etcd/client.go
```diff
@@ -0,0 +1,77 @@
...
+	for _, m := range resp.Members {
+		for _, url := range m.PeerURLs {
+			if url == peerURL {
+				memberID = m.ID
+				break
+			}
+		}
+	}
```
HomayoonAlimohammadi, line 53: In an ideal world, we won't need to iterate through members, right? Only iterating through the PeerURLs of a single member should suffice.

src/k8s/pkg/client/etcd/client.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if memberID == 0 {
+		return fmt.Errorf("no etcd member found with peer URL: %s", peerURL)
+	}
+
+	// Remove the member
+	_, err = c.etcdClient.MemberRemove(ctx, memberID)
+	if err != nil {
+		return fmt.Errorf("failed to remove etcd member: %w", err)
+	}
```
HomayoonAlimohammadi: nit: Maybe inline error?

src/k8s/pkg/client/etcd/client.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/server/v3/embed"
+)
+
+type externalClient struct {
```
HomayoonAlimohammadi: Why is this called `external` client?

src/k8s/pkg/etcd/etcd.go
```diff
@@ -0,0 +1,61 @@
...
+	tlsConfig, err := tlsConfig{
+		CAFile:   registerConfig.TrustedCAFile,
+		CertFile: registerConfig.CertFile,
+		KeyFile:  registerConfig.KeyFile,
+	}.ClientConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize client TLS config: %w", err)
+	}
+
+	if err := os.MkdirAll(config.Dir, 0o700); err != nil {
```
HomayoonAlimohammadi: nit: Do we want to let everyone read these?

src/k8s/pkg/etcd/register.go
```diff
@@ -0,0 +1,81 @@
...
+	return strings.TrimSpace(string(b)) == e.peerURL, nil
+}
+
+func (e *etcd) createSentinel() error {
+	return os.WriteFile(e.sentinelFile, []byte(e.peerURL), 0o600)
+}
+
+func (e *etcd) ensurePeerInCluster(ctx context.Context) (string, error) {
+	if e.config.ClusterState == "new" {
```
HomayoonAlimohammadi: Maybe we can add constants for these states?
Edit: Seems like there are already consts for these in the etcd package. `config.ClusterStateFlagNew`.

src/k8s/pkg/etcd/register.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+func (e *etcd) hasValidSentinelFile() (bool, error) {
```
HomayoonAlimohammadi: What is the usage of this sentinel file? From what I've read so far, it seems like we're only storing our peerURL into that file. Is it similar to the `cluster.yaml` file in dqlite, but we only store a single peer URL into this one?

src/k8s/pkg/etcd/register.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	members, err := client.Cluster.MemberAdd(ctx, []string{e.peerURL})
+	if err != nil {
+		return "", fmt.Errorf("failed to add cluster member with peerURL %q: %w", e.peerURL, err)
+	}
+
+	defer func() {
+		if err := e.createSentinel(); err != nil {
+			panic(fmt.Sprintf("failed to create sentinel file: %v", err.Error()))
```
HomayoonAlimohammadi: Do we need to panic here? How about an error?

src/k8s/pkg/etcd/register.go
```diff
@@ -0,0 +1,81 @@
...
+		return "", fmt.Errorf("failed to list cluster members: %w", err)
+	}
+
+	for _, member := range resp.Members {
+		for _, url := range member.GetPeerURLs() {
+			if e.peerURL == url {
+				return "", nil
+			}
+		}
+	}
```
HomayoonAlimohammadi: If we know that the peer is in cluster by checking the peer URLs, shouldn't we also create the sentinel file before returning since it's not already available? (we have a check for this file above). I guess `initialClusterMembers` should also be returned from this one as well.

src/k8s/pkg/etcd/register.go
```diff
@@ -0,0 +1,81 @@
...
+		return "", fmt.Errorf("failed to add cluster member with peerURL %q: %w", e.peerURL, err)
+	}
+
+	defer func() {
+		if err := e.createSentinel(); err != nil {
+			panic(fmt.Sprintf("failed to create sentinel file: %v", err.Error()))
+		}
+	}()
+
+	initialClusterParts := make([]string, 0, len(members.Members))
```
HomayoonAlimohammadi: ~~I think this is not `initial` anymore, since the `members.Members` include the peer that was just added by the `MemberAdd()` call~~ EDIT: This follows Etcd naming format.
Why are we returning this `name=URL` string only in here? I think this can be returned in some other places in this function as well, when there _is_ a cluster with some members.

src/k8s/pkg/etcd/register.go
```diff
@@ -0,0 +1,81 @@
...
+	for _, member := range members.Members {
+		name := member.GetName()
+		for _, url := range member.GetPeerURLs() {
+			if url == e.peerURL {
+				name = e.config.Name
```
HomayoonAlimohammadi: Why are we overwriting the name of this specific peer? Might `e.config.Name` be different than `member.GetName()`? In that case, don't we need to update the etcd members to reflect the correct name?

src/k8s/pkg/etcd/shutdown.go
```diff
@@ -0,0 +1,22 @@
...
+	}
+	e.instance.Close()
+
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("timed out waiting for server to stop: %w", ctx.Err())
+	case <-e.instance.Server.StopNotify():
+	}
+
+	close(e.mustStopCh)
```
HomayoonAlimohammadi: I think we need to close this channel before actually closing the instance to make sure we shut down gracefully?

src/k8s/pkg/etcd/start.go
```diff
@@ -0,0 +1,32 @@
...
+		return fmt.Errorf("failed to start node: %w", err)
+	} else {
+		e.instance = instance
+	}
+
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("etcd did not start in time: %w", ctx.Err())
+	case <-e.instance.Server.ReadyNotify():
+		return nil
```
HomayoonAlimohammadi: nit: Maybe we can log.info something like "etcd ready" here. I think this can be useful for future debugging.

src/k8s/pkg/etcd/register.go
```diff
@@ -0,0 +1,81 @@
...
+		return false, err
+	}
+	return strings.TrimSpace(string(b)) == e.peerURL, nil
+}
+
+func (e *etcd) createSentinel() error {
+	return os.WriteFile(e.sentinelFile, []byte(e.peerURL), 0o600)
+}
+
+func (e *etcd) ensurePeerInCluster(ctx context.Context) (string, error) {
```
HomayoonAlimohammadi: I think adding a doc to this function can really help. IMO it's not obvious from the name and signature of the function what it returns.

src/k8s/pkg/k8sd/app/cluster_util.go
```diff
@@ -16,7 +16,12 @@ func startControlPlaneServices(ctx context.Context, snap snap.Snap, datastore st
 		if err := snaputil.StartK8sDqliteServices(ctx, snap); err != nil {
 			return fmt.Errorf("failed to start control plane services: %w", err)
 		}
+	case "etcd":
```
HomayoonAlimohammadi, line 19: nit: Maybe for a future work, but I think we can extract these to consts.

src/k8s/pkg/k8sd/app/cluster_util.go
```diff
@@ -16,7 +16,12 @@ func startControlPlaneServices(ctx context.Context, snap snap.Snap, datastore st
...
+		if err := snaputil.StartEtcdServices(ctx, snap); err != nil {
+			return fmt.Errorf("failed to start control plane services: %w", err)
```
HomayoonAlimohammadi: nit: The error looks a bit odd to me. I can see that it's the same for `k8s-dqlite`, but how about `failed to start etcd services`?

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -208,6 +233,29 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+			if member.Name == s.Name() {
+				// skip self
+				continue
+			}
+			clientURLs = append(clientURLs, fmt.Sprintf("https://%s", utils.JoinHostPort(member.Address.Addr().String(), cfg.Datastore.GetEtcdPort())))
+		}
+
+		clientURL := fmt.Sprintf("https://%s", utils.JoinHostPort(nodeIP.String(), cfg.Datastore.GetEtcdPort()))
+		peerURL := fmt.Sprintf("https://%s", utils.JoinHostPort(nodeIP.String(), cfg.Datastore.GetEtcdPeerPort()))
+		if err := setup.Etcd(snap, s.Name(), clientURL, peerURL, clientURLs, joinConfig.ExtraNodeK8sDqliteArgs); err != nil {
```
HomayoonAlimohammadi: I think this should be `ExtraNodeEtcdArgs`

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -42,13 +42,46 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
...
+			members, err := leader.GetClusterMembers(ctx)
+			if err != nil {
+				return types.ClusterConfig{}, nil, fmt.Errorf("failed to retrieve cluster members: %w", err)
+			}
+			nodeIPs := make([]string, 0, len(members))
+			for _, member := range members {
+				nodeIPs = append(nodeIPs, member.Address.Addr().String())
+			}
+
+			return cfg, nodeIPs, nil
```
HomayoonAlimohammadi: IMO it's a bit odd that we only return nodeIPs if we're using etcd. I know that with dqlite we probably don't need it, but returning an empty slice can be tricky since it basically means there are no node IPs. I lean towards returning it anyways.

louiseschmidtgen (commented):
Thank you for all the work on this Berkay! I have a few comments for you:

src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
```diff
@@ -77,9 +77,19 @@ func (c *ControlPlaneConfigurationController) Run(ctx context.Context, getCluste
...
+func (c *ControlPlaneConfigurationController) reconcile(ctx context.Context, config types.ClusterConfig, nodeIPs []string) error {
 	// kube-apiserver: external datastore
-	if config.Datastore.GetType() == "external" {
+	switch config.Datastore.GetType() {
+	case "etcd":
+		updateArgs, deleteArgs := config.Datastore.ToKubeAPIServerArguments(c.snap, nodeIPs)
+
+		// NOTE(neoaggelos): update kube-apiserver arguments in case cluster nodes have changed, but do not
+		// restart kube-apiserver, to avoid downtime for existing cluster nodes. The next time kube-apiserver
+		// restarts on this node, they will use the updated arguments.
```
HomayoonAlimohammadi: But what if the apiserver does not update for a while? Isn't it going to use outdated node IPs (which might cause downtime itself since they translate to `--etcd-servers`)?

src/k8s/pkg/k8sd/controllers/node_label.go
```diff
@@ -80,33 +80,40 @@ func (c *NodeLabelController) reconcileFailureDomain(ctx context.Context, node *
...
+	if datastoreType == "k8s-dqlite" {
+		k8sDqliteStateDir := c.snap.K8sDqliteStateDir()
 
-	if modified {
-		log.Info("Updated k8s-dqlite failure domain", "failure domain", failureDomain, "availability zone", availabilityZone)
-		if err = c.snap.RestartServices(ctx, []string{"k8s-dqlite"}); err != nil {
-			return fmt.Errorf("failed to restart k8s-dqlite to apply failure domain: %w", err)
+		modified, err := snaputil.UpdateDqliteFailureDomain(failureDomain, k8sDqliteStateDir)
+		if err != nil {
+			return err
```
HomayoonAlimohammadi: nit: Maybe wrap?

src/k8s/pkg/k8sd/controllers/node_label.go
```diff
@@ -80,33 +80,40 @@ func (c *NodeLabelController) reconcileFailureDomain(ctx context.Context, node *
...
+		if modified {
+			log.Info("Updated k8s-dqlite failure domain", "failure domain", failureDomain, "availability zone", availabilityZone)
+			if err = c.snap.RestartServices(ctx, []string{"k8s-dqlite"}); err != nil {
+				return fmt.Errorf("failed to restart k8s-dqlite to apply failure domain: %w", err)
+			}
 		}
 	}
 
-	modified, err = snaputil.UpdateDqliteFailureDomain(failureDomain, k8sdDbStateDir)
+	modified, err := snaputil.UpdateDqliteFailureDomain(failureDomain, k8sdDbStateDir)
```
HomayoonAlimohammadi, line 116: Why do we have `UpdateDqliteFailureDomain` two times? I think this one needs to be removed as it's already available in the if block above.

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,144 @@
...
+		c.CAKey = key
+	}
+
+	cert, key, err := pkiutil.LoadCertificate(c.CACert, c.CAKey)
+	if err != nil {
+		return fmt.Errorf("failed to parse etcd CA: %w", err)
+	}
+
+	// Generate etcd server certificate
+	if c.ServerCert == "" && c.ServerKey == "" {
```
HomayoonAlimohammadi, line 90: don't we want to fail hard of one is available but the other is not? like the cacert and cakey

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,144 @@
...
+		if key == nil {
+			return fmt.Errorf("using an external etcd CA with specifying an etcd server certificate is not possible")
+		}
+		template, err := pkiutil.GenerateCertificate(pkix.Name{CommonName: "kube-etcd"}, c.notBefore, c.notAfter, false, append(c.dnsSANs, c.hostname), append(c.ipSANs, net.IP{127, 0, 0, 1}))
+		if err != nil {
+			return fmt.Errorf("failed to generate k8s-dqlite certificate: %w", err)
```
HomayoonAlimohammadi:
```suggestion
			return fmt.Errorf("failed to generate etcd certificate: %w", err)
```

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		cert, key, err := pkiutil.SignCertificate(template, 2048, cert, &key.PublicKey, key)
+		if err != nil {
+			return fmt.Errorf("failed to self-sign k8s-dqlite certificate: %w", err)
```
HomayoonAlimohammadi:
```suggestion
			return fmt.Errorf("failed to self-sign etcd certificate: %w", err)
```

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// Generate etcd peer server certificate
+	if c.ServerPeerCert == "" && c.ServerPeerKey == "" {
+		if key == nil {
+			return fmt.Errorf("using an external etcd CA with specifying an etcd server peer certificate is not possible")
+		}
+
+		template, err := pkiutil.GenerateCertificate(pkix.Name{CommonName: "kube-etcd-peer"}, c.notBefore, c.notAfter, false, append(c.dnsSANs, c.hostname), append(c.ipSANs, net.IP{127, 0, 0, 1}))
+		if err != nil {
+			return fmt.Errorf("failed to generate k8s-dqlite certificate: %w", err)
```
HomayoonAlimohammadi:
```suggestion
			return fmt.Errorf("failed to generate etcd certificate: %w", err)
```

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		cert, key, err := pkiutil.SignCertificate(template, 2048, cert, &key.PublicKey, key)
+		if err != nil {
+			return fmt.Errorf("failed to self-sign k8s-dqlite certificate: %w", err)
```
HomayoonAlimohammadi:
```suggestion
			return fmt.Errorf("failed to self-sign etcd certificate: %w", err)
```

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// Generate kube-apiserver etcd client certificate
+	if c.APIServerClientCert == "" && c.APIServerClientKey == "" {
+		if key == nil {
+			return fmt.Errorf("using an external etcd CA with specifying an etcd apiserver client certificate is not possible")
+		}
+
+		template, err := pkiutil.GenerateCertificate(pkix.Name{CommonName: "kube-apiserver-etcd-client"}, c.notBefore, c.notAfter, false, nil, nil)
+		if err != nil {
+			return fmt.Errorf("failed to generate k8s-dqlite certificate: %w", err)
```
HomayoonAlimohammadi:
```suggestion
			return fmt.Errorf("failed to generate etcd certificate: %w", err)
```

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		cert, key, err := pkiutil.SignCertificate(template, 2048, cert, &key.PublicKey, key)
+		if err != nil {
+			return fmt.Errorf("failed to self-sign k8s-dqlite certificate: %w", err)
```
HomayoonAlimohammadi:
```suggestion
			return fmt.Errorf("failed to self-sign etcd certificate: %w", err)
```

src/k8s/pkg/k8sd/pki/k8sdqlite.go
```diff
@@ -29,7 +29,6 @@ type K8sDqlitePKIOpts struct {
 	NotBefore         time.Time
 	NotAfter          time.Time
 	AllowSelfSignedCA bool
-	Datastore         string
```
HomayoonAlimohammadi, line 32: Not used I guess?

src/k8s/pkg/k8sd/setup/etcd.go
```diff
@@ -0,0 +1,104 @@
...
+	CertFile      string   `yaml:"cert-file,omitempty"`
+	KeyFile       string   `yaml:"key-file,omitempty"`
+	TrustedCAFile string   `yaml:"trusted-ca-file,omitempty"`
+}
+
+func newEtcdConfig(snap snap.Snap, name, clientURL, peerURL string, clientURLs []string) etcdConfig {
+	clusterState := "new"
+	if len(clientURLs) > 0 {
+		clusterState = "existing"
+	}
```
HomayoonAlimohammadi: nit: Use `config.ClusterStateFlagNew|Existing`

src/k8s/pkg/k8sd/setup/etcd.go
```diff
@@ -0,0 +1,104 @@
...
+		ClientURLs:    clientURLs,
+		TrustedCAFile: filepath.Join(snap.EtcdPKIDir(), "ca.crt"),
+		CertFile:      filepath.Join(snap.EtcdPKIDir(), "server.crt"),
+		KeyFile:       filepath.Join(snap.EtcdPKIDir(), "server.key"),
+	}
+}
+
+func Etcd(snap snap.Snap, name, clientURL, peerURL string, clientURLs []string, extraArgs map[string]*string) error {
+	if b, err := yaml.Marshal(newEtcdConfig(snap, name, clientURL, peerURL, clientURLs)); err != nil {
+		return fmt.Errorf("failed to create etcd.yaml file for name=%q address=%q: %w", name, peerURL, err)
```
HomayoonAlimohammadi: I think the error is wrong here. We're not creating any files.

src/k8s/pkg/k8sd/setup/etcd.go
```diff
@@ -0,0 +1,104 @@
...
+	} else if err := os.WriteFile(filepath.Join(snap.EtcdDir(), "etcd.yaml"), b, 0o600); err != nil {
+		return fmt.Errorf("failed to write etcd.yaml config for name=%q address=%q: %w", name, peerURL, err)
+	}
+
+	if b, err := yaml.Marshal(newEtcdRegisterConfig(snap, peerURL, clientURLs)); err != nil {
+		return fmt.Errorf("failed to create register.yaml file for name=%q address=%q: %w", name, peerURL, err)
```
HomayoonAlimohammadi: I think the error can be improved here as well

src/k8s/pkg/k8sd/setup/etcd.go
```diff
@@ -0,0 +1,104 @@
...
+	} else if err := os.WriteFile(filepath.Join(snap.EtcdDir(), "register.yaml"), b, 0o600); err != nil {
```
HomayoonAlimohammadi: From what I understand, `register.yaml` is basically a subset of `etcd.yaml` except for the `clientURLs`, right?
And what is the usage of the sentinel file here since it only includes the peer URL? Isn't it included here in the `register.yaml` as well?

HomayoonAlimohammadi (commented):
Amazing work @berkayoz! Did a first pass, but didn't go through the tests. Left some comments and questions.

src/k8s/cmd/etcd/etcd.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var rootCmdOpts struct {
+	logDebug   bool
+	logVerbose bool
+	logLevel   int
+	stateDir   string
+}
+
+func addCommands(root *cobra.Command, group *cobra.Group, commands ...*cobra.Command) {
```
HomayoonAlimohammadi: I think this is not used. However, we should have a linter for these things.

src/k8s/pkg/etcd/register.go
berkayoz: If this call fails, the sentinel file won't be created and the `etcd` service will fail. This means that systemd service restarts will retry this call on failure. I believe we don't need to orchestrate this here.

src/k8s/cmd/etcd/etcd.go
berkayoz: I'll replace the use of `logrus`. The main reason we do not exit is to perform proper etcd server shutdown / cleanup and handle signals.

src/k8s/pkg/etcd/etcd.go
berkayoz: This is inline with the `EnsureAllDirectories` call which also uses perm  `0o700`.

src/k8s/pkg/etcd/register.go
berkayoz: This sentinel file is used to orchestrate the member add operation / joining nodes. Sentinel file is used to ensure that the peer addition is only performed once. I've added a doc string to the function to explain this.

src/k8s/pkg/etcd/register.go
berkayoz: Error returns are always weird in deferred functions. Panic feels simpler and fair for a file write operation. I guess the alternative is named returns and setting the error variable there, which has shadowing/overwriting issues on it's own.

src/k8s/pkg/etcd/register.go
berkayoz: AFAIK the `initial-*` flags do not matter after a cluster is created / node is joined. From the etcd wiki

> The command line parameters starting with `--initial-cluster` will be ignored on subsequent runs of etcd. 

This is a check to make sure `MemberAdd` is not called even if the sentinel file disappears. If the node is in the member list these flags do not matter anymore.

src/k8s/pkg/etcd/register.go
berkayoz: See above comment I've made, this is only used on joining nodes at the time of joining. It will not be used after member successfully joins.

src/k8s/pkg/etcd/register.go
berkayoz: IIRC the member add function is performed with only the `peerURL` and the peer will not have a name set yet. This is why we override the name for the joining peer, rest of the members will get its value through `member.GetName`.

src/k8s/pkg/etcd/shutdown.go
berkayoz: I believe we already wait for the etcd server to gracefully stop with `case <-e.instance.Server.StopNotify():`.

If I understand correctly this should only matter when the `cmd` gets a signal, we call the `Shutdown`, and make sure to close this channel to not leak it?

src/k8s/pkg/k8sd/controllers/node_label.go
berkayoz: One of them is for k8s-dqlite and other one is for k8sd which we should update for any `datastore-type`.

src/k8s/pkg/client/etcd/client.go
berkayoz: I'm not sure if I understand this well. We do not know the member ids and have to do a simple search to fetch it. The member limit is low enough that this doesn't matter performance wise.

src/k8s/pkg/client/etcd/client.go
berkayoz: This is something we have to determine at runtime. Etcd cluster size is already low enough that the simple loop search shouldn't matter in terms of performance. There is no point in storing/caching this either as it's not performed very often and it would add complexity to keep it synced with etcd state.

src/k8s/pkg/k8sd/pki/k8sdqlite.go
berkayoz: Yeah this is unused as far as I can tell, which I stumbled upon when looking at `pki`.

src/k8s/pkg/k8sd/setup/etcd.go
berkayoz: Yes, it is a subset used for joining the cluster. I've explained above why the sentinel file is used.

src/k8s/pkg/k8sd/pki/etcd.go
berkayoz: This is similar to how k8s-dqlite `pki` handles it, we won't generate anything unless both are empty?

src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
berkayoz: I've mixed feelings about this approach, I've been looking into this. I don't see a reason to not use `127.0.0.1:2380` directly since we are doing a stacked etcd deployment. I initially interpreted this as a requirement which might not hold true. I'm going to test out the loopback approach and see.

For the question as long as the apiserver has its own IP it should never lose communication, which will be the case from the start.

src/k8s/pkg/snap/snap.go
berkayoz: Good catch, thanks!

src/k8s/pkg/client/etcd/client.go
HomayoonAlimohammadi: I mean, why do we need to iterate through PeerURLs of _every member_? I suspect iterating through the PeerURLs of a single member should be enough (let's say `resp.Members[0].PeerURLs`).
But there's no performance issue here as the number of members is definitely not significant and we do have a `break`. I just asked this out of curiosity.

src/k8s/cmd/etcd/etcd.go
HomayoonAlimohammadi: Yea but those two basically do the same thing right? I think `logrus.Fatal` also calls `os.Exit(1)`. If we don't want to exit I think we should remove that part.

src/k8s/pkg/etcd/register.go
HomayoonAlimohammadi: But we're retrieving the `MemberList` right after that, right? Shouldn't the peer be available in that list?

src/k8s/pkg/etcd/register.go
HomayoonAlimohammadi: Got it. I still have existential problems with the sentinel file, but those aside, I think if we look at the `ensurePeerInCluster` as a standalone function, it's better for consistency's sake that it returns the initialClusterMembers list even after the node is joined. Regardless of whether or not this returned value is being ignored by Etcd. But this is a nit.

src/k8s/pkg/etcd/register.go
HomayoonAlimohammadi: I think you're right. But that's very odd. We're not specifying a name in the `MemberAdd` call, but doing this using `etcdctl` does [require](https://etcd.io/docs/v3.3/op-guide/runtime-configuration/) a name...

src/k8s/pkg/etcd/shutdown.go
HomayoonAlimohammadi: Yea but I think there's a dependency cycle going on here.
```go
func (e *etcd) Shutdown(ctx context.Context) error {
	if e.instance == nil {
		return nil
	}
	e.instance.Close()

	select {
	case <-ctx.Done():
		return fmt.Errorf("timed out waiting for server to stop: %w", ctx.Err())
	case <-e.instance.Server.StopNotify():
	}

	close(e.mustStopCh)
	return nil
}
```
and
```go
			select {
			case <-ch:
			case <-instance.MustStop():
			}
			cancel()

			// Create a separate context with 30 seconds to cleanup
			stopCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
			defer cancel()

			if err := instance.Shutdown(stopCtx); err != nil {
				cmd.PrintErrf("Error: Failed to shutdown etcd: %v", err)
			}
```
If we get a signal, `<-ch` is going to return, which will call `Shutdown`, which will `close(mustStopCh)`. This scenario is OK, but the `mustStopCh` is useless here.
But let's consider the other scenario in that select case.
How is the `mustStopCh` going to return first? It needs to get closed, which happens in `Shutdown` call. But that call is blocked by this select statement which is blocked on the `mustStopCh`. This is a dependency and IMO is meaningless.

src/k8s/pkg/k8sd/setup/etcd.go
HomayoonAlimohammadi: So I think sentinelFile is not necessary as the peerURL is available in the `MemberList` call that we have.
With that out of the way, why do we need a `register.yaml` and an `etcd.yaml`? I see that the only thing that's available in the `register.yaml` and is not available in the `etcd.yaml`, is the `clientURLs`. I think there should be a way to obtain that without the need to store it in a file? But this is a nit.

src/k8s/cmd/etcd/etcd.go
```diff
@@ -0,0 +1,71 @@
...
+			if err := instance.Start(ctx); err != nil {
+				cmd.PrintErrf("Error: Failed to start etcd: %v", err)
+			}
+
+			// Cancel context if we receive an exit signal
+			ch := make(chan os.Signal, 1)
+			signal.Notify(ch, unix.SIGPWR)
+			signal.Notify(ch, unix.SIGINT)
+			signal.Notify(ch, unix.SIGQUIT)
+			signal.Notify(ch, unix.SIGTERM)
```
mateoflorido: IIRC, the signature of `Notify` allows you to pass multiple signals:
```suggestion
			signal.Notify(ch, unix.SIGPWR, unix.SIGINT, unix.SIGQUIT, unix.SIGTERM)
```

src/k8s/cmd/etcd/etcd.go
```diff
@@ -0,0 +1,71 @@
...
+			if err != nil {
+				cmd.PrintErrf("Error: Failed to initialize etcd: %v", err)
+				env.Exit(1)
+				return
+			}
+
+			ctx, cancel := context.WithCancel(cmd.Context())
+			if err := instance.Start(ctx); err != nil {
+				cmd.PrintErrf("Error: Failed to start etcd: %v", err)
+			}
```
mateoflorido: Maybe we should exit if the instance couldn't start?
```suggestion
			if err := instance.Start(ctx); err != nil {
				cmd.PrintErrf("Error: Failed to start etcd: %v", err)
				env.Exit(1)
				return
			}
```

src/k8s/cmd/etcd/etcd.go
```diff
@@ -0,0 +1,71 @@
...
+			// Cancel context if we receive an exit signal
+			ch := make(chan os.Signal, 1)
+			signal.Notify(ch, unix.SIGPWR)
+			signal.Notify(ch, unix.SIGINT)
+			signal.Notify(ch, unix.SIGQUIT)
+			signal.Notify(ch, unix.SIGTERM)
+
+			select {
+			case <-ch:
+			case <-instance.MustStop():
```
mateoflorido: Can we add a couple logs here to at least surface why we are exiting?

src/k8s/cmd/etcd/etcd.go
```diff
@@ -0,0 +1,71 @@
...
+			cancel()
+
+			// Create a separate context with 30 seconds to cleanup
+			stopCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer cancel()
+
+			if err := instance.Shutdown(stopCtx); err != nil {
+				cmd.PrintErrf("Error: Failed to shutdown etcd: %v", err)
+			}
+		},
```
mateoflorido: Same here
```suggestion
			if err := instance.Shutdown(stopCtx); err != nil {
				cmd.PrintErrf("Error: Failed to shutdown etcd: %v", err)
				env.Exit(1)
				return
			}
		},
```

src/k8s/cmd/etcd/etcd.go
mateoflorido: Also, if we receive a `MustStop`, should we carry the error so we exit the process with a different error code than `rc: 0`?

mateoflorido (commented):
Great work Berkay! I've left a few comments:

src/k8s/pkg/client/etcd/client.go
```diff
@@ -0,0 +1,82 @@
...
+func (c *etcdClient) RemoveNodeByAddress(ctx context.Context, peerURL string) error {
+	resp, err := c.client.MemberList(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to list etcd members: %w", err)
+	}
+
+	// Find the member with the matching peerURL
+	var memberID uint64
+	for _, m := range resp.Members {
+		for _, url := range m.PeerURLs {
```
mateoflorido, line 47: Quick question about `PeerURLs`. IIRC, these URLs are only intended for internal features (e.g., consensus), but the ClientURLs are used for Client facing API endpoints. What's the rationale for using the PeerURL instead of the ClientURL while removing the nodes.

src/k8s/pkg/k8sd/setup/etcd.go
berkayoz: This comment is no longer relevant since we shifted to a different approach.

src/k8s/pkg/k8sd/app/hooks_start.go
berkayoz: We've shifted to the stacked approach where the apiserver now talks to etcd on the local node.

src/k8s/pkg/client/etcd/client.go
berkayoz: Ah I see, honestly I'm not entirely sure. To me the bigger question would be why do we have peer urls per member, this approach feels on the safer side though may be useless.

src/k8s/pkg/client/etcd/client.go
berkayoz: You're correct that it is used "internally" for raft. Main reason to me is that etcd membership operations (in the client) like `MemberAdd` are also performed with the peer url.

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// CompleteCertificates generates missing or unset certificates. If only a certificate is set and not a key, we assume that the cluster is using managed certificates.
+func (c *EtcdPKI) CompleteCertificates() error {
+	// Fail hard if keys of self-signed certificates are set without the respective certificates
+	switch {
+	case c.CACert == "" && c.CAKey != "":
+		return fmt.Errorf("etcd CA certificate key set without a certificate, fail to prevent further issues")
+	case c.CACert != "" && c.CAKey == "":
+		return fmt.Errorf("etcd CA certificate set without a key, fail to prevent further issues")
```
rapour: Have we dropped support for managed certificates? It seems the method description suggests we support certificates without keys but we enforce providing at least the key PEM

src/k8s/pkg/utils/localhost.go
```diff
@@ -3,25 +3,39 @@ package utils
...
-		memberAddress := clusterMember.Address.Addr().String()
-		nodeIP := net.ParseIP(memberAddress)
-		if nodeIP == nil {
-			return "", fmt.Errorf("failed to parse node IP address %q", memberAddress)
-		}
+// GetLocalhostAddress returns the loopback address (IPv4 or IPv6) of the local machine.
+// It checks the loopback interface "lo" and returns "127.0.0.1" if available,
+// or "::1" if only IPv6 is available. If neither is found, it returns an error.
+func GetLocalhostAddress() (net.IP, error) {
+	iface, err := net.InterfaceByName("lo")
```
rapour, line 12: nit: I don't know if we ever come across a situation where the name of the interface could by anything other than "lo" but if you think that's a probabilty then we can switch to searching for interfaces with the loopback flag set.

src/k8s/pkg/k8sd/setup/etcd.go
```diff
@@ -0,0 +1,80 @@
...
+	listenUrls := []string{
+		fmt.Sprintf("https://%s", utils.JoinHostPort(nodeIP.String(), clientPort)),
+	}
+
+	localhostAddress, err := utils.GetLocalhostAddress()
+	if err != nil {
+		return fmt.Errorf("failed to get localhost address: %w", err)
+	}
+
+	if nodeIP != nil && !nodeIP.IsLoopback() {
```
rapour, line 24: Will it ever be the case that `nodeIP` is `nil`?

src/k8s/pkg/k8sd/setup/util_kubeconfig.go
```diff
@@ -75,3 +81,24 @@ func SetupControlPlaneKubeconfigs(kubeConfigDir string, localhostAddress string,
 	}
 	return nil
 }
+
```
rapour, line 84: nit:
```suggestion
// SetupWorkerKubeconfigs writes kubeconfig files for the worker node components.
```

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -96,6 +91,30 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+		})
+
+		certificates.CACert = cfg.Datastore.GetEtcdCACert()
+		certificates.CAKey = cfg.Datastore.GetEtcdCAKey()
+		certificates.ServerCert = joinConfig.GetEtcdServerCert()
+		certificates.ServerKey = joinConfig.GetEtcdServerKey()
+		certificates.ServerPeerCert = joinConfig.GetEtcdServerPeerCert()
+		certificates.ServerPeerKey = joinConfig.GetEtcdServerPeerKey()
+		certificates.APIServerClientCert = cfg.Datastore.GetEtcdAPIServerClientCert()
+		certificates.APIServerClientKey = cfg.Datastore.GetEtcdAPIServerClientKey()
```
rapour, line 110: Why are we treating the client certificates differently from other etcd certificates? Is there a reason we are storing them in the state instead of letting each control plane node create its own set of client certificates?

src/k8s/pkg/client/etcd/client.go
```diff
@@ -0,0 +1,68 @@
...
+
+	return &Client{
+		client,
+	}, nil
+}
+
+func newClient(pkiDir string, endpoints []string) (*clientv3.Client, error) {
+	certFile := fmt.Sprintf("%s/server.crt", pkiDir)
+	keyFile := fmt.Sprintf("%s/server.key", pkiDir)
+	caFile := fmt.Sprintf("%s/ca.crt", pkiDir)
```
rapour, line 18: Is it possible to switch to `apiserver-etcd-client.crt` for client calls? I'm thinking if user provides their own server certificates without `client auth` usage defined in its config, then the client call may fail.

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -208,6 +226,41 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+		if err != nil {
+			return fmt.Errorf("failed to get microcluster leader: %w", err)
+		}
+		members, err := leader.GetClusterMembers(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get microcluster members: %w", err)
+		}
+		initialClusterMembers := make(map[string]string)
+
+		endpoints := make([]string, 0, len(members))
```
rapour: nit:
```suggestion
		endpoints := make([]string, 0, len(members)-1)
```

rapour (commented):
Great work with lots of different parts put together neatly @berkayoz, left a few questions

src/k8s/pkg/utils/localhost.go
berkayoz: I believe this is a hard coded default in the kernel, I'd say it is safe to assume this interface name.

src/k8s/pkg/k8sd/pki/etcd.go
berkayoz: Good point, updated the CA handling.

src/k8s/pkg/k8sd/setup/etcd.go
berkayoz: Under normal circumstances we don't expect it to, this is a pre-emptive check so we don't panic

src/k8s/pkg/k8sd/app/hooks_join.go
berkayoz: I don't see a reason for all apiserver instances to have a different client certificate. It is easier to orchestrate a single client certificate for the entire kube-apiserver stack.

src/k8s/pkg/client/etcd/client.go
berkayoz: Isn't what you're suggesting reverse of what should solve this problem? If client auth is "disabled" we won't have a "client" cert like the one for kube-apiserver then doesn't it only leave us with the etcd node/server certificates? I think regardless this is easier to orchestrate through the etcd pki dir.

src/k8s/pkg/k8sd/setup/etcd.go
rapour: Then would it make sense to add similar checks for `nodeIP.String()`s in this function? It's only relevant if a nil nodeIP is a possibility

src/k8s/pkg/k8sd/app/hooks_join.go
rapour: Makes sense, thank you

src/k8s/pkg/client/etcd/client.go
rapour: My understanding is that the `server.crt` is used as a client certificate here to make API calls and by default is created by us with `client auth` EKU set. Now what happens if user provides an externally managed server certificate without `client auth` set instead of relying on us to procure it. In that case the API calls by this client should fail. I suppose this is a valid assumption since `server.crt` is a server certificate by nature and it's normal for an external CA to not set the `client auth` for it. Do you think that's a possibility?

src/k8s/pkg/client/etcd/client.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create etcd client: %w", err)
+	}
+
+	return &Client{
+		client,
+	}, nil
+}
+
+func newClient(pkiDir string, endpoints []string) (*clientv3.Client, error) {
```
bschimke95: `NewClient` and `newClient` feels weird. Why not one function - what do we gain from this abstraction?

src/k8s/pkg/client/etcd/client.go
```diff
@@ -0,0 +1,68 @@
...
+				break
+			}
+		}
+	}
+
+	if memberID == 0 {
+		return fmt.Errorf("no etcd member found with peer URL: %s", peerURL)
+	}
+
+	// Remove the member
```
bschimke95: nit: Unnecessary comment

src/k8s/pkg/client/etcd/tls.go
bschimke95, line 1: This would better fit in `src/k8s/pkg/utils/pki`

src/k8s/pkg/k8sd/api/certificates_refresh.go
```diff
@@ -474,20 +466,9 @@ func refreshCertsRunWorker(s state.State, r *http.Request, snap snap.Snap) respo
 		return response.InternalError(fmt.Errorf("failed to parse node IP address %q", s.Address().Hostname()))
 	}
 
-	var localhostAddress string
```
bschimke95, line 477: Are you sure that we can remove this - and even if we can I don't think this should be part of this PR, should it?

src/k8s/pkg/k8sd/api/extensions.go
```diff
@@ -0,0 +1,3 @@
+package api
+
+var Extensions = []string{"k8s_datastore_etcd"}
```
bschimke95: IIRC we cannot do this API extension approach, right?

src/k8s/pkg/k8sd/pki/etcd.go
```diff
@@ -0,0 +1,147 @@
...
+		notBefore:         opts.NotBefore,
+		ipSANs:            opts.IPSANs,
+		dnsSANs:           opts.DNSSANs,
+	}
+}
+
+// CompleteCertificates generates missing or unset certificates. If only a certificate is set and not a key, we assume that the cluster is using managed certificates.
+func (c *EtcdPKI) CompleteCertificates() error {
+	// Fail hard if keys of self-signed certificates are set without the respective certificates
+	switch {
```
bschimke95: please address gocritic

src/k8s/pkg/k8sd/pki/etcd_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/canonical/k8s/pkg/k8sd/pki"
+	. "github.com/onsi/gomega"
+)
+
+func TestEtcdPKI(t *testing.T) {
```
bschimke95: Should we have the same test coverage as in `k8sdqlite_test.go`?

src/k8s/pkg/client/etcd/client.go
berkayoz: The `server.crt` and `server.key` is the pair used for serving the https API endpoint of the node, it should not be `client auth` related. Additionally I don't think we have to support disabling of client auth.

src/k8s/pkg/k8sd/api/certificates_refresh.go
berkayoz: The first commit in this PR is the refactor so that localhost address handling is simplified. For `SetupControlPlaneKubeconfigs` and `SetupWorkerKubeconfigs` the localhost address is not passed as a parameter anymore instead we perform `localhostAddress, err := utils.GetLocalhostAddress()`.

This refactoring simplifies things considerably. Do you mean to have a separate PR for this refactoring?

bschimke95: Discussed offline - it is fine to keep it here.

tests/integration/tests/test_k8s_dqlite.py
bschimke95, line 1: that one seems obsolete and is now in `test_clustering_dqlite`

PR #1564, opened 2025-06-17
fix: Decide same helm values after merge

HomayoonAlimohammadi:
### Overview

When deciding whether or not we're trying to apply the same values in the helm client, we forgot that we might be doing full comparisons for maps, one of which is a subset of the other. This PR merges the values provided by the user with old release config first, and then performs this comparison. 
The exact same process happens in the helm code as well. But helm proceeds with the upgrade regardless of changes.


src/k8s/pkg/client/helm/client.go
```diff
@@ -117,7 +119,9 @@ func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, v
 		// NOTE(Angelos): oldConfig and values are the previous and current values. they are compared by checking their respective JSON, as that is good enough for our needs of comparing unstructured map[string]any data.
 		// NOTE(Hue) (KU-3592): We are ignoring the values that are overwritten by the user.
 		// The user can change some values in the chart, but we will revert them back upon an upgrade.
-		sameValues := jsonEqual(oldConfig, values)
+		// NOTE(Hue): We clone the values to avoid modifying the original values map passed by the caller.
+		mergedValues := chartutil.CoalesceTables(maps.Clone(values), oldConfig)
```
berkayoz: I believe `maps.Clone` is [shallow](https://pkg.go.dev/maps#Clone), which means sub maps, slices and other references will be assigned ordinarily. Won't we need some sort of a deep clone?

HomayoonAlimohammadi: I noticed that since it's a `map[string]any`, a _real_ deep copy can be very complicated (we need a recursive map copy that takes care of all map types, which should probably use reflection). Cloning and copying the map here is not important, so I'm just ignoring it. Are you okay with removing the clone?

berkayoz: If we're sure `values` (dst) is not affected by the `CoalesceTables` call I'm fine with skipping it. For the original discussion the usual approach I see for deep clone is to `json.Marshal` and then `json.Unmarshal`, not the best performance but simple and good enough.

tests/integration/tests/test_user_config.py
```diff
@@ -63,3 +66,85 @@ def test_user_config(instances: List[harness.Instance]):
...
+
+    # Get initial revision of ck-network chart
+    initial_revision = get_helm_release_revision("ck-network", "kube-system")
+    LOG.info("Initial ck-network chart revision: %s", initial_revision)
+
+    # Call 'k8s set' with the same configuration multiple times
+    for i in range(3):
+        LOG.info("Running k8s set command (iteration %d)", i + 1)
+        instance.exec(
+            "k8s set network.enabled=true gateway.enabled=true ingress.enabled=true".split()
```
berkayoz: Any reason to not do this for all the features? If not I feel like it's best we test it all.

src/k8s/pkg/client/helm/client.go
HomayoonAlimohammadi: The values _will_ get changed by the `CoalesceTables`. The return value is just basically the changed `values` (I think it's also returned just for convenience). I'll implement the approach with Marshal, thanks for mentioning it.

tests/integration/tests/test_user_config.py
```diff
@@ -63,3 +66,85 @@ def test_user_config(instances: List[harness.Instance]):
...
+        )
+
+        # NOTE(Hue): I can't think of a better way to make sure the features are reconciled.
+        sleep(10)
```
berkayoz: Good point, since we don't have a way to tell should we just do more iterations? Like 7 tries, even if this `sleep` misses, we'll catch release differences in the next iterations

We could check for logs, e.g. `no changes detected, skipping upgrade` exists or check other logs do not but this can get messy.

tests/integration/tests/test_user_config.py
```diff
@@ -63,3 +66,85 @@ def test_user_config(instances: List[harness.Instance]):
...
+        def check_features_enabled(p: CompletedProcess):
+            status = json.loads(p.stdout.strip())
+            return (
+                status.get("network", {}).get("enabled") is True
+                and status.get("gateway", {}).get("enabled") is True
+                and status.get("ingress", {}).get("enabled") is True
+            )
+
+        util.stubbornly(retries=10, delay_s=2).until(check_features_enabled).exec(
+            ["k8s", "status"],
```
berkayoz: Since we have `json.loads` in `check_features_enabled`, shouldn't this be `k8s status --output-format json` ?

berkayoz (commented):
Great work, LGTM overall left some more comments

tests/integration/tests/test_user_config.py
HomayoonAlimohammadi: I agree. I've increased the number of iterations. I think checking the logs is a bit too much maybe?

src/k8s/pkg/client/helm/client.go
berkayoz: After evaluation for the performance hit vs an existing library, see https://gist.github.com/berkayoz/a9de6d477bc0048d486dc168826826c8

We're observing the following
```
goos: linux
goarch: amd64
pkg: github.com/berkayoz/test-go
cpu: AMD Ryzen 9 9950X3D 16-Core Processor          
BenchmarkCopystructureCopy-32              34914         33129 ns/op       27488 B/op         857 allocs/op
BenchmarkJSONMarshalUnmarshal-32           87315         13947 ns/op       15651 B/op         347 allocs/op
BenchmarkMapsCopy-32                    13968646            85.31 ns/op           0 B/op           0 allocs/op
PASS
ok      github.com/berkayoz/test-go    4.145s
```

ktsakalozos (approved):
LGTM +1

PR #1576, opened 2025-06-19
fix: retry the wait command for load balancer

rapour:
## Description

In the `_test_loadbalancer` we are waiting for `metallb-controller` deployment to become available immediately after we enable the load-balancer. This results in flakiness since sometimes we run the `kubectl wait` command before the deployment object is created:
```
Error from server (NotFound): deployments.apps "metallb-controller" not found
```
This PR suggests adding a retry logic to account for such scenario

### DIFF
--- a/tests/integration/tests/test_util/util.py
+++ b/tests/integration/tests/test_util/util.py
@@ -412,8 +412,8 @@ def wait_for_network(instance: harness.Instance):
 def wait_for_load_balancer(instance: harness.Instance):
     """Wait for the load balancer to be ready."""
     LOG.info("Waiting for load balancer to be ready")
-    instance.exec(
-        [
+    stubbornly(retries=3, delay_s=1).on(instance).exec(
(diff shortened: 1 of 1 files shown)


tests/integration/tests/test_util/util.py
```diff
@@ -412,7 +412,7 @@ def wait_for_network(instance: harness.Instance):
 def wait_for_load_balancer(instance: harness.Instance):
     """Wait for the load balancer to be ready."""
     LOG.info("Waiting for load balancer to be ready")
-    instance.exec(
```
ktsakalozos, line 415: The delay of 1 sec may also not be enough. Could you please make it 5 seconds just to be on the safe side?

rapour: Done, thank you

ktsakalozos (approved):
LGTM +1

PR #1618, opened 2025-07-03
fix: Make helm operations atomic

HomayoonAlimohammadi:
### Overview

This PR makes helm operations atomic so that Helm will wait for the result and rolls back instead of a failure. Uninstall is only waitable and can not be rolled back.
We also add a timeout to helm commands to wait for.
Changed the exponential backoff of feature reconciliation retries to linear backoff to make reconciliations happen faster. Exponential backoffs could have caused the cluster to reach a stable state much slower.

### DIFF
--- a/tests/integration/tests/test_smoke.py
+++ b/tests/integration/tests/test_smoke.py
@@ -5,9 +5,9 @@ import json
 import logging
 import re
 import subprocess
-from typing import List
 import time
 from pathlib import Path
+from typing import List
 
 import pytest
 from test_util import config, harness, tags, util


src/k8s/pkg/client/helm/client.go
```diff
@@ -96,6 +96,8 @@ func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, v
 	case !isInstalled && desired == StatePresent:
 		// there is no release installed, so we must run an install action
 		install := action.NewInstall(cfg)
+		install.Atomic = true
+		install.Timeout = h.timeout
```
bschimke95, line 97: What is the default timeout that we set for the client? 
The helm apply will also wait for e.g. the image pull, right? So on a slow network the user might always run into a timeout for large images (e.g. cilium) and never actually succeed if the timeout is set too low.

src/k8s/pkg/utils/time/backoff.go
```diff
@@ -5,13 +5,16 @@ import (
 	"time"
 )
 
-// ExponentialBackoff returns a duration that grows exponentially with attempt number,
-// capped at maxDelay, and with ±50% jitter.
-func ExponentialBackoff(attempt int, baseDelay, maxDelay time.Duration) time.Duration {
-	exp := min(max(0, attempt), 32) // cap at 32 to avoid overflow
-	// exponential: base * 2^exp
-	d := min(baseDelay<<exp, maxDelay)
+var randInt64N = rand.Int64N
```
bschimke95, line 8: why do you set this global variable?

HomayoonAlimohammadi: So that I can change it in the tests. See the first line of the test function. Is there a better way to achieve testability?

src/k8s/pkg/client/helm/client.go
HomayoonAlimohammadi: Hmm, this is a good point. Let me investigate this a bit further.

HomayoonAlimohammadi: I can't see any defaults being set in the Helm client code... The [cli defaults this to 5 minutes](https://github.com/helm/helm/blob/main/pkg/cmd/install.go#L198) tho.

HomayoonAlimohammadi: We're also setting the same timeout on our client: https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/snap/snap.go#L376

bschimke95: Hm, I guess if we don't provide a way to override this timeout then my point stand, does it? 5min *seems* long enough but it may not in some rare cases.

src/k8s/pkg/utils/time/backoff.go
bschimke95: you could do something like this and introduce an (optional) opts struct

```
type BackoffOpts {
   strategy Linear/Exonential
   randFunc func(){}
   ...
}

func Backoff(attempt, baseDelay, ...opts BackoffOpts) time.Duration{}
```

tests/integration/tests/test_smoke.py
```diff
@@ -139,6 +139,6 @@ def status_output_matches(p: subprocess.CompletedProcess) -> bool:
         return True
 
     LOG.info("Verifying the output of `k8s status`")
-    util.stubbornly(retries=15, delay_s=10).on(instance).until(
+    util.stubbornly(retries=100, delay_s=10).on(instance).until(
```
bschimke95: huh? Where does this come from?

src/k8s/pkg/client/helm/client.go
louiseschmidtgen: +1 can we make the timeout configurable?

src/k8s/pkg/client/helm/client.go
```diff
@@ -146,6 +148,8 @@ func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, v
 
 		// there is already a release installed, so we must run an upgrade action
 		upgrade := action.NewUpgrade(cfg)
+		upgrade.Atomic = true
+		upgrade.Timeout = h.timeout
```
louiseschmidtgen, line 149: Are all upgrades going to be atomic? Do we need this option or will it always be true?

src/k8s/pkg/client/helm/client.go
```diff
@@ -22,8 +22,8 @@ import (
 type client struct {
 	restClientGetter func(string) genericclioptions.RESTClientGetter
 	manifestsBaseDir string
-	// applyTimeout is the timeout for apply operations.
-	applyTimeout time.Duration
+	// timeout for the helm operations.
+	timeout time.Duration
```
louiseschmidtgen:
```suggestion
	helmTimeout time.Duration
```
Let's remove the comment and rename the variable to provide the info.

src/k8s/pkg/client/helm/client.go
```diff
@@ -160,6 +164,8 @@ func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, v
 	case isInstalled && desired == StateDeleted:
 		// run an uninstall action
 		uninstall := action.NewUninstall(cfg)
+		uninstall.Timeout = h.timeout
+		uninstall.Wait = true
```
louiseschmidtgen, line 165: Where does this get used?

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -260,9 +260,10 @@ func (c *FeatureController) reconcileLoop(
 					continue
 				}
 
-				log.Info("Retrying feature reconciliation", "attempts", fmt.Sprintf("%d/%s", attempts, maxAttempts))
-				// notify triggerCh after 3-15 seconds to retry
-				time.AfterFunc(timeutils.ExponentialBackoff(attempts, 3*time.Second, 5*time.Minute), func() { utils.MaybeNotify(triggerCh) })
+				retryAfter := timeutils.LinearBackoff(attempts, 5*time.Second, 1*time.Minute)
+				log.Info(fmt.Sprintf("Retrying feature reconciliation in %f seconds", retryAfter.Seconds()), "attempts", fmt.Sprintf("%d/%s", attempts, maxAttempts))
```
louiseschmidtgen: This sounds like it will blow up our logs. Can we make it a debug log or only log the last retry?

src/k8s/pkg/utils/time/backoff.go
```diff
@@ -5,13 +5,16 @@ import (
...
-// ExponentialBackoff returns a duration that grows exponentially with attempt number,
+var randInt64N = rand.Int64N
+
+// LinearBackoff returns a duration that grows linearly with attempt number,
 // capped at maxDelay, and with ±50% jitter.
-func ExponentialBackoff(attempt int, baseDelay, maxDelay time.Duration) time.Duration {
-	exp := min(max(0, attempt), 32) // cap at 32 to avoid overflow
-	// exponential: base * 2^exp
-	d := min(baseDelay<<exp, maxDelay)
+func LinearBackoff(attempt int, baseDelay, maxDelay time.Duration) time.Duration {
```
louiseschmidtgen, line 12: Do we need a util? Why are we not using https://pkg.go.dev/github.com/cenkalti/backoff/v4@v4.3.0#ConstantBackOff?

tests/integration/tests/test_smoke.py
HomayoonAlimohammadi: This was for testing purposes only to see if the tests pass. I changed the PR to draft.

src/k8s/pkg/utils/time/backoff.go
HomayoonAlimohammadi: Good idea. It's just that I've seen similar patterns before in grpc-go (IIRC). I'll change this in the final version of the PR.

PR #1619, opened 2025-07-04
feat: system limit tuning

louiseschmidtgen:
## Description

An issue surfaced in an internal Kubernetes environment showing errors in kube-apiserver-proxy related to reaching max_user_watches.

## Solution

We should bump the systemd limits on the snap to allow for a seamless experience for our k8s.
When the snap is removed we should revert the adjustments made. I am choosing the same limits that worked for microk8s.

### Details
A persistent file xxx-k8s.config is created at the highest order, overwriting system settings:
```
ubuntu@louise-dev2:~/k8s-snap$ sudo ls /etc/sysctl.d/
10-console-messages.conf  10-magic-sysrq.conf       10-zeropage.conf       99-sysctl.conf
10-ipv6-privacy.conf      10-network-security.conf  100-k8s.conf           99-zzz-override_cilium.conf
10-kernel-hardening.conf  10-ptrace.conf            99-cloudimg-ipv6.conf  README.sysctl
```

Tuning parameters are added to the file and applied if system settings are not sufficient (on the node joining the cluster and on k8sd start):
```
ubuntu@louise-dev2:~/k8s-snap$ sudo cat /etc/sysctl.d/100-k8s.conf 
# Auto-generated by the k8s snap for system tuning.
# Delete this file to disable auto-tuning.
# Overrides defaults via 'sysctl --system' on k8sd service start.
fs.inotify.max_user_instances=1024
fs.inotify.max_user_watches=1048576
```

System settings have been applied:
```
ubuntu@louise-dev2:~/k8s-snap$ sysctl fs.inotify.max_user_watches
fs.inotify.max_user_watches = 1048576
ubuntu@louise-dev2:~/k8s-snap$ sysctl fs.inotify.max_user_instances
fs.inotify.max_user_instances = 1024
```

Upon snap removal the file is removed and the system settings are re-applied.

#### DisableSystemTuning flag:

The disable-system-tuning flag can be specified on bootstrap via the config or in the worker/control-plane join configuration. If the flag is set no tuning file xxx-k8s.conf will be created. If the file is added manually at a later time the system will perform auto-tuning. If the file is removed at a later time the system stops auto-tuning.

## Issue

related to https://github.com/canonical/k8s-snap/pull/1558
N/A internal.

## Backport

Not planned

## Checklist

- [x] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed


src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -27,6 +28,8 @@ import (
 	"github.com/canonical/microcluster/v2/state"
 )
 
+const systemConfigPath = "/etc/sysctl.d/"
```
berkayoz: I'm not sure if bootstrap hook is the best place to define this.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -535,3 +558,37 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
...
+		return err
+	}
+	defer f.Close()
+
+	_, err = f.WriteString("# Auto-generated by the k8s snap for system tuning.\n# Delete this file to disable auto-tuning.\n# Overrides defaults via 'sysctl --system' on k8sd service start.\n")
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
berkayoz: I feel like these should be moved under setup, e.g. `setup_tuning.go`

src/k8s/pkg/utils/file.go
```diff
@@ -76,9 +78,216 @@ func ParseArgumentFile(path string) (map[string]string, error) {
...
+	for i := 0; i < len(lines); i++ {
+		if _, err := file.WriteString(lines[i] + "\n"); err != nil {
+			return fmt.Errorf("failed to write updated lines to configuration file %w", err)
+		}
+	}
+	return nil
+}
+
+// Deep copy of a map.
+func DeepCopyMap(original map[string]string) map[string]string {
```
berkayoz: AFAIK This is not a full deep copy. Sub structs, maps and slices might have the same reference. The approach we took in helm client was to json marshal and unmarshal.

src/k8s/pkg/utils/file.go
```diff
@@ -76,9 +78,216 @@ func ParseArgumentFile(path string) (map[string]string, error) {
...
+	for k, v := range original {
+		copied[k] = v
+	}
+	return copied
+}
+
+// GetFileMatch returns the path of the file in a dir matching the regex or "" if no
+// match was found.
+func GetFileMatch(path string, regex string) (string, error) {
+	re := regexp.MustCompile(regex)
```
berkayoz: It's usually more idiomatic(and more performant though negligible for this use case) to define a regex under the module scope and call `MustCompile` once there

k8s/lib.sh
```diff
@@ -74,6 +74,25 @@ k8s::remove::containerd() {
...
+# Cleanup systemd overrides
+k8s::remove::system_tuning() {
+  if ! k8s::common::is_strict; then
+    # Find files matching pattern: digits followed by '-k8s.conf'
+    files_to_remove=$(find /etc/sysctl.d/ -maxdepth 1 -type f -regextype posix-extended -regex '.*/[0-9]+-k8s\.conf')
+
+
+    if [ -n "$files_to_remove" ]; then
+      echo "Removing the following custom sysctl parameter files:"
+      echo "$files_to_remove"
```
bschimke95: Where will this be logged? 
IIRC `snap remove` will not print this.

src/k8s/pkg/utils/file.go
louiseschmidtgen: Ah great point! I will adjust :)

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -320,3 +332,38 @@ func (a *App) markNodeReady(ctx context.Context, s state.State) error {
...
+	}
+
+	// skip tuning if the tuning file does not exist
+	if confPath == "" {
+		log.Info(fmt.Sprintf("skipping system tuning: no sysctl config file %s", confPath))
+		return nil
+	}
+
+	// parse config files to check minimum requirements are set
+	newConfig := utils.MinConfigFileDiff(confLocs, minConfig)
```
bschimke95: As a side note, this implementation assumes that "minConfig" always means >minValue. I can totally scenarios where "minimum" requirements means <minValue.
I'm fine tackling that problem once we get there, though.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
bschimke95: This should be part of the snap interface

src/k8s/pkg/k8sd/app/app.go
```diff
@@ -17,12 +18,23 @@ import (
...
 	"github.com/canonical/k8s/pkg/log"
 	"github.com/canonical/k8s/pkg/snap"
+	"github.com/canonical/k8s/pkg/utils"
 	"github.com/canonical/k8s/pkg/utils/control"
 	"github.com/canonical/microcluster/v2/client"
 	"github.com/canonical/microcluster/v2/microcluster"
 	"github.com/canonical/microcluster/v2/state"
 )
 
+var (
```
bschimke95: Those configurations should be part of the snap interface so that we can properly mock them.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -535,3 +558,37 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
 	a.NotifyUpdateNodeConfigController()
 	return nil
 }
+
+func (a *App) createSystemTuningConfigFile(ctx context.Context) error {
```
bschimke95: Let's keep the naming consitent with the other setup steps and call this `ensureSystemTuningConfigFile` since we don't always create the file in this function.

Also, please move it as Berkay suggested.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -535,3 +558,37 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
...
+	if err != nil {
+		log.Error(err, "Failed to check for existing system tuning file")
+		return err
+	}
+	if confPath != "" {
+		return nil
+	}
+
+	// Find highest order file
+	// Match patterns like: 10-xyz.conf
```
bschimke95: nit: this comment explains an implementation detail and is already explained in the function. I would just remove it.

src/k8s/pkg/k8sd/app/hooks_bootstrap.go
```diff
@@ -535,3 +558,37 @@ func (a *App) onBootstrapControlPlane(ctx context.Context, s state.State, bootst
...
+	maxOrder, err := utils.GetHighestConfigFileOrder(systemConfigPath)
+	if err != nil {
+		return err
+	}
+
+	f, err := os.Create(filepath.Join(systemConfigPath, fmt.Sprintf("%d-k8s.conf", maxOrder+1)))
```
bschimke95: We have `utils.WriteFile` for creating and writing files. The "partial writes" issue that this function solves is not necessarily required here but it shortens the create/write logic a bit.

src/k8s/pkg/utils/file.go
```diff
@@ -76,9 +78,216 @@ func ParseArgumentFile(path string) (map[string]string, error) {
 			args[a] = v
 		}
 	}
+
 	return args, nil
 }
 
+// ParseConfigFile parses a configuration file and returns a map of key-value pairs.
+// The file is expected to have lines in the format "key=value". Comments (lines starting with '#') are ignored.
+func ParseConfigFile(path string) (map[string]string, error) {
```
bschimke95: We already have `util.ParseArgumentFile` which looks like the same implementation without a "ignore comments" part. Let's just reuse that function and add the comments part there, wdyt?

src/k8s/pkg/utils/file.go
```diff
@@ -76,9 +78,216 @@ func ParseArgumentFile(path string) (map[string]string, error) {
...
+		}
+	}
+
+	return newConfig
+}
+
+// UpdateConfigFile takes a map with new configurations in the format "key=value" and
+// adjusts the file to reflect the new configuration. The file is expected to have
+// lines in the format "key=value". Comments (lines starting with '#') are ignored.
+func UpdateConfigFile(path string, newConfig map[string]string) error {
```
bschimke95: I feel like reading the whole config file, updating the required values and then writing the whole file is more idiomatic then this. 
Right now, this util is basically not reusable and only fits one purpose which means it does not fit into the utils package.
With the above approach, you could just use the existing `ParseArgumentFile` and `SerializeArgumentFile` (maybe with minor tweaks) and we would not need those functions at all. WDYT?

src/k8s/pkg/utils/file.go
louiseschmidtgen: Here, I create a distinct new map and the values are copied there without a reference, there is a test case in file_test.go to prove this.
Ah yes, I [found it](https://github.com/canonical/k8s-snap/blob/af0b3e1980e89b3d7bc63005a37b1add6a93ade3/src/k8s/pkg/client/helm/client.go#L181) `cloneMap(m map[string]any)`, however this function adds additional performance overhead with the json marshal and unmarshal that is not necessary for this case. Of course it is nice to streamline the util and use one for all.
For what it is worth, I see [helm uses a buffer for their deep-copy](https://github.com/helm/helm/pull/6661/files).
Let me know what you prefer, I am happy to adjust.

bschimke95: This.

As a side note, this is not a file util operation and would probably better fit in another utils file.

bschimke95: Also please stick to Godoc conventions

```
// DeepCopyMap creates ...
func DeepCopyMap()
```

src/k8s/pkg/utils/file.go
```diff
@@ -76,9 +78,216 @@ func ParseArgumentFile(path string) (map[string]string, error) {
...
+	copied := make(map[string]string, len(original))
+	for k, v := range original {
+		copied[k] = v
+	}
+	return copied
+}
+
+// GetFileMatch returns the path of the file in a dir matching the regex or "" if no
+// match was found.
+func GetFileMatch(path string, regex string) (string, error) {
```
bschimke95: What does this function do if multiple files match the pattern?
Does it follow directories?

src/k8s/pkg/utils/file.go
```diff
@@ -76,9 +78,216 @@ func ParseArgumentFile(path string) (map[string]string, error) {
...
+		if match := re.FindString(entry.Name()); match != "" {
+			return filepath.Join(path, match), nil
+		}
+	}
+	return "", nil
+}
+
+// GetFileMatch returns the path of the file in a dir matching a file like 10-xyz.conf
+// using the regex `^(\d+)-.*\.conf$` or returns 0s if no match was found.
+func GetHighestConfigFileOrder(path string) (int, error) {
```
bschimke95: This is not really a file util. Please move this specific logic to the `setup_tuning.go` file.

bschimke95 (commented):
Did a first pass - great work @louiseschmidtgen !

src/k8s/pkg/utils/file.go
louiseschmidtgen: That is true, the only issue is keeping the comments in place. The map does not hold our comments which is the beginning comment or any commented out keys. If we scrap all comments we can use it without any problems. However, in our spec we said that we would like comments.
Wdyt?

src/k8s/pkg/utils/file.go
louiseschmidtgen: We are using the first match: "FindString returns a string holding the text of the leftmost match in s of the regular expression." 
Do you think we should have different behavior? We are not creating more than one k8s.config file with our automation, so if there are two that would be the users fault.

k8s/lib.sh
louiseschmidtgen: I removed them :)

src/k8s/pkg/k8sd/app/app.go
louiseschmidtgen: That's right we could extend the function if we need this functionality in the future.

src/k8s/pkg/utils/file.go
bschimke95: How about introducing an optional config?

```
type SerializeArgumentFileOpts struct {
   HeaderComment string
   <possible other opts>
}

func SerializeArgFile(arg1, args2, ...opts) {
}
```

louiseschmidtgen: How do we know where the comment goes in the file? Unless we only support header comments and no commented out keys later on:
```
# my beautiful header comment
key1=value1
# key2=value2
key3=value3
```

bschimke95: I don't think this is a use-case and would even add something like "AUTOGENERATED - DO NOT EDIT" as part of the top. If users really want to override stuff then they should create a new file with higher prio.

Anyway, if you feel we should keep that separated then keep that function but I would at least move it to a more appropriate (non-generic) location.

src/k8s/pkg/utils/file.go
bschimke95: I think we need to update this function but I leave the solution to you:
* rename the function and update the comment to make clear that this only returns the first match
* Return a list of strings/matches and update the name/comment accordingly (I'd prefer this)
* Move this specific function to a non-generic place.

src/k8s/pkg/utils/file.go
louiseschmidtgen: I'm happy with adding `AUTOGENERATED - DO NOT EDIT`, then people can't expect any modifications to stay :)

bschimke95 (approved):
thanks for the changes - lgtm

PR #1624, opened 2025-07-07
test: charm-e2e tests collect the juju crashdumps

addyess:
## Description

Adding juju-crashdump the the charm-e2e tests

## Solution

Running `tox -e integration` will create a crashdump of the model under test
Adding `juju-crashdump` for the controller is nice
Finally assemble all the crashdumps into a github artifact of tht ebuild 

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/.github/workflows/charm-e2e-tests.yaml
+++ b/.github/workflows/charm-e2e-tests.yaml
@@ -26,6 +26,8 @@ jobs:
   charm-e2e-tests:
     name: Test ${{ inputs.arch }} ${{ inputs.artifact }}
     runs-on: ${{ inputs.arch == 'arm64' && 'self-hosted-linux-arm64-jammy-large' || 'self-hosted-linux-amd64-jammy-xlarge' }}
+    env:
+      K8S_OPERATOR: k8s-operator
     steps:
       - name: Check out code
         uses: actions/checkout@v4
@@ -48,12 +50,12 @@ jobs:
           mkdir snap_installation
           mv ${{ steps.download-snap.outputs.snap-path }} snap_installation/k8s.snap
           tar cvzf snap_installation.tar.gz -C snap_installation/ .
-      - name: Install charmcraft
-        run: |
-          sudo snap install charmcraft --classic
       - name: Install juju
         run: |
           sudo snap install juju --classic
+      - name: Install juju-crashdump
+        run: |
+          sudo snap install juju-crashdump --classic
       - name: Download charms ${{ inputs.charm-channel }} ${{ inputs.arch }}
         run: |
           # The operator tests expect the charm file name to have the
@@ -81,7 +83,7 @@ jobs:
         with:
           repository: canonical/k8s-operator
           ref: main
-          path: k8s-operator
+          path: ${{ env.K8S_OPERATOR }}
       - name: Disable lxd ipv6
         run: |
           # The charm tests can't handle ipv6 dualstack.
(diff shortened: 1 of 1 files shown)


.github/workflows/charm-e2e-tests.yaml
```diff
@@ -48,12 +50,12 @@ jobs:
...
           tar cvzf snap_installation.tar.gz -C snap_installation/ .
-      - name: Install charmcraft
-        run: |
-          sudo snap install charmcraft --classic
       - name: Install juju
         run: |
           sudo snap install juju --classic
+      - name: Install juju-crashdump
+        run: |
+          sudo snap install juju-crashdump --classic
```
bschimke95, line 57: Do we need to pin the version or does juju-crashdump work with all juju versions?

.github/workflows/charm-e2e-tests.yaml
```diff
@@ -48,12 +50,12 @@ jobs:
           mkdir snap_installation
           mv ${{ steps.download-snap.outputs.snap-path }} snap_installation/k8s.snap
           tar cvzf snap_installation.tar.gz -C snap_installation/ .
-      - name: Install charmcraft
-        run: |
-          sudo snap install charmcraft --classic
```
bschimke95, line 53: I guess this is just a cleanup of obsolete code?

.github/workflows/charm-e2e-tests.yaml
```diff
@@ -107,3 +110,20 @@ jobs:
             --charm-file $k8sWorkerChamFile \
             --snap-installation-resource $snapInstallRes \
             'tests/integration/test_k8s.py::test_nodes_ready'
+      - name: Collect Juju Status
+        if: ${{ failure() }}
```
bschimke95: nit: You can remove the ${{ }}} - `failure()` should be enough afaik

.github/workflows/charm-e2e-tests.yaml
```diff
@@ -107,3 +110,20 @@ jobs:
...
+        id: collect-juju-status
+        run: |
+          set -x
+          path=${{ env.K8S_OPERATOR }}/tmp
+          mkdir -p $path
+          juju status 2>&1 | tee $path/juju-status.txt
+          juju-crashdump -s -m controller -a debug-layer -a config -o $path/
+          mv ${{ env.K8S_OPERATOR }}/juju-crashdump-* $path/ | true
```
bschimke95: I don't think you want to use the pipe operator `|` here

bschimke95 (approved):
Minor comments, LGTM otherwise

.github/workflows/charm-e2e-tests.yaml
addyess: juju-crashdump hasn't changed in about 2 years... seems to be holding.  And it doesn't have channels.  So it's either this or pinned revision

.github/workflows/charm-e2e-tests.yaml
addyess: JK... turns out its needed just b/c pytest-operator checks to make sure its there.  It really ISN'T needed but its a runtime dependency

.github/workflows/charm-e2e-tests.yaml
addyess: nice, thanks.  This is a copy and paste from elsewhere -- which means you caught two bugs 🎉

.github/workflows/charm-e2e-tests.yaml
bschimke95: TIL

PR #1635, opened 2025-07-10
fix(feature_controller): Fix features getting stuck 

HomayoonAlimohammadi:
### Overview

This PR prevents features from getting stuck. Helm operations might error with `another operation is in progress`  if the API server call from Helm gets interrupted. This PR prevents that by the following:
1. Introducing a lock for the features that operate on the same chart (`network`, `gateway` and `ingress` operate on `ck-network` chart).
2. Upon a Helm operation, if the status is `pending-*`, we will change it to `failed` so that future operations are not blocked by this pending state.
3. Introdues an E2E test to ensure features won't get stuck by API server interruptions.
4. Helm operations have an internal `Timeout` attribute. Seems like operation timeout is based on this attribute rather than the context timeout. This PR populates the `Timeout` attribute instead of passing a context with timeout.


src/k8s/pkg/client/helm/client.go
```diff
@@ -86,6 +84,18 @@ func (h *client) Apply(ctx context.Context, c InstallableChart, desired State, v
...
+	if isInstalled && release.Info.Status.IsPending() {
+		s := releasepkg.StatusFailed
+		log.Info("release is in a pending state, changing status", "status", release.Info.Status, "chart", c.Name, "target_status", s)
+
+		release.Info.Status = s
+		if err := cfg.Releases.Update(release); err != nil {
+			return false, fmt.Errorf("failed to update release %s status: %w", c.Name, err)
+		}
+	}
+
```
berkayoz: Can we add a comment for why we're setting this to failed related to ensuring helm re-applies the manifests where this wouldn't be the case in another approach such as deleting the secret?

berkayoz (approved):
LGTM, one small comment. Thanks a lot!

src/k8s/pkg/k8sd/controllers/feature.go
```diff
@@ -139,14 +142,20 @@ func (c *FeatureController) Run(
 	s := getState()
 
 	go c.reconcileLoop(ctx, getClusterConfig, setFeatureStatus, features.Network, c.triggerNetworkCh, c.reconciledNetworkCh, func(cfg types.ClusterConfig) (types.FeatureStatus, error) {
+		c.ciliumLock.Lock()
+		defer c.ciliumLock.Unlock()
```
berkayoz, line 152: Maybe a comment on why we're locking here would be nice as well. e.g. to ensure we don't actually mess up with an on-going operation for `ck-network`

src/k8s/pkg/client/helm/client.go
HomayoonAlimohammadi: Good point. Done

src/k8s/pkg/k8sd/controllers/feature.go
HomayoonAlimohammadi: Yea makes sense. Done!

PR #1643, opened 2025-07-14
feat(upgrade): Wait for worker nodes to get upgraded as well

HomayoonAlimohammadi:
### Overview

This PR implements a way to wait for the worker nodes in the upgrade process as well, before triggering the feature controllers.

### Implementation
We've introduced a `version.Data` struct that contains all the necessary information about the current version of the installed k8s-snap. The node label controller is responsible for annotating the node with this struct.
The upgrade controller will watch for this annotation and add the node name to the list of `upgradedNodes`.
Once all the nodes (including workers) are available in the `upgradedNodes` list, the upgrade controller will transition the upgrade phase from `NodeUpgrade` to `FeatureUpgrade`, triggering the feature controllers and proceeding with the upgrade.

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_join.go
+++ b/src/k8s/pkg/k8sd/app/hooks_join.go
@@ -15,9 +15,11 @@ import (
 	"github.com/canonical/k8s/pkg/k8sd/types"
 	"github.com/canonical/k8s/pkg/log"
 	"github.com/canonical/k8s/pkg/snap"
+	upgradepkg "github.com/canonical/k8s/pkg/upgrade"
 	"github.com/canonical/k8s/pkg/utils"
 	"github.com/canonical/k8s/pkg/utils/control"
 	"github.com/canonical/k8s/pkg/utils/experimental/snapdconfig"
+	"github.com/canonical/k8s/pkg/version"
 	"github.com/canonical/microcluster/v2/state"
 	versionutil "k8s.io/apimachinery/pkg/util/version"
 )
@@ -422,7 +424,8 @@ func initiateRollingUpgrade(ctx context.Context, snap snap.Snap, s state.State, 
 		strategy = upgradesv1alpha.UpgradeStrategyRollingDowngrade
 	}
 
-	newUpgrade := upgradesv1alpha.NewUpgrade(fmt.Sprintf("cluster-upgrade-to-rev-%s", rev))
+	versionData := version.Info{Revision: rev}
+	newUpgrade := upgradesv1alpha.NewUpgrade(upgradepkg.GetName(versionData))
 	if err := k8sClient.Create(ctx, newUpgrade); err != nil {
 		return fmt.Errorf("failed to create upgrade: %w", err)
 	}
--- a/src/k8s/pkg/k8sd/controllers/node_label.go
+++ b/src/k8s/pkg/k8sd/controllers/node_label.go
@@ -6,10 +6,14 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/canonical/k8s/pkg/client/kubernetes"
 	"github.com/canonical/k8s/pkg/log"
 	"github.com/canonical/k8s/pkg/snap"
 	snaputil "github.com/canonical/k8s/pkg/snap/util"
+	"github.com/canonical/k8s/pkg/version"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
 )
 
 type NodeLabelController struct {
@@ -52,7 +56,7 @@ func (c *NodeLabelController) Run(ctx context.Context, getDatastoreType func(ctx
 
 		if err := client.WatchNode(
 			ctx, nodeName, func(node *v1.Node) error {
-				err := c.reconcile(ctx, node, getDatastoreType)
+				err := c.reconcile(ctx, client, node, getDatastoreType)
 				c.notifyReconciled()
 				return err
 			}); err != nil {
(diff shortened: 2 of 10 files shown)


src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -15,9 +15,11 @@ import (
 	"github.com/canonical/k8s/pkg/k8sd/types"
 	"github.com/canonical/k8s/pkg/log"
 	"github.com/canonical/k8s/pkg/snap"
+	upgradepkg "github.com/canonical/k8s/pkg/upgrade"
```
bschimke95, line 18: Is there a need to rename this?

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -422,7 +424,8 @@ func initiateRollingUpgrade(ctx context.Context, snap snap.Snap, s state.State,
 		strategy = upgradesv1alpha.UpgradeStrategyRollingDowngrade
 	}
 
-	newUpgrade := upgradesv1alpha.NewUpgrade(fmt.Sprintf("cluster-upgrade-to-rev-%s", rev))
+	versionData := version.Data{Revision: rev}
```
bschimke95: super small nit: I feel like `version.Info` would fit better than Data but feel free to ignore this if you think otherwise

src/k8s/pkg/k8sd/app/hooks_join.go
bschimke95: ah nevermind - it makes sense below

src/k8s/pkg/k8sd/controllers/node_label.go
```diff
@@ -131,12 +165,39 @@ func (c *NodeLabelController) updateDqliteFailureDomain(ctx context.Context, fai
...
+	wg.Add(1)
+	go func() {
+		log.FromContext(ctx).Info("Reconciling node version annotation", "nodeName", node.Name)
+		defer wg.Done()
+		if err := c.reconcileVersionAnnotation(ctx, client, node); err != nil {
+			errs = append(errs, fmt.Errorf("failed to reconcile version annotation: %w", err))
+		}
+	}()
+
+	wg.Wait()
```
bschimke95: There are two issues with this code:
1. errs is not protected by a mutex, if `append` is called at the same time, you got yourself a race condition.
2. You may hit the `wg.Done` call of the first go routine before you call `wg.Add` a second time, e.g. **if** the first go routine finishes very quick on workers.
As per docs, this is not allowed and may lead to a panic:
> Add must be called before the WaitGroup is waited on.
Calls to Add must happen before any Wait or Done call on the same WaitGroup for it to be safe.

I wonder if we really need two go routines at all here. Concurrency is always hard and I don't see an immediate reason to use it here. Just run the reconciles synchronously and don't return after the first error. WDYT?

src/k8s/pkg/k8sd/controllers/upgrade/controller.go
```diff
@@ -84,6 +82,10 @@ func NewController(
 func (c *Controller) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&upgradesv1alpha1.Upgrade{}).
+		Watches(
+			&corev1.Node{},
+			&handler.EnqueueRequestForObject{},
```
bschimke95, line 89: I'm curious - what does this exactly mean in this context?

src/k8s/pkg/k8sd/controllers/upgrade/reconcile.go
```diff
@@ -108,18 +174,16 @@ func (c *Controller) reconcileFeatureUpgrade(ctx context.Context, upgrade *upgra
...
+	// (KU-3749) This should be fixed in the future.
+	nodeList := &corev1.NodeList{}
+	if err := c.client.List(ctx, nodeList); err != nil {
+		return false, fmt.Errorf("failed to list nodes: %w", err)
 	}
 
-	if len(upgradedNodes) < len(clusterMembers) {
-		log.Info("Not all nodes have been upgraded.", "clusterMembers", len(clusterMembers), "upgradedNodes", len(upgradedNodes))
+	if len(upgradedNodes) < len(nodeList.Items) {
+		log.Info("Not all nodes have been upgraded.", "allNodes", len(nodeList.Items), "upgradedNodes", len(upgradedNodes))
```
bschimke95: while we are here, should we include the node names in the message as well? Might help debugging.

tests/integration/tests/test_version_upgrades.py
```diff
@@ -344,84 +353,89 @@ def test_feature_upgrades_inplace(instances: List[harness.Instance], tmp_path: P
...
+        if initial_rel["updated"] == new_rel["updated"]:
+            LOG.warning(
+                "Release %s was not updated during upgrade. "
+                "This might be due to a skipped helm apply due to same values or chart versions",
+                name,
+            )
 
 
-def _waiting_for_upgraded_nodes(upgraded_nodes, expected_nodes) -> True:
+def _waiting_for_upgraded_nodes(upgraded_nodes, expected_nodes) -> bool:
```
bschimke95, line 438: ah Python is great to allow me this kind of stupidness :)

bschimke95 (commented):
Great work @HomayoonAlimohammadi 
Did a first pass.

src/k8s/pkg/k8sd/controllers/node_label.go
HomayoonAlimohammadi: "Only use concurrency when it's really necessary". Great comment. I'll fix it.

src/k8s/pkg/k8sd/controllers/upgrade/controller.go
HomayoonAlimohammadi: With this, reconciliations will be triggered if changes happen on both `upgrades` and `nodes`. But with you pointing this out I went back and noticed that we probably can use `For()` twice, instead of using `Watch`. I'll test this out.

src/k8s/pkg/k8sd/controllers/upgrade/reconcile.go
HomayoonAlimohammadi: Done

src/k8s/pkg/k8sd/app/hooks_join.go
HomayoonAlimohammadi: Done

src/k8s/pkg/k8sd/controllers/upgrade/controller.go
HomayoonAlimohammadi: Looks like we can't use `For()` more than once:
```
"Failed to start controller coordinator" err="failed to setup controllers: failed to setup upgrade controller: failed to setup upgrade controller with manager: For(...) should only be called once, could not assign multiple objects for reconciliation" logger="k8sd"
```
Changing this back to Watches()

PR #1649, opened 2025-07-15
build: Build etcd dynamically with cgo enabled

bschimke95:
## Description

Right now, etcd is built statically with CGO disabled. For FIPS, we need to:
* build it dynamically
* enable CGO
* update the binary to load libraries from the core snap instead of the host

In a next step, we can then build etcd with the modified Go toolchain to make it FIPS compliant, as done and explained for the other components here: https://github.com/canonical/k8s-snap/pull/1496

Normally, the modification of the rpath and interpreter is done automatically by snapcraft with `patchelf`. However, this tool has a bug that triggers a segmentation fault when modifing the ELF segments of the etcd binary (see relevant comments in code). Since a fix is unlikely to land any time soon - the issue is open for some years now - I created a custom CLI tool that resembles `patchelf` but builds on the well supported LIEF project.

Note that in etcd v3.4.30, CGO is not configurable. I added the patch from v3.4.31 to make CGO configurable. 

## Backport

FIPS won't be backported

## Checklist

- [ ] Covered by unit tests
- [ ] Covered by integration tests - we only change the build process; the existing tests cover this already.
- [x] CLA signed

### DIFF
--- a/build-scripts/hack/patchelf.py
+++ b/build-scripts/hack/patchelf.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+"""
+ELF Patcher Script using LIEF
+
+This tool inspects and modifies ELF binaries' RPATH and dynamic interpreter (PT_INTERP) fields.
+It serves a similar purpose to the NixOS `patchelf` utility, but avoids several known bugs in
+the upstream `patchelf` tool, particularly issues when modifying or truncating existing RPATH
+segments that can corrupt the binary or silently fail.
+
+To provide a more robust and reliable mechanism for ELF patching, this script uses the
+Library to Instrument Executable Formats(LIEF - https://github.com/lief-project/LIEF) library,
+which gives precise control over ELF structure and allows safe rewriting of RPATH and interpreter fields.
+
+Usage:
+    ./patchelf.py <elf-binary> [--set-rpath <path>] [--set-interpreter <interp>] [--output <out>]
+
+Examples:
+    ./patchelf.py ./my_binary --set-rpath '$ORIGIN/../lib' --output ./patched_binary
+    ./patchelf.py ./tool --set-interpreter /lib64/ld-linux-x86-64.so.2
+
+"""
+
+import lief
+import argparse
+import sys
+
+def print_rpath(binary):
+    rpath = binary[lief.ELF.DynamicEntry.TAG.RPATH]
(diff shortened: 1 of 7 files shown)


build-scripts/hack/patchelf.py
```diff
@@ -0,0 +1,83 @@
...
+To provide a more robust and reliable mechanism for ELF patching, this script uses the
+Library to Instrument Executable Formats(LIEF - https://github.com/lief-project/LIEF) library,
+which gives precise control over ELF structure and allows safe rewriting of RPATH and interpreter fields.
+
+Usage:
+    ./elfpatch.py <elf-binary> [--set-rpath <path>] [--set-interpreter <interp>] [--output <out>]
+
+Examples:
+    ./elfpatch.py ./my_binary --set-rpath '$ORIGIN/../lib' --output ./patched_binary
+    ./elfpatch.py ./tool --set-interpreter /lib64/ld-linux-x86-64.so.2
```
HomayoonAlimohammadi: super small nit:
```suggestion
    ./patchelf.py <elf-binary> [--set-rpath <path>] [--set-interpreter <interp>] [--output <out>]

Examples:
    ./patchelf.py ./my_binary --set-rpath '$ORIGIN/../lib' --output ./patched_binary
    ./patchelf.py ./tool --set-interpreter /lib64/ld-linux-x86-64.so.2
```

HomayoonAlimohammadi (approved):
Great work @bschimke95! I'm not an expert in ELF patches, but the whole thing looks reasonable to me.

snap/snapcraft.yaml
```diff
@@ -77,8 +83,35 @@ parts:
...
+      # Note(Ben): This field must match the base field above.
+      BASE=core22
+
+      $CRAFT_PROJECT_DIR/build-scripts/build-component.sh etcd
+
+      pip3 install -r $CRAFT_PROJECT_DIR/build-scripts/hack/requirements.txt
+      case "$CRAFT_ARCH_TRIPLET_BUILD_FOR" in
+        x86_64-linux-gnu)
+          python3 $CRAFT_PROJECT_DIR/build-scripts/hack/patchelf.py $CRAFT_PART_INSTALL/bin/etcd \
+            --set-rpath /snap/core22/current/lib/x86_64-linux-gnu/ \
```
rapour: I think the `Base` was meant to replace these hardcoded values?

bschimke95: Nice catch - thanks!

PR #1667, opened 2025-07-17
build: Fetch etcd 3.4.30 sources from Ubuntu archive

bschimke95:
## Description

Use the etcd source from the Ubuntu archive to help deduplicate maintenance and security efforts. The snap will use etcd v3.4.30, which is the latest version available in the current LTS Ubuntu release.

## Backport

1.32, 1.33

## Checklist

- [ ] Covered by unit tests - already covered
- [ ] Covered by integration tests - already covered
- [x] CLA signed


build-scripts/components/etcd/version
```diff
@@ -1 +1 @@
-v3.6.1
+applied/3.4.30-3
```
HomayoonAlimohammadi, line 1: Question: How did we figure that this is the latest LTS? I can see `[applied/3.5.16-4](https://git.launchpad.net/ubuntu/+source/etcd/tag/?h=applied/3.5.16-4)` as well.

HomayoonAlimohammadi (approved):
Thanks a lot @bschimke95! LGTM, just a question about the Etcd version.

build-scripts/components/etcd/version
bschimke95: `3.4` is the LTS version as described in the [spec](https://docs.google.com/document/d/1FZb6nHETYlADM-Xd5zQrdRWKD5L6ugWnnZF54eNvrUg/edit?tab=t.0#heading=h.txa600pd5awo). I'm not exactly sure where @berkayoz got this info from but if you e.g. look at the `etcd` snap then you also see that 3.4 is the `latest/stable`

PR #1680, opened 2025-07-22
ci: Use Github token to authenticate GH requests in CI

bschimke95:
We otherwise run into rate-limits, e.g. https://github.com/canonical/k8s-snap/actions/runs/16417613513/job/46387321621

### DIFF
--- a/.github/workflows/update-pre-release-branches.yaml
+++ b/.github/workflows/update-pre-release-branches.yaml
@@ -21,6 +21,8 @@ jobs:
     outputs:
       preRelease: ${{ steps.determine.outputs.preRelease }}
       branch: ${{ steps.determine.outputs.branch }}
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/charm-e2e-tests.yaml
```diff
@@ -103,7 +103,7 @@ jobs:
 
           python3 -m venv env
           source env/bin/activate
-          pip install tox
+          pip install tox tox-uv uv
```
mateoflorido: Let's avoid doing this and instead use the `astral-sh/setup-uv@v6` action. Then, you can install it using `uv tool install tox --with tox-uv`

See: https://github.com/canonical/operator-workflows/blob/dc69c7c0826680bdf5f241106bd9cedf18a67a87/.github/workflows/integration_test_run.yaml#L228-L235

bschimke95: Thanks - moved into a separate PR:
https://github.com/canonical/k8s-snap/pull/1686

PR #1685, opened 2025-07-22
ci: Assign a category to the sarif uploads

bschimke95:
## Description

We run into https://github.blog/changelog/2025-07-21-code-scanning-will-stop-combining-multiple-sarif-runs-uploaded-in-the-same-sarif-file/

## Solution

This adds a specific category as explained in https://github.blog/changelog/2024-05-06-code-scanning-will-stop-combining-runs-from-a-single-upload/

## Backport

We need to backport this as we have the same workflow in the release branches

PR #1687, opened 2025-07-22
ci: Assign a category to the sarif uploads (#1685)

bschimke95:
Backport of #1685


bschimke95:
Actually, we don't need to backport this since the `main` branch also does this checks for 1.32 and 1.33

PR #1702, opened 2025-07-28
ci: Add DISA STIG compliant runners

bschimke95:
Extends the weekly FIPS tests to run on DISA STIG compliant multipass machines. 

### Code changes:
* Added a cloud-init configuration for jammy and noble to setup and configure FIPS & DISA STIG. Some STIG rules are skipped as they would block communication between host and multipass VM 
* Added a `required_ports` marker to specify ports that need to be open for tests.

### DIFF
--- a/tests/integration/templates/cloud-init/ubuntu-pro-fips-stig-jammy.yaml
+++ b/tests/integration/templates/cloud-init/ubuntu-pro-fips-stig-jammy.yaml
@@ -1,0 +1,318 @@
+package_update: true
+package_upgrade: true
+packages:
+  - ubuntu-advantage-tools
+runcmd:
+  # enable fips mode and attach to the ubuntu pro service.
+  # we use a staging token here to avoid load on the production contracts service.
+  - echo "Update the contract URL to the staging environment"
+  - 'echo "contract_url: https://contracts.staging.canonical.com" > /etc/ubuntu-advantage/uaclient.conf'
+  - echo "Attaching to Ubuntu Pro with token"
+  - sudo pro attach $TEST_UBUNTU_PRO_TOKEN --no-auto-enable
+  - echo "Enabling FIPS updates"
+  - sudo pro enable fips-updates --assume-yes
+  # enable stig compliance
+  # add nopasswd for ubuntu user *before* the usg fix and password deletion
+  # This creates a file in /etc/sudoers.d/ that allows 'ubuntu' to run all sudo commands without a password
+  # Each command must be a single string in runcmd's list
+  - echo "ubuntu:mynewpassword" | sudo chpasswd
+  - 'echo "ubuntu ALL=(ALL) NOPASSWD: ALL" | sudo tee /etc/sudoers.d/ubuntu_nopasswd'
+  - sudo chmod 0440 /etc/sudoers.d/ubuntu_nopasswd
+  - sudo pro enable usg --assume-yes
+  - sudo usg fix --tailoring-file /etc/usg/tailoring.xml
+  - sudo passwd -d ubuntu
+  # Configure Firewall rules
+  - sudo sed -i 's/^DEFAULT_FORWARD_POLICY=.*/DEFAULT_FORWARD_POLICY="ACCEPT"/' /etc/default/ufw
+  - sudo sysctl -w net.ipv4.ip_forward=1
+  - sudo ufw allow 6443/tcp
+  - sudo ufw allow 10250/tcp
+  - sudo ufw allow 10257/tcp
+  - sudo ufw allow 10259/tcp
+  - sudo ufw allow 2379/tcp
+  - sudo ufw allow 2380/tcp
+  - sudo ufw allow 6400/tcp
+  - sudo ufw allow 4240/tcp
+  - sudo ufw allow 8472/udp
+  - sudo ufw allow in ssh
+  - sudo ufw allow in 22
+  - sudo ufw enable
+  - sudo reboot
+
+write_files:
+  # This file is used to customize the USG tailoring for the DISA-STIG benchmark.
+  # It is derived from the official DISA-STIG benchmark for Ubuntu 22.04 by running
+  # `sudo usg generate-tailoring disa_stig tailor.xml``
+  # and disabling all rules that may cause issues with block connection to multipass:
+  # xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_audit
+  # xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_deny
+  # xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_interval
+  # xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_silent
+  # xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_unlock_time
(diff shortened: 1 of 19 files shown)


tests/integration/tests/test_util/harness/multipass.py
```diff
@@ -91,19 +93,35 @@ def replace_env_var(match):
...
+                # the cloud-init is done.
                 run(
                     cmd + ["--cloud-init", "-"],
                     input=cloud_init_content.encode(),
                     sensitive_kwargs=True,
                     check=False,
                 )
-                stubbornly(retries=20, delay_s=5).exec(
-                    ["multipass", "exec", instance_id, "--", "echo", "test"],
+                stubbornly(retries=200, delay_s=10).until(
```
bschimke95, line 105: Checking for the cloud status is a more stable check than the previous echo solution.

tests/integration/templates/cloud-init/ubuntu-pro-fips-stig-jammy.yaml
```diff
@@ -0,0 +1,318 @@
...
+  # add nopasswd for ubuntu user *before* the usg fix and password deletion
+  # This creates a file in /etc/sudoers.d/ that allows 'ubuntu' to run all sudo commands without a password
+  # Each command must be a single string in runcmd's list
+  - echo "ubuntu:mynewpassword" | sudo chpasswd
+  - 'echo "ubuntu ALL=(ALL) NOPASSWD: ALL" | sudo tee /etc/sudoers.d/ubuntu_nopasswd'
+  - sudo chmod 0440 /etc/sudoers.d/ubuntu_nopasswd
+  - sudo pro enable usg --assume-yes
+  - sudo usg fix --tailoring-file /etc/usg/tailoring.xml
+  - sudo passwd -d ubuntu
+  # Configure Firewall rules
```
louiseschmidtgen, line 24: Are we going to run this test for k8s-dqlite too? We would need the dqlite port too.

tests/integration/templates/cloud-init/ubuntu-pro-fips-stig-jammy.yaml
```diff
@@ -0,0 +1,318 @@
...
+  # xccdf_org.ssgproject.content_rule_sssd_enable_user_cert
+  # xccdf_org.ssgproject.content_rule_prevent_direct_root_logins
+  # xccdf_org.ssgproject.content_rule_accounts_max_concurrent_login_sessions
+  # xccdf_org.ssgproject.content_rule_account_disable_post_pw_expiration
+  # xccdf_org.ssgproject.content_rule_account_temp_expire_date
+  # xccdf_org.ssgproject.content_rule_ensure_sudo_group_restricted
+  - path: /etc/usg/tailoring.xml
+    owner: root:root
+    permissions: "755"
+    content: |
```
louiseschmidtgen, line 65: Do we have to commit this blob? Can we not generate it in the workflow?

louiseschmidtgen (commented):
Nice work Ben, I only have two comments:

tests/integration/templates/cloud-init/ubuntu-pro-fips-stig-jammy.yaml
HomayoonAlimohammadi, line 1: If I'm reading git correctly, this file seems new, and there's a `ubuntu-pro-fips.yaml` file as well. Since that file is not used anymore, should we remove it?

tests/integration/tests/test_util/harness/base.py
```diff
@@ -99,6 +100,16 @@ def exec(
...
 
+    def open_ports(self, instance_id: str, ports: List[int]):
+        """Open ports on the instance.
+
+        :param instance_id: The instance_id, as returned by new_instance()
+        :param ports: List of ports to open on the instance.
+
+        Ports will be opened on a best effort basis. If the port is already open,
+        or no firewall is installed, no error will be raised.
+        """
```
HomayoonAlimohammadi: nit: `raise NotImplementedError`?

tests/integration/tests/test_util/harness/juju.py
```diff
@@ -191,6 +192,11 @@ def exec(self, instance_id: str, command: list, **kwargs):
             completed.check_returncode()
         return completed
 
+    def open_ports(self, instance_id: str, ports: List[int]):
+        raise NotImplementedError(
+            "Opening ports is not yet supported by the Juju harness. "
```
HomayoonAlimohammadi: nit with guilt:
```suggestion
            "Opening ports is not yet supported by the Juju harness."
```

tests/integration/tests/test_util/harness/multipass.py
```diff
@@ -209,6 +230,24 @@ def exec(self, instance_id: str, command: list, **kwargs):
             **kwargs,
         )
 
+    def open_ports(self, instance_id: str, ports: List[int]):
```
HomayoonAlimohammadi: Why didn't we add this method to the `Instance` type? This might be a dumb question, but is it any different to run the `ufw` command on Multipass VMs compared to Juju or LXD machines?

bschimke95: I assumed that there might be some different/additional steps e.g. for Juju. By making this harness-depended we can add custom implementation if needed.

tests/integration/tests/test_util/harness/juju.py
bschimke95: fixed with guilt

tests/integration/tests/test_util/harness/base.py
bschimke95: Hm, I don't think this is something you document, do you? This is an abstract method, it can also be not implemented. I don't see value in documenting this.

bschimke95: Nevermind... Now I get your comment.
Fixed it.

tests/integration/templates/cloud-init/ubuntu-pro-fips-stig-jammy.yaml
bschimke95: This blob is a custom usg ruleset (see comment above). One can generate the template for such a file with `usg generate-tailoring disa_stig` but we'd still need to modify the rules we want to disable. Hence, it is more transparent if we just write the file directly.

tests/integration/templates/cloud-init/ubuntu-pro-fips-stig-jammy.yaml
bschimke95: See e.g. test_clustering_dqlite. This setup only enables the default ports. Additional ports can be enabled on a per-test basis with pytest markers.

tests/integration/templates/cloud-init/ubuntu-pro-fips-stig-jammy.yaml
bschimke95: Fair point. My rationale was that people might still need this file at some point/test in the future but they can always get it from the history and it does not make sense to keep unused files. 
Removed it.

tests/integration/tests/test_util/harness/multipass.py
bschimke95: See https://github.com/canonical/k8s-snap/pull/1739 for an related issue.

PR #1705, opened 2025-07-29
docs: update homepage docset descriptions

nhennigan:
## Description

We recently received some feedback on the layout and wording of our landing page could be misleading

## Solution

Place the three docs sets (snap,capi,charm) side by side to not insinuate order 
Reword the cards to actually have some relevant info

## Issue

N/A

## Backport

1.32, 1.33

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/docs/canonicalk8s/index.md
+++ b/docs/canonicalk8s/index.md
@@ -72,7 +72,7 @@ The `k8s` charms take care of installing, configuring and managing {{product}} o
 :link: capi/
 ### [Canonical Kubernetes and Cluster API ›](/capi/index)
 
-Using the declarative APIs and tooling provided by Cluster API, deploy and manage multiple {{product}} clusters.
+Using Cluster API's declarative tooling, deploy and manage multiple {{product}} clusters.
 ```
 
 ````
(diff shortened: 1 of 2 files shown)


docs/canonicalk8s/index.md
```diff
@@ -25,9 +25,9 @@ your needs. If you would like to jump straight in, head to the
```
louiseschmidtgen, line 25: I noticed in line 8 we write ingress instead of Ingress like the rest of our docs I believe.

nhennigan: Ingress is one of those things that keeps confusing me :upside_down_face: After looking into it again I think it should be "ingress gateway". We use Gateway the Kubernetes object. We use Cilium Ingress Controller which is just an implementation of the older Kubernetes Ingress. Ingress in "ingress gateway" is an adjective and not referring to the actual Kubernetes Ingress so I think it should be lower case. We are listing the features, the CNI, the DNS services, the networking capabilities i.e get traffic into the cluster. Does that reasoning make sense?

PR #1710, opened 2025-07-31
feat: Included managed-etcd service inspection

addyess:
## Description

Add etcd service to inspection reports

## Solution

Adds another service to the list to be inspected

## Backport

* 1.32
* 1.33

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

No code changes needing unit/integration tests

### DIFF
--- a/k8s/scripts/inspect.sh
+++ b/k8s/scripts/inspect.sh
@@ -293,7 +293,7 @@ mkdir -p "$INSPECT_DUMP"
 
 printf -- 'Collecting service information\n'
 
-control_plane_services=("k8s.containerd" "k8s.kube-proxy" "k8s.k8s-dqlite" "k8s.k8sd" "k8s.kube-apiserver" "k8s.kube-controller-manager" "k8s.kube-scheduler" "k8s.kubelet")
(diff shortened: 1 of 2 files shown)


k8s/scripts/inspect.sh
```diff
@@ -293,7 +293,7 @@ mkdir -p "$INSPECT_DUMP"
 
 printf -- 'Collecting service information\n'
 
-control_plane_services=("k8s.containerd" "k8s.kube-proxy" "k8s.k8s-dqlite" "k8s.k8sd" "k8s.kube-apiserver" "k8s.kube-controller-manager" "k8s.kube-scheduler" "k8s.kubelet")
+control_plane_services=("k8s.containerd" "k8s.etcd" "k8s.kube-proxy" "k8s.k8s-dqlite" "k8s.k8sd" "k8s.kube-apiserver" "k8s.kube-controller-manager" "k8s.kube-scheduler" "k8s.kubelet")
```
rapour, line 296: Does this result in printing 
```
log_warning "Service $service should be running on this node"
```
for both `etcd` and `k8s-dqlite`? If yes, then it might be misleading for the user since we are only running one of them. Is there a way to get the right datastore type and include the relevant service here?

rapour (commented):
Thanks a lot @addyess, left a question

k8s/scripts/inspect.sh
addyess: Yes. It would show both. But currently the report states a non running dqlite service when we're using an external etcd.

PR #1722, opened 2025-08-05
chore: Use go/1.24-fips

HomayoonAlimohammadi:
### Overview

This PR uses `go/1.24-fips/stable` snap and adds necessary changes such as:
* Adding `GOEXPERIMENT=opensslcrypto`
* Removing the TLS1.2 min version workaround for LXD (TLS min version will be 1.3)


louiseschmidtgen (approved):
Thank you Hue!

PR #1723, opened 2025-08-05
tests: Add integration test to verify cluster health after reboot

HomayoonAlimohammadi:
### Overview

This PR adds an integration test to verify cluster health after a reboot.

### DIFF
--- a/tests/integration/tests/test_reboot.py
+++ b/tests/integration/tests/test_reboot.py
@@ -1,0 +1,50 @@
+#
+# Copyright 2025 Canonical, Ltd.
+#
+import logging
+from typing import List
+
+import pytest
+from test_util import harness, tags, util
+
+LOG = logging.getLogger(__name__)
+
+STATUS_PATTERNS = [
+    r"cluster status:\s*ready",
+    r"control plane nodes:\s*(\d{1,3}(?:\.\d{1,3}){3}:\d{1,5})\s\(voter\)",
+    r"high availability:\s*no",
+    r"datastore:\s*etcd",
+    r"network:\s*enabled",
+    r"dns:\s*enabled at (\d{1,3}(?:\.\d{1,3}){3})",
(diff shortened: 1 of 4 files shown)


tests/integration/tests/test_reboot.py
```diff
@@ -0,0 +1,50 @@
...
+#
+import logging
+from typing import List
+
+import pytest
+from test_util import harness, tags, util
+
+LOG = logging.getLogger(__name__)
+
+STATUS_PATTERNS = [
```
louiseschmidtgen, line 12: So we are parsing k8s status... Maybe we should have a --json option. But that is beyond the scope of this PR.

tests/integration/tests/test_reboot.py
```diff
@@ -0,0 +1,50 @@
...
+    r"dns:\s*enabled at (\d{1,3}(?:\.\d{1,3}){3})",
+    r"ingress:\s*disabled",
+    r"load-balancer:\s*disabled",
+    r"local-storage:\s*enabled at /var/snap/k8s/common/rawfile-storage",
+    r"gateway\s*enabled",
+]
+
+
+@pytest.mark.tags(tags.WEEKLY)
+def test_reboot(instances: List[harness.Instance]):
```
louiseschmidtgen, line 28: Should we try to provoke failures in a distributed set-up -> 3 nodes? We can test our fault tolerance.

tests/integration/tests/test_reboot.py
HomayoonAlimohammadi: Good point, thanks! I'll create a separate PR to address this throughout our tests.

tests/integration/tests/test_reboot.py
HomayoonAlimohammadi: Good point. I've updated the tests.

PR #1726, opened 2025-08-06
feat: [Follow up] FIPS-compliant embedded k8s-snap components

HomayoonAlimohammadi:
### Overview

This PR rebases https://github.com/canonical/k8s-snap/pull/1496 on main and bumps Go to `1.24-fips`. The reason behind a brand new PR is to maintain https://github.com/canonical/k8s-snap/pull/1496 in its original form for future reference. Changes on top of the original PR:
* Switch to Go 1.24-fips
* Switch to Kubernetes 1.33.3 (based on current k8s8-snap main)
* Switch to Go 1.24 module


snap/hooks/configure
```diff
@@ -4,6 +4,6 @@
 
 k8s::common::setup_env
 
-k8s::cmd::k8s x-snapd-config reconcile
+k8s::cmd::k8s x-snapd-config reconcile || true
 
-k8s::containerd::ensure_systemd_defaults
+k8s::containerd::ensure_systemd_defaults || true
```
berkayoz: Why do we have this change?

HomayoonAlimohammadi: I don't know, this was in the original PR. @bschimke95 might have a better answer.

.github/workflows/lint_and_integration.yaml
```diff
@@ -151,7 +151,7 @@ jobs:
     name: Charm e2e tests - ${{ matrix.arch }}
     strategy:
       matrix:
-        arch: [amd64]
+        arch: [amd64, arm64]
```
addyess: the e2e tests start LXD containers as VMs right?

i see this failure on arm runners b/c they don't support virtualization.  We run into the same problem with lxd on arm in k8s-operator.  FYI -- this might not be possible

```
Failed creating instance record: Instance type "virtual-machine" is not supported on this server: KVM support is miss...
```

addyess (commented):
Just a think to note

.github/workflows/lint_and_integration.yaml
HomayoonAlimohammadi: Got it. Thanks for the context @addyess. I need to discuss with Ben why we might need this or whether or not we can ignore this failure (at least for now?).

bschimke95: We don't run the LXD containers as VMs, see https://github.com/canonical/k8s-snap/blob/342204987c1467d69ac4c8778277f58960a4ea48/tests/integration/tests/test_util/harness/lxd.py#L82

Hence, we shouldn't run into any ARM virtualization issues. In fact, the arm builds on the initial PR are green.

snap/hooks/configure
bschimke95: I initially added this for debugging the k8s-snap (it otherwise wouldn't install if the k8s binary failed). We can probably remove it.

HomayoonAlimohammadi: Thanks. Removed it

.github/workflows/lint_and_integration.yaml
HomayoonAlimohammadi: I think @addyess is right. We're seeing this failure both in this PR (charms e2e testing) and in the original PR. I think we should remove this for now.

> Hence, we shouldn't run into any ARM virtualization issues. In fact, the arm builds on the initial PR are green.

Yes, ARM _builds_ are fine, but the charm e2e tests fails on ARM (on both PRs).

HomayoonAlimohammadi:
Converting to draft to make sure the new images work as expected.

berkayoz (approved):
LGTM, amazing work!

PR #1729, opened 2025-08-07
fix: Use -ck tags for metallb frr and rawfile-localpv rocks

HomayoonAlimohammadi:
### Overview

This PR adds a `-ck` suffix to metallb frr and rawfile-localpv image tags which should've had this suffix before. These tags DO NOT include FIPS changes.

NOTES:
* `-ck1` for rawfile-localpv was discussed with the team internally.
* `-ck1` for metallb frr is based on the [images history](https://github.com/canonical/metallb-rocks/pkgs/container/frr). `-ck2` was created last week with the FIPS changed, and `-ck1` is a month old with the contents prior to FIPS changes.


berkayoz (approved):
LGTM

PR #1739, opened 2025-08-11
test: Make instance restart harness-dependent

bschimke95:
## Description

The multipass tests were failing because rebooting needs to happen via `multipass restart` instead of an `sudo reboot` from within the instance. See https://github.com/canonical/k8s-snap/actions/runs/16878895760/job/47819464883?pr=1702

## Solution

* Refactored the reboot and delete functions to call the underlying harness function to restart/delete an instance
* Renamed reboot to restart to match lxd/multipass naming conventions

### DIFF
--- a/tests/integration/tests/test_util/harness/base.py
+++ b/tests/integration/tests/test_util/harness/base.py
@@ -106,6 +106,15 @@ class Harness:
         """
         raise NotImplementedError
 
+    def restart_instance(self, instance_id: str):
+        """Restart an previously created instance.
+
+        :param instance_id: The instance_id, as returned by new_instance()
+
+        If the operation fails, a HarnessError is raised.
+        """
+        raise NotImplementedError
+
     def delete_instance(self, instance_id: str):
         """Delete a previously created instance.
 
--- a/tests/integration/tests/test_util/harness/base.py
+++ b/tests/integration/tests/test_util/harness/base.py
@@ -3,6 +3,7 @@
 #
 import subprocess
 from functools import cached_property, partial
+from typing import List
 
 
 class HarnessError(Exception):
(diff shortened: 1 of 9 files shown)


tests/integration/tests/test_util/harness/juju.py
```diff
@@ -191,6 +192,16 @@ def exec(self, instance_id: str, command: list, **kwargs):
             completed.check_returncode()
         return completed
 
+    def restart_instance(self, instance_id):
+        raise NotImplementedError(
+            "Restarting instances is not yet supported by the Juju harness."
+        )
```
addyess: Could this work?

```sh
        self.exec(instance_id, ["sudo", "reboot"])
```

tests/integration/tests/test_util/harness/juju.py
```diff
@@ -191,6 +192,16 @@ def exec(self, instance_id: str, command: list, **kwargs):
...
+
+    def open_ports(self, instance_id: str, ports: List[int]):
+        raise NotImplementedError(
+            "Opening ports is not yet supported by the Juju harness."
+        )
```
addyess: how about this:

```sh
        for port in ports:
            self.exec(instance_id, ["open-port", port])
```

tests/integration/tests/test_util/harness/base.py
```diff
@@ -103,6 +107,25 @@ def exec(
...
+    def restart_instance(self, instance_id: str):
+        """Restart an previously created instance.
+
+        :param instance_id: The instance_id, as returned by new_instance()
+
+        If the operation fails, a HarnessError is raised.
+        """
+        raise NotImplementedError
+
+    def open_ports(self, instance_id: str, ports: List[int]):
```
HomayoonAlimohammadi, line 123: I'm ok with keeping this in the same PR, but I suspect the "open ports"-related logic was not supposed to be in this PR? Anyways, just a small nit:
```python
raise NotImplementedError
```

HomayoonAlimohammadi (approved):
Thanks for fixing the reboot logic @bschimke95. LGTM overall. A small nit would be to break down the 3 things that we're doing in 3 separate commits and rebase them instead of squashing (I don't think separate PRs is necessary). These 3 would be: changing the instance deletion logic, changing the instance restart logic, and adding the open ports logic.

PR #1754, opened 2025-08-14

cdkbot:


bschimke95:
@louiseschmidtgen it looks like this tries to revert #1750.
I think we were a bit to fast with that PR. We align the CNI with the upstream version: https://github.com/canonical/k8s-snap/blob/6575710b1673146d6af7a5cd8d3793c0d54b0d83/build-scripts/hack/update-component-versions.py#L94-L95

So this should automatically be updated with the 1.34 version bump.

bschimke95 (approved):

PR #1756, opened 2025-08-14
fix: derive initial-cluster from etcd member list

berkayoz:
## Description

Derives the `initial-cluster` flag from the etcd member list instead of the k8sd state to avoid including nodes that are not part of the etcd cluster. This issue manifests when 2 nodes are joined in parallel.

## Backport

`release-1.32`
`release-1.33`

## Checklist

- [x] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_join.go
+++ b/src/k8s/pkg/k8sd/app/hooks_join.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"slices"
@@ -21,6 +22,7 @@ import (
 	"github.com/canonical/k8s/pkg/utils/experimental/snapdconfig"
 	"github.com/canonical/k8s/pkg/version"
 	"github.com/canonical/microcluster/v2/state"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	versionutil "k8s.io/apimachinery/pkg/util/version"
 )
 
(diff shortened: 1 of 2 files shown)


src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -256,6 +254,21 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
...
+		}
+
+		// Build initial cluster members map from etcd members
+		initialClusterMembers := make(map[string]string)
+		for _, member := range memberListResp.Members {
+			if !member.IsLearner && len(member.PeerURLs) > 0 && member.Name != "" {
+				// Use the first peer URL for each member
+				initialClusterMembers[member.Name] = member.PeerURLs[0]
+			}
+		}
```
mateoflorido: nit: Could we use a more explicit approach for the skipping conditions? Maybe we could improve the skipping clause with a `log.Debug` to improve the o11y during the join hook.
```suggestion
		// Build initial cluster members map from etcd members
		initialClusterMembers := make(map[string]string)
		for _, member := range memberListResp.Members {
			if member.IsLearner || len(member.PeerURLs) == 0 || member.Name == "" {
				continue
			}
			// Use the first peer URL for each member
			initialClusterMembers[member.Name] = member.PeerURLs[0]
		}
```

addyess (approved):
After seeing today's demo i can't wait to land this

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -256,10 +256,27 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
 		}
 		defer etcdClient.Close()
 
-		if _, err := etcdClient.MemberAdd(ctx, []string{fmt.Sprintf("https://%s", utils.JoinHostPort(nodeIP.String(), cfg.Datastore.GetEtcdPeerPort()))}); err != nil {
+		memberAddResp, err := etcdClient.MemberAdd(ctx, []string{fmt.Sprintf("https://%s", utils.JoinHostPort(nodeIP.String(), cfg.Datastore.GetEtcdPeerPort()))})
```
louiseschmidtgen, line 259: Does it make sense to retry the member add operation on failure, or does it get retried on etcd's side?
Otherwise, we might want to retry if the cause for the error is temporary like a re-election, or an ongoing membership operation. Wdyt?

louiseschmidtgen (commented):
Great work Berkay! One little comment:

src/k8s/pkg/k8sd/app/hooks_join.go
berkayoz: If I understand correctly, according to https://etcd.io/docs/v3.6/learning/design-client/ the client itself has a retry mechanism built in for what etcd consider transient errors. I've seen this a couple of times in member remove where etcd client retried after `leader changed`.

louiseschmidtgen: Thanks for double checking! Yes, the etcd should by design handle these cases already.

PR #1773, opened 2025-08-18
ci: add more clustering test cases

louiseschmidtgen:
## ci: add more clustering test cases

This PR introduces more test cases on our nightly CI to test membership operations under more conditions: (parallel joins/removals, mixed join and removal operations) as well as disruptions such as snap restarts.

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/tests/integration/tests/test_clustering.py
+++ b/tests/integration/tests/test_clustering.py
@@ -124,44 +124,70 @@ def test_disa_stig_clustering(instances: List[harness.Instance]):
     assert "worker" in util.get_local_node_status(joining_worker)
 
 
-@pytest.mark.node_count(4)
+@pytest.mark.node_count(3)
 @pytest.mark.tags(tags.NIGHTLY)
-def test_concurrent_membership_operations(instances: List[harness.Instance]):
+def test_concurrent_cp_membership_operations(instances: List[harness.Instance]):
     cluster_node = instances[0]
     joining_cp_A = instances[1]
     joining_cp_B = instances[2]
-    joining_worker_C = instances[3]
 
     util.wait_until_k8s_ready(cluster_node, [cluster_node])
 
-    with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
+    with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
         future_A = executor.submit(join_node_with_retry, cluster_node, joining_cp_A)
         future_B = executor.submit(join_node_with_retry, cluster_node, joining_cp_B)
-        future_C = executor.submit(
-            join_node_with_retry, cluster_node, joining_worker_C, worker=True
-        )
-        concurrent.futures.wait([future_A, future_B, future_C])
+        concurrent.futures.wait([future_A, future_B])
 
     util.wait_until_k8s_ready(cluster_node, instances)
 
     for node in [cluster_node, joining_cp_A, joining_cp_B]:
         assert "control-plane" in util.get_local_node_status(node)
-    assert "worker" in util.get_local_node_status(joining_worker_C)
 
-    with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
+    with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
         future_A = executor.submit(remove_node_with_retry, cluster_node, joining_cp_A)
         future_B = executor.submit(remove_node_with_retry, cluster_node, joining_cp_B)
-        future_C = executor.submit(
-            remove_node_with_retry, cluster_node, joining_worker_C
-        )
-        concurrent.futures.wait([future_A, future_B, future_C])
+        concurrent.futures.wait([future_A, future_B])
 
     util.wait_until_k8s_ready(cluster_node, [cluster_node])
 
     nodes = util.ready_nodes(cluster_node)
     assert (
         len(nodes) == 1
-    ), "two control-plane nodes, one worker should have been removed from cluster"
+    ), "two control-plane nodes, should have been removed from cluster"
+
+    assert cluster_node.id in [node["metadata"]["name"] for node in nodes]
+
+
+@pytest.mark.node_count(3)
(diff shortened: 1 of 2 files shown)


tests/integration/tests/test_clustering.py
```diff
@@ -229,6 +435,37 @@ def _check_cert(instance, cert_fname):
     util.wait_until_k8s_ready(cluster_node, instances)
 
 
+def join_node_with_retry(cluster_node, joining_node, retries=5, delay_s=1):
```
bschimke95: Is there a reason to not use the existing `stubborn.retry` mechanism?

tests/integration/tests/test_clustering.py
```diff
@@ -122,6 +124,210 @@ def test_disa_stig_clustering(instances: List[harness.Instance]):
...
+    util.wait_until_k8s_ready(cluster_node, [cluster_node])
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+        future_A = executor.submit(join_node_with_retry, cluster_node, joining_cp_A)
+        future_B = executor.submit(join_node_with_retry, cluster_node, joining_cp_B)
+        concurrent.futures.wait([future_A, future_B])
+
+    util.wait_until_k8s_ready(cluster_node, instances)
+
+    assert "control-plane" in util.get_local_node_status(cluster_node)
```
bschimke95: nit:
```
for node in [cluster_node, joining_cp_A, joining_cp_B]:
    assert "control-plane" in util.get_local_node_status(node)
```

bschimke95 (approved):
Generally LGTM. 
I think it would be interesting to throw in a worker clustering operation to validate that this code-path works concurrently as well.

tests/integration/tests/test_clustering.py
louiseschmidtgen: Yes, `stubborn.retry` is designed to retry shell commands not python functions. I want to retry a function that gets a token and joins the node here.

tests/integration/tests/test_clustering.py
```diff
@@ -122,6 +124,216 @@ def test_disa_stig_clustering(instances: List[harness.Instance]):
...
+        future_C = executor.submit(
+            join_node_with_retry, cluster_node, joining_worker_C, worker=True
+        )
+        concurrent.futures.wait([future_A, future_B, future_C])
+
+    util.wait_until_k8s_ready(cluster_node, instances)
+
+    for node in [cluster_node, joining_cp_A, joining_cp_B]:
+        assert "control-plane" in util.get_local_node_status(node)
+    assert "worker" in util.get_local_node_status(joining_worker_C)
```
rapour: nit: we can add assertion error massages here too for verbosity

tests/integration/tests/test_clustering.py
```diff
@@ -122,6 +124,216 @@ def test_disa_stig_clustering(instances: List[harness.Instance]):
...
+    assert join_token_A != join_token_B, "Join tokens should be different"
+
+    util.join_cluster(joining_cp_A, join_token_A)
+    util.wait_until_k8s_ready(cluster_node, [cluster_node, joining_cp_A])
+
+    assert "control-plane" in util.get_local_node_status(cluster_node)
+    assert "control-plane" in util.get_local_node_status(joining_cp_A)
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+        executor.submit(cluster_node.exec, ["k8s", "remove-node", joining_cp_A.id])
```
rapour: Are we intentionally using two CP nodes here? IIRC dqlite could still form quorum if a node lost from a 2 node cluster but etcd needed both to form quorum. Is etcd functionality is also being tested here?

tests/integration/tests/test_clustering.py
```diff
@@ -122,6 +124,216 @@ def test_disa_stig_clustering(instances: List[harness.Instance]):
...
+    joining_cp_C.exec(["k8s", "remove-node", cluster_node_id, "--force"])
+    nodes = util.ready_nodes(joining_cp_C)
+    assert (
+        len(nodes) == 3
+    ), "three control plane nodes should be ready, original node is removed"
+
+    assert (
+        joining_cp_A.id in [node["metadata"]["name"] for node in nodes]
+        and joining_cp_B.id in [node["metadata"]["name"] for node in nodes]
+        and joining_cp_C.id in [node["metadata"]["name"] for node in nodes]
```
rapour: nit: Maybe we can evaluate the node names once and then check the ones we care. Something like:

```suggestion
   node_names = {node["metadata"]["name"] for node in nodes}
   assert {joining_cp_A.id, joining_cp_B.id, joining_cp_C.id}.issubset(node_names)
```

tests/integration/tests/test_clustering.py
louiseschmidtgen: Yes, the purpose of this test is to test **concurrent** membership operations and ensure the node membership is removed from microcluster and etcd.
You're right about the 2 node case for etcd, however, I am not after functionality testing in this scenario.

rapour: I thought since we are removing a node from a two node cluster there might be situations where the number of CP nodes goes from 2 (Original, A)  ----> 1 (Original) ----> 2 (Original,B) and having a single CP node before adding the new node (B) can cause the etcd to reject the new membership operation since it cannot form a quorum. If it's never going to happen then we are good to go

rapour (approved):
Great work @louiseschmidtgen, thanks

tests/integration/tests/test_clustering.py
louiseschmidtgen: I think the difference here would be the fact that we are properly removing the other healthy member in etcd. Meaning the new majority is one as there are one members in the cluster. 
Does that make sense?

rapour: That makes perfect sense, thank you

PR #1777, opened 2025-08-20
[release-1.33] Update component versions

cdkbot:
[release-1.33] Update component versions


addyess (approved):
Patch rev fix -- nice LGTM

PR #1790, opened 2025-08-27
chore: disable 1.34 feature gates incompatible with k8s-dqlite 

berkayoz:
Disables feature gates;
- ListFromCacheSnapshot
- SizeBasedListCostEstimate
- SizeBasedListCostEstimate
that is incompatible with k8s-dqlite.

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_post_refresh.go
+++ b/src/k8s/pkg/k8sd/app/hooks_post_refresh.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net"
 	"slices"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -171,3 +177,27 @@ func (a *App) updateNodeIPAddresses(ctx context.Context, s state.State) error {
 
 	return nil
 }
+
+func (a *App) updateK8sDqliteFeatureGates(ctx context.Context) error {
+	snap := a.Snap()
+
+	args := map[string]string{"--feature-gates": "ListFromCacheSnapshot=false,SizeBasedListCostEstimate=false,DetectCacheInconsistency=false"}
+	mustRestart, err := snaputil.UpdateServiceArguments(snap, "kube-apiserver", args, nil)
```
berkayoz: Self review, I've to double check if this is performed as a list-merge vs overriding the entire flag.

We might need to manually do that here if `UpdateServiceArguments` does not handle it

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
```diff
@@ -171,3 +177,27 @@ func (a *App) updateNodeIPAddresses(ctx context.Context, s state.State) error {
 
 	return nil
 }
+
+func (a *App) updateK8sDqliteFeatureGates(ctx context.Context) error {
```
HomayoonAlimohammadi: nit: Maybe a comment and a log about what we're doing here might be useful.

HomayoonAlimohammadi (approved):
LGTM, thanks for the fix @berkayoz!

src/k8s/pkg/k8sd/app/hooks_post_refresh.go
berkayoz: As suspected it does overwrite the entire flag like this, pushed an update.

src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
```diff
@@ -106,6 +109,56 @@ func (c *ControlPlaneConfigurationController) reconcile(ctx context.Context, con
...
+	} else if config.Datastore.GetType() == "k8s-dqlite" {
+		// Get existing feature gates to merge our changes
+		featureGates, err := snaputil.GetServiceArgument(c.snap, "kube-apiserver", "--feature-gates")
+		if err != nil {
+			return fmt.Errorf("failed to get kube-apiserver feature gates: %w", err)
+		}
+
+		featureGatesMap := make(map[string]bool)
+		if featureGates != "" {
+			pairs := strings.Split(featureGates, ",")
```
bschimke95, line 123: We have a similar key=val logic [here](https://github.com/canonical/k8s-snap/blob/358904bc6db19f78bd6ac997a0b23313061fb3b4/src/k8s/pkg/snap/util/arguments.go#L45). 
I think it makes sense to have one "parseKeyValFromString" (naming is hard) util that can handle this parsing.

src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
```diff
@@ -106,6 +109,56 @@ func (c *ControlPlaneConfigurationController) reconcile(ctx context.Context, con
...
+					boolVal, err := strconv.ParseBool(stringVal)
+					if err != nil {
+						return fmt.Errorf("failed to parse feature gate value %q for key %q: %w", stringVal, kv[0], err)
+					}
+					featureGatesMap[gate] = boolVal
+				}
+			}
+		}
+
+		// Disable feature gates incompatible with k8s-dqlite introduced in kubernetes 1.34
```
bschimke95: nit: Do we have a card to follow-up on the missing features in k8s-dqlite? If yes, let's add the todo/reference here.

bschimke95 (approved):
Generally LGTM, only minor comments. Feel free to merge once they are addressed.

src/k8s/pkg/k8sd/controllers/control_plane_configuration.go
berkayoz: I'll create a card to consolidate merging logic for argument values which is trickier to handle across the board, we can take a look into it later on.

PR #1806, opened 2025-08-29
feat: add resource files on snap install

louiseschmidtgen:
## Description

This PR adds resources available on snap install for users. These resources currently include disa-stig related bootstrap configuration, join-configurations and audit policies.

Note: I used the install hook to copy the files since we want to make the bootstrap configuration templates available before bootstrap.

## Solution

The new proposed file organization is:
```
$ sudo snap install ./k8s_v1.34.0_arm64.snap --classic --dangerous
k8s v1.34.0 installed
$ sudo tree /var/snap/k8s/common/etc/
├── configurations
│     ├── audit-policy-kube-system.yaml
│     └── audit-policy.yaml
│     └── pod-security-admission-baseline.yaml
│     └── pod-security-admission-restricted.yaml
└── templates
    └── disa-stig
        ├── bootstrap.yaml
        ├── control-plane.yaml
        └── worker.yaml

4 directories, 10 files
```

## Backport

NA

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/tests/integration/tests/test_clustering.py
+++ b/tests/integration/tests/test_clustering.py
@@ -94,7 +94,9 @@ def test_disa_stig_clustering(instances: List[harness.Instance]):
     cp_file = config.COMMON_ETC_DIR + "/templates/disa-stig/control-plane.yaml"
     join_token_cp = util.get_join_token(cluster_node, joining_cp)
 
-    cp_file_content = joining_cp.exec(["cat", cp_file], capture_output=True, text=True).stdout
+    cp_file_content = joining_cp.exec(
+        ["cat", cp_file], capture_output=True, text=True
+    ).stdout
     cp_data = yaml.safe_load(cp_file_content)
     joining_cp.exec(["sysctl", "-w", "vm.overcommit_memory=1"])
     joining_cp.exec(["sysctl", "-w", "kernel.panic=10"])
@@ -105,7 +107,9 @@ def test_disa_stig_clustering(instances: List[harness.Instance]):
     worker_file = config.COMMON_ETC_DIR + "/templates/disa-stig/worker.yaml"
     join_token_worker = util.get_join_token(cluster_node, joining_worker, "--worker")
 
-    worker_file_content = joining_worker.exec(["cat", worker_file], capture_output=True, text=True).stdout
+    worker_file_content = joining_worker.exec(
+        ["cat", worker_file], capture_output=True, text=True
+    ).stdout
     worker_data = yaml.safe_load(worker_file_content)
     joining_worker.exec(["sysctl", "-w", "vm.overcommit_memory=1"])
     joining_worker.exec(["sysctl", "-w", "kernel.panic=10"])


k8s/resources/configurations/audit-policy.yaml
```diff
@@ -0,0 +1,5 @@
+# Log all requests at the Metadata level.
```
ktsakalozos, line 1: Although not clear, the disa stig [1] states that 
```
The policy file must look like this:

# Log all requests at the RequestResponse level.
apiVersion: audit.k8s.io/vX (Where X is the latest apiVersion)
kind: Policy
rules:
- level: RequestResponse
````
So could we please have this specific policy used just to be on the safe side. Thanks

[1] https://stigviewer.com/stigs/kubernetes/2024-06-10/finding/V-242403

File: k8s/lib.sh [hunk 2]
```diff
@@ -43,6 +43,12 @@ k8s::common::is_strict() {
   fi
 }
 
+k8s::common::resources() {
+  mkdir -p "$SNAP_COMMON/etc/"
+  cp -r "$SNAP/etc/templates" "$SNAP_COMMON/etc/"
+  cp -r "$SNAP/etc/configurations" "$SNAP_COMMON/etc/"
```
ktsakalozos, line 49: Under `configurations` could we have a disa-stig directory. That way we could then extend this approach to CIS hardening too.

k8s/resources/templates/disa-stig-bootstrap.yaml
```diff
@@ -0,0 +1,29 @@
+cluster-config:
+  network:
+    enabled: true
+  dns:
+    enabled: true
+  ingress:
+    enabled: true
+  load-balancer:
+    enabled: true
```
bschimke95: We don't enable all features by default. Let's stick to the defaults.
```
ubuntu@my-k8s-vm:~$ sudo k8s status
cluster status:           ready
control plane nodes:      10.90.77.24:6400 (voter)
high availability:        no
datastore:                etcd
network:                  enabled
dns:                      enabled at 10.152.183.50
ingress:                  disabled
load-balancer:            disabled
local-storage:            enabled at /var/snap/k8s/common/rawfile-storage
gateway                   enabled
```

k8s/resources/templates/disa-stig-bootstrap.yaml
bschimke95, line 1: Please add a basic clustering test that verifies a three node (cp, cp, wk) cluster comes up just fine with those configuration options.

tests/integration/tests/test_util/config.py
```diff
@@ -11,6 +11,7 @@
 DEFAULT_WAIT_RETRIES = int(os.getenv("TEST_DEFAULT_WAIT_RETRIES") or 120)
 DEFAULT_WAIT_DELAY_S = int(os.getenv("TEST_DEFAULT_WAIT_DELAY_S") or 5)
 
+COMMON_ETC_DIR = DIR / "var/snap/k8s/common/etc"
```
bschimke95: DIR will resolve to "<parents>/test_utils" so this path seems to be wrong. 
It should probably just be `/var/snap...`

deuberger:
@louiseschmidtgen @ktsakalozos @bschimke95 why are the auditing files in a disa_stig folder? They aren't specific to stig at all, right? One is a config to audit everything and the other is to audit just kube-system, right? Why wouldn't those be generic config options?

bschimke95 (approved):

ktsakalozos (approved):
LGTM +1

PR #1820, opened 2025-09-03
fix: Fix Invalid `iproute2` JSON output in arm64

mateoflorido:
## Description

Fix an invalid `iproute2` JSON output in arm64

## Solution

Given that all JSON is a [valid](https://stackoverflow.com/questions/21584985/what-valid-json-files-are-not-valid-yaml-1-1-files) YAML we can use its parser to avoid errors caused by invalid JSON.

## Issue
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2121908

## Backport

## Checklist

- [x] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed


src/k8s/pkg/utils/netlink.go
```diff
@@ -63,3 +69,21 @@ func RemoveLink(name string) error {
...
+// VNI value with the fan-map extension, resulting in invalid JSON.
+func fixInvalidIproute2JSON(input []byte) ([]byte, error) {
+	output := string(input)
+
+	// Target the specific case where numeric VXLAN VNI values are concatenated
+	// with text (like "0fan-map") without proper JSON quoting
+	re, err := regexp.Compile(`"id":\s*([0-9]+[a-zA-Z][a-zA-Z0-9_-]*)`)
+	if err != nil {
+		return nil, fmt.Errorf("failed to compile invalid VXLAN VNI regexp: %w", err)
+	}
```
addyess: wow, that's quite a narrow work-around

rapour:
@mateoflorido Thanks a lot for the workaround and LGTM. I think this can catch what we've seen so far. Even if it doesn't catch every edge case, a valid json value cannot start with a number and continue with a string, so it should be quite benign for any valid response and shouldn't do any harm. I'll leave the final approval to @addyess and @berkayoz

src/k8s/pkg/utils/netlink.go
```diff
@@ -30,7 +31,12 @@ func ListVXLANInterfaces() ([]VXLANInterface, error) {
 		return vxlanDevices, fmt.Errorf("running ip command failed: %s", string(out))
 	}
 
-	if err := json.Unmarshal(out, &ipLinks); err != nil {
+	// NOTE: (mateoflorido) Parsing as YAML cleans up invalid JSON output
+	// produced by the iproute2 command in arm64. Currently, the Ubuntu package
+	// combines the VXLAN VNI value with the fan-map extension, resulting in
+	// invalid JSON, but a valid YAML.
```
addyess, line 37: 😍

addyess (approved):
Yep, this is lovely and simple

PR #1839, opened 2025-09-10
test: Configurable runners for each test

bschimke95:
We’ve been encountering out-of-disk errors during the upgrade tests, which has highlighted the need for larger runners. Since we generally rely on GitHub-hosted runners, we can’t simply move all tests to self-hosted runners. To address this, we’ve introduced a new parameter, runner-tags, in the e2e workflow. This allows us to override the default runner tags on a per-test basis.

With this approach, we gain fine-grained control over the runner requirements for each test, ensuring that only the tests that truly need larger or specialized runners use self-hosted resources, while minimizing the overall usage of self-hosted runners.

### DIFF
--- a/.github/workflows/lint_and_integration.yaml
+++ b/.github/workflows/lint_and_integration.yaml
@@ -136,7 +136,6 @@ jobs:
       os: ubuntu:24.04
       test-tags: ${{ needs.get-e2e-tags.outputs.test-tags }}
       artifact: k8s-${{ matrix.arch }}.snap
-      extra-test-args: "-k test_version_upgrades_1"
       parallel: true
 
   security-scan:


.github/workflows/e2e-tests.yaml
```diff
@@ -110,6 +121,13 @@ jobs:
           python-version: "3.12"
       - name: Install tox
         run: sudo apt-get install -y tox
+      - name: Compute test runners
+        id: compute-test-runners
+        run: |
+          # Github Actions does not support dynamic concatination of lists, so we need to add the arch tag here.
+          # See https://github.com/orgs/community/discussions/11401
+          TEST_RUNNERS=$(jq -c -n '${{ inputs.runner-tags }}' | jq -c 'with_entries(.value += ["${{ inputs.arch }}"])')
```
HomayoonAlimohammadi, line 129: Verified this: ✔️ 
```
export TEST_RUNNERS=$(jq -c -n '{"tests/test_version_upgrades.py::test_version_upgrades": ["self-hosted", "linux", "jammy", "large"]}' | jq -c 'with_entries(.value += ["amd64"])')
echo $TEST_RUNNERS           
{"tests/test_version_upgrades.py::test_version_upgrades":["self-hosted","linux","jammy","large","amd64"]}
```

bschimke95: Yeah, way to much bash magic for my taste but I didn't see a good way around that if GH actions doesn't ship that by default.

PR #1865, opened 2025-09-16
fix: test_build 

rapour:
## Description

Currently the test_build is failing with the following error:
```
ERROR tests/test_build.py - TypeError: is_fips_enabled() missing 1 required positional argument
```

Additionally the `tests/test_airgapped.py::test_airgapped_with_image_mirror` test is switched to self-hosted runner because of disk limitations on GH runners.


rapour:
The `Test Branch Management` failure is irrelevant to the changes introduced in this PR

PR #1869, opened 2025-09-17
fix: error PRs where components change major versions on release branches

addyess:
## Description

Automated PRs raised by bots have the ability to step components on release branches pasts their maj.min revisions.  We want to only allow patch revisions to be updated on release branches. 

## Solution

This PR establishes a PR action to confirm the diff doesn't step upgrades any of the "components" past their maj.min revisions.

## Checklist

- [X] PR title formatted as `type: title`
- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/.github/workflows/component-version-check.yaml
+++ b/.github/workflows/component-version-check.yaml
@@ -86,7 +86,7 @@ jobs:
               )
             })
 
-      #Fail the job if invalid
+      # Fail the job if invalid
       - name: Fail job if invalid
         if: steps.validate.outputs.invalid == 'true'
         run: exit 1
--- a/build-scripts/hack/check-component-versions.py
+++ b/build-scripts/hack/check-component-versions.py
@@ -1,28 +1,28 @@
 #!/usr/bin/env python3
 import sys
 import re
-from packaging.version import Version
+import argparse
+from packaging.version import InvalidVersion, Version
 
-VERSION_FILE_PATTERN = re.compile(r'^(\+\+\+|---) a?\.?/build-scripts/components/[^/]+/version')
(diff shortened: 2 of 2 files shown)


build-scripts/hack/check-component-versions.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
+import sys
+import re
+from packaging.version import Version
+
+VERSION_FILE_PATTERN = re.compile(r'^(\+\+\+|---) a?\.?/build-scripts/components/[^/]+/version')
+
+def parse_version(line):
+    line = line.strip()[1:].strip()  # Remove leading + or -
```
bschimke95: What if the version does not start with +/-? This line would remove the first char regardless, right?

bschimke95: Ah, this is checked outside of the function already. I feel like this check should be part of this function but I might be wrong.

.github/workflows/component-version-check.yaml
```diff
@@ -0,0 +1,92 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: (
+                "❌ Invalid change detected in PR diff. Please remove it.\n" +
+                "Label this PR with 'skip-component-version-check' to ignore\n\n" +
+                "${{ steps.validate.outputs.stderr }}\n\n" +
+              )
+            })
+
+      #Fail the job if invalid
```
bschimke95: nit: whitespace after # (sorry)

build-scripts/hack/check-component-versions.py
addyess: Thanks @bschimke95 , i think this is a clearer solution.  Less lines too

PR #1871, opened 2025-09-18
fix: sanitize not-applicable feature gates in configure hook

berkayoz:
## Description

We've disabled certain feature gates in 1.34 #1790 via setting `Gate=False` in the apiserver arguments for `k8s-dqlite` backed cluster. On revert/downgrade these flags were not sanitized since they reside in `SNAP_COMMON` which results in the kube-apiserver crashing.

## Solution

Sanitizes feature gates that do not exist on the current kbuernetes version

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/k8s/lib.sh
+++ b/k8s/lib.sh
@@ -284,52 +284,56 @@ k8s::containerd::ensure_systemd_defaults() {
 }
 
 # Sanitize feature gates in kube-apiserver arguments
-# Usage: k8s::apiserver::sanitize_feature_gates "flag1" "flag2" "flag3"
-# Example: k8s::apiserver::sanitize_feature_gates "SizeBasedListCostEstimate" "OtherFlag"
+# This removes any feature gates that are not present in the current apiserver version
+# from the --feature-gates argument in /var/snap/k8s/common/args/kube-apiserver.
+# Usage: k8s::apiserver::sanitize_feature_gates
 k8s::apiserver::sanitize_feature_gates() {
   local args_file="/var/snap/k8s/common/args/kube-apiserver"
-  
+
   # Check if the args file exists
   if [ ! -f "$args_file" ]; then
     return 0
   fi
-  
-  # Get the list of feature gates to remove from function arguments
-  local flags_to_remove=("$@")
-  
-  # If no flags specified, return
-  if [ ${#flags_to_remove[@]} -eq 0 ]; then
-    return 0
-  fi
-  
+
   # Process the feature gates line if it exists
   if grep -q "^--feature-gates=" "$args_file"; then
+    # Get the list of supported feature gates from kube-apiserver
+    local supported_gates=""
+    if [ -x "/snap/k8s/current/bin/kube-apiserver" ]; then
+      # Extract feature gate names from help output (format: kube:FeatureName=true|false)
+      supported_gates=$(/snap/k8s/current/bin/kube-apiserver --help 2>/dev/null | awk '/^ *kube:/{print $1}' | sed 's/^kube://' | sed 's/=.*//')
+    fi
+
+    # If we couldn't get supported gates, return without changes
+    if [ -z "$supported_gates" ]; then
+      return 0
+    fi
+
+    # Convert supported gates to array
+    declare -A supported_gates_map
+    while IFS= read -r gate; do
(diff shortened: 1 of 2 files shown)


k8s/lib.sh
```diff
@@ -187,3 +187,69 @@ k8s::containerd::ensure_systemd_defaults() {
   fi
 
 }
+
+# Sanitize feature gates in kube-apiserver arguments
```
bschimke95, line 286: Please expand on this/add an example what this exactly means.

bschimke95 (commented):
Generally LGTM but please add a bit more context to the function (and maybe to the invocation as well). I don't think it is immediately clear what sanitizing feature gates means in this context.

Also, change the target to `main` and add the backport labels before merging this.

snap/hooks/configure
```diff
@@ -7,3 +7,6 @@ k8s::common::setup_env
 k8s::cmd::k8s x-snapd-config reconcile
 
 k8s::containerd::ensure_systemd_defaults
+
+# Sanitize kube-apiserver feature gates that don't exist in k8s 1.33
+k8s::apiserver::sanitize_feature_gates "DetectCacheInconsistency" "ListFromCacheSnapshot" "SizeBasedListCostEstimate"
```
addyess: i think i understand that this removes these feature gates (on the 1.33 snap) if they've been set by a previous install (previously 1.34) 

This is similar to giving a caveman instructions on which tools to avoid when repairing his tractor.  Doesn't it feel odd to tell a 1.33 cluster to avoid a specific set of 1.34 features?

I know i'm not offering an improvement here.  I guess there's no way to query from the 1.33 executable a list of "Available" features to enabled, and remove any that are no longer in that list of available features?


```
/snap/k8s/current/bin/kube-apiserver --help | awk '/^ *kube:/{print $1}'
```

provides a pretty compelling list -- but i'm not sure it's complete.

addyess (commented):
Is there a way to sanitize that's not so revision specific?

snap/hooks/configure
berkayoz: Good suggestion!

k8s/lib.sh
```diff
@@ -282,3 +282,73 @@ k8s::containerd::ensure_systemd_defaults() {
...
+k8s::apiserver::sanitize_feature_gates() {
+  local args_file="/var/snap/k8s/common/args/kube-apiserver"
+
+  # Check if the args file exists
+  if [ ! -f "$args_file" ]; then
+    return 0
+  fi
+
+  # Process the feature gates line if it exists
+  if grep -q "^--feature-gates=" "$args_file"; then
```
rapour: nit: Using a negative check may help with reading the rest of the conditions

k8s/lib.sh
```diff
@@ -282,3 +282,73 @@ k8s::containerd::ensure_systemd_defaults() {
...
+    local updated_gates=""
+
+    # Split by comma and filter out unsupported gates
+    IFS=',' read -ra gates <<< "$feature_gates_value"
+    for gate in "${gates[@]}"; do
+      local gate_name="${gate%%=*}"
+      local gate_value="${gate#*=}"
+
+      # Check if this gate is supported using associative array lookup
+      if [[ -n "${supported_gates_map[$gate_name]}" ]]; then
```
rapour: nit: The feature gate can be both in the format of `component:key=value` and `key=value` (which defaults to `kube` component) but the `supported_gates_map` has stripped the component and therefore the following argument would not pass:
```
--feature-gates 'kube:featureC=true'
```
even if `featureC` is in the `supported_gates_map`. If we are the only ones who modify these arguments a comment on the code to omit the component when defining the feature gates would suffice, otherwise it could be a good idea to check both formats.

addyess (approved):
Nicely done.  bash can do this job, but it doesn't make it easy

k8s/lib.sh
berkayoz: I've not seen that form utilized commonly, and mainly we care about the gates we add and want to remove them. Ultimately in the future with the  implementation of the new service args handling, we should not adjust the user provided config and let them be handled manually.

rapour: Fair enough, thank you

PR #1876, opened 2025-09-30
chore: add MTU detection integration test in a dual NIC environment

rapour:


tests/integration/tests/test_networking.py
```diff
@@ -172,3 +174,119 @@ def test_ipv6_only_on_dualstack_infra(instances: List[harness.Instance]):
...
+
+    util.wait_until_k8s_ready(cp_instance, instances)
+    util.wait_for_network(cp_instance)
+
+    util.set_node_labels(cp_instance, cp_instance.id, {"kubernetes.io/role": "master"})
+    util.set_node_labels(
+        cp_instance, worker_instance.id, {"kubernetes.io/role": "worker"}
+    )
+
+    manifest = MANIFESTS_DIR / "nginx-sticky-pod.yaml"
```
louiseschmidtgen, line 203: IIRC there is the existing nginx-pod.yaml from other test cases. Can we use it and its ready check?

tests/integration/tests/test_networking.py
```diff
@@ -172,3 +174,119 @@ def test_ipv6_only_on_dualstack_infra(instances: List[harness.Instance]):
...
+
+    # make sure the netshoot pod is scheduled on the control plane node
+    # while the nginx is scheduled on the worker node
+    cp_instance.exec(
+        [
+            "k8s",
+            "kubectl",
+            "run",
+            "netshoot",
+            "--image=ghcr.io/nicolaka/netshoot:v0.14",
```
louiseschmidtgen, line 230: Is this version going to be subject to much change? If yes we should make it configurable.

tests/integration/tests/test_util/harness/lxd.py
```diff
@@ -67,6 +67,22 @@ def __init__(self):
             ),
         )
 
+        self._configure_network(
+            config.LXD_JUMBO_NETWORK,
+            "ipv4.address=auto",
+            "ipv6.address=auto",
+            "ipv4.nat=true",
+            "ipv6.nat=true",
+            "bridge.mtu=9000",
```
louiseschmidtgen, line 76: Why choose 9000?

tests/integration/tests/test_networking.py
```diff
@@ -172,3 +174,119 @@ def test_ipv6_only_on_dualstack_infra(instances: List[harness.Instance]):
...
+            "-s",
+            "8000",
+            f"{nginx_pod_ip}",
+        ],
+        capture_output=True,
+    )
+
+    # 50 packets transmitted, 40 received, 20% packet loss, time 9109ms
+    assert "50 packets transmitted" in result.stdout.decode()
+    packet_loss_percentage = int(result.stdout.decode().split(", ")[1].split(" ")[0])
```
ktsakalozos: probably not the right variable name :)

.github/workflows/e2e-tests.yaml
```diff
@@ -47,7 +47,21 @@ on:
           (e.g. use ["self-hosted", "linux", "jammy", "xlarge"] instead of ["self-hosted-linux-amd64-noble-2xlarge"])
         type: string
         required: false
-        default: '{"tests/test_version_upgrades.py::test_version_upgrades": ["self-hosted", "linux", "jammy", "large"], "tests/test_airgapped.py::test_airgapped_with_image_mirror": ["self-hosted", "linux", "jammy", "large"]}'
+        default: |
```
ktsakalozos, line 50: In a different PR please.

tests/integration/tests/test_networking.py
rapour: This one is using `nodeSelector` to make sure it's scheduled on the worker node. Since it's a hard limit, setting it on the nginx-pod.yaml would make it unschedulable in other tests.

tests/integration/tests/test_networking.py
rapour: I don't think so, since it's only used for pinging the nginx pod. However, I would appreciate if you have a more prominent image in mind to use instead of this.

.github/workflows/e2e-tests.yaml
rapour: Done

tests/integration/tests/test_networking.py
rapour: Thanks a lot, fixed

tests/integration/tests/test_util/harness/lxd.py
rapour: It's actually arbitrary but 9000 is [the most popular size](https://en.wikipedia.org/wiki/Jumbo_frame) used for jumbo frames.

PR #1886, opened 2025-10-06
fix: Force removing lost nodes from the cluster

bschimke95:
## Description

The `--force` flag for `k8s remove-node` did not work for lost nodes because it required a node to be available to remove it.

This PR ensures that all mentions of the leaving node will be removed (etcd, microcluster, k8s-dqlite, Kubernetes) and not fail if the node does not exist already in one of them.

## Solution

1. Moved all membership operations from the `pre-remove` hook (which runs on the leaving node) to the k8sd `cluster/remove` API (which runs on any other cluster node)
2. Use the `name` instead of the `address` to remove a member from etcd. The existing logic resolved the address from the name in microcluster. If the cluster membership diverged and the node is already removed from microcluster but not etcd, this would not work, hence we resolve this by name.

## Commits

This PR is split into three distinctive commits:
1. Adds the code functionality
2. Updates/enables the test
3. Update docs

This makes a potential backport easier and makes it easier to review while still covering the full feature in one PR. I will keep those commits and not squash them on merge. 

## Backport

yes

## Checklist

- [ ] Covered by unit tests - already
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/docs/canonicalk8s/snap/howto/install/uninstall.md
+++ b/docs/canonicalk8s/snap/howto/install/uninstall.md
@@ -6,7 +6,9 @@ snap from your system.
 ## Remove the node from the cluster
 
 ```{important}
-You must remove the node from the cluster **before** deleting the snap. Deleting the snap while the node is part of the cluster can cause the node to enter an unreachable state.
+You must remove the node from the cluster **before** deleting the snap.
+Deleting the snap while the node is part of the cluster can cause the node to enter an unreachable state.
+In this case, the node needs to be removed with the `--force` flag as explained in the [troubleshooting page].
 ```
 
 From any control plane node:
@@ -18,7 +20,7 @@ sudo k8s remove-node <NODE_NAME>
 Ensure the node has been removed from the cluster:
 
 ```
-sudo k8s kubectl get nodes 
+sudo k8s kubectl get nodes
 ```
 
 ## Remove the k8s snap
@@ -51,3 +53,6 @@ This command should produce an output similar to:
 ```
 error: no matching snaps installed
 ```
+
+<!-- Links -->
+[troubleshooting page]: ../../reference/troubleshooting.md#remove-a-permanently-lost-node-from-the-cluster
--- a/docs/canonicalk8s/snap/reference/troubleshooting.md
+++ b/docs/canonicalk8s/snap/reference/troubleshooting.md
@@ -317,6 +317,36 @@ Verify the Cilium pod has restarted and is now in the running state:
 sudo k8s kubectl get pods -n kube-system
 ````
 
+## Remove a permanently lost node from the cluster
+
+A node that is permanently lost cannot be removed with `k8s remove-node`:
+
+```sh
+Error: Failed to remove node "t1" from the cluster.
+
+The error was: failed after potential retry: wait check failed: failed to POST /k8sd/cluster/remove:
+failed to delete cluster member t1: Post "https://10.23.245.80:6400/core/internal/hooks/pre-remove?target=t1":
+Unable to connect to "10.23.245.80:6400": dial tcp 10.23.245.80:6400: connect: no route to host
+```
+
+````{dropdown} Explanation
+By default, `k8s remove-node` attempts to contact the node being removed
+to execute cleanup routines.
+If the node is unreachable, the command fails with the error shown above.
+This can also happen if the node membership tracked by `k8sd` becomes
+inconsistent with the Kubernetes datastore.
+
+````
+
+````{dropdown} Solution
+
+Use the `--force` flag to forcibly remove the node:
+
+```
+sudo k8s remove-node --force <node-name>
+```
+````
+
 <!-- Links -->
 
 [troubleshooting how-to guide]: ../howto/troubleshooting.md


src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,172 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
-		return response.SyncResponse(true, nil)
+	// NOTE(hue): node removal process in CAPI might fail, we figured that the context passed to
+	// `DeleteClusterMember` is somehow getting canceled but couldn't figure out why or by which component.
+	// The cancellation happens after the `RunPreRemoveHook` call and before the `DeleteCoreClusterMember` call
+	// in `clusterMemberDelete` endpoint of microcluster. This is a workaround to avoid the cancellation.
+	// keep in mind that this failure is flaky and might not happen in every run.
+	deleteCtx, deleteCancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer deleteCancel()
+	log.Info("Deleting node from Microcluster cluster")
+	if err := c.DeleteClusterMember(deleteCtx, nodeName, force); err != nil {
```
louiseschmidtgen, line 295: If we are in a partially joined microcluster state does this call still work to remove any mention of the node in core_cluster_members, truststore, go-dqlite and dqlite?

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,172 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
 
-	cfg, err := databaseutil.GetClusterConfig(ctx, s)
+	return nil
+}
+
+func removeNodeFromMicrocluster(ctx context.Context, s state.State, nodeName string, force bool) error {
+	log := log.FromContext(ctx).WithValues("name", nodeName)
+
+	log.Info("Waiting for node to not be pending")
+	control.WaitUntilReady(ctx, func() (bool, error) {
```
louiseschmidtgen: This might be problematic if microcluster is in a bad state (the member is not in core_cluster_members but in dqlite). Hopefully, my fix will prevent that but we should ideally be able to surpass mc remove hook and nuke everything ourselves.

src/k8s/pkg/k8sd/app/hooks_remove.go
```diff
@@ -66,136 +89,41 @@ func (a *App) onPreRemove(ctx context.Context, s state.State, force bool) (rerr
...
-			if err != nil {
-				log.Error(err, "Failed to create Kubernetes client", err)
+		// NOTE(claudiub): We should only remove the certificates only if we're stopping the Kubernetes
+		// services as well. Removing them without stopping the services will result in the services
+		// being paralyzed and unable to continue their function, including potential Pod evictions
+		// started by CAPI.
+		if _, ok := cfg.Annotations.Get(apiv1_annotations.AnnotationSkipStopServicesOnRemove); !ok {
+			// Perform all cleanup steps regardless of if this is a worker node or control plane.
+			// Trying to detect the node type is not reliable as the node might have been marked as worker
+			// or not, depending on which step it failed.
```
louiseschmidtgen, line 71: Nice, thanks Ben!

tests/integration/tests/test_clustering.py
```diff
@@ -278,12 +278,12 @@ def test_concurrent_membership_restart_operations(instances: List[harness.Instan
 
 @pytest.mark.node_count(4)
 @pytest.mark.tags(tags.NIGHTLY)
-@pytest.mark.xfail(
-    reason="The node join while there is a dead node does not work currently due to a microcluster bug."
-)
 def test_node_join_succeeds_when_original_control_plane_is_down(
```
louiseschmidtgen, line 281: Nice Ben! Can we also check that if microcluster is in a broken state that half joined node can also be removed?

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,253 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
-			var notPending bool
-			if err := s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
-				member, err := cluster.GetCoreClusterMember(ctx, tx, req.Name)
+		if err := removeNodeFromDatastore(ctx, s, snap, req.Name, cfg); err != nil {
+			if req.Force {
+				// With force=true, we want to cleanup all out-of-sync mentions of this node.
+				// So we log the error, but continue.
+				log.Error(err, "Failed to remove node from datastore, but continuing due to force=true", "datastore", cfg.Datastore.GetType())
+			} else {
+				return response.InternalError(fmt.Errorf("failed to delete node from microcluster: %w", err))
```
HomayoonAlimohammadi: nit:

```suggestion
				return response.InternalError(fmt.Errorf("failed to delete node from datastore: %w", err))
```

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,253 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
+			}
+		}
```
HomayoonAlimohammadi, line 73: Maybe we can add a log here as well

```suggestion
		} else {
		    log.Info("Node removed from datastore")
		}
```

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,253 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
+		if err := removeNodeFromMicrocluster(ctx, s, req.Name, req.Force); err != nil {
+			if req.Force {
+				log.Error(err, "Failed to remove node from microcluster, but continuing due to force=true")
+			} else {
+				return response.InternalError(fmt.Errorf("failed to delete node from microcluster: %w", err))
+			}
+		}
+		log.Info("Node removed from microcluster")
```
HomayoonAlimohammadi: I think this log should happen in case of no error:

```suggestion
		} else {
		    log.Info("Node removed from microcluster")
		}
```

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,253 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
+			if req.Force {
+				// With force=true, we want to cleanup all out-of-sync mentions of this node.
+				// It might be that the node is already gone from k8s, but not from microcluster.
+				// So we log the error, but continue.
+				log.Error(err, "Failed to remove node from Kubernetes, but continuing due to force=true")
+			} else {
+				return response.InternalError(fmt.Errorf("failed to remove node from Kubernetes: %w", err))
+			}
+		}
+		log.Info("Node removed from Kubernetes cluster")
```
HomayoonAlimohammadi: Same thing goes here:

```suggestion
		} else {
		    log.Info("Node removed from Kubernetes cluster")
		}
```

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,253 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
+		if err := removeNodeFromEtcd(ctx, snap, s, clusterConfig, nodeName); err != nil {
+			log.Error(err, "Failed to remove node from etcd cluster")
+		}
+	case "external":
+		// The admin is responsible for cleaning up the external datastore membership.
+	default:
+	}
+
+	return nil
+}
```
HomayoonAlimohammadi, line 121: I think since this will never return an error, we won't be stopped if `--force` is not provided.

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,253 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
+
+		log.WithValues("kubernetes-members", node.Status.Addresses, "k8s-dqlite-members", members).Info("Matching Kubernetes node addresses with k8s-dqlite members")
+		for _, addr := range node.Status.Addresses {
+			for _, member := range members {
+				host, _, err := net.SplitHostPort(member.Address)
 				if err != nil {
-					log.Error(err, "Failed to get member")
-					return nil
+					log.Error(err, "Failed to split host and port from k8s-dqlite member address", "address", member.Address)
+				}
```
HomayoonAlimohammadi, line 159: Let's continue here:

```suggestion
					log.Error(err, "Failed to split host and port from k8s-dqlite member address", "address", member.Address)
					continue
				}
```

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,253 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
+
+	if nodeAddress == "" {
+		return fmt.Errorf("failed to resolve node address for node %q", nodeName)
+	}
+
+	// Remove port if present in the address
+	host, _, err := net.SplitHostPort(nodeAddress)
+	if err == nil && host != "" {
+		nodeAddress = host
+	}
```
HomayoonAlimohammadi: Shouldn't we handle `err != nil` as well?

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,253 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
+
+	nodeAddress = net.JoinHostPort(nodeAddress, fmt.Sprintf("%d", clusterConfig.Datastore.GetK8sDqlitePort()))
+
+	log.Info("Removing node from k8s-dqlite using address", "address", nodeAddress)
+	if err := client.RemoveNodeByAddress(ctx, nodeAddress); err != nil {
+		log.Error(err, "failed to remove node from k8s-dqlite cluster", "address", nodeAddress)
```
HomayoonAlimohammadi: Maybe we can also return an error here

src/k8s/pkg/k8sd/api/cluster_remove.go
```diff
@@ -35,77 +40,253 @@ func (e *Endpoints) postClusterRemove(s state.State, r *http.Request) response.R
...
+			if err != nil {
+				log.Error(err, "Failed to get member")
+				return nil
+			}
+			log.WithValues("role", member.Role).Info("Current node role")
+			notPending = member.Role != cluster.Pending
+			return nil
+		}); err != nil {
+			log.Error(err, "Transaction to check cluster member role failed")
+		}
```
HomayoonAlimohammadi, line 274: I think this transaction never fails (we're not returning any errors)

src/k8s/pkg/k8sd/api/cluster_remove.go
bschimke95: No, this is a bit more involved and required aligning with the microcluster team. I'll address this in follow-up work next.

src/k8s/pkg/k8sd/api/cluster_remove.go
bschimke95: Great catch, fixed!

tests/integration/tests/test_clustering.py
bschimke95: I have a test for that but decided to move this into a separate PR due to the dependency on microcluster (see comment above).

src/k8s/pkg/k8sd/api/cluster_remove.go
bschimke95: fixed, great catch

PR #1888, opened 2025-10-06
docs: Security explanation rework

nhennigan:
## Description

We need to update our docs to include the latest info on DISA STIG and FIPS.  This is the first PR where I will tackle the explanation pages. 

## Solution

- Moving all security pages into one to avoid duplication and make navigation easier. The pages are essentially the same with a few minor tweaks for flow and making one consistent document
- I have also added FIPS and DISA STIG compliance sections. 
- There is a TODO in the link section that I will complete once I have the other PRs up for review. 

## Issue

n/a

## Backport

1.34. The FIPS work is only available from 1.34 onwards

## Checklist

- [ ] PR title formatted as `type: title`
- [ ] Covered by unit tests
- [ ] Documentation updated
- [ ] CLA signed


docs/canonicalk8s/snap/explanation/security.md
```diff
@@ -0,0 +1,387 @@
...
+and Technology ([NIST]). All of our components including the built-in features
+such as networking or load-balancer can be configured to use host systems
+FIPS compliant libraries instead of the non-compliant internal go cryptographic
+modules. When building workloads on top of {{ product }}, it is essential that
+organizations build these in a FIPS compliant manner to comply with the FIPS
+security requirements. In addition, FIPS 140-3 has additional requirements to
+the system and hardware that have to be met in order to be fully FIPS compliant.
+
+If you would like to enable FIPS in your Kubernetes cluster see our
+[Canonical Kubernetes FIPS installation] guide.
```
louiseschmidtgen: Should we swap these two around? Usually, you would apply FIPS before DISA STIG?

louiseschmidtgen (commented):
Nice work @nhennigan! It certainly condenses the security pages.

One thought on making the page size a bit more manageable would be to split it into "security" and "enterprise-security" where we would put all the hardening explanations.

Just a suggestion, let me know what you think!

nhennigan:
Thanks for the review @rapour. I agree that it can definitely be refined for the charm narrative. Currently we do not offer DISA STIG compliance for the k8s operator but I did feel compliance was an important part of the security story for our product so I tried to make that distinction clear. I have updated the reference to k8s config to say at the snap level, was there any other things I missed? I know there are plans to improve the overall security docs for the charm in the backlog.

PR #1894, opened 2025-10-08
fix: Ensure snap services are running after a revert

HomayoonAlimohammadi:
### Overview

Snap services remain `disabled` after a `snap revert`. This PR ensures that services are running on both control-plane and worker nodes.

### DIFF
--- a/src/k8s/pkg/k8sd/app/hooks_start.go
+++ b/src/k8s/pkg/k8sd/app/hooks_start.go
@@ -19,16 +19,16 @@ import (
 )
 
 func (a *App) onStart(ctx context.Context, s state.State) error {
+	if err := a.ensureRunningServices(ctx); err != nil {
+		return fmt.Errorf("failed to ensure running services: %w", err)
+	}
+
 	// start a goroutine to mark the node as running
 	go func() {
 		if err := a.markNodeReady(ctx, s); err != nil {
 			log.FromContext(ctx).Error(err, "Failed to mark node as ready")
 		}
 	}()
-
-	if err := a.ensureRunningServices(ctx); err != nil {
-		return fmt.Errorf("failed to ensure running services: %w", err)
-	}
 
 	// tune system settings if enabled and necessary
 	if !a.snap.Strict() {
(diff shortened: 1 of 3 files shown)


src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -24,6 +26,10 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
 		}
 	}()
 
+	if err := a.ensureRunningServices(ctx); err != nil {
```
rapour: Does it make sense to call `markNodeReady` after ensuring the services are running?

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -142,3 +148,61 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
...
+
+	// Start services
+	// This may fail if the node controllers try to restart the services at the same time, hence the retry.
+	if isWorker {
+		log.Info("Starting worker services")
+		if err := control.RetryFor(ctx, 5, 5*time.Second, func() error {
+			if err := snaputil.StartWorkerServices(ctx, a.snap); err != nil {
+				return fmt.Errorf("failed to start services: %w", err)
+			}
+			return nil
```
rapour: Since we are wrapping the error on the outer block and have `failed to start worker services after retry`, it might be a good idea to remove the inner wrap for the sake of simplicity.
```suggestion
			return snaputil.StartWorkerServices(ctx, a.snap)
```

src/k8s/pkg/k8sd/app/hooks_start.go
```diff
@@ -142,3 +148,61 @@ func (a *App) onStart(ctx context.Context, s state.State) error {
...
+			return fmt.Errorf("failed to retrieve cluster configuration: %w", err)
+		}
+
+		log.Info("Starting control-plane services")
+
+		if err := control.RetryFor(ctx, 5, 5*time.Second, func() error {
+			if err := startControlPlaneServices(ctx, a.snap, cfg.Datastore.GetType()); err != nil {
+				return fmt.Errorf("failed to start services: %w", err)
+			}
+			return nil
```
rapour: Same as the worker node services

tests/integration/tests/test_snap_services.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Test that snap services are running after a `snap revert` instances.
+    """
+
+    cp = instances[0]
+    worker = instances[1]
+    token = util.get_join_token(cp, worker, "--worker")
+    util.join_cluster(worker, token)
+
+    refresh_to = "1.33-classic"
```
rapour: Is there a way we can avoid hardcoding the channel here? For example using something like `snap.get_most_stable_channels` like we do in version upgrades test?

HomayoonAlimohammadi: Ah this one slipped, thanks for pointing this out. Let me fix it.

PR #1899, opened 2025-10-14
test: move common e2e inputs to an anchor

rapour:
## Description

Currently the debugging workflow [would fail](https://github.com/canonical/k8s-snap/actions/runs/18497011176) because some of the inputs are missing from the `workflow_dispatch`. 

This PR 
1) Adds a common input to be used in both  `workflow_call` and `workflow_dispatch`. 
2) Removes some of the tests running on self hosted runners since they are now fine to be run on GH runners


.github/workflows/e2e-tests.yaml
```diff
@@ -111,43 +77,18 @@ on:
...
-        description: Substrate to use when running e2e tests
-        default: lxd
-        type: string
+      arch: *arch
+      os: *os
+      substrate: *substrate
       channel:
         description: k8s snap channel
         type: string
         default: latest/edge
```
rapour, line 90: Should we do the same for channel since one of them doesn't have a default value?

ktsakalozos (approved):
Less code less bugs! LGTM +1

PR #1904, opened 2025-10-16
chore: update cilium image tag

rapour:
## Description

Previous Cilium rock image was failing to remove temporary files in case the CNI binary call failed for whatever reason. This PR updates the rock image tag that contains the fix.


berkayoz (approved):
I assume this includes the fix from https://github.com/canonical/cilium-rocks/pull/27, lgtm!

PR #1890, opened 2025-10-06
docs: Security howto rework

nhennigan:
## Description

We need to update our docs to include the latest info on DISA STIG and FIPS. This is the first PR where I will tackle the how-to pages.

## Solution

- Added install in fips mode page 
- Added install with disa stig page 
-  Removed disa stig assessment page as that is now in reference (other pr)
- Removed CIS audit steps as that is also now in reference (other pr)

## Issue

n/a

## Backport

1.34

## Checklist

- [ ] PR title formatted as `type: title`
- [ ] Covered by unit tests
- [ ] Covered by integration tests
- [ ] Documentation updated
- [ ] CLA signed
- [ ] Backport label added if necessary 

If any item on the checklist is not complete, please provide justification why.

PR #1889, opened 2025-10-06
docs: Security refrence rework

nhennigan:
## Description

We need to update our docs to include the latest info on DISA STIG and FIPS. This is the first PR where I will tackle the reference pages.

## Solution

- I have moved the CIS manual audit to reference 
- I have moved the DISA STIG manual audit to reference. This is taken from @louiseschmidtgen PR #1882 disa-stig-assessment.md
- I have updated the links in the security reference page (There are TODOs but I will fill those in once I have all PRs up)

## Issue

N/A

## Backport

1.34 only

## Checklist

- [ ] PR title formatted as `type: title`
- [ ] Covered by unit tests
- [ ] Covered by integration tests
- [ ] Documentation updated
- [ ] CLA signed
- [ ] Backport label added if necessary 

If any item on the checklist is not complete, please provide justification why.

PR #1911, opened 2025-10-16
docs: fips+stig documentation release branch

louiseschmidtgen:
## Fips+stig documentation release branch

This PR holds changes from multiple PRs all related to the fips/stig documentation.
These include:

- [how to](https://github.com/canonical/k8s-snap/pull/1890)
-  [reference](https://github.com/canonical/k8s-snap/pull/1889)
- [explanation](https://github.com/canonical/k8s-snap/pull/1888) 
- updated 1.34 release notes

❗ PR should be rebased and merged NOT squashed ❗ 

## Backport
1.34

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/docs/canonicalk8s/snap/howto/install/disa-stig.md
+++ b/docs/canonicalk8s/snap/howto/install/disa-stig.md
@@ -5,11 +5,10 @@ Information System Agency (DISA) and are comprehensive frameworks of security
 requirements designed to protect U.S. Department of Defense (DoD) systems and
 networks from cybersecurity threats.
 
-The [Kubernetes STIG] contains guidelines on how to check and remediate various
-potential security concerns for a Kubernetes deployment, both on the host and
-within the cluster itself. This guide goes through the steps needed to deploy
-{{product}} with our dedicated security configuration files to meet DISA STIG
-hardening requirements.
+{{product}} aligns by default with many of these recommendations as they are
+expected to benefit most users. The additional recommendations are covered in
+this guide as well as the end-to-end procedure to be DISA STIG compliant at all
+layers including FIPS 140-3 compliance and host hardening.
 
 ## Prerequisites
 
@@ -18,10 +17,22 @@ This guide assumes the following:
 - Ubuntu machine with at least 4GB of RAM and 30 GB disk storage
 - You have root or sudo access to the machine
 - Internet access on the machine
-- You have Ubuntu Pro enabled on your system. For more information, see
-  [Ubuntu Pro documentation]
-- You have FIPS enabled on your machine. See our [FIPS installation guide] for
-guidance
+
+## Firewall configuration
+
+It is recommended to enable a host firewall (such as UFW) for DISA STIG
+compliance, but this is not done automatically. If you choose to enable a
+firewall, ensure you apply the required rules as described in the
+[firewall configuration] guide to avoid connectivity issues.
+
+## Install {{product}} in FIPS mode
+
+DISA STIG compliance also incorporates FIPS compliance. Start by following our
+[FIPS installation guide], but stop after installing {{product}} and do not
+follow the steps to bootstrap the cluster. Instead continue here as there
+are additional steps and configuration needed for DISA STIG compliance. The
+guide also covers enabling [Ubuntu Pro], which is needed for some of the
+steps below.
 
 ## DISA STIG host compliance
 
@@ -42,8 +53,7 @@ sudo usg audit disa_stig
 
 ```{warning}
 The following command applies rule [V-270714] which will cause issues using
-accounts with an empty password to access your machine. To avoid being locked
-out use an RSA key to access your machine.
+accounts with an empty password to access your machine.
 ```
 
 To automatically apply the recommended hardening changes:
@@ -80,16 +90,17 @@ overridden by another higher order file.
 ## Deploy and bootstrap STIG-compliant nodes
 
 For your convenience, we have provided template configuration files that can be
-used to configure Kubernetes service arguments that align with DISA STIG
-requirements.
+used to configure the additional Kubernetes service arguments that are needed to
+fully align with DISA STIG requirements.
 
 ### Bootstrap the first control-plane node
 
-Install {{product}}:
-
-```{literalinclude} ../../../_parts/install.md
-:start-after: <!-- snap start -->
-:end-before: <!-- snap end -->
+```{attention}
+Before bootstrapping the first control-plane node, review the available bootstrap
+configuration templates to ensure you select the one that best fits your
+requirements (for example, stricter or more permissive control-plane settings).
+Once a node is bootstrapped, changing certain settings may require re-deploying
+the node. See Alternative configurations below for more details.
 ```
 
 Initialize the first control plane node with the necessary arguments for
@@ -100,31 +111,25 @@ sudo k8s bootstrap --file /var/snap/k8s/common/etc/templates/disa-stig/bootstrap
 sudo k8s status --wait-ready
 ```
 
-Through this configuration file, the following rules are applied:
-| STIG | Summary |
-|------|---------|
-|[V-242434] |Kubernetes Kubelet must enable kernel protection|
-| [V-245541] |Kubernetes Kubelet must not disable timeouts|
-| [V-242402], [V-242403], [V-242461], [V-242462], [V-242463], [V-242464], [V-242465] |The Kubernetes API Server must have an audit log configured|
-| [V-254800] | Kubernetes must have a Pod Security Admission control file configured |
-| [V-242400] | The Kubernetes API server must have Alpha APIs disabled|
-| [V-242384] | The Kubernetes Scheduler must have secure binding |
-| [V-242385] | The Kubernetes Controller Manager must have secure binding |
+This configuration file applies the following rules:  
+
+| STIG                                                                               | Summary                                                               |
+| ---------------------------------------------------------------------------------- | --------------------------------------------------------------------- |
+| [V-242434]                                                                         | Kubernetes Kubelet must enable kernel protection                      |
+| [V-245541]                                                                         | Kubernetes Kubelet must not disable timeouts                          |
+| [V-242402], [V-242403], [V-242461], [V-242462], [V-242463], [V-242464], [V-242465] | The Kubernetes API Server must have an audit log configured           |
+| [V-254800]                                                                         | Kubernetes must have a Pod Security Admission control file configured |
+| [V-242400]                                                                         | The Kubernetes API server must have Alpha APIs disabled               |
+| [V-242384]                                                                         | The Kubernetes Scheduler must have secure binding                     |
+| [V-242385]                                                                         | The Kubernetes Controller Manager must have secure binding            |
 
 ### Join additional control plane nodes
 
-To join another control plane node to the cluster, first retrieve the join token
-from an existing control plane node:
+On the additional control plane node, join the cluster with the DISA STIG
+control plane configuration file:  
 
 ```
 sudo k8s get-join-token <joining-node-hostname>
-```
-
-On the joining control plane node install {{product}}:
-
-```{literalinclude} ../../../_parts/install.md
-:start-after: <!-- snap start -->
-:end-before: <!-- snap end -->
 ```
 
 and join the cluster with the DISA STIG control plane configuration file:
@@ -133,8 +138,8 @@ and join the cluster with the DISA STIG control plane configuration file:
 sudo k8s join-cluster --file=/var/snap/k8s/common/etc/templates/disa-stig/control-plane.yaml <join-token>
 ```
 
-Through this configuration file, the same rules that were applied to the first
-control plane node have also been applied.
+This configuration file applies the same rules shown above when bootstrapping
+the first control plane node.
 
 ### Join worker nodes
 
@@ -145,28 +150,22 @@ existing control plane node:
 sudo k8s get-join-token <joining-node-hostname> --worker
 ```
 
-On the joining worker node, install {{product}}:
-
-```{literalinclude} ../../../_parts/install.md
-:start-after: <!-- snap start -->
-:end-before: <!-- snap end -->
-```
-
-and join the cluster with the DISA STIG worker configuration file:
+On the worker node, join the cluster with the DISA STIG worker configuration
+file:
 
 ```
 sudo k8s join-cluster --file=/var/snap/k8s/common/etc/templates/disa-stig/worker.yaml <join-token>
 ```
 
-Through this configuration file, the following rules are applied:
+This configuration file applies the following rules:  
 
-| STIG | Summary |
-|------|---------|
-| [V-242434]| Kubernetes Kubelet must enable kernel protection |
-| [V-245541]| Kubernetes Kubelet must not disable timeouts |
+| STIG       | Summary                                          |
+| ---------- | ------------------------------------------------ |
+| [V-242434] | Kubernetes Kubelet must enable kernel protection |
+| [V-245541] | Kubernetes Kubelet must not disable timeouts     |
 
 If SSH is not needed to access the worker nodes it is recommended you disable
-the SSH:
+the SSH service:
 
 ```
 sudo systemctl disable ssh.service ssh.socket
@@ -175,8 +174,8 @@ sudo systemctl disable ssh.service ssh.socket
 ```{note}
 According to rule [V-242393] and [V-242394] Kubernetes Worker Nodes must not have
 sshd service running or enabled. The host STIG rule [V-270665] on the other hand
-expects sshd to be installed on the host. To comply with both rules, leave SSH
-installed, but disable the service.
+expects sshd to be installed on the host. To comply with both rules, we
+recommend leaving SSH installed, but disabling the service.  
 ```
 
 ## Alternative configurations
@@ -186,18 +185,27 @@ installed, but disable the service.
 To comply with rule [V-254800], you must configure a Pod Security Admission
 control file for your Kubernetes cluster. This file defines the Pod Security
 Standards (PSS) that are enforced at the namespace level. The STIG templates we
-provide to bootstrap/join nodes configure the Pod Security admission controller
-to comply with these recommendations. For your own cluster, assess if:
+provide to bootstrap/join control plane nodes configure the Pod Security
+admission controller to comply with these recommendations.
 
-- The default config at
-`/var/snap/k8s/common/etc/configurations/pod-security-admission-baseline.yaml`
-meets your needs or if it needs to be adjusted
-- Using the more restrictive config we provide at
-`/var/snap/k8s/common/etc/configurations/pod-security-admission-restricted.yaml`
-is more suited to your needs
-- You need to create your own audit policy based on the
-[upstream instructions] and adjust the `--admission-control-config-file` path
-used when you bootstrap/join nodes.
+By default, the bootstrap configuration template will point to
+`/var/snap/k8s/common/etc/configurations/pod-security-admission-baseline.yaml`,
+which sets the pod security policy to “baseline”, a minimally restrictive
+policy that prevents known privilege escalations.
+
+This policy may be insufficient or impractical in some situations, in which case
+the settings would need to be adjusted by doing one of the following:
+
+1. Adjust the `--admission-control-config-file` path used when you
+   bootstrap/join nodes to
+   `/var/snap/k8s/common/etc/configurations/pod-security-admission-restricted.yaml`
+   rather than the file above. This sets a more restrictive policy.
+2. Edit
+   `/var/snap/k8s/common/etc/configurations/pod-security-admission-baseline.yaml`
+   to suit your needs based on the [upstream instructions].
+3. Create your own audit policy based on the [upstream instructions] and adjust
+   the `--admission-control-config-file` path used when you bootstrap/join
+   nodes.
 
 For more details, see the [Kubernetes Pod Security Admission documentation],
 which provides an overview of Pod Security Standards (PSS), their enforcement
@@ -205,72 +213,66 @@ levels, and configuration options.
 
 ### Kubernetes API Server audit log
 
-This applies to rules [V-242402], [V-242403], [V-242461], [V-242462],
-[V-242463],[V-242464] and [V-242465]. The STIG templates we provide to
-bootstrap/join nodes configures the Kubernetes
+To comply with rules [V-242402], [V-242403], [V-242461], [V-242462],
+[V-242463],[V-242464] and [V-242465] you must configure the Kubernetes API
+Server audit log. The STIG templates we provide to
+bootstrap/join control plane nodes configures the Kubernetes
 API servers audit settings and policy to comply with these recommendations.
 
-The default level of logging may be impractical for some situations, in which
-case the settings would need to be adjusted and an exception put in place. For
-your own cluster, assess if:
-
-- The default logging configuration set at
+By default, the bootstrap configuration template will point to
 `/var/snap/k8s/common/etc/configurations/audit-policy.yaml`, which configures
 logging of all (non-resource) events with request metadata, request body, and
-response body is suitable for your needs or if it needs to be adjusted
-- Using the audit policy we provide at
-`/var/snap/k8s/common/etc/configurations/audit-policy-kube-system.yaml` which
-configures the same level of logging but only for events in the kube-system
-namespace is more suited to your needs
-- You need to create your own audit policy based on the
-[upstream audit instructions] and adjust the `--audit-policy-file` path used
-when you bootstrap/join nodes to use it.
+response body as recommended by [V-242403].
 
-In the default bootstrap configuration, Canonical Kubernetes does not enable
-audit logging as it may incur
-performance penalties in the form of increased disk I/O, which can lead to
-slower response times and reduced overall cluster efficiency, especially under
-heavy workloads.
+This level of logging may be impractical for some situations, in which case the
+settings would need to be adjusted and an exception put in place. To adjust the
+audit settings, do one of the following:
 
-## Manual checks
+1. Adjust the `--audit-policy-file` path used when you bootstrap/join nodes to
+   use `/var/snap/k8s/common/etc/configurations/audit-policy-kube-system.yaml`
+   rather than the file above. This configures the same level of logging but
+   only for events in the kube-system namespace.
+2. Edit `/var/snap/k8s/common/etc/configurations/audit-policy.yaml` to suit
+   your needs based on the [upstream audit instructions] for this policy
+   file.
+3. Create your own audit policy based on the
+   [upstream audit instructions] and adjust the `--audit-policy-file` path
+   used when you bootstrap/join nodes to use it.
 
-There are certain rules that must be completed manually in order to achieve
-DISA STIG compliance.
+## Required Documentation
 
-- [V-242383]  User-managed resources must be created in dedicated namespaces so
-you must manually inspect all services in the default namespace
-- [V-242410], [V-242411], [V-242412], [V-242413] The Kubernetes API Server,
-Scheduler, Controllers and etcd must enforce ports, protocols, and services
-(PPS) that adhere to the Ports, Protocols, and Services Management Category
-Assurance List (PPSM CAL). The {{product}} [ports and services] must be
-manually reviewed in accordance with this policy
-- [V-242414] The Kubernetes cluster must use non-privileged host ports for user
-pods so you must manually inspect all pods in the default namespace
-- [V-242415] Secrets in Kubernetes must not be stored as environment variables
-so you must manually inspect all pods in the default namespace
-- [V-242417] Kubernetes must separate user functionality
-so you must manually inspect all pods in the default namespace
+[V-242410], [V-242411], [V-242412], and [V-242413] requires that the Kubernetes
+API Server, Scheduler, and Controllers as well as etcd enforce ports, protocols,
+and services (PPS) that adhere to the Ports, Protocols, and Services Management
+Category Assurance List (PPSM CAL). The {{product}} [ports and services] must be
+manually audited in accordance with this policy. These ports, protocols, and
+services will need to be added to your specific PPSM list and the list will need
+to be updated anytime the list of ports, protocols, and services used by your
+cluster changes. For instance, this list will need to be updated each time a
+new service is exposed externally.
 
+## Post-Deployment Requirements
 
-## Next steps
+In addition to the above deployment steps, there are some guidelines that must
+be followed by users and administrators throughout the life of the cluster in
+order to maintain DISA STIG compliance.
 
-Please assess your cluster for compliance using the [DISA STIG auditing page].
-Review all findings and apply any necessary remediations to be fully DISA STIG
-compliant. Be aware that some rules need to be upheld when you add workloads
-to your cluster.
+- [V-242383]  User-managed resources must be created in dedicated namespaces
+- [V-242414] User pods must only use non-privileged host ports
+- [V-242415] Secrets must not be stored as environment variables
+- [V-242417] User functionality must be separate from management functions
+  meaning all user pods must be in user specific namespaces rather than system
+  namespaces
 
 <!-- Links -->
 [ports and services]: /snap/reference/ports-and-services/
 [FIPS installation guide]: fips.md
-[Kubernetes STIG]: https://stigviewer.com/stigs/kubernetes
-[DISA STIG auditing page]: /snap/reference/disa-stig-audit.md
+[firewall configuration]: /snap/howto/networking/ufw.md
 [USG tool]: https://documentation.ubuntu.com/security/docs/compliance/usg/
-[Ubuntu Pro documentation]: https://documentation.ubuntu.com/pro/start-here/#start-here
+[Ubuntu Pro]: https://documentation.ubuntu.com/pro/start-here/#start-here
 [Kubernetes Pod Security Admission documentation]: https://kubernetes.io/docs/concepts/security/pod-security-admission/
 [upstream instructions]: https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-admission-controller/
 [upstream audit instructions]: https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/
-[V-270714]: https://stigviewer.com/stigs/canonical_ubuntu_24.04_lts/2025-02-18/finding/V-270714
-[V-270665]: https://stigviewer.com/stigs/canonical_ubuntu_24.04_lts/2025-02-18/finding/V-270665
 [V-242403]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242403
 [V-242434]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242434
 [V-245541]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-245541
@@ -284,8 +286,6 @@ to your cluster.
 [V-242400]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242400
 [V-242384]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242384
 [V-242385]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242385
-[V-242393]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242393
-[V-242394]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242394
 [V-242383]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242383
 [V-242410]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242410
 [V-242411]: https://stigviewer.com/stigs/kubernetes/2025-02-20/finding/V-242411
--- a/docs/canonicalk8s/snap/howto/install/fips.md
+++ b/docs/canonicalk8s/snap/howto/install/fips.md
@@ -1,13 +1,14 @@
 # How to install a FIPS compliant Kubernetes cluster
 
-[FIPS 140-3] (Federal Information Processing Standards) ensures security
-compliance crucial for US government and regulated industries. This
-how-to guide provides the steps to set up a FIPS compliant Kubernetes
-cluster using the {{ product }} snap.
+The [Federal Information Processing Standard] (FIPS) 140-3 is a US government
+security standard regulating the use of cryptography. Compliance is crucial for
+US government and regulated industries. This how-to guide provides the steps to
+set up a FIPS compliant Kubernetes cluster using the
+{{ product }} snap.
 
-Please note that FIPS is only available in the `k8s` release 1.34 and later.
-If you are using an earlier version, you will need to upgrade to the latest
-version of the snap to use FIPS support.
+Please note that FIPS is only available in the `k8s` snap release 1.34 and
+later. If you are using an earlier version, you will need to upgrade to
+a newer version of the snap to use FIPS mode.
 
 ## Prerequisites
 
@@ -70,25 +71,19 @@ If this section leaves open any further questions consult the [enable FIPS with 
 guide for more detailed instructions.
 ```
 
-## Firewall configuration for Kubernetes
-
-{{ product }} requires certain firewall rules and guidelines to
-ensure its operation. Additional firewall rules may also be necessary based on
-user deployed workloads and services. Please follow the steps in the
-[firewall configuration] guide.
-
 ## Ensure runtime with FIPS-certified libraries
 
-Install the [core22] runtime with FIPS-certified libraries. The core22 snap
-offers the `fips-updates` track, which contains NIST-certified packages along
-with [security patches].
+Install the [core22] runtime with FIPS-certified libraries from the
+`fips-updates` track, which contains NIST-certified packages along with
+[security patches].
 
 ```
 sudo snap install core22 --channel=fips-updates/stable
 ```
 
-In case you have core22 already installed, perform a snap refresh to update it
-to the latest version:
+If core22 is already installed, a message will be displayed:
+`snap "core22" is already installed, see 'snap help refresh'`.
+In this case, use the refresh command instead of install.
 
 ```
 sudo snap refresh core22 --channel=fips-updates/stable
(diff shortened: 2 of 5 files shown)


nhennigan (approved):
Thank you for all your hard work on this 😸

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
+# How to install a FIPS compliant Kubernetes cluster
+
+[FIPS 140-3] (Federal Information Processing Standards) ensures security
+compliance crucial for US government and regulated industries. This
+how-to guide provides the steps to set up a FIPS compliant Kubernetes
+cluster using the {{ product }} snap.
+
+Please note that FIPS is only available in the `k8s` release 1.34 and later.
```
deuberger:
```suggestion
Please note that FIPS is only available in the `k8s` snap release 1.34 and later.
```

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
+# How to install a FIPS compliant Kubernetes cluster
+
+[FIPS 140-3] (Federal Information Processing Standards) ensures security
```
deuberger: Tt was strange to me to see the abbreviation expanded after the numeric designation. I don't know if that we need to expand it at all, but I always though the correct way would have been "FIPS (Federal Information Processing Standards) 140-3". Maybe I'm wrong?

deuberger:
```suggestion
[FIPS 140-3] (Federal Information Processing Standards) is a US government security standard
```

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+compliance crucial for US government and regulated industries. This
```
deuberger:
```suggestion
regulating the use of cryptography. This
```

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+how-to guide provides the steps to set up a FIPS compliant Kubernetes
+cluster using the {{ product }} snap.
+
+Please note that FIPS is only available in the `k8s` release 1.34 and later.
+If you are using an earlier version, you will need to upgrade to the latest
```
deuberger:
```suggestion
If you are using an earlier version, you will need to upgrade to a newer
```

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+version of the snap to use FIPS support.
```
deuberger:
```suggestion
version of the snap to use FIPS mode.
```

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+If the output is `1`, your host machine is running in FIPS mode.
+
+``` {note}
+If this section leaves open any further questions consult the [enable FIPS with Ubuntu]
+guide for more detailed instructions.
+```
+
+## Firewall configuration for Kubernetes
```
deuberger: I don't think we need this in the FIPS how to right @ktsakalozos @bschimke95? It was in the end to end guide because ufw is enabled as a part of DISA STIG. But this guide doesn't include STIG, so we can assume either it's off or properly configured, but it's out of scope of this FIPS how to.

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+```
+sudo snap install core22 --channel=fips-updates/stable
+```
+
+In case you have core22 already installed, perform a snap refresh to update it
+to the latest version:
+
+```
+sudo snap refresh core22 --channel=fips-updates/stable
+```
```
deuberger, line 90: This section needs updated to match what's in the end-to-end guide.

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+
+Install {{ product }} on your FIPS host:
+
+```{literalinclude} /_parts/install.md
+:start-after: <!-- snap start -->
+:end-before: <!-- snap end -->
+```
+
+The k8s snap can leverage the host's FIPS compliant
+cryptography. The components will automatically detect if the system is
```
deuberger:
```suggestion
The installed components will automatically detect if the system is
```
I suggest we just delete the first sentence because it's a bit confusing and doesn't really add anything. We had a different version in the end to end guide, but I didn't like it either and removed it.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+kernel.panic_on_oops=1
+EOF
+sudo sysctl --system
+```
+
+```{note}
+Please ensure that the configuration of `/etc/sysctl.d/99-kubelet.conf` is not
+overridden by another higher order file.
+```
+
```
deuberger, line 89: FW config is missing from this guide. It looks like it ended up in the FIPS guide when it really belongs here. So probably that section should be moved there. The language needs updated though to explain why/if/when it's needed. IIRC, the host stig recommends enabling UFW, but does not do so by default. Instead it prints something saying it's recommended, but leaves it to the user to do so. One user followed this recommendation and ran into issues which is why we explain the rules that are needed now.

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+running in FIPS mode and activate internal FIPS-related settings
+accordingly.
+
+```{attention}
+If you intend to apply DISA STIG hardening to your cluster, go to the [Canonical Kubernetes DISA STIG deployment guide](disa-stig.md) to get detailed instructions on deploying with a stricter bootstrap configuration file or joining the cluster with a stricter join configuration file.
+```
```
deuberger, line 113:
```suggestion
## Bootstrap the cluster

```{attention}
If you are deploying a DISA STIG hardened cluster, stop here and instead continue following the [the DISA STIG deployment guide](disa-stig.md).
```
```

Having this in a separate section will make it easier to see where to stop if deploying a STIGed machine.

File: docs/canonicalk8s/snap/howto/install/fips.md [hunk 11]
```diff
@@ -0,0 +1,182 @@
...
+After the snap installation completes, you can bootstrap the node as usual:
+
+```
+sudo k8s bootstrap
+```
+
+Then you may wait for the node to be ready, by running:
+
+```
+sudo k8s status
```
deuberger: Why aren't we using ` --wait-ready` like in the end to end guide? Should we? Maybe it's not necessary since we don't have join instructions here?

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+Your Kubernetes cluster is now ready for workload deployment and
+additional node integrations. Please ensure that your workloads and
+underlying system and hardware are FIPS compliant as well, to
+maintain the security standards required by FIPS. For example,
+ensure that your container images used for your applications can
+be used with the hosts FIPS compliant libraries.
```
deuberger:
```suggestion
You now have a single node Kubernetes cluster operating in FIPS mode and can add additional nodes or begin deploying workloads.
```

I removed the comments about underlying system and workload compliance since this guide covers the underlying system and workload compliance is out of scope. In the future, when we have clear information about how to make workloads compliant, we could add it back and include a link here. It would be great to have links for adding nodes or deploying workloads but we don't have guides focused on those right now (we have a tutorial for adding/removing nodes, but it's based on multipass so it's not a good fit here).

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+
+
+## Disable FIPS on an Ubuntu host machine
```
deuberger: I was on the fence about whether to include this section and I think I'm leaning towards removing it. The problem is it's not recommended meaning it might break and therefore we can't reliably test it or support it.  And I don't really want docs for things we don't test or support. Advanced users can always go digging to try to find ways to undo things. Maybe one day we could have a section of docs for advanced steps that aren't tested, supported, or recommended, but I don't really want to spend time on that now. Why not just say this isn't supported and that instead users should create a new cluster if they don't want FIPS.

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -0,0 +1,182 @@
...
+and will revert to the default non-FIPS settings.
+
+<!-- LINKS -->
+[FIPS 140-3]: https://csrc.nist.gov/pubs/fips/140-3/final
+[Ubuntu Pro]: https://ubuntu.com/pro
+[Ubuntu Pro subscription dashboard]: https://ubuntu.com/pro/dashboard
+<!-- markdownlint-disable MD053 -->
+[enable FIPS with Ubuntu]: https://ubuntu.com/tutorials/using-the-ubuntu-pro-client-to-enable-fips#1-overview
+<!-- markdownlint-enable MD053 -->
+[firewall configuration]: /snap/howto/networking/ufw.md
```
deuberger: I think this isn't need now either. Also, all the other k8s snap links are inlined rather than down her so maybe we should be consistent with this one?

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
+# How to install {{ product }} with DISA STIG hardening
```
deuberger, line 1: @nhennigan more a thought for the future, but maybe we should recast the "install" section as a "deployment" section. Really we are doing the install in FIPS and doing the bootstrap and DISA STIG in this part. I think the key point is that you can't rebootstrap. So any guide that requires bootstrapping should go in this section and you can only pick one of them. We could also call it "Bootstrap". We can discuss more later; out of scope for this PR I think.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+This guide assumes the following:
+
+- Ubuntu machine with at least 4GB of RAM and 30 GB disk storage
+- You have root or sudo access to the machine
+- Internet access on the machine
+- You have Ubuntu Pro enabled on your system. For more information, see
+  [Ubuntu Pro documentation]
+- You have FIPS enabled on your machine. See our [FIPS installation guide] for
+guidance
```
deuberger:
```suggestion

## Install {{product}} in FIPS mode

DISA STIG compliance also incorporates FIPS compliance. Start by following our [FIPS installation guide], but stop after installing {{product}} and do not follow the steps bootstrap the cluster. Instead continue here as there additional steps and configuration needed for DISA STIG compliance. The guide also covers enabling Ubuntu Pro, which is needed for some of the steps below.
```

I think this needs a little more explanation and might make sense in its own section vs being just a bullet in the pre-requesitte list. Ubuntu Pro really shouldn't be its own bullet since it's part of the FIPS guide.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+Security Technical Implementation Guides (STIGs) are developed by the Defense
+Information System Agency (DISA) and are comprehensive frameworks of security
+requirements designed to protect U.S. Department of Defense (DoD) systems and
+networks from cybersecurity threats.
+
+The [Kubernetes STIG] contains guidelines on how to check and remediate various
+potential security concerns for a Kubernetes deployment, both on the host and
+within the cluster itself. This guide goes through the steps needed to deploy
+{{product}} with our dedicated security configuration files to meet DISA STIG
+hardening requirements.
```
deuberger:
```suggestion
{{product}} aligns by default with many of these recommendations as they are expected to benefit most users. The additional recommendations are covered in this guide as well as the end-to-end procedure to be DISA STIG compliant at all layers including FIPS 140-3 compliance and host hardening.
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+To generate a compliance audit report (without applying changes):
+
+```
+sudo usg audit disa_stig
+```
+
+```{warning}
+The following command applies rule [V-270714] which will cause issues using
+accounts with an empty password to access your machine. To avoid being locked
+out use an RSA key to access your machine.
```
deuberger: 2 things:
1. We shouldn't specify the type of SSH key to use as that's irrelevant and RSA isn't necessarily the best option and might not be allowed by some organizations.
2. I'm pretty sure SSH keys will not solve this problem. The issue is the account is locked out unless a password is set and it's common not to set passwords for accounts that are only going to be accessed via SSH (e.g. multipass machines). I solve this by setting a password. There's probably other solutions, but I'm not sure off the top of my head what to recommend here. It might be best just to remove the extra sentence and leave it as an exercise for the user; it's not really in scope for this guide.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+requirements.
+
+### Bootstrap the first control-plane node
+
+Install {{product}}:
+
+```{literalinclude} ../../../_parts/install.md
+:start-after: <!-- snap start -->
+:end-before: <!-- snap end -->
+```
```
deuberger:
```suggestion
```
This will be done already as a part of the FIPS guide.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+On the joining control plane node install {{product}}:
+
+```{literalinclude} ../../../_parts/install.md
+:start-after: <!-- snap start -->
+:end-before: <!-- snap end -->
+```
+
+and join the cluster with the DISA STIG control plane configuration file:
```
deuberger:
```suggestion
On the additional control plane node, join the cluster with the DISA STIG control plane configuration file:
```
Again, the install will be done already as a part of the FIPS guide.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+Initialize the first control plane node with the necessary arguments for
+DISA STIG compliance:
+
+```
+sudo k8s bootstrap --file /var/snap/k8s/common/etc/templates/disa-stig/bootstrap.yaml
+sudo k8s status --wait-ready
+```
+
+Through this configuration file, the following rules are applied:
```
deuberger:
```suggestion
This configuration file applies the following rules:
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+and join the cluster with the DISA STIG control plane configuration file:
+
+```
+sudo k8s join-cluster --file=/var/snap/k8s/common/etc/templates/disa-stig/control-plane.yaml <join-token>
+```
+
+Through this configuration file, the same rules that were applied to the first
+control plane node have also been applied.
```
deuberger:
```suggestion
This configuration file applies the same rules shown above when bootstrapping the first control plane node.
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+On the joining worker node, install {{product}}:
+
+```{literalinclude} ../../../_parts/install.md
+:start-after: <!-- snap start -->
+:end-before: <!-- snap end -->
+```
+
+and join the cluster with the DISA STIG worker configuration file:
```
deuberger:
```suggestion
On the worker node, join the cluster with the DISA STIG worker configuration file:
```
Again, the install will be done already as a part of the FIPS guide.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+```
+sudo k8s join-cluster --file=/var/snap/k8s/common/etc/templates/disa-stig/worker.yaml <join-token>
+```
+
+Through this configuration file, the following rules are applied:
```
deuberger:
```suggestion
This configuration file applies the following rules:
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+| STIG | Summary |
+|------|---------|
+| [V-242434]| Kubernetes Kubelet must enable kernel protection |
+| [V-245541]| Kubernetes Kubelet must not disable timeouts |
+
+If SSH is not needed to access the worker nodes it is recommended you disable
+the SSH:
```
deuberger:
```suggestion
the SSH service:
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+```
+sudo systemctl disable ssh.service ssh.socket
+```
+
+```{note}
+According to rule [V-242393] and [V-242394] Kubernetes Worker Nodes must not have
+sshd service running or enabled. The host STIG rule [V-270665] on the other hand
+expects sshd to be installed on the host. To comply with both rules, leave SSH
+installed, but disable the service.
```
deuberger:
```suggestion
expects sshd to be installed on the host. To comply with both rules, we recommend leaving SSH
installed, but disabling the service.
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+```{note}
+Please ensure that the configuration of `/etc/sysctl.d/99-kubelet.conf` is not
+overridden by another higher order file.
+```
+
+## Deploy and bootstrap STIG-compliant nodes
+
+For your convenience, we have provided template configuration files that can be
+used to configure Kubernetes service arguments that align with DISA STIG
+requirements.
```
deuberger:
```suggestion
For your convenience, we have provided template configuration files that can be
used to configure the additional Kubernetes service arguments that are needed to fully align with DISA STIG
requirements.
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+- The default config at
+`/var/snap/k8s/common/etc/configurations/pod-security-admission-baseline.yaml`
+meets your needs or if it needs to be adjusted
+- Using the more restrictive config we provide at
+`/var/snap/k8s/common/etc/configurations/pod-security-admission-restricted.yaml`
+is more suited to your needs
+- You need to create your own audit policy based on the
+[upstream instructions] and adjust the `--admission-control-config-file` path
+used when you bootstrap/join nodes.
```
deuberger: This was adjusted from what's in the doc, but I found it less clear. Read literally, this says to assess whether any of the 3 options are needed, but doesn't say what to do. It also doesn't say anything about why someone would want to make these changes.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+For more details, see the [Kubernetes Pod Security Admission documentation],
+which provides an overview of Pod Security Standards (PSS), their enforcement
+levels, and configuration options.
+
+### Kubernetes API Server audit log
+
+This applies to rules [V-242402], [V-242403], [V-242461], [V-242462],
```
deuberger: The first sentence in this section and the above section are inconsistent for no good reason. I think we should make them consistent.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+Install {{product}}:
+
+```{literalinclude} ../../../_parts/install.md
+:start-after: <!-- snap start -->
+:end-before: <!-- snap end -->
+```
+
+Initialize the first control plane node with the necessary arguments for
+DISA STIG compliance:
+
```
deuberger, line 108: I worry that users are going to follow this guide top to bottom and get to the end and realized they wanted an alternative bootstrap template and it will be too late. That's why in the original end to end guide there was a short explanation and link about this at the beginning. Unless there's a good reason for cutting (e.g. doc practice guideline around it), I think we should put it back. Though since the only alternatives are for CP configuration, I would adjust it and put it in this section rather than at the top.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+## Alternative configurations
+
+### Pod Security Admission control file
+
+To comply with rule [V-254800], you must configure a Pod Security Admission
+control file for your Kubernetes cluster. This file defines the Pod Security
+Standards (PSS) that are enforced at the namespace level. The STIG templates we
+provide to bootstrap/join nodes configure the Pod Security admission controller
```
deuberger:
```suggestion
provide to bootstrap/join control plane nodes configure the Pod Security admission controller
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+For more details, see the [Kubernetes Pod Security Admission documentation],
+which provides an overview of Pod Security Standards (PSS), their enforcement
+levels, and configuration options.
+
+### Kubernetes API Server audit log
+
+This applies to rules [V-242402], [V-242403], [V-242461], [V-242462],
+[V-242463],[V-242464] and [V-242465]. The STIG templates we provide to
+bootstrap/join nodes configures the Kubernetes
```
deuberger:
```suggestion
bootstrap/join control plane nodes configures the Kubernetes
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+`/var/snap/k8s/common/etc/configurations/audit-policy.yaml`, which configures
+logging of all (non-resource) events with request metadata, request body, and
+response body is suitable for your needs or if it needs to be adjusted
+- Using the audit policy we provide at
+`/var/snap/k8s/common/etc/configurations/audit-policy-kube-system.yaml` which
+configures the same level of logging but only for events in the kube-system
+namespace is more suited to your needs
+- You need to create your own audit policy based on the
+[upstream audit instructions] and adjust the `--audit-policy-file` path used
+when you bootstrap/join nodes to use it.
```
deuberger: Again, this literally says to make an assessment, but doesn't say to do anything. I would fix or revert to what we had in the doc.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+In the default bootstrap configuration, Canonical Kubernetes does not enable
+audit logging as it may incur
+performance penalties in the form of increased disk I/O, which can lead to
+slower response times and reduced overall cluster efficiency, especially under
+heavy workloads.
```
deuberger:
```suggestion
```
This is the only place on this page where we explain why something isn't configured by default, so it's out of place. We either need to explain this everywhere or nowhere. I thought we were going to have this information in the DISA STIG audit reference page; that page needs a lot of work though.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+Scheduler, Controllers and etcd must enforce ports, protocols, and services
+(PPS) that adhere to the Ports, Protocols, and Services Management Category
+Assurance List (PPSM CAL). The {{product}} [ports and services] must be
+manually reviewed in accordance with this policy
+- [V-242414] The Kubernetes cluster must use non-privileged host ports for user
+pods so you must manually inspect all pods in the default namespace
+- [V-242415] Secrets in Kubernetes must not be stored as environment variables
+so you must manually inspect all pods in the default namespace
+- [V-242417] Kubernetes must separate user functionality
+so you must manually inspect all pods in the default namespace
```
deuberger:
```suggestion
## Required Documentation

[V-242410], [V-242411], [V-242412], and [V-242413] requires that the Kubernetes API Server,
Scheduler, and Controllers as well as etcd enforce ports, protocols, and services
(PPS) that adhere to the Ports, Protocols, and Services Management Category
Assurance List (PPSM CAL). The {{product}} [ports and services] must be
manually audited in accordance with this policy. These ports, protocols, and services will need added to your specific PPSM list and the list will need updated anytime the list of ports, protocols, and services used by your cluster changes. For instance, this list will need to be updated each time a new service is exposed externally.

## Post-Deployment Requirements

In addition to the above deployment steps, there are some guidelines that must be followed by users and administrators throughout the life of the cluster in order to maintain DISA STIG compliance.

- [V-242383]  User-managed resources must be created in dedicated namespaces
- [V-242414] User pods must only use non-privileged host ports
- [V-242415] Secrets must not be stored as environment variables
- [V-242417] User functionality must be separate from management functions meaning all user pods must be in user specific namespaces rather than system namespaces
```

I split this into two sections for clarity and because I thought more detail was needed for the PPSM CAL whereas less is needed for the other items. I don't think it's helpful to have the additional generic statement saying to audit those items without providing the details. 

Maybe it would be better to have the ongoing obligations documented on a separate page?

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+
+
+## Next steps
+
+Please assess your cluster for compliance using the [DISA STIG auditing page].
+Review all findings and apply any necessary remediations to be fully DISA STIG
+compliant. Be aware that some rules need to be upheld when you add workloads
+to your cluster.
```
deuberger:
```suggestion
## Auditing
If you would like to manually audit your cluster, see our [DISA STIG auditing page], where you will find detailed instructions and tools to evaluate your Kubernetes cluster against DISA STIG compliance requirements.
```
The way it was made it sound like a required step to be fully compliant. Really it's optional. A lot of me thinks we should just delete this section entirely. It feels out of place.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Also, it looks like the FW page was merged to main already a few weeks ago. Some issues I noticed though:

1. "Package forwarding" rather than "Packet forwarding"
2. Etcd port 2379 is missing from the instructions
3. The section about IP forwarding is confusing and maybe wrong. It says to edit sysctl.conf, but doesn't say to reboot and doesn't load the config file. But then it says, "Otherwise..." set the setting directly, but it never really said to reboot. See the doc for one way to handle this.

CC: @ktsakalozos

docs/canonicalk8s/snap/reference/disa-stig-audit.md
deuberger, line 1: Could we add a card to do another round of improvements and review on this file. It's not really consistent with what we've done with the FIPS/STIG story elsewhere.

docs/canonicalk8s/snap/reference/versions/1.34.md
```diff
@@ -18,6 +18,11 @@ k8s-dqlite will continue to function without changes during upgrades.
 - **Improved feature controller reliability** - Enhanced feature upgrade 
 process with worker node coordination.
 
+- **FIPS 140-3 compliance** - {{product}} snap can now be deployed with
+[FIPS 140-3](Federal Information Processing
+Standards) compliance, which ensures security compliance crucial for US
+government and regulated industries.
```
deuberger:
```suggestion
Standards) compliance for US government and regulated industries.

- **DISA STIG enhancements** - we now support running on a STIG hardened host and have improved the UX for applying the DISA STIG recommendations for Kubernetes.
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -0,0 +1,296 @@
...
+Initialize the first control plane node with the necessary arguments for
+DISA STIG compliance:
+
+```
+sudo k8s bootstrap --file /var/snap/k8s/common/etc/templates/disa-stig/bootstrap.yaml
+sudo k8s status --wait-ready
+```
+
+Through this configuration file, the following rules are applied:
+| STIG | Summary |
```
deuberger: @nhennigan thought for the future. I wonder if it would help to have a separate reference page that covers the available bootstrap templates and which settings are applied for each rather than having those extra details here.

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: Based on what I can find online you are right, fips does not explicitly require the firewall to be configured. I will remove it from here and put it as a prerequisite in the disa-stig install docs.

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: @nhennigan / @ktsakalozos What do you think?
I would rather include the revert steps for the sake of people trying fips out in test environments. We could further extend the warning to emphasize this is unsupported: `Disabling FIPS on a host machine is not recommended and unsupported`.

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: Great suggestion, thanks!

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: Could you please be more specific on what should be updated here? We chose to do this tutorial with Ubuntu 22.04 not 24.04. These steps are in line with what is needed for deployments on Ubuntu 22.04 AFAIK.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: On point 1. not all keys are allowed under fips such as ed25519 keys. https://csrc.nist.gov/csrc/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp2389.pdf
Let me remove the second sentence from here, do you think we should add a note in the fips docs?

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: +1 I removed the firewall from the fips docs and moved them into a section in the disa-stig guide capturing what you said.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: I think that would make sense once we have more botstrap templates available. Currently, we only have the ones designated for DISA STIG.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Does the attention note address your concern adequately?

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: @nhennigan you wanted to trim this section. Are you onboard with bringing the playbook wording back?

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Thank you for improving this section.
I agree that we should move this section out into a separate page eventually. Right now, based on the length of the section, I think it makes sense to have everything "conveniently" in one place for users making this a streamlined experience without too many hops around our docs. 

CC: @nhennigan You might find this interesting.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Sure, let's remove this note. It is optional and more of a nice to know.

docs/canonicalk8s/snap/reference/disa-stig-audit.md
louiseschmidtgen: I can do that. Could you please be more specific in what points we are inconsistent with the rest of the fips/stig story.
The rules are already updated to refer to the bootstrap config yaml for example.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Right, but still that's 3 different files and two lists of what they apply. It's okay to leave them on this page for now, but I think there's enough content already that a page on its own wouldn't hurt.

docs/canonicalk8s/snap/howto/install/fips.md
deuberger: @louiseschmidtgen it might be better to have a section for it with a little more guidance rather than just listing it as a pre-requisite. The usg tool does not enable the FW, but the STIG requires it and the usg says something about this which we might want to explain. From a technical perspective the FW is optional so users could decide to skip it. Another option would be to list it as a prerequisite and have a note explaining the warning when running the usg tool; this might be better but I haven't tested that it works this way.

docs/canonicalk8s/snap/howto/install/fips.md
deuberger: The text here is old and needs the updates from the "Install the FIPS core base snap" section in the end to end guide.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Removing second sentence works for me. Good point re types of keys supported; I suppose FIPS supports more than just RSA keys though? I don't think we need a note in the FIPS docs. This may vary for different environments/needs and isn't really in scope of this doc.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Leaving it on the same page is fine for now.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Sounds good.

docs/canonicalk8s/snap/reference/disa-stig-audit.md
```diff
@@ -1,22 +1,12 @@
...
-
-This guide assumes the following:
-
-- You have a bootstrapped {{product}} cluster (see the [getting started] guide)
-- You have root or sudo access to the machine
-- You have reviewed the [post-deployment hardening] guide and have applied the
-  hardening steps that are relevant to your use-case
+potential security concerns for a Kubernetes deployment. Here we have tailored
+these checks to work for our {{product}} cluster that is deployed using the
+[deploy a DISA STIG hardened cluster] guide.
```
deuberger:
```suggestion
guide to [deploy a DISA STIG hardened cluster].
```

docs/canonicalk8s/snap/reference/disa-stig-audit.md
```diff
@@ -1,22 +1,12 @@
...
 
 ## Guideline classes overview
```
deuberger, line 11: This section needs updated. The link is wrong now. Some of the "not applicable" rules are actually handled. "Manual" should name and description should be updated to match what's in the other guide and some of these are no longer manual. I think we should split the "Deployment" section into a "Default" and "Bootstrap" section to be more clear.

docs/canonicalk8s/snap/reference/disa-stig-audit.md
deuberger: Actually, I added comments on the key issues that should be addressed under the current epic and I captured the follow on work as a part of https://warthogs.atlassian.net/browse/PF-6137. So if you address all the comments/suggestions, you should be good; no need to add a card.

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: I think the right way would probably be Federal Information Processing Standards (FIPS) 140-3 with the abbreviation in the brackets. Wdyt?

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: Yes, I ended up turning it into a little section saying it is recommended and you can do it. Does it capture what you are saying fully?

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: Yes, that is a fair point. I think it makes sense to add this page some time in the future. Right now this does not give the user any actionable steps. I accepted your suggestion.

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: Removed the link.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Thank you for the suggestion, I accepted it.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Yes, there are more than just RSA keys (this was just my choice) supported by FIPS, but we can leave it from the FIPS docs.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Chose to remove it.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Thank you for the suggestion, I've applied it. It breaks things up nicely into two sections.
I agree another section would make sense once we have more content to put into a separate page.
For now I would prefer to keep all things in one place to make it easier for users to find. Wdyt?
CC: @nhennigan

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: I think you are right here, it is not essential for the install guide. I have removed it.

docs/canonicalk8s/snap/howto/install/disa-stig.md
ktsakalozos: > "Package forwarding" rather than "Packet forwarding"

Yes, we need to correct this.

> Etcd port 2379 is missing from the instructions

This is not needed for our K8s. Port 2379 is [for clients to connect to etcd](https://etcd.io/docs/v3.1/op-guide/configuration/), but we do not want external clients to connect to 2379. All clients that need to access etcd are local on the host. Maybe we need to update the playbook.

>  The section about IP forwarding is confusing and maybe wrong. It says to edit sysctl.conf, but doesn't say to reboot and doesn't load the config file. But then it says, "Otherwise..." set the setting directly, but it never really said to reboot. See the doc for one way to handle this.

Indeed we should make this improvement:

```
Use sysctl to apply the forwarding rules immediately without rebooting the system:

sudo sysctl -w net.ipv4.ip_forward=1

To have the rules persists through system reboots edit /etc/sysctl.conf and make sure the following is set:

net.ipv4.ip_forward=1

```

docs/canonicalk8s/snap/howto/install/fips.md
ktsakalozos: +1 in removing this section. I think that as soon as you switch core22 to latest/stable our k8s will crash until you reboot.

docs/canonicalk8s/snap/reference/disa-stig-audit.md
louiseschmidtgen: Let's do this in another card.

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: Calling this resolved.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Let's address this in a separate PR.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Mark as resolved I brought back the playbook section.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: calling this addressed in attention note, if this needs further rework it will be part of another PR.

docs/canonicalk8s/snap/reference/disa-stig-audit.md
louiseschmidtgen: Let's double check this in another PR.

docs/canonicalk8s/snap/howto/install/disa-stig.md
nhennigan: I like splitting the two sections. It is definitely clearer to understand. I also agree there is not enough currently to warrant a new page for ongoing expectations but this could be reviewed in the future

docs/canonicalk8s/snap/howto/install/disa-stig.md
nhennigan: This is a good idea, especially as we add more templates for different functions. I'll add it to the backlog

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: I addressed these in #1990 except the etcd issue which I added https://warthogs.atlassian.net/browse/KU-4672 for.

docs/canonicalk8s/snap/reference/disa-stig-audit.md
deuberger: I mostly addressed this in #1990 though I added https://warthogs.atlassian.net/browse/KU-4673 and https://warthogs.atlassian.net/browse/KU-4674 for follow-up.

docs/canonicalk8s/snap/reference/disa-stig-audit.md
deuberger: Done in #1990.

PR #1912, opened 2025-10-16
docs: Remove linkcheck

nhennigan:
## Description

There are intermittent failures in our CI in relation to the linkcheck test. There has been investigation into the issue and there needs to be work done to make the test more suitable for our needs. The issues include being rate limited on certain websites and being denied access which leads to failures and also runner build issues. 

## Solution

 For now, I am removing the check and will reinstate when the better solution is reached

## Issue

N/A

## Backport

This does not affect other branches as much so main only should be ok

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/.github/workflows/docs-checks.yml
+++ b/.github/workflows/docs-checks.yml
@@ -18,6 +18,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


bschimke95:
Let's backport this as it also breaks on 1.32 - 1.34
@nhennigan FYI

PR #1918, opened 2025-10-17
feat: Make CLI Look More Responsive

ethandcosta:
Several k8s snap commands follow a similar pattern of waiting on a blocking RPC to k8sd. Currently, on particularly slow devices, this can appear to users like the terminal session froze because nothing is printed. This PR adds a ticker heartbeat to the following commands
- bootstrap
- disable
- enable
- join_cluster
- set
- status (with wait enabled)

Note that this isn't adding any additional information about the state of the cluster, which is housed in the k8sd client api and isn't currently exposed.

### DIFF
--- a/src/k8s/cmd/k8s/k8s_disable.go
+++ b/src/k8s/cmd/k8s/k8s_disable.go
@@ -97,11 +97,11 @@ func newDisableCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
 			}
 
-			err = cmdutil.WithSpinner(ctx, cmd.ErrOrStderr(), fmt.Sprintf("Disabling %s from the cluster...", strings.Join(args, ", ")), func(ctx context.Context) error {
+			err = cmdutil.WithSpinner(ctx, cmd.ErrOrStderr(), fmt.Sprintf("Disabling %s...", strings.Join(args, ", ")), func(ctx context.Context) error {
 				return client.SetClusterConfig(ctx, apiv1.SetClusterConfigRequest{Config: config})
 			})
 			if err != nil {
-				cmd.PrintErrf("Error: Failed to disable %s from the cluster.\n\nThe error was: %v\n", strings.Join(args, ", "), err)
+				cmd.PrintErrf("Error: Failed to disable %s.\n\nThe error was: %v\n", strings.Join(args, ", "), err)
 				env.Exit(1)
 				return
 			}
--- a/src/k8s/cmd/k8s/k8s_join_cluster.go
+++ b/src/k8s/cmd/k8s/k8s_join_cluster.go
@@ -108,7 +108,7 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
 			}
 
-			err = cmdutil.WithSpinner(cmd.Context(), cmd.ErrOrStderr(), "Joining the cluster. This may take a few seconds, please wait.", func(ctx context.Context) error {
+			err = cmdutil.WithSpinner(cmd.Context(), cmd.ErrOrStderr(), "Joining the cluster. This may take some time, please wait.", func(ctx context.Context) error {
 				return client.JoinCluster(ctx, apiv1.JoinClusterRequest{
 					Name:    opts.name,
 					Address: address,
--- a/src/k8s/cmd/k8s/k8s_remove_node.go
+++ b/src/k8s/cmd/k8s/k8s_remove_node.go
@@ -54,7 +54,7 @@ func newRemoveNodeCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 
 			name := args[0]
 
-			err = cmdutil.WithSpinner(cmd.Context(), cmd.ErrOrStderr(), fmt.Sprintf("Removing %q from the cluster. This may take a few seconds, please wait.", name), func(ctx context.Context) error {
+			err = cmdutil.WithSpinner(cmd.Context(), cmd.ErrOrStderr(), fmt.Sprintf("Removing %q from the cluster. This may take some time, please wait.", name), func(ctx context.Context) error {
 				return client.RemoveNode(ctx, apiv1.RemoveNodeRequest{Name: name, Force: opts.force, Timeout: opts.timeout})
 			})
 			if err != nil {
--- a/src/k8s/cmd/k8s/k8s_status.go
+++ b/src/k8s/cmd/k8s/k8s_status.go
@@ -48,7 +48,7 @@ func newStatusCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 			}
 
 			var response apiv1.ClusterStatusResponse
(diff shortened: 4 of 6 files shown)


src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -120,6 +123,9 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
```
louiseschmidtgen, line 123: What happens if we return before calling `stopHB()`? Does the spinner continue?

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -108,6 +108,9 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 			}
 
 			cmd.PrintErrln("Joining the cluster. This may take a few seconds, please wait.")
+
+			stopHB := cmdutil.StartSpinner(cmd.Context(), cmd.ErrOrStderr(), "Still working...")
```
louiseschmidtgen: "Still working..." seems a bit too generic. Could we please have messages specific to the command? Something along the lines of "Joining cluster" here.

Same applies to the other commands.

src/k8s/cmd/k8s/k8s_join_cluster.go
ethandcosta: The spinner stops because the OS terminates all goroutines regardless on env.Exit(), but in practice it's a good idea to add a StopHB() here since no clean-up runs when the process is killed- so in effect, the line is preserved in what it's last state was before the thread was interrupted.

src/k8s/cmd/k8s/k8s_join_cluster.go
ethandcosta: good point, I'll change the messages to be more informative

src/k8s/cmd/k8s/k8s_join_cluster.go
ethandcosta: a better solution would be to have the env.exit process call StopHB() but that might be more involved

HomayoonAlimohammadi (commented):
This looks great! Thanks a lot @ethandcosta! I was thinking about a slightly different approach. What if we add a wrapper of some sort, like `cmdutil.WithSpinner(...)` that wraps the execution context? This way we won't be burdened with remembering to stop the spinner. WDYT?

src/k8s/cmd/k8s/k8s_disable.go
```diff
@@ -84,11 +84,10 @@ func newDisableCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
 			}
 
-			cmd.PrintErrf("Disabling %s from the cluster. This may take a few seconds, please wait.\n", strings.Join(args, ", "))
 			ctx, cancel := context.WithTimeout(cmd.Context(), opts.timeout)
 			cobra.OnFinalize(cancel)
 
-			if _, initialized, err := client.NodeStatus(cmd.Context()); err != nil {
+			if _, initialized, err := client.NodeStatus(ctx); err != nil {
```
bschimke95, line 90: nice catch!

src/k8s/cmd/util/progress.go
```diff
@@ -0,0 +1,63 @@
...
+	"io"
+	"strings"
+	"sync"
+	"time"
+)
+
+// StartSpinner displays a message with an animated spinner that updates in-place.
+// The spinner continues until either the context is cancelled or the returned
+// stop function is called.
+func StartSpinner(ctx context.Context, w io.Writer, msg string) func() {
```
bschimke95: nit: I don't see `StartSpinner` being used anywhere. Let's make it private (`startSpinner`) to indicate that use of `WithSpinner` is preferred. Maybe also move the `WithSpinner` on top.

src/k8s/cmd/k8s/k8s_disable.go
```diff
@@ -98,7 +97,10 @@ func newDisableCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
 			}
 
-			if err := client.SetClusterConfig(ctx, apiv1.SetClusterConfigRequest{Config: config}); err != nil {
+			err = cmdutil.WithSpinner(ctx, cmd.ErrOrStderr(), fmt.Sprintf("Disabling %s from the cluster...", strings.Join(args, ", ")), func(ctx context.Context) error {
```
bschimke95: nit: I'd remove the `... from the cluster` part from the message (here and below)

src/k8s/cmd/k8s/k8s_join_cluster.go
```diff
@@ -107,14 +108,16 @@ func newJoinClusterCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
 
-			cmd.PrintErrln("Joining the cluster. This may take a few seconds, please wait.")
-			if err := client.JoinCluster(cmd.Context(), apiv1.JoinClusterRequest{
-				Name:    opts.name,
-				Address: address,
-				Token:   token,
-				Config:  joinClusterConfig,
-				Timeout: opts.timeout,
-			}); err != nil {
+			err = cmdutil.WithSpinner(cmd.Context(), cmd.ErrOrStderr(), "Joining the cluster. This may take a few seconds, please wait.", func(ctx context.Context) error {
```
bschimke95: Let's be consistent and use the same wording as in the bootstrapping.

src/k8s/cmd/k8s/k8s_status.go
```diff
@@ -47,8 +47,12 @@ func newStatusCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 				return
 			}
 
-			response, err := client.ClusterStatus(ctx, opts.waitReady)
-			if err != nil {
+			var response apiv1.ClusterStatusResponse
+			if err := cmdutil.WithSpinner(ctx, cmd.ErrOrStderr(), "Waiting for cluster to become ready...", func(ctx context.Context) error {
```
bschimke95: nit: `the cluster`

src/k8s/cmd/k8s/k8s_remove_node.go
```diff
@@ -53,8 +54,10 @@ func newRemoveNodeCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 
 			name := args[0]
 
-			cmd.PrintErrf("Removing %q from the Kubernetes cluster. This may take a few seconds, please wait.\n", name)
-			if err := client.RemoveNode(cmd.Context(), apiv1.RemoveNodeRequest{Name: name, Force: opts.force, Timeout: opts.timeout}); err != nil {
+			err = cmdutil.WithSpinner(cmd.Context(), cmd.ErrOrStderr(), fmt.Sprintf("Removing %q from the cluster. This may take a few seconds, please wait.", name), func(ctx context.Context) error {
```
bschimke95: same here.

bschimke95 (commented):
Great work @ethandcosta. I did a first pass and added a few suggestions but already looks great overall.

PR #1914, opened 2025-10-16
chore (release-1.32): Bump Helm version across the snap

mateoflorido:
## Description

Bump Helm versions across the different dependencies.

## Solution

We have detected multiple CVEs in the Helm dependencies. This pull request updates the version for the Helm binary, the target toolchain, and the Go dependencies. Example of the CVEs found:
- CVE-2025-22874
- CVE-2025-47907
- CVE-2025-53547

## Backport

Other releases will be patched in different pull requests.

## Checklist

- [x] PR title formatted as `type: title`
- [x] Covered by unit tests
- [x] Covered by integration tests
- [ ] Documentation updated
- [x] CLA signed
- [x] Backport label added if necessary

PR #1929, opened 2025-10-21
fix: ck-loadbalancer schema and value mapping

bschimke95:
We bumped the Helm and go-helm version in #1914 

go-helm 3.18.5 introduced a new json schema validator that is stricter against the schemas and caused the `ck-loadbalancer` values to get rejected.
The helm maintainers suggest to modify the json schema: https://github.com/helm/helm/issues/31177#issuecomment-3380147813

Note that there seems to be a bug (or at least an oddity) that requires us to cast the value slices to `any` before the new validation for now, see https://github.com/santhosh-tekuri/jsonschema/issues/238


bschimke95:
Closing in favor of #1942

PR #1931, opened 2025-10-22
feat(kube-apiserver-proxy): Retry endpoint retrieval

HomayoonAlimohammadi:
### Overview

Upon removal of a control plane node, the worker node might get stuck in a situation that it's not able to communicate with the remaining API servers (the kubelet) and hence its status is declared as `NotReady` in the cluster.


HomayoonAlimohammadi:
Closed in favor of https://github.com/canonical/k8s-snap/pull/1939

PR #1933, opened 2025-10-22
fix: improve log cleanup and rotation policies

louiseschmidtgen:
## Description

Currently, we have a significant resource cost for our current logging configuration.

This PR attempts to reduce the resource footprint of our logs by:

1. cleaning up "orphaned" pods logs on snap removal
2. adding instructions on how to tune system settings regarding system logs

### Details on cleaning up "orphaned" pods logs on snap removal
The logs are under `/var/log/pods` and `/var/log/containers` and [managed by kubelet](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/). When we remove kubelet no one is "responsible" for these logs anymore and they don't become cleaned up. Therefore, we remove them on the removal hook

See also: https://kubernetes.io/docs/concepts/cluster-administration/system-logs/#log-location, 

### Details on adding instructions on how to tune system settings regarding system logs

In my multipass VM I noticed that the default journalctl log policies are not ideal. No maximum log file limits, unbounded log retention time, and log duplication of etcd and k8sd in journalctl and syslog make it possible to consume a lot of disk space.

We should not tune user's sytem settings, therefore, I've added a docs section instead that helps users do this on their own.

## Backport

## Checklist

- [x] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/docs/canonicalk8s/.sphinx/.wordlist.txt
+++ b/docs/canonicalk8s/.sphinx/.wordlist.txt
@@ -23,6 +23,8 @@ IAM
 installable
 JSON
 Juju
+journalctl
+journald
 Kubeflow
 Kubernetes
 Launchpad
@@ -55,6 +57,7 @@ RTD
 subdirectories
 subfolders
 subtree
+syslog
 TODO
 Ubuntu
 UI


rapour (approved):
Great catch @louiseschmidtgen, thank you

Do you think a log rotation mechanism (e.g., logrotate) is in order too?

docs/canonicalk8s/snap/reference/troubleshooting.md
```diff
@@ -172,6 +172,43 @@ sudo snap restart snap.k8s.k8s-dqlite
 
 ````
 
+## High disk usage for log files
+
+When using {{product}} for a longer period of time, the disk usage for
```
bschimke95, line 177: This should not happen if log-rotation is in place. Do you still see this behavior when you enable the etcd logrotation?

bschimke95: `journallog` should have enabled log-rotation enabled by default. I'm a bit confused that we'd need to tune this.

louiseschmidtgen: For reference copying my answer from MM:
My multipass VM was running out of disk space so I had a look at the main consumers. This is my "default" config from my multipass VM which configures the system's log rotation:
```
#SystemMaxUse=
#SystemKeepFree=
#SystemMaxFileSize=
#SystemMaxFiles=100
#RuntimeMaxUse=
#RuntimeKeepFree=
#RuntimeMaxFileSize=
#RuntimeMaxFiles=100
#MaxRetentionSec=
#MaxFileSec=1month
#ForwardToSyslog=yes
```

PR #1939, opened 2025-10-23
feat(k8s-apiserver-proxy): Use current endpoints as serverName to get new endpoints

HomayoonAlimohammadi:
### Overview

Scenario: Bootstrap a cluster with `SkipServiceCleanup` and `SkipNodeRemoval` annotations. 2 CPs and 1 worker node.
remove 1 CP from the cluster, and you MIGHT notice (there's a chance) that the worker node can not communicate with the API server anymore, and hence will become `NotReady`. If you get into this situation, it won't get resolved automatically. That's because the API server endpoints can not get updated by the `k8s-apiserver-proxy`. This PR solves that issue querying the API server through the actual endpoints, rather than localhost (going through proxy).
Fixes: https://github.com/canonical/cluster-api-k8s/issues/183

### DIFF
--- a/src/k8s/pkg/proxy/endpoints.go
+++ b/src/k8s/pkg/proxy/endpoints.go
@@ -12,11 +12,10 @@ import (
 )
 
 func getKubernetesEndpoints(ctx context.Context, kubeconfigFile string, server string) ([]string, error) {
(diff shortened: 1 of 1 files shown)


bschimke95 (approved):
LGTM, I feel like this is a fix we'd need to backport. Please add the appropriate labels before merge.

PR #1940, opened 2025-10-23
fix: Spinner CLI only triggers on k8s status if --wait-ready is enabled

ethandcosta:
fixes a minor bug where the spinner would appear when calling `k8s status` even when --wait-ready wasn't specified.


bschimke95 (approved):
LGTM

PR #1946, opened 2025-10-24
test: Extend nightly test matrix and move CI logic to Python CLI

bschimke95:
This pull request does two things:
1. It extends the testing matrix we do for nightly runs
2. For that, it introduces a Python-based CI toolkit for aggregating and posting test results to Mattermost, along with workflow refactors to use it.

## CI Toolkit and Python Environment

Added a new Python CLI (`k8s-ci.py`) with a `mattermost` subcommand to aggregate per-test results and post structured summaries to Mattermost via webhook. The main idea here is to have a common entry point for (complex) ci scripts that would became a nightmare to maintain in `bash`. Moving those to Python makes it easier to read, maintain and test locally.
The `k8s-ci` tox environment follows the standards we already define for the integration testing.

## GitHub Actions Workflow Updates

* Enhanced e2e-tests.yaml to upload per-test JSON artifacts with channel-specific naming. This is disabled by default as it mostly (only) makes sense if this is invoked in a matrix and adds a little overhead to the workflow runs
* Limit the amount of parallel runs to 10 per matrix entry - otherwise we'd kick of ~800 jobs each night which would make IS slightly unhappy. Also, it is fine if those runs take longer as they don't block anything. 
* Extended nightly workflows to test multiple release channels, collect results, and post summaries via k8s-ci mattermost results-message.
* Replaced legacy notification logic with the Python-based system.

## Future work

* The branches are still hardcoded for now. Eventually, those should be obtained dynamically which is where the new `k8s-ci` CLI can play a part in the future.
* I expect every new ci step that contains significant logic to be moved in the `k8s-ci`


.github/workflows/e2e-tests.yaml
```diff
@@ -206,7 +228,6 @@ jobs:
           TEST_GH_BASE_REF: ${{ github.base_ref }}
           TEST_UBUNTU_PRO_TOKEN: ${{ secrets.UBUNTU_PRO_TOKEN }}
           TEST_LAUNCHPAD_K8S_CI_BOT_PASSWORD: ${{ secrets.LAUNCHPAD_K8S_CI_BOT_PASSWORD }}
```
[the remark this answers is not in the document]
bschimke95: This should not have been added in the first place and probably was forgotten to remove after debugging in another PR.

bschimke95:
@berkayoz @HomayoonAlimohammadi for k8s-snap integration
@addyess for looking at my elementary school Python code.

.github/workflows/e2e-tests.yaml
```diff
@@ -29,7 +29,7 @@ on:
         description: Integration test filter tags (e.g. pull_request, up_to_weekly)
         default: pull_request
         type: string
-      flavor: &flavor
+      flavor:
```
bschimke95, line 32: There is a hard limit of 10 for workflow parameters. We don't use flavors right now, so I removed it from the workflow dispatch.

ci/tox.ini
```diff
@@ -0,0 +1,46 @@
...
+passenv =
+    PYTHONPATH
+
+[testenv:format]
+description = Apply coding style standards to code
+deps = -r {toxinidir}/requirements-ci.txt
+commands =
+    licenseheaders -t {toxinidir}/.copyright.tmpl -cy -o 'Canonical, Ltd' -d {toxinidir}
+    isort {toxinidir} --profile=black
+    black {toxinidir} --verbose
```
addyess, line 19: black and isort are python tools that are fine -- but they're slow.  We should consider using `ruff` where possible since it's written in rust, it's fast, and simpler to deal with..

but for now, this is fine

ci/tox.ini
```diff
@@ -0,0 +1,46 @@
...
+
+[testenv:lint]
+description = Check code against coding style standards
+deps = -r {toxinidir}/requirements-ci.txt
+commands =
+    codespell --skip .tox {toxinidir}
+    flake8 {toxinidir}
```
addyess, line 26: same for flake8 -- ruff is so much faster

ci/cmds/mattermost.py
```diff
@@ -0,0 +1,373 @@
...
+                        child_conn = "└──" if fi == len(failed_runs) - 1 else "├──"
+                        run_link = _determine_run_link(fe)
+                        run_part = f" [Run]({run_link})" if run_link else " Run"
+                        name = (
+                            fe.get("test").split("::")[-1]
+                            if fe.get("test")
+                            else "Unnamed Test"
+                        )
+                        lines.append(f"{child_base}{child_conn} {name}{run_part}")
+    return "\n".join(lines)
```
addyess, line 188: wow, this looks -- like it was written by an AI agent

ci/tox.ini
bschimke95: yes, we need to do this eventually. Also for the test folders. I'll change both in a separate PR.

ci/tox.ini
bschimke95: same - will do in a separate PR along with the tests folder

ci/cmds/mattermost.py
bschimke95: Indeed!

PR #1950, opened 2025-10-24
docs: ceph-csi how-to should avoid resource collisions by default

addyess:
## Description

The ceph-csi how-to default deployment creates colliding kubernetes resources and instructs users how to resolve them.  Instead, the instructions should provide a conflict installation best guess, while remaining instructive on how to resolve those collisions.

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/docs/canonicalk8s/charm/howto/ceph-csi.md
+++ b/docs/canonicalk8s/charm/howto/ceph-csi.md
@@ -20,7 +20,7 @@ an adequate amount of resources must be allocated:
 :end-before: <!-- juju vm end -->
 ```
 
-## Deploying Ceph
+## Deploy Ceph
 
 Deploy a Ceph cluster containing one monitor and one storage unit
 (OSDs). In this example, a limited amount of resources is being allocated.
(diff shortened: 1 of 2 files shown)


docs/canonicalk8s/charm/howto/ceph-csi.md
```diff
@@ -165,6 +169,10 @@ juju deploy -n 1 ceph-osd-alt ceph-osd \
 juju deploy ceph-csi-alt ceph-csi \
     --config create-namespace=true \
     --config namespace=${CEPH_NS_ALT} \
+    --config ceph-xfs-storage-class-name-formatter="ceph-xfs-{app}" \
+    --config ceph-ext4-storage-class-name-formatter="ceph-ext4-{app}" \
+    --config cephfs-storage-class-name-formatter="cephfs-{name}-{app}" \
```
nhennigan, line 212: I think including the {name} and {app} here could be a tad confusing because we don't explicitly call it out. Users may believe they can copy and paste the entire code block as they were able to in the previous block. We set the namespace alt be default.

nhennigan (commented):
Thanks Adam, from what I remember when we last updated this page, there was a reason we did not provide all these configs from the outset as you are proposing now, or was this just an oversight? Is there any issue with configuring these parameters on creation?

docs/canonicalk8s/charm/howto/ceph-csi.md
addyess: These changes were the result of a discussion with @deuberger  over this topic:

> It does seem to me that the howto is overly complex. Why not just include the formatters (and namespace) in the instructions to begin with and then just explain that things will go wrong if they are missing.
> Docs Section [Relate to multiple ceph cluster](https://documentation.ubuntu.com/canonical-kubernetes/release-1.32/charm/howto/ceph-csi/#relate-to-multiple-ceph-clusters) doesn’t need to lead to user to believe there are guaranteed resource collisions
> If you follow the docs, there should be no collisions
> If you didn’t follow all the docs, their COULD be collisions


It seems that the docs deliberately lead you down a road to intended collisions that must be repaired -- but we could make these docs adjustments to intentionally AVOID duplicates.

nhennigan: This has been addressed with PR #1973

PR #1954, opened 2025-10-27
fix: Upped expected helm version to 3.18

ethandcosta:
Cdk bot tracks helm version 3.14 even though we use 3.18.


addyess (approved):
Thank you so much @ethandcosta

PR #1978, opened 2025-11-04
feat: Make kubectl independent of k8sd

HomayoonAlimohammadi:
### Overview

* Make `k8s kubectl` independent of the `k8sd` service (and the underlying Microcluster), allowing cluster access even when `k8sd` is stopped or unavailable. 
* Allow `k8s kubectl` on worker nodes if the `KUBECONFIG` environment variable is provided.
* Add an integration test to verify the requirements and functionality

### DIFF
--- a/src/k8s/cmd/k8s/k8s_kubectl.go
+++ b/src/k8s/cmd/k8s/k8s_kubectl.go
@@ -2,7 +2,6 @@ package k8s
 
 import (
 	"os"
-	"os/exec"
 	"path/filepath"
 	"syscall"
 
@@ -17,13 +16,7 @@ func newKubectlCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 		DisableFlagParsing: true,
 		PreRun:             chainPreRunHooks(hookRequireRoot(env)),
 		Run: func(cmd *cobra.Command, args []string) {
-			binary, err := exec.LookPath("kubectl")
-			if err != nil {
-				cmd.PrintErrln("Error: kubectl binary not found")
-				env.Exit(1)
-				return
-			}
-
+			binary := filepath.Join(env.Snap.K8sBinDir(), "kubectl")
 			kubeconfigEnvKey := "KUBECONFIG"
 			adminKubeconfigPath := filepath.Join(env.Snap.KubernetesConfigDir(), "admin.conf")
 			kubeletKubeconfigPath := filepath.Join(env.Snap.KubernetesConfigDir(), "kubelet.conf")


src/k8s/cmd/k8s/k8s_kubectl.go
```diff
@@ -17,38 +18,37 @@ func newKubectlCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
 		DisableFlagParsing: true,
 		PreRun:             chainPreRunHooks(hookRequireRoot(env)),
 		Run: func(cmd *cobra.Command, args []string) {
-			client, err := env.Snap.K8sdClient("")
+			binary, err := exec.LookPath("kubectl")
```
bschimke95: We ship the kubectl binary at `/snap/k8s/current/bin/kubectl`. I wonder how this works. Do you happen to know?

tests/integration/tests/test_kubectl.py
```diff
@@ -0,0 +1,125 @@
...
+    Test kubectl behavior before bootstrap, after bootstrap, and when k8sd is stopped, on both
+    control-plane and worker nodes.
+    """
+    cp, worker = instances
+
+    # kubectl should fail when cluster is not bootstrapped
+    LOG.info("Testing kubectl before bootstrap...")
+    result = cp.exec(
+        ["k8s", "kubectl", "get", "nodes"],
+        capture_output=True,
```
bschimke95, line 27: capture_output and text are not used/required

bschimke95 (commented):
Did a first pass.

src/k8s/cmd/k8s/k8s_kubectl.go
HomayoonAlimohammadi: Good point. I think this will grab the first `kubectl` binary that it can find in the PATH (which might be different from what we are shipping), right? If we explicitly want to grab our own included `kubectl`, then we'll have to grab the binary path in a different way. WDYT?

tests/integration/tests/test_kubectl.py
HomayoonAlimohammadi: Nice catch, thanks!

src/k8s/cmd/k8s/k8s_kubectl.go
```diff
@@ -17,38 +18,41 @@ func newKubectlCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
+			adminKubeconfigPath := filepath.Join(env.Snap.KubernetesConfigDir(), "admin.conf")
+
+			kubeconfigInEnv := false
+			for _, envKV := range env.Environ {
+				parts := strings.SplitN(envKV, "=", 2)
+				if len(parts) >= 2 && parts[0] == kubeconfigEnvKey && parts[1] != "" {
+					kubeconfigInEnv = true
+					break
+				}
 			}
```
rapour: Could we pack this logic into a utility function or method receiver of, for example, `ExecutionEnvironment` and also have some tests for it?

src/k8s/cmd/k8s/k8s_kubectl.go
```diff
@@ -17,38 +18,41 @@ func newKubectlCmd(env cmdutil.ExecutionEnvironment) *cobra.Command {
...
+					env.Exit(1)
+					return
+				}
 			}
 
 			command := append([]string{"kubectl"}, args...)
 			environ := cmdutil.EnvironWithDefaults(
 				env.Environ,
-				"KUBECONFIG", filepath.Join(env.Snap.KubernetesConfigDir(), "admin.conf"),
+				kubeconfigEnvKey, adminKubeconfigPath,
```
rapour: If we are in a worker node and `KUBECONFIG` env is set, doesn't this overwrite the value in `env.Environ` since `adminKubeconfigPath` is not empty, only the config file is not in that path?

HomayoonAlimohammadi: According to [`EnvironWithDefaults`](https://github.com/canonical/k8s-snap/blob/main/src/k8s/cmd/util/environ.go#L80-L85), if `KUBECONFIG` is set, we won't overwrite it with the `adminKubeconfigPath`. Did I understand your question correctly?

rapour: Ah! right, I though the key/values get merged with the `os.Environ`

src/k8s/cmd/k8s/k8s_kubectl.go
HomayoonAlimohammadi: Done, thanks!

src/k8s/cmd/k8s/k8s_kubectl.go
berkayoz: Well we do have the wrapper/util that sets the necessary path information/priority, though there is a chance that this might be sourced from the host if the path priority is messed up or we fail in shipping kubectl.

bschimke95: Ok, let's make it explicit then.

PR #1980, opened 2025-11-04
feat: Add support for s390x and ppc64el

addyess:
## Description
This pull request updates the Snap packaging configuration to improve architecture support and refines the build and prime stages for the `etcd`  and `containerd` components. The main changes are focused on enabling additional architectures and ensuring custom ELF patching is applied correctly per architecture.

## Solution
Architecture support updates:

* Added support for the `s390x` and `ppc64el` architectures

Build and prime stage improvements:

* Changed `build-attributes` from `no-patchelf` to `enable-patchelf` and moved the custom build script for `etcd` and `containerd` to the `override-build` step for clarity and consistency.
* Refactored the `override-prime` step to ensure the custom ELF patching script runs per architecture, with Python requirements installed only for supported architectures (`x86_64-linux-gnu` and `aarch64-linux-gnu`). For unknown architectures, the default craftctl behavior is used instead of failing the build.

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed


addyess:
thanks so much @kwmonroe.  Did i get your +1 for any specific reason. 

I need to test this out first to make sure the snap will startup with all the features disabled as the CNI and CSI etc images aren't yet available

snap/snapcraft.yaml
```diff
@@ -97,28 +100,31 @@ parts:
...
       case "$CRAFT_ARCH_TRIPLET_BUILD_FOR" in
         x86_64-linux-gnu)
+          pip3 install -r $CRAFT_PROJECT_DIR/build-scripts/hack/requirements.txt
           python3 $CRAFT_PROJECT_DIR/build-scripts/hack/patchelf.py $CRAFT_PART_INSTALL/bin/etcd \
             --set-rpath /snap/$BASE/current/lib/x86_64-linux-gnu/ \
             --set-interpreter /snap/$BASE/current/lib64/ld-linux-x86-64.so.2
           ;;
         aarch64-linux-gnu)
+          pip3 install -r $CRAFT_PROJECT_DIR/build-scripts/hack/requirements.txt
           python3 $CRAFT_PROJECT_DIR/build-scripts/hack/patchelf.py $CRAFT_PART_INSTALL/bin/etcd \
```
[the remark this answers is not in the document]
addyess: thats intentional

build-scripts/hack/patch-elf-with-lief.sh
bschimke95, line 1: Please move this script in the `build-scripts/hack` directory which is the directory for arbitrary scripts. (I'm not sure if I like the naming but we align here with [upstream](https://www.reddit.com/r/golang/comments/b95sd6/whats_the_recent_pattern_of_a_hack_directory_in/))

build-scripts/patch-elf-with-lief.sh
```diff
@@ -0,0 +1,30 @@
...
+
+BASE="${1}"
+BIN="${2}"
+ARCH_TRIPLET="${3}"
+
+# Note(ben): patchelf messes up the segments when patching the binary
+# due to https://github.com/NixOS/patchelf/issues/446.
+# Instead, we use a custom script based on LIEF to manually patch the rpath and interpreter.
+echo "Patching ELF file: $BIN with LIEF"
+# Only install Python deps on supported arches
```
bschimke95: nit: This comment does not seem to match - the deps are installed regardless of the arch.

build-scripts/hack/patch-elf-with-lief.sh
```diff
@@ -0,0 +1,30 @@
...
+# Instead, we use a custom script based on LIEF to manually patch the rpath and interpreter.
+echo "Patching ELF file: $BIN with LIEF"
+# Only install Python deps on supported arches
+pip3 install -r $DIR/hack/requirements.txt
+
+if [ "$ARCH_TRIPLET" = "x86_64-linux-gnu" ]; then
+python3 $DIR/hack/patchelf.py "$BIN" \
+    --set-rpath /snap/$BASE/current/lib/x86_64-linux-gnu/ \
+    --set-interpreter /snap/$BASE/current/lib64/ld-linux-x86-64.so.2
+else
```
bschimke95, line 30: This will fall back to ARM settings in all cases that are not x86

Shouldn't we rather use an `else if` and fail if the arch is neither AMD or ARM?
I see in the code below that we don't invoke the script with any other arch right now but it would avoid accidents in the future.

snap/snapcraft.yaml
```diff
@@ -303,6 +282,7 @@ parts:
     after:
       - cni
       - containerd
+      - etcd
```
bschimke95, line 287: good catch

bschimke95 (approved):

PR #2006, opened 2025-11-11
chore: Remove CODEOWNERS file

bschimke95:
The codeowners are required for all PRs and are auto-assigned automatically. This causes a lot of notification spam with little value. We should be more mindful about who is added to a PR and select individuals manually.

## Backport

Yes, otherwise old releases would not be affected and spam the team every time there is a backport created.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,0 @@
-*               @canonical/kubernetes


omar-selo (approved):
Thank you!

PR #2011, opened 2025-11-12
docs: DISA STIG docs review and updates

ktsakalozos:
## Description

* broken links in the explanations page
* remove some uncertainty in a couple of sentencies
* use the term of "bootstrap configuration file" instead of introducing a new term "template"

## Backport

1.34

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/docs/canonicalk8s/snap/howto/install/disa-stig.md
+++ b/docs/canonicalk8s/snap/howto/install/disa-stig.md
@@ -282,8 +282,7 @@ audit settings, do one of the following:
     use it.
 
 <!-- Links -->
-<!-- Angle brackets preserve query parameters with & characters -->
-[supported version]: <https://ubuntu.com/about/release-cycle?product=kubernetes&release=canonical+kubernetes&version=all>
+[supported version]: https://ubuntu.com/about/release-cycle?product=kubernetes&release=canonical+kubernetes&version=all 
 [ports and services]: /snap/reference/ports-and-services/
 [FIPS installation guide]: fips.md
 [configure UFW]: /snap/howto/networking/ufw.md


docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -46,6 +46,7 @@ with empty passwords to access this machine.
 
 You can check whether the current account has an empty password by running
 `passwd --status` and looking for "NP" in the second field of the output.
+To add a password to your account use `passwd` and follow the prompts.
```
louiseschmidtgen: Ideally, we should keep the install guide short and to the point. Could you combine this into the previous sentence.
I am thinking something along the lines of "and if the output shows "NP" set a password using `passwd`.
Wdyt?

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -174,16 +178,18 @@ instance each time a new service is exposed externally).
 
 ## Appendix
```
louiseschmidtgen: The templates should be moved to a reference page

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -174,16 +178,18 @@ instance each time a new service is exposed externally).
...
+configuration file for 
+[bootstrapping](#bootstrap-the-first-control-plane-node) the first node
+of a cluster and
 `/var/snap/k8s/common/etc/templates/disa-stig/control-plane.yaml` is the
-template for [joining additional control plane
-nodes](#join-control-plane-nodes). Both of these templates apply configuration
-to align with the following recommendations:
+bootstrap configuration file for [joining additional control plane
+nodes](#join-control-plane-nodes). Both of these bootstrap files
+apply configuration to align with the following recommendations:
```
louiseschmidtgen:
```suggestion
apply configurations to align with the following recommendations:
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -174,16 +178,18 @@ instance each time a new service is exposed externally).
 
 ## Appendix
 
-### Templates Information
+### Bootstrap Configuration Template Files
 
-#### Control plane templates
+#### Control plane bootstrap configuration template file
```
louiseschmidtgen:
```suggestion
#### Control plane bootstrap template
```
This is a bit long for a header

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -195,9 +201,10 @@ to align with the following recommendations:
 | [V-245541]                                                                         | Kubernetes Kubelet must not disable timeouts                          |
 | [V-254800]                                                                         | Kubernetes must have a Pod Security Admission control file configured |
 
-#### Worker node templates
+#### Worker node bootstrap configuration template file
```
louiseschmidtgen:
```suggestion
#### Worker node bootstrap template
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -195,9 +201,10 @@ to align with the following recommendations:
...
 
-`/var/snap/k8s/common/etc/templates/disa-stig/worker.yaml` is the template
+`/var/snap/k8s/common/etc/templates/disa-stig/worker.yaml` is the
+bootstrap configuration file template
```
louiseschmidtgen: I think we can keep the original here, the `bootstrap configuration file template` is quite the mouth-full.

louiseschmidtgen (commented):
Thank you for the improvements Kos, I've left a few suggestions:

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: I'm not sure it's a good idea to make this recommendation at all because it's a potential security issue. Accounts that don't have passwords are often missing them for a reason and adding a password changes the security posture (it can be used for things that would have otherwise been blocked). And potentially there's different/better ways to address the issue depending on the environment. Our main goal here is to alert them to the  problem; I don't think we need to solve it int his case.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -85,22 +86,25 @@ overridden by another higher order file.
 
 ## Apply Kubernetes STIG
 
-{{product}} provides [templates](#templates-information) to apply additional
+{{product}} provides [bootstrap configuration template files]
```
deuberger: These are not just for bootstrap, they're also for joining.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -85,22 +86,25 @@ overridden by another higher order file.
...
 ### Set up control plane nodes
 
 ```{attention}
 Before bootstrapping or joining control-plane nodes, review the
-[templates](#control-plane-templates) and
-[alternative configurations](#alternative-control-plane-configurations).
+[bootstrap configuration template
+files](#control-plane-bootstrap-configuration-template-file)
+and [alternative configurations](#alternative-control-plane-configurations).
 Once a node is bootstrapped, changing certain settings is more difficult
```
deuberger: Again these are not just for bootstrap, but also for joining.

deuberger: I suppose "once a node is bootstrapped" may also need clarified to include the idea of joining.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -129,7 +134,7 @@ First retrieve a join token from an existing control plane node:
 sudo k8s get-join-token <joining-node-hostname> --worker
 ```
 
-Then join the new worker node using the template:
+Then join the new worker node using the bootstrap configuration template file:
```
deuberger: Looks like a search and replace mistake. This is the wrong name here.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -115,7 +119,8 @@ First retrieve a join token from an existing control plane node:
 sudo k8s get-join-token <joining-node-hostname>
 ```
 
-Then join the new control plane node using the template:
+Then join the new control plane node using the
```
deuberger, line 125: Wrong name / search and replace error.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Maybe "initialization"

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -85,22 +86,25 @@ overridden by another higher order file.
...
+and [alternative configurations](#alternative-control-plane-configurations).
 Once a node is bootstrapped, changing certain settings is more difficult
 and may require re-deploying the node or cluster.
 ```
 
 #### Bootstrap the first control-plane node
 
-To initialize the first control plane node using the template:
+To initialize the first control plane node using the
+bootstrap configuration template file:
```
deuberger: What is the reasoning for this change? It makes it more verbose and I'm not sure what it helps with?

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -146,8 +151,7 @@ sudo systemctl disable ssh.service ssh.socket
 According to rule [V-242393] and [V-242394] Kubernetes worker nodes must not
 have sshd service running or enabled. The host STIG rule [V-270665] on the
 other hand expects sshd to be installed on the host. To comply with both
-rules, leave SSH installed, but disable the service. It is probably acceptable
-however to remove SSH if it is not needed.
```
deuberger, line 150: @ktsakalozos what's the reason for this change? It's sort of silly to have a package installed and disabled and increases the attack surface. Also, I think the point of the rule in the OS STIG is that you should use ssh rather than telnet or some other insecure protocol. And I don't think it's intended to mean you must have SSH installed even though that is what it says.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -174,16 +178,18 @@ instance each time a new service is exposed externally).
...
-for [bootstrapping](#bootstrap-the-first-control-plane-node) and
+`/var/snap/k8s/common/etc/templates/disa-stig/bootstrap.yaml` is the
+configuration file for 
+[bootstrapping](#bootstrap-the-first-control-plane-node) the first node
+of a cluster and
 `/var/snap/k8s/common/etc/templates/disa-stig/control-plane.yaml` is the
-template for [joining additional control plane
-nodes](#join-control-plane-nodes). Both of these templates apply configuration
-to align with the following recommendations:
+bootstrap configuration file for [joining additional control plane
```
deuberger:
```suggestion
configuration file for [joining additional control plane
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -174,16 +178,18 @@ instance each time a new service is exposed externally).
...
+nodes](#join-control-plane-nodes). Both of these bootstrap files
```
deuberger:
```suggestion
nodes](#join-control-plane-nodes). Both of these bootstrap files
```
```suggestion
nodes](#join-control-plane-nodes). Both of these configuration files
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: We should remove bootstrap too since this is for joining.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Agreed. I think this is another search and replace error. And we're not bootstrapping here either.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -206,33 +213,36 @@ It applies configuration to align with the following recommendations:
 | [V-242434] | Kubernetes Kubelet must enable kernel protection |
 | [V-245541] | Kubernetes Kubelet must not disable timeouts     |
 
-### Alternative control plane configurations
+### Tuning the control plane configuration
```
deuberger: Links and language elsewhere need updated to account for this change.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -206,33 +213,36 @@ It applies configuration to align with the following recommendations:
...
 
-The STIG templates provided to [set up control plane
-nodes](#set-up-control-plane-nodes) may need adjusted to suit your specific
-needs.
+The STIG bootstrap configuration template files provided
```
deuberger: Another search and replace error...

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -206,33 +213,36 @@ It applies configuration to align with the following recommendations:
...
+to [set up control plane nodes](#set-up-control-plane-nodes) can be
+adjusted to suit your specific needs.
 
 #### Pod Security Admission control file
 
 To comply with rule [V-254800], you must configure a Pod Security Admission
 control file for your Kubernetes cluster. This file defines the Pod Security
 Standards (PSS) that are enforced at the namespace level. By default, the
-templates point to
+bootstrap configuration template files point to
```
deuberger: Another search and replace error...

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -206,33 +213,36 @@ It applies configuration to align with the following recommendations:
...
 `/var/snap/k8s/common/etc/configurations/pod-security-admission-baseline.yaml`,
 which sets the pod security policy to “baseline”, a minimally restrictive
 policy that prevents known privilege escalations.
 
 This policy may be insufficient or impractical in some situations, in which
-case the templates would need to be adjusted by doing one of the following:
+case the bootstrap configuration file templates
```
deuberger: Another search and replace error...

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -206,33 +213,36 @@ It applies configuration to align with the following recommendations:
...
+would need to be adjusted by doing one of the following:
 
-1. Adjust the `--admission-control-config-file` path in the templates to
+1. Adjust the `--admission-control-config-file` path in
+    the bootstrap configuration file to
```
deuberger: Another search and replace error...

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -206,33 +213,36 @@ It applies configuration to align with the following recommendations:
...
     `/var/snap/k8s/common/etc/configurations/pod-security-admission-restricted.yaml`
     rather than the file above. This sets a more restrictive policy.
 2. Edit
     `/var/snap/k8s/common/etc/configurations/pod-security-admission-baseline.yaml`
     to suit your needs based on the [upstream instructions].
 3. Create your own audit policy based on the [upstream instructions] and
-    adjust the `--admission-control-config-file` path used in the templates.
+    adjust the `--admission-control-config-file` path used in the
+    bootstrap configuration file templates.
```
deuberger: Another search and replace error...

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -242,7 +252,8 @@ levels, and configuration options.
 
 To comply with rules [V-242402], [V-242403], [V-242461], [V-242462],
 [V-242463],[V-242464] and [V-242465] you must configure the Kubernetes API
-Server audit log. The STIG templates we provide to bootstrap/join control
+Server audit log. The STIG bootstrap configuration file
```
deuberger: Another search and replace error...

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -21,11 +21,10 @@ This guide assumes the following:
...
 Canonical K8s uses the core22 base snap which includes certified crypto
-libraries from Ubuntu 22.04. Strictly speaking FIPS compliance requires
-deploying on a matching certified kernel (Ubuntu 22.04). In practice auditors
-sometimes accept mixing different kernel and user space library versions as long
-as both are certified. From a technical perspective, FIPS mode should work on
-other OS versions just like the k8s snap.
+libraries from Ubuntu 22.04. FIPS compliance requires
+deploying on a matching certified kernel (Ubuntu 22.04). It is common to mix
+different kernel and user space library versions as long
+as both are certified.
```
deuberger:
```suggestion
deploying on a matching certified kernel (Ubuntu 22.04). However, the k8s snap including FIPS mode does work on other OS versions and auditors may accept mixing
different kernel and user space library versions as long
as both are certified.
```

deuberger: I agree that maybe this can be more concise, but we want to be careful not to miss/change the details. I don't know if it's "common" to mix versions, but I know it is sometimes accepted by auditors. Also, I think it's important to note that it should work from a technical perspective and we are just addressing whether it's compliant.

deuberger (commented):
Thanks for the suggestions. I made a few comments/suggestions. Also, I don't understand why the below change is being applied across the board. Was it was search and replace error or am I misunderstanding something?

> use the term of "bootstrap configuration file" instead of introducing a new term "template"

https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/reference/config-files/control-plane-join-config/ doesn't call it this, so using "bootstrap" everywhere is too much. Also, the file path already has the term "template" so we're inroducing the term either way. I agree that it's a bit messy to add a new term, but I like that "template" implies that it's a starting point. We could say "example" instead, but then people might think it's not production worthy. We could remove it from the path altogether but that would be more then a doc change and might make people think it's used by default.

docs/canonicalk8s/snap/howto/install/disa-stig.md
ktsakalozos: As a reader, I am reading this note and I do not know what do do about it:

> You can check whether the current account has an empty password by running
> `passwd --status` and looking for "NP" in the second field of the output.

Sure I can check this but what do I need to do if I see a "NP". "NP" does not even tell me if this is a good thing or not. We do say these instructions work for accounts that have passwords so if my account is paswordless what should I do? Does my journey end here?

docs/canonicalk8s/snap/howto/install/disa-stig.md
ktsakalozos: No these are not search and replace @deuberger please see the comment https://github.com/canonical/k8s-snap/pull/2011#issuecomment-3522637599

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Are you saying just the "Templates Information" section or also the alternative/tuning sections? I would just do the former I think, but maybe we need to discuss deeper with @nhennigan. In either case, I would do it under a separate card/PR.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Apologies. See https://github.com/canonical/k8s-snap/pull/2011#issuecomment-3522958734.

docs/canonicalk8s/snap/howto/install/disa-stig.md
ktsakalozos: I removed the suggestion to add a password.

docs/canonicalk8s/snap/howto/install/disa-stig.md
ktsakalozos: I like how this page is self contained so I would not put this into a separate reference page. We already have one DISA page for auditing.

I think the title "Appendix" seems too vague/generic. How would you feel about "Advance Configuration"?

docs/canonicalk8s/snap/howto/install/disa-stig.md
ktsakalozos: My concern here is with the uncertainty of the phase "It is probably acceptable". 

Readers visiting this page would want to know how to configure their system based on DISA STIG recommendations. If we are our instructions are "probably acceptable" then we show we are not confident about the result.

What we state above, saying that there are two STIGs that are slightly conflicting/overlapping, should be enough for the reader to make a call.

docs/canonicalk8s/snap/howto/install/fips.md
ktsakalozos: The previous version made me question if what I am reading will give me a FIPS compliant system. We were saying "auditors sometimes accept mixing different kernel and user space library versions" which reads as if we (and the reader) is in the mercy of the auditor. This is not a position I want to be in. As a reader/admin I want to apply the configuration and get to a FIPS compliant setup.

On the word "common", maybe this is not so accurate, I am going to change this to "common practice". What we want to say here is that it is ok to have a certified applications running under a certified kernel, right?

docs/canonicalk8s/snap/howto/install/disa-stig.md
ktsakalozos: In general, forward references within the same page as we have in this case are candidates for separate pages. It is best we avoid forward references.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: I think both points are valid. 
I do think "how-to's" are intended to be more actionable. 
We could phrase it as more of a suggestion than a definite approach. "Consider setting a password via `passwd` if this alligns with your security posture."

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Let's wait on @nhennigan to make the final call. I see the value of having all information in one places but on the other hand the information fits better in a reference from a Diataxis standpoint.

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -21,11 +21,10 @@ This guide assumes the following:
 
 ```{note}
 Canonical K8s uses the core22 base snap which includes certified crypto
-libraries from Ubuntu 22.04. Strictly speaking FIPS compliance requires
-deploying on a matching certified kernel (Ubuntu 22.04). In practice auditors
-sometimes accept mixing different kernel and user space library versions as long
-as both are certified. From a technical perspective, FIPS mode should work on
-other OS versions just like the k8s snap.
+libraries from Ubuntu 22.04. FIPS compliance requires
+deploying on a matching certified kernel (Ubuntu 22.04). It is common
```
bschimke95: I would keep the initial phrasing here or at least make the claim less bold. Auditors might accept this approach but it is definitely not common practice.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -172,18 +174,20 @@ instance each time a new service is exposed externally).
...
    namespaces
 
-## Appendix
+## Advance Configuration
 
-### Templates Information
+### Configuration Files
 
-#### Control plane templates
+#### Control plane & bootstrap configuration files
```
bschimke95: That header sounds a bit weird, e.g. what are control plane files?
Maybe just "Control plane configuration files"?

docs/canonicalk8s/snap/howto/install/fips.md
ktsakalozos: Thank you for this comment. Please have a look at the revised version. Also [the comment](https://github.com/canonical/k8s-snap/pull/2011#discussion_r2522498818) has the rational regarding this change.

nhennigan (approved):
There are a few minor things that I would like to change on this page but I have created a card for next pulse and will follow up there. For now, thanks for greatly improving this page!

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -146,8 +148,7 @@ sudo systemctl disable ssh.service ssh.socket
 According to rule [V-242393] and [V-242394] Kubernetes worker nodes must not
 have sshd service running or enabled. The host STIG rule [V-270665] on the
 other hand expects sshd to be installed on the host. To comply with both
-rules, leave SSH installed, but disable the service. It is probably acceptable
-however to remove SSH if it is not needed.
+rules, leave SSH installed, but disable the service.
```
deuberger:
```suggestion
rules, leave SSH installed, but disable the service. Alternatively, SSH
can be removed and the exception documented.
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: I suggested an alternative.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: I agree it would be better to have a suggestion, but if we're going to make one, we should make a good one that is solid from a security perspective. By setting a password, we're basically enabling password authentication for the account which is generally weaker and not always desirable. Instead, we should find a way to keep an empty password, keep the STIG recommendation to remove nullok support, and still allow SSH key based login.  Searching "pam_unix.so allow ssh key login with no password without using nullok" yielded some promising results, but I haven't tried them yet. I'd be very surprised if there's no way to make it work. We have to test it if we're going to suggest it though.

docs/canonicalk8s/snap/howto/install/disa-stig.md
deuberger: Agreed we can defer to Niamh. I wouldn't just remove the forward references though.

docs/canonicalk8s/snap/howto/install/fips.md
```diff
@@ -21,11 +21,10 @@ This guide assumes the following:
...
 Canonical K8s uses the core22 base snap which includes certified crypto
-libraries from Ubuntu 22.04. Strictly speaking FIPS compliance requires
-deploying on a matching certified kernel (Ubuntu 22.04). In practice auditors
-sometimes accept mixing different kernel and user space library versions as long
-as both are certified. From a technical perspective, FIPS mode should work on
-other OS versions just like the k8s snap.
+libraries from Ubuntu 22.04. Strict FIPS compliance requires deploying on a
+matching certified kernel and user space libraries (e.g., both from Ubuntu 22.04).
+However, in practice, it is common to mix different certified kernel and user space
+library versions, as long as all components are FIPS-certified.
```
deuberger:
```suggestion
However, mixing different kernel and user space versions is supported from a technical perspective and may be acceptable to auditors in some situations.
```

docs/canonicalk8s/snap/howto/install/fips.md
deuberger: See updated suggestion.

docs/canonicalk8s/snap/howto/install/fips.md
deuberger: The goal here is to make it clear that it's technically supported, but officially not compliant. On the other hand some users do want to do this  and we're trying to clue them in that it might be acceptable though ultimately that's between them and their auditor.

docs/canonicalk8s/snap/howto/install/fips.md
deuberger: See updated suggestion.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -85,22 +85,23 @@ overridden by another higher order file.
...
-[alternative configurations](#alternative-control-plane-configurations).
-Once a node is bootstrapped, changing certain settings is more difficult
+Before bootstrapping or joining control plane nodes, review the
+respective [configuration files](#configuration-files) as well as the
+alternative configuration options for [audit logs and PSS](#audit-logs-and-pss-configuration).
+Once a node is configured, changing certain settings is more difficult
 and may require re-deploying the node or cluster.
 ```
 
 #### Bootstrap the first control-plane node
```
louiseschmidtgen:
```suggestion
#### Bootstrap the first control plane node
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -195,44 +198,47 @@ to align with the following recommendations:
 | [V-245541]                                                                         | Kubernetes Kubelet must not disable timeouts                          |
 | [V-254800]                                                                         | Kubernetes must have a Pod Security Admission control file configured |
 
-#### Worker node templates
+#### Worker node join configuration file
```
louiseschmidtgen: Let's please be consistent either we use `node-join` configuration file everywhere or `node join`

docs/canonicalk8s/snap/howto/install/fips.md
louiseschmidtgen: As @omar-selo pointed out, we know what this means but it is confusing for the user. 
We need something simpler: fips is supported on 22.04 and (anticipated) on 24.04. The mix and match and likely the auditors won't mind is a bit unclear. 

This may be out of scope for this PR.

docs/canonicalk8s/snap/howto/install/disa-stig.md
ktsakalozos: Thank you. Accepted.

docs/canonicalk8s/snap/howto/install/fips.md
ktsakalozos: I see there is no consensus over this change. I am reverting it to the original wording. We can improve this part in a future PR as Louise suggested.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -85,22 +85,23 @@ overridden by another higher order file.
...
 
 ### Set up control plane nodes
 
 ```{attention}
-Before bootstrapping or joining control-plane nodes, review the
-[templates](#control-plane-templates) and
-[alternative configurations](#alternative-control-plane-configurations).
-Once a node is bootstrapped, changing certain settings is more difficult
+Before bootstrapping or joining control plane nodes, review the
+example [configuration files](#configuration-files) as well as the
```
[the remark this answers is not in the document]
deuberger: Good catch copilot. Also here we say "Example configuration files" whereas the seciton header is "configuration example files" We should be consistent and fix the link @ktsakalozos @louiseschmidtgen.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -172,18 +174,20 @@ instance each time a new service is exposed externally).
    meaning all user pods must be in user specific namespaces rather than system
    namespaces
 
-## Appendix
+## Advanced Configuration
 
-### Templates Information
+### Configuration Example Files
```
deuberger: See comment above. Let's make this consistent.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -195,44 +199,47 @@ to align with the following recommendations:
...
+to [set up control plane nodes](#set-up-control-plane-nodes) can be
+adjusted to suit your specific needs.
 
 #### Pod Security Admission control file
 
 To comply with rule {ref}`254800`, you must configure a Pod Security Admission
 control file for your Kubernetes cluster. This file defines the Pod Security
 Standards (PSS) that are enforced at the namespace level. By default, the
-templates point to
+bootstrap and control plane configuration files point to
```
deuberger:
```suggestion
example configuration files point to
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -195,44 +199,47 @@ to align with the following recommendations:
...
-1. Adjust the `--admission-control-config-file` path in the templates to
+1. Set the `--admission-control-config-file` path in
+    the bootstrap and control plane configuration files to
     `/var/snap/k8s/common/etc/configurations/pod-security-admission-restricted.yaml`
-    rather than the file above. This sets a more restrictive policy.
+    rather than the baseline one. This sets a more restrictive policy.
 2. Edit
     `/var/snap/k8s/common/etc/configurations/pod-security-admission-baseline.yaml`
     to suit your needs based on the [upstream instructions].
 3. Create your own audit policy based on the [upstream instructions] and
```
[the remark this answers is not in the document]
deuberger: @ktsakalozos @louiseschmidtgen I think copilot is right here.

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -242,11 +249,11 @@ levels, and configuration options.
 
 To comply with rules {ref}`242402`, {ref}`242403`, {ref}`242461`, {ref}`242462`,
 {ref}`242463`, {ref}`242464`, and {ref}`242465` you must configure the 
-Kubernetes API Server audit log. The STIG templates we provide to 
-bootstrap/join control plane nodes configures the Kubernetes API servers audit 
-settings and policy to comply with these recommendations.
+Kubernetes API Server audit log. The STIG configuration files we provide to
+bootstrap and join control plane nodes configure the Kubernetes API
+servers audit settings and policy to comply with these recommendations.
```
deuberger:
```suggestion
Kubernetes API Server audit log. The example configuration files we provide configure the Kubernetes API
servers audit settings and policy to comply with these recommendations.
```

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: This will be addressed in https://github.com/canonical/k8s-snap/pull/2129, marking as resolved.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Kos chose "Once a node is configured" marking as resolved.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: This seems fine to me, otherwise the code block floats without an explanation. Marking as resolved.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: This looks outdated.

louiseschmidtgen (approved):
lgtm

nhennigan (approved):

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -271,10 +284,11 @@ audit settings, do one of the following:
     use it.
 
 <!-- Links -->
-[supported version]: https://ubuntu.com/about/release-cycle#canonical-kubernetes-release-cycle
+[supported version]: https://ubuntu.com/about/release-cycle?product=kubernetes&release=canonical+kubernetes&version=all
```
deuberger: @louiseschmidtgen this still doesn't quite work. Do we have a follow-up ticket?

docs/canonicalk8s/snap/howto/install/disa-stig.md
```diff
@@ -177,67 +185,72 @@ instance each time a new service is exposed externally).
...
-### Templates Information
+The following configuration files apply DISA STIG settings for control
+plane nodes:
 
-#### Control plane templates
+- `bootstrap.yaml`: Used for
+  [bootstrapping](#bootstrap-the-first-control-plane-node) the first
+  control plane node
+- `control-plane.yaml`: Used for
+  [joining additional control plane nodes](#join-control-plane-nodes)
```
deuberger: Two issues here:
1. It's less now where these files are actually located.
2. This set of changes was not applied to the worker section below which still has the full file path so the sections feel inconsistent now.

deuberger (commented):
Two nits, but good otherwise. Thanks.

docs/canonicalk8s/snap/howto/install/disa-stig.md
nhennigan: I looked into this before they updated the page. The first update they made that made it break initially was caused by them removing the page anchor that we could link to. Now that is has been updated we can provide a link to the filtered page https://ubuntu.com/about/release-cycle?product=kubernetes&release=canonical+kubernetes&version=all which will show the Canonical Kubernees page. I am conscious though that this link is very long with the filters applied and I would not be surprised if this was to change. When I get the linkcheck running however it should catch these things

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: I've reverted to the original.

docs/canonicalk8s/snap/howto/install/disa-stig.md
louiseschmidtgen: Let's follow-up on the link @nhennigan, something is not working about the link when you follow it from the readthedocs page. I suspect the page parameter is not working due to a readthedocs rendering issue.

PR #2018, opened 2025-11-12
fix: bump sqlite to 3.50.2

ethandcosta:
Sqlite version 3.45.2 has a high vulnerability CVE (https://github.com/advisories/GHSA-2m69-gcr7-jv3q). To fix, sqlite has been bumped to 3.50.2


bschimke95:
Looks like the k8s-dqlite tests are not happy with that change (I just re-ran this to see if it is just a flake - it is not.)

PR #2048, opened 2025-11-18
chore: Skip automated downgrades

bschimke95:
The update-components workflow pulls the newest upstream version and overwrites the local value. Certain cases require keeping a specific version, such as when one component (runc) must align with another (containerd) that is already end-of-life but still needs a controlled version pin.

The change updates the workflow to perform replacements only when the upstream version is newer than the current one. It does not handle the opposite scenario; the workflow will continue proposing upgrades even when an older version is intentionally retained.

# Backports

We need to backport this to all releases since they run the script from the branch when checking for updates.

### DIFF
--- a/build-scripts/hack/update-component-versions.py
+++ b/build-scripts/hack/update-component-versions.py
@@ -159,9 +159,25 @@ def update_component_versions(dry_run: bool):
         LOG.info("Updating version for %s", component)
         version: str = get_version()
         path = COMPONENTS / component / "version"
+        existing = Path(path)
+        existing_version_text = existing.read_text().strip() if existing.exists() else None
+        upstream_version_text = version.strip()
+
+        existing_parsed = parse_version(existing_version_text) if existing_version_text else None
+        upstream_parsed = parse_version(upstream_version_text)
+
+        # If both versions parse and the existing one is greater than upstream, skip update.
+        if existing_parsed and upstream_parsed and existing_parsed > upstream_parsed:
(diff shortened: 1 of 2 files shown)


build-scripts/hack/update-component-versions.py
```diff
@@ -159,9 +159,25 @@ def update_component_versions(dry_run: bool):
...
+        upstream_parsed = parse_version(upstream_version_text)
+
+        # If both versions parse and the existing one is greater than upstream, skip update.
+        if existing_parsed and upstream_parsed and existing_parsed > upstream_parsed:
+            LOG.info(
+            "Existing version %s is greater than upstream %s; keeping existing version",
+            existing_version_text,
+            upstream_version_text,
+            )
+            continue
```
HomayoonAlimohammadi, line 176: QQ: Do we want to stay consistent with the upstream version? If so, can we downgrade in case the existing is higher than the upstream?

bschimke95: Good question Hue. That is exactly the behavior we want to avoid with this PR (see second part of PR description).
For example, we manually  override the [runc version](https://github.com/canonical/k8s-snap/pull/2047). Without this check, the automation would always try to create a PR to downgrade this.

HomayoonAlimohammadi: Only if I've paid more attention to the PR title "Skip automated downgrades" :/

PR #2051, opened 2025-11-19
test: Don't setup nodes before airgapped test

bschimke95:
The airgapped test is marked as `disable_k8s_bootstrapping` which will not run `k8s bootstrap` on the node but still install the snap on the node. The airgapped test then later tries to actually setup the node within the test. Since the snap is already installed, this causes a snap refresh/upgrade from a potentially non-supported upgrade path. Marking this test as `no_setup` gives full setup control to the test itself and avoids this scenario.

## Issue

https://github.com/canonical/k8s-snap/actions/runs/19484748307/job/55764474151

## Backport

Yes, to all releases


berkayoz (approved):
LGTM

PR #2077, opened 2025-11-24
fix: Cluster recovery flags logic and docs

bschimke95:
This pull request updates the documentation and code for the `k8sd cluster-recover` command to clarify backup procedures and improve the handling of optional state directories. 

Note: Both fixes belong together (the docs only work with the fix) and need to be backported together which is why I put both commits into one PR. I will not squash them.

**Documentation improvements:**

* Added a note clarifying that the backup command only creates a snapshot for `k8sd`, and that users should follow the upstream procedure for `etcd` backups.
* Fixed the recovery commands

**CLI fixes:**

* Updated pre-check logic in `recoveryCmdPrechecks` to only require state directory flags if their corresponding components are not being skipped. 
* Removed the unused `logDebugf` helper function from `k8sd_cluster_recover.go`.

## Backport

Yes, to 1.32 - 1.34

### DIFF
--- a/docs/canonicalk8s/snap/howto/restore-quorum.md
+++ b/docs/canonicalk8s/snap/howto/restore-quorum.md
@@ -1,6 +1,6 @@
 # How to recover a cluster after quorum loss
 
-Highly available {{product}} clusters are designed to tolerate the loss of 
+Highly available {{product}} clusters are designed to tolerate the loss of
 one or more nodes. Both [etcd] and [Dqlite] use [Raft] protocol
 where an elected leader holds the definitive copy of the database, which is
 then replicated on two or more secondary nodes.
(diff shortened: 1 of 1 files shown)


docs/canonicalk8s/snap/howto/restore-quorum.md
```diff
@@ -81,23 +85,22 @@ sudo /snap/k8s/current/bin/k8sd cluster-recover \
 `````
 
 ```{note}
-By default, the command will recover both the cluster and Kubernetes datastore
```
nhennigan, line 84: Having three notes so close together in one section takes away from the importance of the note. Maybe this one can be removed and the info added as plain text before the tabs?

nhennigan (commented):
Thank you Ben for being mindful of the docs impact with your update. I have left a small comment on the over use of notes but I think this is definitely clearer now. 😸

docs/canonicalk8s/snap/howto/restore-quorum.md
bschimke95: Good point! Done

nhennigan (approved):

ktsakalozos (approved):
LGTM +1

PR #2102, opened 2025-12-01
feat: coredns rebalancing controller

louiseschmidtgen:
## Description

CoreDNS is deployed with two replicas and anti-affinity/topology spread constraints to encourage pods to run on separate nodes. When a second node joins the cluster, both CoreDNS pods remain on the original node—they are not automatically redistributed. Only after manually deleting a pod does the scheduler place it on the new node. 

Kubernetes does not automatically rebalance or evict existing pods when new nodes join. Anti-affinity and topology spread constraints only affect new pod scheduling, not already-running pods. 

As soon as a new node is added, CoreDNS pods should be automatically distributed (one per node) without requiring manual pod deletion or a rolling restart.

## Solution

We can use a reconciler controller watching for nodes to become ready, if necessary this controller restarts the deployment of coreDNS which will re-distribute the pod equally across nodes once we have at least two ready nodes.

Our scheduling logic will take care of balancing the CoreDNS deployments that are scheduled in a multi-node cluster.
A sensible rollout strategy is already defined in the [CoreDNS helm chart](https://github.com/coredns/helm/blob/master/charts/coredns/values.yaml#L28).

Using `matchLabelKeys` set to `pod-template-hash` to prevent “old” pods from counting into the topology constraints.

⚠️ Warning: This change requires setting `"whenUnsatisfiable": "DoNotSchedule",` which forces pods to be scheduled on different nodes at all times. This is opposed to the `ScheduleAnyways` strategy. 

## Backport

NA

## Checklist

- [x] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/src/k8s/pkg/k8sd/controllers/dnsrebalancer/controller.go
+++ b/src/k8s/pkg/k8sd/controllers/dnsrebalancer/controller.go
@@ -12,7 +12,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-type Controller struct {
+type controller struct {
 	logger           logr.Logger
 	client           client.Client
 	getClusterConfig func(context.Context) (types.ClusterConfig, error)
@@ -24,8 +24,8 @@ func NewController(
 	client client.Client,
 	getClusterConfig func(context.Context) (types.ClusterConfig, error),
 	snap snap.Snap,
-) *Controller {
-	return &Controller{
+) *controller {
+	return &controller{
 		logger:           logger,
 		client:           client,
 		getClusterConfig: getClusterConfig,
@@ -34,7 +34,7 @@ func NewController(
 }
 
 // SetupWithManager sets up the controller with the manager.
-func (r *Controller) SetupWithManager(mgr ctrl.Manager) error {
+func (r *controller) SetupWithManager(mgr ctrl.Manager) error {
 	// Watch Node resources, trigger reconciliation when node status changes
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Node{}).
--- a/src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
+++ b/src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
@@ -13,18 +13,21 @@ import (
 )
 
 // Reconcile implements the reconciliation loop.
-func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := log.FromContext(ctx).WithValues("node", req.Name)
 
+	if r.getClusterConfig == nil {
+		log.Info("getClusterConfig is nil")
+		return ctrl.Result{}, nil
+	}
+
 	// skip reconcile if dns is disabled
-	if r.getClusterConfig != nil {
-		config, err := r.getClusterConfig(ctx)
-		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to get cluster config: %w", err)
-		}
-		if !config.DNS.GetEnabled() {
-			return ctrl.Result{}, nil
-		}
+	config, err := r.getClusterConfig(ctx)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get cluster config: %w", err)
+	}
+	if !config.DNS.GetEnabled() {
+		return ctrl.Result{}, nil
 	}
 
 	// Count ready nodes
@@ -39,7 +42,7 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	readyCount := countReadyNodes(nodeList)
 	if readyCount < 2 {
 		log.V(1).Info("Less than 2 nodes ready, skipping rebalancing check", "readyCount", readyCount)
-		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+		return ctrl.Result{}, nil
 	}
 
 	log.V(1).Info("Sufficient nodes ready, checking CoreDNS distribution", "readyCount", readyCount)
@@ -71,7 +74,7 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 }
 
 // coreDNSNeedsRebalancing checks if CoreDNS pods need rebalancing.
-func (r *Controller) coreDNSNeedsRebalancing(ctx context.Context) (bool, error) {
+func (r *controller) coreDNSNeedsRebalancing(ctx context.Context) (bool, error) {
 	pods := &corev1.PodList{}
 	if err := r.client.List(ctx, pods, client.InNamespace("kube-system"), client.MatchingLabels{"k8s-app": "coredns", "app.kubernetes.io/instance": "ck-dns"}); err != nil {
 		return false, err
--- a/src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile_test.go
+++ b/src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile_test.go
@@ -3,7 +3,6 @@ package dnsrebalancer
 import (
 	"context"
 	"testing"
-	"time"
 
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
@@ -35,7 +34,7 @@ func TestReconcile_LessThanTwoNodesReady(t *testing.T) {
 		WithObjects(node).
 		Build()
 
-	reconciler := &Controller{
+	reconciler := &controller{
 		logger: ctrl.Log.WithName("test"),
 		client: fakeClient,
 		snap:   nil,
@@ -43,7 +42,7 @@ func TestReconcile_LessThanTwoNodesReady(t *testing.T) {
 
 	result, err := reconciler.Reconcile(ctx, ctrl.Request{})
 
-	g.Expect(result).To(Equal(ctrl.Result{RequeueAfter: 30 * time.Second}))
+	g.Expect(result).To(Equal(ctrl.Result{}))
 	g.Expect(err).ToNot(HaveOccurred())
 }
 
@@ -102,7 +101,7 @@ func TestReconcile_CoreDNSAlreadyBalanced(t *testing.T) {
 		WithObjects(&nodes[0], &nodes[1], &pods[0], &pods[1]).
 		Build()
 
-	reconciler := &Controller{
+	reconciler := &controller{
 		logger: ctrl.Log.WithName("test"),
 		client: fakeClient,
 		snap:   nil,
@@ -142,7 +141,7 @@ func TestCoreDNSNeedsRebalancing_AllPodsSameNode(t *testing.T) {
 		WithObjects(&pods[0], &pods[1]).
 		Build()
 
-	reconciler := &Controller{
+	reconciler := &controller{
 		logger: ctrl.Log.WithName("test"),
 		client: fakeClient,
 		snap:   nil,
@@ -182,7 +181,7 @@ func TestCoreDNSNeedsRebalancing_Distributed(t *testing.T) {
 		WithObjects(&pods[0], &pods[1]).
 		Build()
 
-	reconciler := &Controller{
+	reconciler := &controller{
 		logger: ctrl.Log.WithName("test"),
 		client: fakeClient,
 		snap:   nil,
@@ -223,7 +222,7 @@ func TestCoreDNSNeedsRebalancing_IgnoresUnscheduledPods(t *testing.T) {
 		WithObjects(&pods[0], &pods[1]).
 		Build()
 
-	reconciler := &Controller{
+	reconciler := &controller{
 		logger: ctrl.Log.WithName("test"),
 		client: fakeClient,
 		snap:   nil,
@@ -264,7 +263,7 @@ func TestCoreDNSNeedsRebalancing_AllPodsPending(t *testing.T) {
 		WithObjects(&pods[0], &pods[1]).
 		Build()
 
-	reconciler := &Controller{
+	reconciler := &controller{
 		logger: ctrl.Log.WithName("test"),
 		client: fakeClient,
 		snap:   nil,
@@ -284,7 +283,7 @@ func TestCoreDNSNeedsRebalancing_NoPods(t *testing.T) {
 		WithScheme(scheme.Scheme).
 		Build()
 
-	reconciler := &Controller{
+	reconciler := &controller{
 		logger: ctrl.Log.WithName("test"),
 		client: fakeClient,
 		snap:   nil,


src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -342,6 +343,11 @@ func (a *App) onPostJoin(ctx context.Context, s state.State, initConfig map[stri
 		log.Error(err, "Failed to apply custom CRDs")
 	}
 
+	// Rebalance CoreDNS pods when transitioning from to an HA (3 node) cluster
+	if err := rebalanceCoreDNSIfNeeded(ctx, k8sClient); err != nil {
+		log.Error(err, "Failed to rebalance CoreDNS deployment")
+	}
+
```
berkayoz: Does this trigger for worker nodes / how do we handle worker joins?

src/k8s/pkg/k8sd/app/hooks_join.go
```diff
@@ -390,6 +396,31 @@ func getNodeVersion(ctx context.Context, snap snap.Snap) (*versionutil.Version,
...
+	nodes, err := k8sClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list nodes: %w", err)
+	}
+
+	nodeCount := len(nodes.Items)
+	log.Info("Checking if CoreDNS rebalance is needed", "nodeCount", nodeCount)
+
+	// Only rebalance when we have exactly 3 nodes (2→3 transition)
+	if nodeCount == 3 {
```
berkayoz: This feels fragile in terms of nodes doing parallel join/leave operations, especially for workers where we can't gate this.

louiseschmidtgen: Agreed, I scrapped this in favour of watching for 2 nodes in a Ready state + check if CoreDNS pods need re-balancing.

src/k8s/pkg/k8sd/app/hooks_join.go
louiseschmidtgen: This should be handled in the new controller approach

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	// Check if minimum 2 nodes are Ready
+	if err := control.WaitUntilReady(ctx, func() (bool, error) {
+		readyCount, err := c.getNodesReadyCount(ctx, k8sClient)
+		if err != nil {
+			log.V(1).Info("Failed to get control plane counts while waiting", "error", err)
+			return false, nil
+		}
+		log.V(1).Info("Checking control plane readiness", "readyControlPlaneNodes", readyCount)
```
berkayoz: Checking control plane only feels off to me.
What happens on a cluster that has 1 cplane multi workers?
What happens when control plane nodes are tainted / pods can not be scheduled?

We probably should check for workers as well.

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to check CoreDNS pods distribution: %w", err)
+	}
+	if !needsRebalancing {
+		log.V(1).Info("CoreDNS pods are already balanced across control plane nodes")
+		return nil
+	}
+
+	log.Info("Triggering CoreDNS deployment rollout restart to rebalance pods across control plane nodes")
+	if err := k8sClient.RestartDeployment(ctx, "coredns", "kube-system"); err != nil {
```
berkayoz: I believe we need leadership / leader election for this controller. This running on multiple nodes could create a large window of interruptions.

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
```diff
@@ -0,0 +1,102 @@
...
+			// Pods are on different nodes - no rebalancing needed
+			return false, nil
+		}
+	}
+	// All pods are on the same node - rebalancing needed
+	return true, nil
+}
+
+func (c *DNSRebalancerController) getNodesReadyCount(ctx context.Context, k8sClient *kubernetes.Client) (readyNodeCount int, err error) {
+	nodes, err := k8sClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
berkayoz: This lists all nodes right? Not just control plane, if so above messages need to be updated and my comment above is not applicable.

louiseschmidtgen: +1 yes I forgot to update the log after I changed it

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
louiseschmidtgen: Agreed, I added a lease to protect the critical section.

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
```diff
@@ -0,0 +1,149 @@
...
+	for _, node := range nodes.Items {
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == "Ready" && condition.Status == "True" {
+				readyNodeCount++
+				break
+			}
+		}
+	}
+	return readyNodeCount, nil
+}
```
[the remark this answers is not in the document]
louiseschmidtgen: You're right I added a test.

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
```diff
@@ -0,0 +1,163 @@
...
+
+func NewDNSRebalancerController(snap snap.Snap, waitReady func()) *DNSRebalancerController {
+	return &DNSRebalancerController{
+		snap:      snap,
+		waitReady: waitReady,
+	}
+}
+
+func (c *DNSRebalancerController) Run(ctx context.Context) error {
```
bschimke95: So let me see if understand this workflow.

1. A new node joins the cluster and start the controller
2. The controller checks the amount of nodes in the cluster
3. a) NodesReady >= 2 -> continue to 4.
   b) NodesReads < 2 -> wait and retry
4. A leader election run will happen and whoever gets there first is the leader.
5. The `OnStartLeading`callback kicks in on the leader node and restarts the CoreDNS deployment if necessary.

A few questions:
* All nodes will run for the leader election at about the same time since they wait in step 3b - a new leader will get elected and try to rebalance. The other controllers will [wait here](https://github.com/canonical/k8s-snap/pull/2102/files#diff-9ba3130d975ff59b35d959195afb0a33ac52d4c185d7bf0728e58d8f555a93a7R105) until it is their turn to check for a rebalance, right? So if the rebalancing fails on the first node, the second node kicks in and tries the rebalancing?
* The LeaseDuration is 15seconds. Will the existing leader automatically try to refresh the lease? 
* If not, what happens if the redeploying takes longer than 15s? The next node would also try to do a rebalancing, right?

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
```diff
@@ -0,0 +1,163 @@
+package controllers
```
bschimke95: Can we add a very simple integration test (or add it to another test that already includes > one node). That verifies that the rebalancing happens?

bschimke95 (commented):
Did a first pass, looks good. A few questions for my understanding.

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
louiseschmidtgen: 1. Not quite, if the first node fails to restart the deployment the other nodes will only try to rebalance when the leader node gives up its lease. (-> This can probably be improved with a retry+backoff or eventual leadership handoff).
2. Yes that happens within 10 seconds: `RenewDeadline: 10 * time.Second,`
3. That doesn't matter the lease will be refreshed. So the next node only rebalances if the leader node doesn't refresh the lease (it died) or if it gave up its lease (when can add that).

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
louiseschmidtgen: Yes, I've added one.

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
louiseschmidtgen: @bschimke95 Should we add any of this in a v2 or do you think it is critical to have in this first version?

louiseschmidtgen: In my last commit, I added one more return statement in the critical section, the node will hand-over if it fails anywhere in the critical section. Another node can then try.

bschimke95 (approved):

src/k8s/pkg/k8sd/controllers/dns_rebalancer.go
```diff
@@ -0,0 +1,163 @@
...
+			Name:      "coredns-rebalancer",
+			Namespace: "kube-system",
+		},
+		Client: k8sClient.CoordinationV1(),
+		LockConfig: resourcelock.ResourceLockConfig{
+			Identity: c.snap.Hostname(),
+		},
+	}
+
+	leaderElector, err := leaderelection.NewLeaderElector(leaderelection.LeaderElectionConfig{
```
berkayoz: Why don't we use the coordinator similar to `csrsigning` and `upgrade` controllers? We can use the controller runtime this way as well(that was my intention when I mentioned leader election, sorry I should've been more clear) . We can watch `Node` resource to trigger the Reconcile loop. It would be ideal if we move away from home-baked controllers.

louiseschmidtgen: Done, thanks for the input!

src/k8s/pkg/k8sd/controllers/dnsrebalancer/controller.go
```diff
@@ -0,0 +1,55 @@
...
+func NewController(
+	logger logr.Logger,
+	client client.Client,
+	getClusterConfig func(context.Context) (types.ClusterConfig, error),
+	snap snap.Snap,
+) *Controller {
+	return &Controller{
+		logger:           logger,
+		client:           client,
+		getClusterConfig: getClusterConfig,
```
berkayoz, line 31: Do we use this function, if not I believe we can remove it from here and the call stack.

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+	log := log.FromContext(ctx).WithValues("node", req.Name)
+
+	// Count ready nodes
+	nodeList := &corev1.NodeList{}
+	if err := r.client.List(ctx, nodeList); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		log.Error(err, "Failed to list nodes")
+		return ctrl.Result{}, err
```
berkayoz: I don't remember exactly how this works in k8sd but

```suggestion
		return ctrl.Result{}, fmt.Errorf("Failed to list nodes: %w", err)
```

wouldn't this log the error on it's own?

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	log.V(1).Info("Sufficient nodes ready, checking CoreDNS distribution", "readyCount", readyCount)
+
+	// Check if rebalancing is needed
+	needsRebalancing, err := r.coreDNSNeedsRebalancing(ctx)
+	if err != nil {
+		log.Error(err, "Failed to check CoreDNS pods distribution")
+		return ctrl.Result{}, err
```
berkayoz: Same as above comment, we can reduce it to a return if controller-runtime already logs the returned `err`

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+// coreDNSNeedsRebalancing checks if CoreDNS pods need rebalancing.
+func (r *Controller) coreDNSNeedsRebalancing(ctx context.Context) (bool, error) {
+	pods := &corev1.PodList{}
+	if err := r.client.List(ctx, pods, client.InNamespace("kube-system"), client.MatchingLabels{"k8s-app": "coredns"}); err != nil {
+		return false, fmt.Errorf("failed to list CoreDNS pods: %w", err)
+	}
+
+	if len(pods.Items) == 0 {
+		return false, fmt.Errorf("no CoreDNS pods found")
+	}
```
berkayoz, line 85: This returns an error in the Reconcile function, do we want to keep retrying until there are coredns pods available? Or is this a return early / skip? (This might be a problem if dns feature is disabled)

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+	log.Info("CoreDNS pods need rebalancing, triggering deployment rollout restart")
+
+	// Get kubernetes client to restart deployment
+	k8sClient, err := r.snap.KubernetesClient("")
+	if err != nil {
+		log.Error(err, "Failed to get Kubernetes client")
+		return ctrl.Result{}, err
+	}
+
+	if err := k8sClient.RestartDeployment(ctx, "coredns", "kube-system"); err != nil {
```
berkayoz, line 68: Do we need to check if the dns feature is enabled before the checks and restart? If this returns an error when `coredns` deployment does not exist, we could be stuck in a retry loop unnecessarily.

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile_test.go
```diff
@@ -0,0 +1,296 @@
...
+		logger: ctrl.Log.WithName("test"),
+		client: fakeClient,
+		snap:   nil,
+	}
+
+	result, err := reconciler.Reconcile(ctx, ctrl.Request{})
+
+	g.Expect(result).To(Equal(ctrl.Result{}))
+	g.Expect(err).ToNot(HaveOccurred())
+}
```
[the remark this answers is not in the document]
louiseschmidtgen: The integration test covers this.

src/k8s/pkg/k8sd/controllers/dnsrebalancer/controller.go
```diff
@@ -0,0 +1,55 @@
...
+			if !ok {
+				return false
+			}
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionTrue {
+					return true
+				}
+			}
+			return false
+		})).
```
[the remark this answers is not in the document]
louiseschmidtgen: Unnecessary complexity

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+	// Filter scheduled pods (NodeName != ""). Pending pods can cause false positives.
+	scheduled := make([]corev1.Pod, 0, len(pods.Items))
+	for _, pod := range pods.Items {
+		if pod.Spec.NodeName != "" {
+			scheduled = append(scheduled, pod)
+		}
+	}
+
+	// If fewer than 2 pods are scheduled, we cannot assess imbalance yet.
+	if len(scheduled) < 2 {
```
berkayoz, line 95: Since reconcile is triggered by `Ready` nodes, if we return reconcile early here what happens if scheduling happens later on? Should this return an error/retry until 2 or more pods are scheduled?

We could also configure the controller to watch both the nodes and coredns pods to trigger reconcile instead of returning an error to retry here. (Returning a retry feels simpler since we configure it to have min 2 replicas)

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if err := k8sClient.RestartDeployment(ctx, "coredns", "kube-system"); err != nil {
+		log.Error(err, "Failed to restart CoreDNS deployment, will retry on next reconciliation")
+		return ctrl.Result{}, err
+	}
+
+	log.Info("Successfully triggered CoreDNS deployment restart")
+	return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
louiseschmidtgen: PR description was outdated

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// coreDNSNeedsRebalancing checks if CoreDNS pods need rebalancing.
+func (r *Controller) coreDNSNeedsRebalancing(ctx context.Context) (bool, error) {
+	pods := &corev1.PodList{}
+	if err := r.client.List(ctx, pods, client.InNamespace("kube-system"), client.MatchingLabels{"k8s-app": "coredns"}); err != nil {
```
berkayoz: Do we have a finer-grained selector/more labels (like something to check for `ck-dns` etc.) to make sure we only apply this controller to our own coredns deployment?

src/k8s/pkg/k8sd/controllers/dnsrebalancer/controller.go
louiseschmidtgen: On it :)

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
louiseschmidtgen: The `RestartDeployment` function already performs this check before marking the pods for restart.
If for some reason our pods disappeared since the rebalance check this function will return an error.
Is that your question? Maybe I am missing something.

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
louiseschmidtgen: Yes, good point- I'll add the ck-dns check :)

src/k8s/pkg/k8sd/controllers/dnsrebalancer/controller.go
louiseschmidtgen: Actually, we do need it to check if DNS is disabled in which case we can skip the reconcile loop.

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
louiseschmidtgen: I chose the option to return an error and retry.

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
louiseschmidtgen: I've also added a check for DNS being disabled at the start of the reconcile logic.

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
louiseschmidtgen: Now we check if CoreDNS is enabled, if there are no pods we return an error and perform a retry.

src/k8s/pkg/k8sd/features/coredns/coredns.go
```diff
@@ -113,20 +114,24 @@ func ApplyDNS(ctx context.Context, snap snap.Snap, dns types.DNS, kubelet types.
...
+						"k8s-app":                    "coredns",
 					},
 				},
+				"matchLabelKeys": []string{"pod-template-hash"},
 			},
 			{
 				"maxSkew":           1,
 				"topologyKey":       "kubernetes.io/hostname",
-				"whenUnsatisfiable": "ScheduleAnyway",
+				"whenUnsatisfiable": "DoNotSchedule",
```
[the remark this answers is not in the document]
louiseschmidtgen: True, this is the trade-off to force the pods to rebalance.

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
```diff
@@ -0,0 +1,120 @@
...
+	// skip reconcile if dns is disabled
+	if r.getClusterConfig != nil {
+		config, err := r.getClusterConfig(ctx)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get cluster config: %w", err)
+		}
+		if !config.DNS.GetEnabled() {
+			return ctrl.Result{}, nil
+		}
+	}
```
berkayoz, line 31: We could move the `getClusterConfig` nil check to `NewController`, return an error there and not worry about it. If we want to keep it here maybe we can invert the check like 

```suggestion
	if r.getClusterConfig == nil {
        log.Info("getClusterConfig is nil")
        return ctrl.Result{}, nil
	}
	
    // skip reconcile if dns is disabled
	config, err := r.getClusterConfig(ctx)
	if err != nil {
		return ctrl.Result{}, fmt.Errorf("failed to get cluster config: %w", err)
	}
	if !config.DNS.GetEnabled() {
		return ctrl.Result{}, nil
	}
```

src/k8s/pkg/k8sd/controllers/dnsrebalancer/controller.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/canonical/k8s/pkg/k8sd/types"
+	"github.com/canonical/k8s/pkg/snap"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+type Controller struct {
```
berkayoz: We could make this unexported to prevent people from using the struct directly instead of using `NewController`

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
```diff
@@ -0,0 +1,120 @@
...
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	readyCount := countReadyNodes(nodeList)
+	if readyCount < 2 {
+		log.V(1).Info("Less than 2 nodes ready, skipping rebalancing check", "readyCount", readyCount)
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
berkayoz: I think we don't need to requeue here since a node ready event will retrigger this naturally

berkayoz (approved):
LGTM, we can merge after the last set of minor comments.

src/k8s/pkg/k8sd/controllers/dnsrebalancer/reconcile.go
louiseschmidtgen: I thought I should keep it in here in case the user turns dns on/off. I'll do as you've suggested.

PR #2110, opened 2025-12-03
fix: Go component update doesn't run in pre-release and has issues

omar-selo:
## Description

- Fix go update component workflow to include fips track
- Modify go update component workflow to change version in go.mod too
- Update go component in pre-release branches too

## Backport

Probably to all versions.

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed


build-scripts/k8s_release.py
```diff
@@ -10,6 +10,7 @@
 
 import requests
 from packaging.version import Version
+from hack.update_component_versions import update_go_version
```
bschimke95: I feel like this dependency can become very brittle.
People might change the `update_component.py` script and don't realize the dependency.

I'd propose to:
* move the k8s_release.py into the `hack` folder to have all Python modules in one folder
* move the update-go-version into a separate utils file
* import the function from this util into the two scripts.

build-scripts/k8s_release.py
```diff
@@ -194,6 +195,8 @@ def _update_prerelease_k8s_component(project_basedir: str, k8s_version: str):
     with open(k8s_component_path, "w") as f:
         f.write(k8s_version)
 
+    update_go_version(False)
```
bschimke95: nit: let's make this a named call
```
update_go_version(dry-run=False)
```
Otherwise it is hard to reason about this boolean without looking into the function signature.

build-scripts/hack/update_component_versions.py
```diff
@@ -196,21 +202,39 @@ def update_go_version(dry_run: bool):
...
 
+def _update_go_version_in_go_mod(go_version: str, dry_run: bool):
+    go_mod = GO_MOD.read_text()
+    if f"go {go_version}" in go_mod:
+        LOG.info("go.mod already contains go version %s", go_version)
+        return
+
+    LOG.info("Update go version to %s in %s", go_version, GO_MOD)
+    if not dry_run:
+        updated = re.sub(r"go \d+\.\d+\.\d+", f"go {go_version}", go_mod)
```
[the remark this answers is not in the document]
bschimke95: Good point

build-scripts/hack/update_component_versions.py
```diff
@@ -196,21 +202,39 @@ def update_go_version(dry_run: bool):
...
+    go_snap = f"go/{'.'.join(go_version.split('.')[:2])}-fips/stable"
     snapcraft_yaml = SNAPCRAFT.read_text()
     if f"- {go_snap}" in snapcraft_yaml:
         LOG.info("snapcraft.yaml already contains go version %s", go_snap)
         return
 
     LOG.info("Update go snap version to %s in %s", go_snap, SNAPCRAFT)
     if not dry_run:
-        updated = re.sub(r"- go/\d+\.\d+/stable", f"- {go_snap}", snapcraft_yaml)
+        updated = re.sub(r"- go/\d+\.\d+-fips/stable", f"- {go_snap}", snapcraft_yaml)
```
bschimke95: We don't use the fips track e.g. in 1.32. Let's make this optional.

build-scripts/k8s_release.py
omar-selo: I thought about this, and there is actually already a [util.py](https://github.com/canonical/k8s-snap/blob/main/build-scripts/hack/util.py) file in `hack` folder. But looking at the contents of that file, it didn't sound right to place this function there. This function is very much about updating component versions, and in fact it reuses some of the constants defined there. So it makes more sense for it to live in that file.

bschimke95: I understand your point but disagree.
The update_components.py is basically a CLI/script for a special purpose. It is fine to have utlitiy function in there **as long as they are only used in this context**. If we start using that function somewhere else, we need to move it out of the script into a library. Devs will not expect another script to use this function and may silently break it (we don't have a CI test that would catch this).
I'm not a big fan of generic `util.py` files, anyways. Maybe we can meet in the middle and create a new `component_utils.py` or `update_utils.py` file?

omar-selo: Alright, I made the change

build-scripts/k8s_release.py
```diff
@@ -194,6 +195,8 @@ def _update_prerelease_k8s_component(project_basedir: str, k8s_version: str):
     with open(k8s_component_path, "w") as f:
         f.write(k8s_version)
 
+    update_go_version(dry_run=False)
```
bschimke95, line 198: nit: you could introduce a `--dry-run` flag as we have for the [other commands](https://github.com/canonical/k8s-snap/pull/2110/files#diff-53cac1323bdc99ba53bf4af2b56faa8a0d569c041154c782697c9f3c9c2d831eR346) but feel free to not do it.

bschimke95 (approved):
A minor suggestions but LGTM otherwise

PR #2114, opened 2025-12-03
feat: Use vxlan encapsulation for ipv6 only networking

rapour:
## Description

Up until now, Cilium's encapsulation modes have assumed an IPv4 underlay. With Cilium 1.18, encapsulation routing mode now supports an IPv6 underlay. This means the physical network can operate entirely over IPv6, while Cilium continues to encapsulate pod-to-pod traffic using VXLAN or Geneve as needed.

## Changes

We used to automatically switch to direct routing mode whenever the cluster was operating on an IPv6 underlay network. Now we can operate on VXLAN so the switch is removed. 

### Note

This PR is dependent to the changes introduced in #2095. This PR will be rebased once it merges.

### DIFF
--- a/src/k8s/pkg/k8sd/features/cilium/network.go
+++ b/src/k8s/pkg/k8sd/features/cilium/network.go
@@ -119,16 +119,6 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, s state.State, apiserver 
 		}, err
 	}
 
-	defaultCidr, err := utils.FindCIDRForIP(nodeIP)
-	if err != nil {
-		err = fmt.Errorf("failed to find cidr of default interface: %w", err)
-		return types.FeatureStatus{
-			Enabled: false,
-			Version: CiliumAgentImageTag,
-			Message: fmt.Sprintf(NetworkDeployFailedMsgTmpl, err),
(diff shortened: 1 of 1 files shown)


src/k8s/pkg/k8sd/features/cilium/network.go
```diff
@@ -222,13 +212,6 @@ func ApplyNetwork(ctx context.Context, snap snap.Snap, s state.State, apiserver
 		"tunnelPort": config.tunnelPort,
 	}
 
-	// If we are deploying with IPv6 only, we need to set the routing mode to native
-	if ipv4CIDR == "" && ipv6CIDR != "" {
-		values["routingMode"] = "native"
```
berkayoz, line 227: I believe we'll need to set this explicitly to `vxlan` since we have `--reuse-values` configured for the helm client, so that existing clusters can make the switch as well.

rapour: That's quite right, thanks a lot for catching it, I changed the values to reflect the default chart values.

rapour:
@berkayoz I completely agree, maybe going towards having a declarative state could help with this

PR #2130, opened 2025-12-05
fix: coredns should tolerate control-plane:NoSchedule role

addyess:
This pull request adds support for tolerations to the CoreDNS deployment, ensuring that pods can be scheduled on control-plane nodes. It also introduces corresponding test coverage to validate the new toleration configuration.

CoreDNS toleration configuration:

* Added a `tolerations` field to the CoreDNS deployment spec in `coredns.go`, allowing pods to tolerate the `node-role.kubernetes.io/control-plane` taint with the `NoSchedule` effect.

Test coverage for tolerations:

* Updated `coredns_test.go` to validate the presence and correctness of the new tolerations, checking that the toleration matches the expected key, operator, and effect.

### DIFF
--- a/build-scripts/components/containerd/build.sh
+++ b/build-scripts/components/containerd/build.sh
@@ -27,7 +27,7 @@ for bin in containerd; do
 done
 
 # Shims can be built statically as they do not contain any crypto functions
-for bin in ctr containerd containerd-shim-runc-v2; do
+for bin in ctr containerd-shim-runc-v2; do
   export STATIC=1
   export CGO_ENABLED=0
   export GO_BUILDTAGS=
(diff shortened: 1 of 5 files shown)


addyess:
it turns out in the env I was testing, there are 2 nodes, 1 Control-plane and 1 worker.  

The CP is tained with `node-role.kubernetes.io/control-plane:NoExecute` to keep pesky workloads from showing up on precious CP nodes. 

So, the Controllers replicates to 2 Pods
```
coredns-859b649bd6-cjp96
coredns-859b649bd6-cd88b
```

One of these pods lands happily on the worker, and one is unschedulable b/c of the combo of the taint and the topoSpread


In this case, i do believe that the right course of action is to allow CoreDNS to run on control-plane nodes and this fix is still correct

louiseschmidtgen (approved):
Thank you Adam!

PR #2131, opened 2025-12-08
fix: CoreDNS soft topology constraints

louiseschmidtgen:
## Description

This PR changes the hard topology constraint to a soft topology constraint. 
The controller will eventually ensure that the pods are being rescheduled once more than two nodes are ready.
Until then pods should not be un-schedulable but can live on the same node until being balanced.

Updating the e2e test to use the pod-template hash to make sure we are waiting on the "new" balanced pods not "old" ones.

## Testing

The e2e test confirms that we can soften the topology constraint. Manual testing confirms this as well. Pods eventually get re-scheduled when the second node becomes ready.

## Backport

NA

## Checklist

- [x] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/tests/integration/tests/test_dns.py
+++ b/tests/integration/tests/test_dns.py
@@ -89,6 +89,7 @@ def test_dns_ha_rebalancing(instances: List[harness.Instance]):
 
     # Wait for initial cluster to be ready
     util.wait_until_k8s_ready(initial_node, [initial_node])
+    util.wait_for_dns(initial_node)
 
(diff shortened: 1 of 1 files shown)


louiseschmidtgen:
The failures on `test_version_upgrades` are unrelated to this PR and currently under investigation.
I will be merging this PR in favor of removing flakiness and unblocking CI from the `test_dns:test_dns_ha_rebalancing` failures.

PR #2132, opened 2025-12-08
docs: default dns reference page

louiseschmidtgen:
## Description

This PR adds a default dns reference page that describes the feature's behavior.

## Backport

NA

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed


docs/canonicalk8s/snap/explanation/default-dns.md
```diff
@@ -0,0 +1,33 @@
+# Default DNS
```
nhennigan, line 1: The other networking sections do not specify default. To keep things consistent, can we change to dns.md and DNS as the title?

louiseschmidtgen: I think you're referring to the `ingress.md` etc pages. Those explanations are "generic" in nature whereas this is specific for our DNS feature configuration which is why I went with default DNS. Does that make sense or would you still like me to change it?

nhennigan: Good point. Also considering we should be creating one overview page in the future I think it is fine to leave as you have it for now. Thanks Louise

nhennigan (approved):

PR #2156, opened 2025-12-12
test: add containerd sideload img test

louiseschmidtgen:
## Description

This test ensures that containerd works as expected. In particular we want to confirm that the image side-loading functionality works as expected.

## Backport

NA

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/tests/integration/tests/test_containerd.py
+++ b/tests/integration/tests/test_containerd.py
@@ -22,7 +22,7 @@ def test_containerd(instances: List[harness.Instance]):
             "k8s.io",
             "images",
             "pull",
-            "docker.io/library/nginx:latest",
+            "docker.io/library/nginx:1.29",
         ],
         capture_output=True,
         text=True,
@@ -38,7 +38,7 @@ def test_containerd(instances: List[harness.Instance]):
             "image",
             "export",
             "/tmp/nginx-export.tar",
-            "docker.io/library/nginx:latest",
+            "docker.io/library/nginx:1.29",
         ],
         capture_output=True,
         text=True,
@@ -52,7 +52,7 @@ def test_containerd(instances: List[harness.Instance]):
             "k8s.io",
             "image",
             "rm",
-            "docker.io/library/nginx:latest",
+            "docker.io/library/nginx:1.29",
         ],
         capture_output=True,
         text=True,
@@ -78,7 +78,7 @@ def test_containerd(instances: List[harness.Instance]):
         capture_output=True,
         text=True,
     )
-    assert "nginx:latest" in result.stdout, (
+    assert "nginx:1.29" in result.stdout, (
         f"nginx image not found after sideloading\n"
         f"ctr output: {result.stdout}"
         f"ctr error: {result.stderr}"
@@ -91,7 +91,7 @@ def test_containerd(instances: List[harness.Instance]):
             "kubectl",
             "run",
             "nginx-test",
-            "--image=docker.io/library/nginx:latest",
+            "--image=docker.io/library/nginx:1.29",
             "--image-pull-policy=Never",
             "-l",
             "run=my-nginx",


tests/integration/tests/test_containerd.py
```diff
@@ -0,0 +1,66 @@
...
+        [
+            "/snap/k8s/current/bin/ctr",
+            "images",
+            "pull",
+            "docker.io/library/nginx:latest",
+        ],
+        capture_output=True,
+        text=True,
+    )
+    assert result.returncode == 0, "Failed to pull nginx image"
```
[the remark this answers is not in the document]
louiseschmidtgen: Good point

tests/integration/tests/test_containerd.py
```diff
@@ -0,0 +1,88 @@
...
+import logging
+from typing import List
+
+import pytest
+from test_util import config, harness, tags, util
+
+LOG = logging.getLogger(__name__)
+
+
+@pytest.mark.bootstrap_config((config.MANIFESTS_DIR / "bootstrap-all.yaml").read_text())
```
bschimke95: nit: Do you require all features? Otherwise, you can remove this line and the default config will be used for bootstrapping.

tests/integration/tests/test_containerd.py
```diff
@@ -0,0 +1,88 @@
...
+@pytest.mark.tags(tags.PULL_REQUEST)
```
bschimke95: Is there a specific reason why you want this as part of the pull_request test suite? I feel like nightly could be sufficient but if you got a good reason for it than pull_request would be fine for me as well.

tests/integration/tests/test_containerd.py
```diff
@@ -0,0 +1,88 @@
...
+            "image",
+            "import",
+            "/tmp/nginx-export.tar",
+        ],
+        capture_output=True,
+        text=True,
+    )
+    assert result.returncode == 0, "Failed to import nginx image via sideloading"
+
+    # Verify the image is available after sideloading
```
bschimke95: It would probably make sense to actually have a deployment use the image and check if it comes up.

bschimke95 (commented):
Great work. Added a couple of comments.

tests/integration/tests/test_containerd.py
louiseschmidtgen: If side-loading fails, the image would just get pulled from remote. Therefore, I didn't include the check as it doesn't really prove anything.

bschimke95: You can just set the imagePullPolicy to never to avoid this 

https://spacelift.io/blog/kubernetes-imagepullpolicy#image-pull-policy

tests/integration/tests/test_containerd.py
```diff
@@ -0,0 +1,124 @@
...
+            "--image=docker.io/library/nginx:latest",
+            "--image-pull-policy=Never",
+            "-l",
+            "run=my-nginx",
+            "--port=80",
+        ],
+        capture_output=True,
+        text=True,
+    )
+    assert result.returncode == 0, "Failed to create nginx test pod"
```
bschimke95: nit: This is not required. You can just do this:
```
instance.exec(
        [
            "k8s",
            "kubectl",
            "run",
            "nginx-test",
            "--image=docker.io/library/nginx:latest",
            "--image-pull-policy=Never",
            "-l",
            "run=my-nginx",
            "--port=80",
        ],
    )
```
`exec` sets "check=True" by default, which will raise if the returncode is != 0

bschimke95: Same for the invocations above. You don't need to set capture_output and text if you don't use the output.

bschimke95 (approved):
A small cleanup, then it is good to go.
I'll approve it so that you can merge it right after applying the change.

PR #2269, opened 2026-01-14
build: Update containerd to v1.7.30

bschimke95:
This fixes CVE-2024-25621 (https://avd.aquasec.com/nvd/cve-2024-25621)

PR #2270, opened 2026-01-14
build: Update containerd to v1.7.30

bschimke95:
This fixes CVE-2024-25621 (https://avd.aquasec.com/nvd/cve-2024-25621)


src/k8s/go.mod
```diff
@@ -17,12 +17,13 @@ require (
...
-	golang.org/x/net v0.46.0
-	golang.org/x/sync v0.17.0
-	golang.org/x/sys v0.37.0
+	golang.org/x/net v0.47.0
+	golang.org/x/sync v0.18.0
+	golang.org/x/sys v0.38.0
 	gopkg.in/yaml.v2 v2.4.0
 	helm.sh/helm/v3 v3.18.6
 	k8s.io/api v0.34.0
+	k8s.io/apiextensions-apiserver v0.33.3
```
louiseschmidtgen, line 26: I see for 1.33 we're on `k8s.io/apiextensions-apiserver v0.34.0`?

bschimke95: That is the same scenario as in https://github.com/canonical/k8s-snap/pull/2269#discussion_r2689863557
Maybe one of the dependencies stripped out the k8s.io/apiextensions-apiserver` from the dependencies and now they only need the lower version?

I only ran `go get github.com/containerd/containerd@v1.7.30` and `go mod tidy` o this branch.

PR #2276, opened 2026-01-14
ci: Move linkcheck logic out for weekly tests

nhennigan:
## Description

We have changed linkcheck to only run on the changed files in the individual pr. This means that there could be files that have out of date links that we are not aware of until someone updates the page.

## Solution

This PR extracts the linkcheck logic into its own linkcheck.yaml so it can be reusable from week-test.yaml and docs-check.yaml. There will be a follow on PR that sets up weekly linkcheck tests on all our versions. 

## Issue

Issue I have noticed internally

## Backport

1.32, 1.33, 1.34, 1.35

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed

### DIFF
--- a/.github/workflows/linkcheck.yaml
+++ b/.github/workflows/linkcheck.yaml
@@ -59,10 +59,13 @@ jobs:
         uses: actions/setup-python@v5
       - name: Run linkcheck 
         run: |
-          echo "Changed files: ${{ inputs.changed_files }}"
+          CHANGED_FILES="${{ inputs.changed_files }}"
+          echo "Running linkcheck with files: $CHANGED_FILES"
+
+          # echo "Changed files: ${{ inputs.changed_files }}"
           python3 docs/canonicalk8s/.sphinx/linkcheck.py docs/canonicalk8s \
             --install_target install \
             --linkcheck_target linkcheck \
             --makefile Makefile \
-            "${{ inputs.changed_files }}"
+            "$CHANGED_FILES"
         shell: bash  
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-checks.yml
```diff
@@ -10,21 +10,17 @@ permissions:
...
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
-
       - name: Get changed files list with git diff
```
louiseschmidtgen, line 25: A nice to have would be to move this into an action. (I think we use it in multiple places).
If you have time please do it, otherwise a follow-up would be great.

.github/workflows/weekly-test.yaml
```diff
@@ -70,14 +70,30 @@ jobs:
...
+      ref: ${{ matrix.ref }}
 
   Mattermost:
     name: Notify Mattermost
     # Notify on success or failure but only if the event is a scheduled run.
     # We don't want to ping people of failed PRs.
     if: ${{ always() && github.event_name == 'schedule' }}
     # Note: Update results check below if you change the "needs" list.
-    needs: [test-on-fips]
+    needs: [test-on-fips,docs]
```
louiseschmidtgen:
```suggestion
    needs: [test-on-fips, docs]
```

.github/workflows/docs-checks.yml
nhennigan: I will create a follow up card. Thanks Louise

PR #167, opened 2024-02-28
Move k8s-dqlite out of tree

neoaggelos:
### Summary

Move `k8s-dqlite` out of tree

### Changes

- Build `dqlite` libraries once and use them when building `k8s`, `k8sd` and `k8s-dqlite`
- Remove `k8s-dqlite` code from the repository
- Adjust CI

PR #2296, opened 2026-01-19
build: Move k8sd out of the k8s-snap

bschimke95:
This PR moves the k8sd code into its own [repository](https://github.com/canonical/k8sd).
We do this for the following reasons:
* There is no inherent requirement for k8sd to share the release lifecycle of k8s-snap. A new snap release often occurs without changes to the daemon logic. Separating them ensures that a new k8s-snap release only implies a new k8sd version if there are actual updates to the daemon.
* We utilize a defined build-component workflow for external components included in the snap (e.g., containerd, etcd). Currently, k8sd deviates from this standard. Moving it to its own repository allows us to treat it like any other component.

See internal spec KU-222 for more details.

### DIFF
--- a/ci/cmds/docs.py
+++ b/ci/cmds/docs.py
@@ -15,6 +15,11 @@ import sys
 import tempfile
 from pathlib import Path
 
+# Constants
+K8S_SNAP_ROOT = Path(__file__).parent.parent.parent
+K8SD_VERSION_FILE = K8S_SNAP_ROOT / "build-scripts/components/k8sd/version"
+DOCS_OUTPUT_PATH = K8S_SNAP_ROOT / "docs/canonicalk8s/_parts"
+
 
 def add_docs_cmds(parser: argparse.ArgumentParser) -> None:
     """
@@ -23,9 +28,7 @@ def add_docs_cmds(parser: argparse.ArgumentParser) -> None:
     Args:
         parser: The parent argparse.ArgumentParser to which subcommands will be added.
     """
-    docs_parser = parser.add_parser(
-        "docs", help="Documentation-related commands."
-    )
+    docs_parser = parser.add_parser("docs", help="Documentation-related commands.")
     docs_sub = docs_parser.add_subparsers(
         dest="docs_command", required=True, title="docs commands"
     )
@@ -37,8 +40,13 @@ def add_docs_cmds(parser: argparse.ArgumentParser) -> None:
     )
     p.add_argument(
         "--k8sd-version-file",
-        default="build-scripts/components/k8sd/version",
-        help="Path to the k8sd version file (default: build-scripts/components/k8sd/version)",
+        default=K8SD_VERSION_FILE,
+        help=f"Path to the k8sd version file (default: {K8SD_VERSION_FILE})",
+    )
+    p.add_argument(
+        "--docs-output-path",
+        default=DOCS_OUTPUT_PATH,
+        help=f"Path for DOCS_OUTPUT_PATH environment variable (default: {DOCS_OUTPUT_PATH})",
     )
     p.set_defaults(func=cmd_update_k8sd_api)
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-checks.yml
```diff
@@ -67,5 +68,24 @@ jobs:
         with:
           working-directory: docs/canonicalk8s
 
+  k8sd-api-docs:
```
bschimke95, line 71: Added the k8sd API check here (was in `go.yaml` before) and moved it into a `k8s-ci` command

build-scripts/components/k8sd/version
```diff
@@ -0,0 +1 @@
+main
```
bschimke95, line 1: We'll keep k8sd on main during this transition period to catch issues early on.
This will be changed to a k8sd tag before the 1.36 latest (probably way earlier).

build-scripts/components/k8sd/build.sh
HomayoonAlimohammadi, line 1: QQ: Deprecating dqlite will happen in a future work, right?

snap/snapcraft.yaml
```diff
@@ -60,19 +60,21 @@ parts:
...
       if [ -d "${DQLITE_STAGING_DIR}" ]; then
         rm -rf "${DQLITE_STAGING_DIR}"
       fi
       mkdir "${DQLITE_STAGING_DIR}"
-      cp -rv * "${DQLITE_STAGING_DIR}"
+      cp -rv ./hack/* "${DQLITE_STAGING_DIR}"
 
       # Build dqlite libraries on the staging area
-      "${DQLITE_STAGING_DIR}/static-dqlite.sh"
+      "${DQLITE_STAGING_DIR}/dynamic-dqlite.sh"
```
berkayoz, line 77: Why don't we build dqlite based on the upstream repository instructions, this always felt a bit weird and is weirder now. @louiseschmidtgen wdyt?

berkayoz (approved):
LGTM overall, one comment we can handle in a separate PR as well.

snap/snapcraft.yaml
louiseschmidtgen: Yes, this is weird. 
Looks like we used to build it normally until Angelos changed it in this PR: https://github.com/canonical/k8s-snap/pull/167/changes. 
So the idea seemed to be to build it once and use it for "k8s, k8sd and k8s-dqlite".

I think we should fully move dqlite out of the k8s-snap into k8sd.
We won't need Dqlite for k8s-dqlite anymore and we don't use it anywhere else but in microcluster.

bschimke95: Yes, that is correct. The idea was/is that you otherwise need to build dqlite twice. I did not remove the part in this PR because it would mean we need to touch the k8s-dqlite build as well (and the PR is already large enough).

Instead, I removed it as part of the k8s-dqlite removal, see [here](https://github.com/canonical/k8s-snap/pull/2314/changes#diff-56759910381a014fecfd7556dd72ddd68c747d922a5b7df2044b9ce7c552f5f5L60)

build-scripts/components/k8sd/build.sh
bschimke95: If you mean `k8s-dqlite` then yes, I'll do it [here](https://github.com/canonical/k8s-snap/pull/2314). (Still in draft until this PR is merged.

Dqlite itself is required for microcluster, so we cannot remove it from the build

snap/snapcraft.yaml
louiseschmidtgen: I wouldn't be opposed to building it twice for now until we remove the second build with k8s-dqlite next week.
But it's up to you - I can live with both.

build-scripts/components/k8sd/build.sh
HomayoonAlimohammadi: Ah, my bad! Yea I meant k8s-dqlite. Thanks!

louiseschmidtgen (approved):
lgtm, as discussed async we will move Dqlite out of the k8s-snap fully with the k8s-dqlite removal.

snap/snapcraft.yaml
bschimke95: dqlite will not be built twice. The k8sd step checks if dqlite is already built and skip the internal build step if so.

PR #2323, opened 2026-01-27
docs: clarified cos-lite integration page

ethandcosta:
Added note for where to install yq from and that the COS setup tutorial is only internally accessible.

### DIFF
--- a/docs/canonicalk8s/.sphinx/.wordlist.txt
+++ b/docs/canonicalk8s/.sphinx/.wordlist.txt
@@ -333,3 +333,4 @@ snapcrafting
 subcluster
 swrast
 zSystems
+yq


docs/canonicalk8s/charm/howto/cos-lite.md
```diff
@@ -130,7 +134,10 @@ url: http://10.246.154.87/cos-lite-grafana
 The username for this credential is `admin`.
 
 You’ve successfully gained access to a comprehensive observability stack. Visit
-the URL and use the credentials to log in.
+the URL and use the credentials to log in. This Grafana endpoint is only 
+accessible from within the cluster. To get external access, you will need
+ to deploy a [load balancer][load-balancer] so that the network connection
```
nhennigan, line 139: We don't have a how-to guide here yet for enabling a loadbalancer on the charm like we do for the snap. Is it just a matter of juju config k8s load-balancer=true or something along those lines? Do we expect a juju user to know how to expose an endpoint once they have been made aware it is not accessible outside the cluster? We already have a note about it being important so I just wonder what level of guidance do you think our users need

ethandcosta: To my understanding there isn't one unified approach because of the different cloud substrates you can have, but I think it's generally straightforward. For public clouds, all you have to do is deploy the integrator charm and integrate it with k8s.

For bare metal deployments it's a bit more complicated, but I think this is well captured in the [source I linked to](https://ubuntu.com/kubernetes/charmed-k8s/docs/metallb).  

And then to connect it to traefik, you just set juju config traefik external_hostname="monitoring-site.com" and I believe that's it

nhennigan: Ok so it seems like a step that a general juju and k8s user would know how to figure out. Thanks for clarifying!

PR #2405, opened 2026-02-16
fix: (release-1.32) use self-hosted runners for `20.04`

mateoflorido:
## Description

The latest version of the Linux kernel in `24.04` creates some issues with LXD containers.

## Solution

Our self-hosted runners have moved to the previous version of the kernel in order to mitigate this impact. We are moving the affected instances (i.e., those using `20.04`) for now.

## Backport

No

## Checklist

- [x] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -92,7 +92,7 @@ jobs:
   run-tests:
     name: ${{ matrix.test }}
     needs: prepare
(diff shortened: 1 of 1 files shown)


berkayoz (approved):
I'm generally okay with the change, however I wonder if we'll overwhelm self hosted runners with our parallelization of 1 test per runner.

PR #2438, opened 2026-02-27
docs: added hw requirements page

ethandcosta:
This PR adds a consolidated reference page in the docs to see the minimum hardware requirements for all of our Kubernetes products.

### DIFF
--- a/docs/canonicalk8s/snap/reference/hw-requirements.md
+++ b/docs/canonicalk8s/snap/reference/hw-requirements.md
@@ -1,10 +1,13 @@
 # Hardware Requirements
 
-The following are the suggested minimum hardware requirements for running Canonical's 
-various Kubernetes products. Note that running with these resources will only allow
-for basic usage, and extra resource allotment is expected for complex use cases.
+The following are the suggested minimum hardware requirements 
+for running Canonical's various Kubernetes products. Note 
+that running with these resources will only allow for 
+basic usage, and extra resource allotment is expected for 
+complex use cases.
 
 ## Canonical Kubernetes (Snap)
+
 | Component | Specification | Notes |
 | --- | --- | --- |
 | CPU | 2 core | amd64/arm64 only |
--- a/docs/canonicalk8s/snap/reference/hw-requirements.md
+++ b/docs/canonicalk8s/snap/reference/hw-requirements.md
@@ -1,0 +1,36 @@
+# Hardware Requirements
+
+The following are the suggested minimum hardware requirements for running Canonical's 
+various Kubernetes products. Note that running with these resources will only allow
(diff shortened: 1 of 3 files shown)


nhennigan (commented):
Thanks Ethan. I think we can remove the Microk8s and Charmed Kubernetes sections as they are not part of this docset. Other than that great job!

docs/canonicalk8s/snap/reference/hw-requirements.md
```diff
@@ -0,0 +1,23 @@
...
+basic usage, and extra resource allotment is expected for 
+complex use cases.
+
+## Canonical Kubernetes (Snap)
+
+| Component | Specification | Notes |
+| --- | --- | --- |
+| CPU | 2 cores | amd64/arm64 only |
+| RAM | 8 GiB | |
+| Root Disk | 20 GiB of free space | SSD |
```
louiseschmidtgen, line 15: I'm a bit sceptical of these values. If we look at [etcd's minimum requirements](https://etcd.io/docs/v3.6/op-guide/hardware/) that already puts us at 2 CPU, 8GB Memory for a "small" cluster. That's just etcd- no other components in the k8s-snap.

I would propose 4 cores 16 GiB 40GiB for the snap installation. (Let's also hear opinions from the team).

docs/canonicalk8s/snap/reference/hw-requirements.md
```diff
@@ -0,0 +1,23 @@
+# Hardware requirements
+
+The following are the suggested minimum hardware requirements 
+for running Canonical's various Kubernetes products. Note
```
louiseschmidtgen, line 4:
```suggestion
for running a "small" Kubernetes cluster. Note 
```

louiseschmidtgen (commented):
Let's discuss these values with the rest of the team:

docs/canonicalk8s/snap/reference/hw-requirements.md
bschimke95: I think "minimum" hardware requirements already imply that, doesn't it?

docs/canonicalk8s/snap/reference/hw-requirements.md
bschimke95: We could add minimum and production requirements.
You can totally run the k8s-snap on 2core/8GB machines ([kubeadm](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#before-you-begin) even only requires 2GB RAM). The etcd requirements that you've linked are not the minimum requirements but recommendations. If you use less RAM then performance might suffer but etcd will still be able to serve requests.

PR #2522, opened 2026-04-09

bschimke95:

### DIFF
--- a/docs/canonicalk8s/about.md
+++ b/docs/canonicalk8s/about.md
@@ -1,3 +1,9 @@
+---
+myst:
+  html_meta:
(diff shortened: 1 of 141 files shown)


build-scripts/components/cni/version
```diff
@@ -1 +1 @@
-v1.8.0
+v1.9.1
```
HomayoonAlimohammadi: Just making sure that this is not breaking anything

HomayoonAlimohammadi (approved):

build-scripts/components/cni/version
bschimke95: Discussed offline: The CNI version follows upstream. If upstream bumps it then we should, too.

Issue #2448, opened 2026-03-06
x509 client certificates authentication is not working


pamudithaA:
### Summary

x509 client certificates based authentication is not working with Canonical K8s snap. 
If we create a `CertificateSigningRequest` and approve using `kubectl certificate approve`, it will be signed by `kubernetes-ca` certificate. 
However, k8s-apiserver is validating the certificates against `kubernetes-ca-client` CA certificate. 

Running kubectl commands failed with below error. 

```
E0305 13:43:23.006350   12337 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
E0305 13:43:23.027180   12337 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
E0305 13:43:23.046835   12337 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
E0305 13:43:23.069306   12337 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
E0305 13:43:23.088084   12337 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials"
error: You must be logged in to the server (the server has asked for the client to provide credentials)

````

According to the kube-apiserver configuration, `client-ca-file` is set to `/etc/kubernetes/pki/client-ca.crt`, eventhough CSR is signed using `/etc/kubernetes/pki/ca.crt`. 

Comparison with admin.conf is below to see the different issuer.  

Custom user certificate
```
openssl x509 -in john.crt -noout -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            32:72:41:ad:a8:a1:4b:a2:c6:50:59:99:70:9e:f4:d1
        Signature Algorithm: sha256WithRSAEncryption
       *Issuer: CN = kubernetes-ca
```

Admin user certificate
```
openssl x509 -in api-unen.crt -noout -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            b5:27:7e:0b:8c:06:1f:57:86:fb:ba:0f:37:50:77:ed
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN = kubernetes-ca-client
```

### What Should Happen Instead?

Admin user should be able to approve the csr and custom client certificates should be accepted as a authentication method. 

### Reproduction Steps

[...]

PR #2581, opened 2026-06-03
fix: x509 client certificates based authentication is not working with Canonical K8s snap

MoeSalah1999:
## Description

Fixes x509 client-certificate authentication for certificates issued through the Kubernetes CSR flow in the Canonical K8s snap.

Previously, CSRs approved with kubectl certificate approve could be signed by the wrong CA for API server client authentication. The API server validates client certificates against /etc/kubernetes/pki/client-ca.crt, so CSR-issued client certificates must be signed by the corresponding client CA, not the general Kubernetes CA.

## Solution

1. Updated the kube-controller-manager service wrapper to pass the client CA as the default cluster signing CA:

- --cluster-signing-cert-file=/etc/kubernetes/pki/client-ca.crt
- --cluster-signing-key-file=/etc/kubernetes/pki/client-ca.key

2. Added guard logic so this override is only applied when:

- the controller-manager args file does not exist,
- the args file does not define cluster signing cert/key flags, or
- the args file still points cluster signing to the general Kubernetes CA:
    - /etc/kubernetes/pki/ca.crt
    - /etc/kubernetes/pki/ca.key

3. Preserved user/custom configuration by not overriding cluster signing flags when they are already explicitly configured to another value.

4. Added an integration regression test that:
- creates a private key and CSR for a test client,
- submits a Kubernetes CertificateSigningRequest using signer kubernetes.io/kube-apiserver-client,
- approves the CSR with kubectl certificate approve,
- waits for .status.certificate because CSR signing is asynchronous,
- decodes the issued certificate,
- verifies it chains to /etc/kubernetes/pki/client-ca.crt,
- builds a kubeconfig using the issued client certificate,
- confirms the issued certificate can authenticate to the API server.

5. Updated certificate documentation to clarify that:
- client-ca is the CA used for client certificates,
- its generated certificate common name is kubernetes-ca-client,
- approved Kubernetes CSRs are signed with this client CA so the resulting certificates are trusted by the API server’s client-ca.crt bundle.

Why this approach?
Instead of patching Kubernetes source defaults, this fixes the snap’s runtime configuration directly through the existing service wrapper mechanism. This avoids carrying fragile Kubernetes source patches and keeps the behavior aligned with the snap’s generated controller-manager args.

The wrapper also avoids overwriting explicit custom cluster-signing settings, so users can still configure a different signing CA if needed.

## Issue

Issue https://github.com/canonical/k8s-snap/issues/2448

## Checklist

- [ ] Covered by unit tests
- [x] Documentation updated
- [x] CLA signed
- [ ] Confirm whether the `release-notes` label should be kept or removed

### DIFF
--- a/docs/canonicalk8s/snap/reference/certificates.md
+++ b/docs/canonicalk8s/snap/reference/certificates.md
@@ -15,6 +15,11 @@ and locations on the disk.
 | `kubernetes-front-proxy-ca` | CA for front-end proxy     | `/etc/kubernetes/pki/front-proxy-ca.crt` | Signing certificates for the front-proxy    |
 | `client-ca`                 | CA for client certificates | `/etc/kubernetes/pki/client-ca.crt`      | Signing certificates for the client         |
 
+The `client-ca` certificate is generated with the common name
+`kubernetes-ca-client`. CertificateSigningRequests approved through
+`kubectl certificate approve` are signed with this client CA, so the resulting
+client certificates are trusted by the API server's `client-ca.crt` bundle.
+
 ## Certificates
 
 This table provides an overview of the certificates currently in use,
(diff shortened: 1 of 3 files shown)


MoeSalah1999:
@bschimke95 Could you please verify this?

PR #2591, opened 2026-06-11
fix: Use local microcluster client for node removal

claudiubelu:
## Description

In microcluster's `clusterMemberDelete`[1], when a non-leader receives a request to remove itself, it locks `clusterDisableMu` before forwarding the request to the leader [2]. That lock is only released after the forwarded request returns and the response has been flushed back to the caller. The lock prevents the node from re-execing its daemon (via `unix.Exec` in `resetClusterMember`) until the caller has received a response.

At the moment, `removeNodeFromMicrocluster` is calling `s.Leader()` to obtain a client, which connects directly to the dqlite leader's network endpoint. By going directly to the leader, k8sd bypasses the local microcluster handler mentioned above entirely. Because of this, `clusterDisableMu` is never locked.

When the leader subsequently calls `ResetClusterMember` on the node being removed, the re-exec fires as soon as the leader's PUT request is done, before the leader's DELETE response has been returned to k8sd. The entire k8sd process is replaced by `unix.Exec` while the outbound HTTP connection to the leader is still open, so `k8sd` never receives the response. The original caller (e.g.: CAPI) gets a connection reset instead of a success response.

[1] https://github.com/canonical/microcluster/blob/936197c307c8ebbc82f151f33bd0e1e883abe9be/internal/rest/resources/cluster.go#L414
[2] https://github.com/canonical/microcluster/blob/936197c307c8ebbc82f151f33bd0e1e883abe9be/internal/rest/resources/cluster.go#L454-L462
[3] https://github.com/canonical/k8sd/pull/9

## Solution

By using the local client to connect to microcluster, the issue is resolved.

This issue has been incidentally resolved in k8sd [3], which is why this commit is not a backport.

## Issue

Fixes: https://github.com/canonical/k8s-snap/pull/2591

## Backport

This issue can be seen in release-1.35, release-1.34, release-1.33, release-1.32. It can be backported to all of them.

## Checklist

- [ ] Covered by unit tests
- [ ] Covered by integration tests - existing integration tests should cover this
- [x] CLA signed
- [ ] Backport label added if necessary - Cannot add labels.
- [ ] Confirm whether the `release-notes` label should be kept or removed

### DIFF
--- a/src/k8s/pkg/k8sd/api/cluster_remove.go
+++ b/src/k8s/pkg/k8sd/api/cluster_remove.go
@@ -20,6 +20,7 @@ import (
 	"github.com/canonical/k8s/pkg/utils/node"
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/microcluster/v2/cluster"
+	"github.com/canonical/microcluster/v2/microcluster"
 	"github.com/canonical/microcluster/v2/state"
 )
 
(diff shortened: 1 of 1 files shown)


claudiubelu:
@bschimke95 Hello, PTAL. Also, not sure why the the `Lint and integration` job fails at code checkout.

PR #2597, opened 2026-06-12
ci: pin snapcraft to 8.x for core20 build

bschimke95:
## Problem

The snap build on `release-1.33` fails because the workflow installs `snapcraft --classic` (now **9.0.0**), and snapcraft 9 dropped support for the `core20` base this branch uses:

```
Recommended resolution: Use Snapcraft 8 from the '8.x' channel of snapcraft where 'core20' was last supported.
```

This blocks CI on every PR targeting `release-1.33` (e.g. backport #2594).

## Fix

Pin the snapcraft install to the `8.x` channel, the last series supporting `core20`.

Only `release-1.33` is affected — `release-1.34`/`1.35`/`main` use `core22`, which snapcraft 9 still supports.

### DIFF
--- a/.github/workflows/build-snap.yaml
+++ b/.github/workflows/build-snap.yaml
@@ -33,7 +33,8 @@ jobs:
           channel: '5.21/stable'
       - name: Install snapcraft
         run: |
-          sudo snap install snapcraft --classic
+          # Pin to 8.x: snapcraft 9 dropped core20 support, which this snap uses.
+          sudo snap install snapcraft --classic --channel=8.x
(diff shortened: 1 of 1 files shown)


bschimke95:
Folding this snapcraft 8.x pin into the backport PR #2594 instead. Standalone, this PR can't merge: any PR targeting a release branch runs the full `up_to_weekly` suite, including nightly tests that call the buggy `previous_track` — which only #2594 (the #2539 backport) fixes. #2594 now carries both the test fix and this build fix.

PR #2598, opened 2026-06-15
docs: update containerd conflict documentation

ethandcosta:
## Description

The current documentation on resolving containerd conflicts with k8s-snap and other providers is vague and needs to be updated in lieu of changes from https://github.com/canonical/k8sd/pull/51

## Solution

This PR updates the documentation to clarify that users should seek to avoid installing containerd to a tmpfs and that they can use a CLI flag to provide, in addition to the existing yaml method.


ethandcosta:
CI will pass when the associated change in k8sd is merged in

docs/canonicalk8s/snap/howto/install/dev-env.md
```diff
@@ -55,15 +58,38 @@ To check the available space on the tmpfs:
...
+that can also cause a failed bootstrap.
+
+### External consumption
+
+When changing the containerd install path, make sure that the configurations of
+external consumers of {{product}} such as operators are also updated. For example,
+in the GPU operator, you will have to update the Helm chart to include the new
+containerd paths.
+
+```bash
```
nhennigan:
```suggestion
```
```
According to our style guide, we try not to include the language in our code blocks unless it is crucial to the output. E.g With yaml, formatting is very important as well as having the key:value pairs aligned. But bash code blocks only introduces colour which can be distracting.

File: docs/canonicalk8s/snap/howto/install/dev-env.md [hunk 2]
```diff
@@ -55,15 +58,38 @@ To check the available space on the tmpfs:
...
+helm install gpu-operator nvidia/gpu-operator -n gpu-operator \
```
nhennigan, line 121: We do not own / maintain the gpu-operator so I would be afraid of this command going out of date. Setting the containerd value is a valuable point and I definitely think we should include it but maybe we can make it more generic?

nhennigan (commented):
Thanks Ethan! I left a few minor comments. 

While we are editing this page can we add a metadata tag?

docs/canonicalk8s/snap/howto/install/dev-env.md
ethandcosta: The trouble with that is that not all operators need this fix, so it's only something that needs to be done on a case-by-case basis and each of the operators is going to have different syntax for doing so, so I don't know how to generalize this. At the same time, I don't think it can be omitted because the gpu operator is one of the most popular add-ons used with k8s and they are all going to run into the same issue

docs/canonicalk8s/snap/howto/install/dev-env.md
```diff
@@ -16,25 +21,16 @@ virtual machine.
...
-upstream projects and operators (e.g.: GPU Operator).
-
-If you already have Docker installed, or another Kubernetes instance that uses
-containerd directly installed on the host, this can cause various conflicts
-with {{product}}.
-
-But, if necessary, {{product}} can be configured to use a custom containerd
-path, like so:
-
-<!-- SPREAD
```
nhennigan, line 35: Please add this back in. It is needed for our spread tests

docs/canonicalk8s/snap/howto/install/dev-env.md
```diff
@@ -49,21 +45,24 @@ cluster-config:
...
-(e.g.: `/ck8s`). {{product}} will then use this base directory for the
-containerd-related files (e.g.: `/ck8s/etc/containerd`,
-`/ck8s/var/run/containerd/containerd.sock`, etc.).
+By doing this, all containerd files will be stored under the parent
+directory specified by `containerd-base-dir`. For example, if
+`containerd-base-dir` is set to `/ck8s`, containerd files will be stored at
+paths such as `/ck8s/etc/containerd` and
+`/ck8s/run/containerd/containerd.sock`.
 
-<!-- SPREAD
```
nhennigan, line 57: Please add this back in also

docs/canonicalk8s/snap/howto/install/dev-env.md
```diff
@@ -76,17 +75,38 @@ To check the available space on the tmpfs:
...
 
-If the space is low and you're experiencing these issues, you can temporarily 
+If the space is low and you're experiencing these issues, you can temporarily
 increase the size of the tmpfs mount to see if it resolves the problem:
 
 ```
 sudo mount -o remount,size=10G /run
 ```
 
-<!-- SPREAD SKIP END -->
```
nhennigan, line 86: Adding this back in will make the tests pass. The missing END comment is what is failing the prepare step

docs/canonicalk8s/snap/howto/install/dev-env.md
```diff
@@ -16,25 +21,16 @@ virtual machine.
...
-
-If you already have Docker installed, or another Kubernetes instance that uses
-containerd directly installed on the host, this can cause various conflicts
-with {{product}}.
-
-But, if necessary, {{product}} can be configured to use a custom containerd
-path, like so:
-
-<!-- SPREAD
-export containerdBaseDir=/ck8s/etc/containerd
```
nhennigan, line 36: Based on your updated content, should the example for our spread testing be set to /ck8s/ ?

docs/canonicalk8s/snap/howto/install/dev-env.md
```diff
@@ -76,17 +75,38 @@ To check the available space on the tmpfs:
...
+When changing the containerd install path, make sure that the configurations of
+external consumers of {{product}} such as operators are also updated. For example,
+in the GPU operator, you will have to update the Helm chart to include the new
+containerd paths.
+
+```
+helm install gpu-operator nvidia/gpu-operator -n gpu-operator \
+ --set operator.defaultRuntime=containerd \
+ --set toolkit.env[0].name=CONTAINERD_CONFIG \
+ --set toolkit.env[0].value={custom_containerd_dir}/etc/containerd/config.toml \
```
nhennigan: Do we want to add this to our spread testing? As is, with no spread skip marker, this will be picked up in our testing as a testable code block. 

We should make the var consistent - either custom_containerd_dir or containerdBaseDir regardless if we want testing or not.

If we do want to test and they are the same, it will be set earlier in the spread comments and then this command should work. We should then add a command to ensure the operator comes up correctly.

nhennigan (commented):
Thanks Ethan. With the spread docs testing merged, there are a few things to make this pass the tests. I've left comments. Let me know if I need to expand more

docs/canonicalk8s/snap/howto/install/dev-env.md
ethandcosta: Fixed the variable naming so it's more consistent. In the future, I think it would be nice to be able to test this (and maybe after the separate effort I have to enable GPU testing, this may be possible), but as of right now our SHR don't have GPUs, so it's impossible to test this. I fixed the code block so it's in a skip block.

nhennigan (approved):
Thanks Ethan! These updates are great. Nice catch on the spread tests suite.

PR #2603, opened 2026-06-17
feat: add integration testing for containerd CLI flag

ethandcosta:
In conjunction with the [added containerd-base-dir CLI flag added in k8sd](github.com/canonical/k8sd/pull/51), this PR adds an e2e test to verify the CLI flag works as expected, changing the install path of containerd files. We also need to change a test that used the old path linting for containerd-base-dir.


bschimke95 (approved):
LGTM, why do you want to backport this test?

ethandcosta:
Ah, I forgot that the k8sd migration was only in this version, removing the backports

PR #2642, opened 2026-07-08
docs: Add scenario based testing to our docs spread tests

nhennigan:
## Description

This PR extends our docs automated testing to allow us to test pages together in one task.yaml spread test. I have introduced a new build_spread_task.py which removes the complexity from the CI. It builds on create_spread_task_file.py and combines pages into one task.yaml if they are specified in the scenarios.yaml file. It also handles standalone files too.

## Solution

- User specifies in the scenarios.yaml file 
  - the name of the scenario
  - which pages to be tested together 
  - which suite  it should be ran in (snap_clean or snap_bootstrapped) 
- At PR time, the CI will take the changed files and make an individual spread test if it is marked with snap_clean or snap_bootstrapped (same functionality as before). Then it will take a second pass and see if any of the altered files are part of a scenario. It will then generate a single task.yaml with all the scenario.yaml files 
  - This PR will create  scenario-cis/task.yaml, snap-howto-security-cis-assessment/task.yaml and snap-howto-security-hardening/task.yaml
- build_spread_task.py now handles suite verification 
- Pages can now be marked with scenario_only. This will tell build_spread_task.py to skip this file during standalone task creation but still include it in the scenarios second pass. 
- The build_spread_task.py is setup to work in CI mode or manually. 
  - `--target` is used locally to specify the name of the scenario or the name of the file to create a task.yaml for e.g `python3 build_spread_task.py --target snap/howto/security/hardening.md` or `python3 build_spread_task.py --target cis`
  - `--detect-scenarios` only to be used by the CI to check all changed files if they belong in a scenario and need a task.yaml created 
- When we have multiple trap exit commands (clean up commands), build_spread_task.py will collect them and print them at the top of the scenario task.yaml to ensure proper clean up
- Pages are marked with SCENARIO_MODE=true/false. This allows us to run certain commands based on if we are in the scenario or not. E.g 
```
<!-- SPREAD 
if [ "${SCENARIO_MODE}" = "true" ]; then
  sudo -E kube-bench --version ck8s-cis-1.24 --config-dir ./kube-bench-ck8s-cfg/cfg/ --config ./kube-bench-ck8s-cfg/cfg/config.yaml | grep -A4 "== Summary total ==" | grep "0 checks FAIL"
fi
-->
```

  In our cis assessment, the grep for 0 fails will error if we do not run the hardening steps first. In the regular snap-howto-security-cis-assessment this is skipped. In scenario-cis, it is ran and will only fail on a genuine error. 

## Issue

N/A

## Backport

No

## Checklist

- [ ] Covered by unit tests
- [ ] Documentation updated
- [x] CLA signed
- [x] Confirm whether the `release-notes` label should be kept or removed

### DIFF
--- a/.github/workflows/docs-spread-tests.yaml
+++ b/.github/workflows/docs-spread-tests.yaml
@@ -39,6 +39,9 @@ jobs:
         with:
           python-version: "3.x"
 
+      - name: Install Python dependencies
+        run: pip install pyyaml==6.0.2
+
       - id: generate
         name: Generate tasks.yaml for changed docs
         shell: bash
@@ -138,4 +141,4 @@ jobs:
           sudo adduser --gecos "" --disabled-password ubuntu || true
           echo "ubuntu:ubuntu" | sudo chpasswd || true
           cd docs
-          spread -vv "github-ci:ubuntu-24.04-64:tests/spread_generated/${{ matrix.suite }}/"
+          spread -vv -repeat 0 "github-ci:ubuntu-24.04-64:tests/spread_generated/${{ matrix.suite }}/"
--- a/.github/workflows/docs-spread-tests.yaml
+++ b/.github/workflows/docs-spread-tests.yaml
@@ -11,6 +11,9 @@ on:
 permissions:
   contents: read
 
+env:
+  OPERATOR_WORKFLOWS_REF: df449f1e3d1b8babbe9df48bbebcff1e58c9fda9  # main as of 2026-05-07
+
 jobs:
   prepare-docs-spread:
     runs-on: ubuntu-latest
@@ -27,7 +30,7 @@ jobs:
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           repository: canonical/operator-workflows
-          ref: df449f1e3d1b8babbe9df48bbebcff1e58c9fda9 # main as of 2026-05-07
+          ref: ${{ env.OPERATOR_WORKFLOWS_REF }}
           sparse-checkout: |
             spread/create_spread_task_file.py
           sparse-checkout-cone-mode: false
--- a/.github/workflows/docs-spread-tests.yaml
+++ b/.github/workflows/docs-spread-tests.yaml
@@ -140,5 +140,8 @@ jobs:
           set -euo pipefail
           sudo adduser --gecos "" --disabled-password ubuntu || true
           echo "ubuntu:ubuntu" | sudo chpasswd || true
+          # Ensure all suite directories defined in spread.yaml exist so spread
+          # does not fail its directory scan even when only one suite has tasks.
+          mkdir -p tests/spread_generated/snap_clean tests/spread_generated/snap_bootstrapped
           cd docs
           spread -vv -repeat 0 "github-ci:ubuntu-24.04-64:tests/spread_generated/${{ matrix.suite }}/"
--- a/.github/workflows/docs-spread-tests.yaml
+++ b/.github/workflows/docs-spread-tests.yaml
@@ -48,10 +48,6 @@ jobs:
           set -euo pipefail
 
           rm -rf tests/spread_generated
-          mkdir -p tests/spread_generated
-          
-          # Create all suite directories so spread doesn't fail reading spread.yaml
-          mkdir -p tests/spread_generated/{snap_clean,snap_bootstrapped}
 
           changed_files_input="${CHANGED_FILES_INPUT:-}"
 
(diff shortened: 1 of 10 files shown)


docs/tools/build_spread_task.py
```diff
@@ -0,0 +1,536 @@
...
+Scenario mode  (--target matches a name in scenarios.yaml):
+    Chains multiple documentation pages into one task.yaml.
+    Injects ``set -e`` and ``export SCENARIO_MODE=true`` at the top so
+    individual pages can detect they are running inside a chain.
+
+Standalone mode  (--target is a path to a .md file):
+    Generates a task.yaml for a single page.
+    Injects ``set -e`` but does NOT set SCENARIO_MODE, so conditional
+    blocks such as ``sudo k8s bootstrap`` execute normally.
+
```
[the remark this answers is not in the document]
nhennigan: Fixed

docs/tools/build_spread_task.py
```diff
@@ -0,0 +1,536 @@
...
+    docs_root = repo_root / "docs" / "canonicalk8s"
+    sections: list[tuple[str, list[str]]] = []
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        for page in pages:
+            doc_file = (docs_root / page).resolve()
+            if not doc_file.is_relative_to(repo_root.resolve()):
+                raise ValueError(
+                    f"Refusing to process {doc_file}: path is outside the repository root"
+                )
```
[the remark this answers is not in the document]
nhennigan: Done

.github/workflows/docs-spread-tests.yaml
```diff
@@ -39,6 +39,9 @@ jobs:
         with:
           python-version: "3.x"
 
+      - name: Install Python dependencies
```
louiseschmidtgen, line 45: Should we have a requirements.txt file?

docs/tools/build_spread_task.py
```diff
@@ -0,0 +1,536 @@
...
+        "Error: PyYAML is required. Install with: pip install pyyaml\n"
+        "       or: sudo apt-get install python3-yaml",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+
+UPSTREAM_SCRIPT_URL = (
+    "https://raw.githubusercontent.com/canonical/operator-workflows"
+    "/df449f1e3d1b8babbe9df48bbebcff1e58c9fda9/spread/create_spread_task_file.py"
+)
```
louiseschmidtgen, line 55: Are we keeping this hardcoded?

louiseschmidtgen (commented):
Overall amazing work here! Great to see this running.
I looked at the CI output and I see that `Suite snap_clean has no generated tasks; skipping.`. Could we please put some examples to validate the snap_clean work please?

docs/tools/build_spread_task.py
nhennigan: Yes for now. Any changes to that script would probably mean manual changes to out build_spread_task.py functionality so I want to keep this pinned for now. Do you see issues with this approach?

.github/workflows/docs-spread-tests.yaml
nhennigan: As there is only one dependency that needs downloading so I decided not to include a requirements.txt file to reduce the number of files needed. I do however understand it is the standard approach for python development. I would include a requirements.txt file in the future if there was more than one dep. Are you happy with this approach or do you think it is better to follow standard practice from day 1?

nhennigan:
Updates:
CIS assessment was changed to scenario_only test suite. This is because it does not make much sense to run the test standalone without hardening as it will always fail. This is also testing the scenario_only functionality
image
A snap_clean scenario was introduced to test all our default networking features together. I added SCENARIO_MODE != true around the disable commands to ensure the cluster still worked correctly in the scenario test
image
All tests ran successfully in CI 
image

.github/workflows/docs-spread-tests.yaml
louiseschmidtgen: We can postpone this to the next PR, that's not a blocking issue for me.

docs/tools/build_spread_task.py
louiseschmidtgen: I think you can also use the workflow's hash to avoid duplication? From here:

```
      - name: Checkout upstream spread script
        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
        with:
          repository: canonical/operator-workflows
          ref: df449f1e3d1b8babbe9df48bbebcff1e58c9fda9 # main as of 2026-05-07
```

louiseschmidtgen (approved):
Really amazing work on this @nhennigan!

docs/tools/build_spread_task.py
nhennigan: Thats a good point. It is best to keep local runs and CI runs on the same version and the duplication could mean this may drift. I have added a function to read the SHA directly from docs_spread_tast.yaml.

PR #2651, opened 2026-07-21
test: fix integration tests broken by kube-proxy → Cilium migration

louiseschmidtgen:
## Problem

After kube-proxy was replaced with Cilium as the default networking solution, 100% of integration tests started failing on both amd64 and arm64 with:

```
AssertionError: Service kube-proxy should be active, but it is inactive
```

## Root cause

`_is_kube_proxy_enabled()` in `test_util/util.py` has two fallback paths that both return `True` when they should return `False`:

1. **`CalledProcessError` handler** — the log message already says *"defaulting to False"* but the code returned `True`
2. **Missing config field** — the docstring already documented this as returning `False`, but the code returned `True`

Since Cilium replaced kube-proxy, the `kube-proxy-enabled` field is absent from the cluster config. Both fallbacks fired, causing `check_snap_services_ready()` to add `kube-proxy` to the expected-active-services set and then assert it is running — which it isn't.

## Changes

- **`util.py`**: Fix both fallback paths in `_is_kube_proxy_enabled` to return `False` (aligning code with existing docstring and log message)
- **`bootstrap-smoke.yaml` / `bootstrap-node-taints.yaml`**: Remove `extra-node-kube-proxy-args` (kube-proxy no longer runs by default)
- **`test_smoke.py`**: Remove the kube-proxy entry from the extra-args file assertion loop

## Testing

This unblocks all integration tests that were failing with `Service kube-proxy should be active, but it is inactive`. Tests that explicitly opted in to kube-proxy testing (with `skip_services=["kube-proxy"]` or by setting `kube-proxy-enabled: true`) are unaffected.

### DIFF
--- a/tests/integration/templates/bootstrap-node-taints.yaml
+++ b/tests/integration/templates/bootstrap-node-taints.yaml
@@ -25,8 +25,6 @@ extra-node-kube-controller-manager-args:
   --leader-elect-retry-period: 3s
 extra-node-kube-scheduler-args:
   --authorization-webhook-cache-authorized-ttl: 11s
-extra-node-kube-proxy-args:
-  --config-sync-period: 14m
 extra-node-kubelet-args:
   --authentication-token-webhook-cache-ttl: 3m
 extra-node-containerd-args:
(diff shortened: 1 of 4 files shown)


HomayoonAlimohammadi (approved):
I wonder why the tests were passing on the original PR. Thanks for the fix Louise!
LGTM

louiseschmidtgen:
@HomayoonAlimohammadi The PR was only triggering PR and nightly tests not the weekly once. I think these just slipped past us.

Issue #2565, opened 2026-05-19
Juju snap upgrades of k8s does not drain nodes before upgrading k8s application


sombrafam:
### Summary

During the upgrade of a Juju-managed Canonical K8s from 1.34 to 1.35, we observed that the process does not drain the node before rolling out the upgrade. 

### What Should Happen Instead?

Considering the potential issues during an upgrade, it would be beneficial to handle this more gracefully by draining workloads before a software update.

It seems that this was standard practice in Charmed K8s.

### Reproduction Steps

I went through an performed an upgrade from 1.34/stable to 1.35/stable on a two node cluster.

I added a test pod beforehand:

```
sudo k8s kubectl run --image nginx:latest test -- sleep 1000000
sudo k8s kubectl get po -o wide
NAME   READY   STATUS    RESTARTS   AGE   IP           NODE                        NOMINATED NODE   READINESS GATES
test   1/1     Running   0          20s   10.1.1.133   juju-55cddd-k8s-upgrade-4   <none>           <none>
```

After the upgrade the pod remained on the same node and there did not appear to be any event indicating a drain of the node:

```
ubuntu@juju-55cddd-k8s-upgrade-4:~$ sudo k8s kubectl get events --field-selector involvedObject.kind=Node | grep drain
ubuntu@juju-55cddd-k8s-upgrade-4:~$
ubuntu@juju-55cddd-k8s-upgrade-4:~$ sudo k8s kubectl get events --all-namespaces --field-selector reason=Evicted
No resources found
NAME   READY   STATUS    RESTARTS   AGE   IP           NODE                        NOMINATED NODE   READINESS GATES
test   1/1     Running   0          52m   10.1.1.133   juju-55cddd-k8s-upgrade-4   <none>           <none>
```

### System information

Canonical k8s 1.34

PR #2665, opened 2026-07-27
docs: Adds phased upgrade for Kubernetes Workers

sombrafam:
Change-Id: I5166f19876f44f23449840afee8e8f045fa89c34

## Description

Currently, the charm-based `k8s` upgrade process refreshes the snap on each node without draining workloads first. This means the snap refresh restarts all Kubernetes services immediately, making the node unavailable to the scheduler with no warning to running pods — which is unacceptable for clusters running sensitive workloads that need graceful pod rescheduling.

## Solution

Until we have the charm to automatically do the draining. We need to document our work around for the process.
Added a new "Kubernetes workers' phased upgrade with node draining" section to the charm upgrade how-to guide, documenting a step-by-step procedure to safely upgrade worker nodes one at a time:

1. **Set the target channel** – clarifies that the `k8s` snap channel versioning differs from the charm channel versioning, and adds a table mapping Kubernetes version → charm channel → snap channel.
2. **Run pre-upgrade-check** – with a caution callout that this must be run *before* `juju refresh`, otherwise units are set to failed.
3. **Drain and upgrade worker nodes one at a time** – cordon the node, drain workloads gracefully, refresh only the snap on that unit (holding future refreshes), wait for the node to be `Ready`, then uncordon and confirm pods reschedule before moving to the next node.
4. **Run `juju refresh`** – updates only the charm code afterward, since snaps are already at the target revision on every unit.
5. **Verify** – confirm all units are `active/idle` with the correct Kubernetes version, and recommends running cluster validation.

Also includes a note that the same phased process applies to control plane nodes (which don't need draining, unless they double as worker nodes).

## Issue

https://github.com/canonical/k8s-snap/issues/2565

## Backport

Not applicable - documentation-only change

## Checklist

- [x] PR title formatted as `type: title` → `docs: Adds phased upgrade for Kubernetes Workers`
- [x] Covered by unit tests → N/A (documentation only)
- [x] Covered by integration tests → N/A (documentation only)
- [x] Documentation updated ✓ (this PR *is* the documentation change)
- [x] CLA signed
- [x] Backport label added if necessary → Not needed
- [x] Confirm whether the `release-notes` label should be kept or removed → Remove (documentation-only, no user-facing product change)

No code, unit, or integration tests apply since this is a documentation-only change.

### DIFF
--- a/docs/canonicalk8s/charm/howto/upgrade.md
+++ b/docs/canonicalk8s/charm/howto/upgrade.md
@@ -278,6 +278,108 @@ version of **Kubernetes** is listed in the application's **Version**.
 It is recommended that you run a [cluster validation][cluster-validation]
 to ensure that the cluster is fully functional.
 
+## Kubernetes workers' phased upgrade with node draining
+
+By default, the charm upgrades the `k8s` snap on each node without draining
+workloads first. The snap refresh stops and restarts all Kubernetes services
+immediately, making the node temporarily unavailable to the scheduler with no
+warning to running pods.
+
+For clusters running sensitive workloads, a phased approach allows you to
+cordon and drain each node before the snap is refreshed, giving pods time to
+reschedule gracefully.
+
+```{caution}
+You must run `pre-upgrade-check` **before** issuing `juju refresh`. Without it,
+the charm will set every unit to failed with `"Unit was upgraded without a
+pre-upgrade-check"`.
+```
+
+```{note}
+The phased process below can also be applied to Kubernetes control plane nodes,
+except that they don't need to be drained. If your control plane nodes
+also run as worker nodes, they will need to be drained.
+```
+
+### 1. Set the target channel
+
+The `k8s` snap channel versioning *is not* the same as the charm channel
+versions. Instead, it follows the following versioning scheme:
+
+| Kubernetes Version | Charm Channel | Snap Channel |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


nhennigan (commented):
Hi @sombrafam, sorry about the delayed response. Thank you for your contribution! After discussing this with the team, we cannot recommend this approach as we have not tested it and it goes against the recommended pattern for upgrades. If you would like to see a change to the way we do upgrades, please submit a PF ticket. 

Your PR did however highlight very important gaps in the guide. The charm/explanation/upgrades.md page explains how the charm orchestrates the snap and juju refresh handles the snap upgrade, but we do not link to it in this how-to guide or mention this information anywhere on the page. You also brought up a good point about cordoning and draining the node. Every deployment is different, so we cannot provide a blanket method for doing this. We should however make the user aware that the High Availability promise is for k8s cluster components only. Workloads must be handled independently. Finally, I like how you included the snap/charm naming conventions. 

I think there is valuable information to be added to this guide from your input but I understand it would require a large rewrite. If you would like to continue working on this, great! Otherwise I can take the points I mentioned above and add this page to our backlog for future improvements.

sombrafam:
Hi @nhennigan,

Thanks for the detailed feedback. I have already created a PF-6704, so please take that in consideration in the next planning road maps.

Erlon

PR #2666, opened 2026-08-02
[release-1.34] Update component versions

bschimke95:
[release-1.34] Update component versions


bschimke95:
test_build's FIPS panic-message assertion fix kept getting wiped from this branch by the daily component-version cron (which force-recreates autoupdate/sync/release-1.34 from release-1.34's tip on every run, discarding any commits added independently). Landed the fix separately on release-1.34 directly instead: #2676. Once that merges, this branch will pick it up automatically on its next cron cycle.

PR #2669, opened 2026-08-02
[release-1.35] Update component versions

bschimke95:
[release-1.35] Update component versions


bschimke95:
Superseded by #2671, which fixes the root cause (the `KUBERNETES_VERSION_MARKER` in `update-component-versions.py` was left pointing at the generic `stable.txt` instead of `stable-1.35.txt` when the branch was cut). That PR includes the correct same-minor Kubernetes bump (v1.35.3 -> v1.35.7). This PR can no longer produce a valid diff and is being closed.

PR #2690, opened 2026-08-11
fix: [1.32] trigger feature controllers when setting annotations

HomayoonAlimohammadi:
This PR triggers feature controllers when setting annotations.


louiseschmidtgen (approved):
lgtm

PR #2694, opened 2026-08-12
docs: add charm cluster annotations reference page

louiseschmidtgen:
## What

Documents how to set cluster annotations via the `k8s` charm's `cluster-annotations` config option.

This was originally a new `charm/reference/annotations.md` page, but the content is procedural (step-by-step setting instructions), not a reference catalog — so it's now folded into the existing **how-to guide** `charm/howto/configure-cluster.md` as a new "Set cluster annotations" section instead.

## Content

- How to set flat `key=value` annotations via `juju config k8s cluster-annotations=...`
- Cross-references the [snap annotations reference](https://github.com/canonical/k8s-snap/blob/main/docs/canonicalk8s/snap/reference/annotations.md) for the full catalog of supported keys, rather than duplicating it
- Documents the workaround for nested annotations (e.g. multi-peer BGP) that the flat charm config can't express, via `juju exec ... sudo k8s set annotations=...`

## Depends on

This assumes **canonical/k8s-operator#977** is merged, which restores the `cluster-annotations` wiring in `assemble_cluster_config()` (currently a no-op on `main` — a regression from #508). Opening as **draft** until that PR merges.

## Testing

Built locally with `make html` in `docs/canonicalk8s` — page builds clean with no new warnings, and both cross-reference links resolve correctly in the generated output.


docs/canonicalk8s/charm/howto/configure-cluster.md
```diff
@@ -38,7 +38,28 @@ The charm's configuration options include:
...
+
+```{note}
+v1alpha annotations are experimental and subject to change or removal in
+future {{product}} releases.
+```
+
+The `cluster-annotations` config option sets the same
+[cluster annotations][snap-annotations] used by the {{product}} snap. It
+accepts a space-separated list of `key=value` pairs, and is only applied when
+non-empty — clearing it (`juju config k8s --reset cluster-annotations`)
```
nhennigan: I am slightly confused with the wording here. If we run juju config k8s --reset cluster-annotations, what does this actually do? From the sounds of this sentence it does nothing?

docs/canonicalk8s/charm/howto/configure-cluster.md
nhennigan: This may be out of scope of your work but this is quite confusing way of wording each option. We use --config to both pass command line config options or a config file so naming this the --config flag option is unclear

nhennigan: This does not render well on RTD. 

Image
I was expecting a command under the "using a YAML file:" . So many : is distracting

nhennigan:
```suggestion

```

nhennigan:
```suggestion

```

nhennigan:
```suggestion

```
Same here. We can remove these lines

nhennigan:
```suggestion

```

nhennigan:
```suggestion
Alternatively, specify the configuration options directly:
```

nhennigan:
```suggestion
## Monitor and verify the configuration
```

nhennigan:
```suggestion
## Charm configuration
```

nhennigan:
```suggestion
including networking, DNS, labels, taints, and more. Review the CharmHub
```

nhennigan: If we are pointing out caveats with the annotations in its own heading section, I believe we can do the same here for bootstrap- configs. See the suggestion below

docs/canonicalk8s/charm/howto/configure-cluster.md
```diff
@@ -38,7 +38,28 @@ The charm's configuration options include:
   ingress):
   - An enable/disable flag (e.g., `dns-enabled`)
   - Feature specific configuration options (e.g., `dns-cluster-domain`)
-- **Cluster wide configurations** (e.g., labels, taints).
+- **Cluster wide configurations** (e.g., labels, taints, and annotations).
+
```
nhennigan, line 35: ### Bootstrap configuration 

When setting up your cluster for the first time, you will have access to
certain configuration settings (prefixed with `bootstrap-`) that cannot be
changed later. Be sure to check the charm's documentation beforehand to
understand the available configuration options.

docs/canonicalk8s/charm/howto/configure-cluster.md
```diff
@@ -38,7 +38,28 @@ The charm's configuration options include:
...
+future {{product}} releases.
+```
+
+The `cluster-annotations` config option sets the same
+[cluster annotations][snap-annotations] used by the {{product}} snap. It
+accepts a space-separated list of `key=value` pairs, and is only applied when
+non-empty — clearing it (`juju config k8s --reset cluster-annotations`)
+leaves any existing cluster annotations untouched rather than deleting them.
+
+Some annotations, such as the [multi-peer BGP annotation][multi-peer-bgp],
```
nhennigan: Are the BGP annotations the only one that need nested YAML? If that is the case I think it would be better to be explicit. 

"The [multi-peer BGP annotation][multi-peer-bgp], requires nested YAML that `cluster-annotations` cannot express. We cannot use `juju deploy` or `juju config`. Set these by
running the `k8s` CLI on a unit instead:..."


Question - can we set multi-peer BGP from deployment?

Or tell the user where to find the nested annotations.

"Some annotations, such as the [multi-peer BGP annotation][multi-peer-bgp], require nested YAML that `cluster-annotations` cannot express. See the [snap annotations reference page](snap-annotations) to understand the formatting of all annotations. Set these nested YAMLs annotations by
running the `k8s` CLI on a unit instead:"

nhennigan (commented):
Thanks Louise 😸 Most of my comments are to do with existing page formatting issues. There are a few clarifications though needed.

docs/canonicalk8s/charm/howto/configure-cluster.md
canonicalmateo: We can reflect the actual behavior as I suggested here: https://github.com/canonical/k8s-operator/pull/977#discussion_r3768110937

louiseschmidtgen: I tried this: juju config k8s --reset cluster-annotations and it nukes all annotations. So I removed that sentence in favour of Mateo's key=- suggestion.

docs/canonicalk8s/charm/howto/configure-cluster.md
```diff
@@ -17,28 +17,49 @@ This guide assumes the following:
...
+- **Cluster wide configurations** (e.g., labels, taints, and annotations).
+
+### Bootstrap configuration
+
 When setting up your cluster for the first time, you will have access to
 certain configuration settings (prefixed with `bootstrap-`) that cannot be
 changed later. Be sure to check the charm's documentation beforehand to
 understand the available configuration options.
+
+## Cluster annotations
```
nhennigan:
```suggestion
### Cluster annotations
```

docs/canonicalk8s/charm/howto/configure-cluster.md
```diff
@@ -17,28 +17,49 @@ This guide assumes the following:
...
+```{note}
+v1alpha annotations are experimental and subject to change or removal in
+future {{product}} releases.
 ```
 
-The charm's configuration options include:
+The `cluster-annotations` config option sets the same
+[cluster annotations][snap-annotations] used by the {{product}} snap. It
+accepts a space-separated list of `key=value` pairs, and is only applied when
+non-empty — clearing it via `key=-` removes an existing cluster annotation.
```
nhennigan: This highlights an important point - we do not provide guidance on how to remove these annotations. This is probably out of scope, and I can do this next pulse, but on /snap/reference/annotations/ we should add an example of how to remove an annotation.
```
sudo k8s set annotations="k8sd/v1alpha1/cilium/tunnel-port=-"
```
Now back to this page, am I correct when I say this is an e.g of how we set cluster annotations (aside from BGP) at juju level 
```
juju config k8s cluster-annotations="k8sd/v1alpha1/cilium/cni-exclusive=true k8sd/v1alpha1/cilium/sctp/enabled=true"
```

If that is the case, I was able to piece that together from our docs. 

For removal, am I correct with:
```
juju config k8s cluster-annotations="k8sd/v1alpha1/cilium/cni-exclusive=- k8sd/v1alpha1/cilium/sctp/enabled=-"
```

Finally, is cluster annotations the only config that is removed this way? 

I believe we need a section at the end how to remove a config with the reset flag. What do you think? This way we are highlighting that annotations are the only one that need special attention. 

## Remove the configuration 

If you would like to return to the default configuration:

```
juju config k8s --reset dns-enabled,dns-cluster-domain
```

docs/canonicalk8s/charm/howto/configure-cluster.md
louiseschmidtgen: Done in ab603df4 — demoted to `### Cluster annotations` so it nests under `## Charm configuration` alongside `### Bootstrap configuration`.

docs/canonicalk8s/charm/howto/configure-cluster.md
louiseschmidtgen: Great questions — all addressed in ab603df4:

1. **Set example** — yes, exactly right. Added it verbatim as the concrete example in the Cluster annotations section.
2. **Removal example** — also correct: `key=-` is the deletion marker (aligned with k8sd's merge semantics in canonical/k8s-operator#977). Documented with `cni-exclusive=-`.
3. **Is it the only config removed this way?** — yes. Every other option is a scalar/string that `--reset` returns to its default. Annotations are special because the charm *merges* them into the cluster's existing annotation map (preserving out-of-band ones set via the k8s CLI), so individual keys need the explicit `-` marker.

Added the `## Remove the configuration` section at the end as you suggested, with the `--reset dns-enabled,dns-cluster-domain` example and a callout that annotations are the exception, linking back to the annotations section.

Agreed the snap annotations reference removal example (`sudo k8s set annotations="k8sd/v1alpha1/cilium/tunnel-port=-"`) is a follow-up — happy to leave that to your next pulse.

docs/canonicalk8s/charm/howto/configure-cluster.md
louiseschmidtgen: Answered both questions in ab603df4:

- **Is BGP the only nested-YAML annotation?** Yes — `k8sd/v1alpha1/metallb/bgp-peers` is the only one; every other annotation in the snap reference is a flat `key=value`. The text now says this explicitly.
- **Can multi-peer BGP be set from deployment?** No — deploy-time config goes through the same `cluster-annotations` option, which can't express nested YAML, so the k8s CLI on a unit is the only route. The paragraph now states this.

Also fixed the grammar in that paragraph (stray comma, `require` → `requires`, `can not` → `cannot`).

nhennigan (approved):
Thanks for being so patient Louise. This looks great 😸

PR #2741, opened 2026-08-19
feat: prep 1.36 release

HomayoonAlimohammadi:
This PR prepares the 1.36 snap release. Note that the charm is not touched right now and is supposed to happen in a later stage. The 1.36 branch should be created after this PR is merged.


.github/workflows/lint_and_integration.yaml
```diff
@@ -113,7 +113,8 @@ jobs:
           EVENT_NAME: ${{ github.event_name }}
           BASE_REF: ${{ github.base_ref }}
         run: |
-          tags="pull_request"
+          # TODO: revert back to "pull_request"
+          tags="up_to_weekly"
```
HomayoonAlimohammadi: revert before merge

docs/canonicalk8s/releases/snap/upgrading.md
```diff
@@ -6,6 +6,72 @@ myst:
 
 # Upgrade notes
 
+## Upgrade 1.35 to 1.36
+
+### Before you upgrade: the k8s-dqlite datastore is removed
```
nhennigan:
```suggestion
### Datastore upgrade considerations
```
We have Upgrade notes as H1 then Upgrade 1.35 -> 1.36 as H2 so I think we should alter it slightly here in H3

docs/canonicalk8s/releases/snap/1.36.md
```diff
@@ -0,0 +1,131 @@
...
+several methods. For specific requirements, see the [Installation guides].
+
+## What's new
+
+- **Kubernetes 1.36** - read more about the upstream release [here].
+
+- **kube-proxy replacement** - Clusters bootstrapped with the default
+  Cilium network no longer run `kube-proxy`. Service routing is handled by
+  Cilium's eBPF datapath instead. Clusters upgraded from earlier releases
+  keep running `kube-proxy` until you opt in with
```
nhennigan:
```suggestion
  will continue to run `kube-proxy` until you opt in with
```

docs/canonicalk8s/releases/snap/1.36.md
```diff
@@ -0,0 +1,131 @@
...
+
+{{product}} can be installed on a variety of operating systems using
+several methods. For specific requirements, see the [Installation guides].
+
+## What's new
+
+- **Kubernetes 1.36** - read more about the upstream release [here].
+
+- **kube-proxy replacement** - Clusters bootstrapped with the default
+  Cilium network no longer run `kube-proxy`. Service routing is handled by
```
nhennigan:
```suggestion
  Cilium network no longer deploy `kube-proxy`. Service routing is handled by
```

docs/canonicalk8s/releases/snap/1.36.md
```diff
@@ -0,0 +1,131 @@
...
+  Cilium's eBPF datapath instead. Clusters upgraded from earlier releases
+  keep running `kube-proxy` until you opt in with
+  `sudo k8s set network.kube-proxy-enabled=false`. See the
+  [service routing](/snap/explanation/networking.md#service-routing-kube-proxy-replacement)
+  explanation for more information.
+
+- **k8s-dqlite datastore removed** - The `k8s-dqlite` datastore, deprecated in
+  1.35, has been removed. Only `etcd` and external datastores are supported.
```
nhennigan: The way this is phrased makes it seem we support external datastores - we support disabling the default datastore and the user taking on the responsibility of running their own 
```suggestion
  1.35, has been removed. The only supported datastore is now `etcd` but users still have the option to deploy an external datastore.
```

docs/canonicalk8s/releases/snap/1.36.md
```diff
@@ -0,0 +1,131 @@
...
+# 1.36
+
+**{{product}} 1.36 LTS - Release notes - 24 August 2026**
+
+Welcome to the second [LTS release] of {{product}}!
+
+## Requirements and compatibility
+
+{{product}} can be installed on a variety of operating systems using
+several methods. For specific requirements, see the [Installation guides].
```
nhennigan:
```suggestion
several methods. For specific requirements, see the [installation guides].
```

docs/canonicalk8s/releases/snap/1.36.md
```diff
@@ -0,0 +1,131 @@
...
+  migration path.
+
+- **Cilium 1.20 and networking updates**
+  - Cilium and Cilium-Operator upgraded to 1.20.0.
+  - Gateway API CRDs upgraded from v1.2.0 to v1.6.1.
+  - The bundled `ck-gateway` GatewayClass now uses
+    `gateway.networking.k8s.io/v1`.
+
+- **MetalLB 0.16.1 and multi-peer BGP** - MetalLB controller and speaker
+  upgraded to v0.16.1. BGP mode can now be configured with multiple peers, each
```
nhennigan:
```suggestion
  is now upgraded to v0.16.1. BGP mode can now be configured with multiple peers, each
```

docs/canonicalk8s/releases/snap/1.36.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Deprecations and API changes
+
+- Upstream - Please review the
+  [upstream release notes][upstream-changelog-1.36], which include deprecation
+  notices and API changes for Kubernetes 1.36.
+- The `k8s-dqlite` datastore option has been removed. There is no update or
+  migration path for deployments using this datastore.
+- `kube-proxy-client-crt`, `kube-proxy-client-key` and
+  `extra-node-kube-proxy-args` are deprecated. They are accepted for backward
```
nhennigan: I am a bit confused here. By they are accepted but have no affect do you mean that when you set them they don't throw an error but they do nothing?

docs/canonicalk8s/releases/snap/upgrading.md
```diff
@@ -6,6 +6,72 @@ myst:
...
+
+```
+snap info k8s
+sudo k8s status --wait-ready
+```
+
+### Optional: switch to the Cilium kube-proxy replacement
+
+Clusters bootstrapped on 1.36 with the default Cilium network do not run
+`kube-proxy`; Cilium handles service routing in eBPF instead. Upgraded clusters
```
nhennigan:
```suggestion
`kube-proxy`. Instead, Cilium handles service routing in eBPF. Upgraded clusters
```

docs/canonicalk8s/releases/snap/upgrading.md
```diff
@@ -6,6 +6,72 @@ myst:
...
+keep `kube-proxy` running so that the refresh does not change the datapath.
```
nhennigan: This could be a gap in my understanding, why would the datapath be changed on refresh?

docs/canonicalk8s/releases/snap/upgrading.md
```diff
@@ -6,6 +6,72 @@ myst:
...
+sudo snap services k8s.kube-proxy
+```
+
+```{note}
+This changes how service traffic is routed on every node. Roll it out during a
+maintenance window and validate your workloads afterwards.
+```
+
+```{warning}
+The switch is one-way while the network feature is enabled: setting
```
nhennigan:
```suggestion
Disabling `kube-proxy` on 1.36 is not reversible while the network feature is enabled. If you try to set
```

docs/canonicalk8s/releases/snap/upgrading.md
```diff
@@ -6,6 +6,72 @@ myst:
...
+`network.kube-proxy-enabled` back to `true` is rejected as long as the default
```
nhennigan:
```suggestion
`network.kube-proxy-enabled` back to `true`, it will be rejected as long as the default
```

nhennigan (commented):
Thanks Hue. Its great that we can link to so many of the docs to support the release notes. I think the spread tests are failing as there is no 1.36 snap they can pull.

.github/workflows/lint_and_integration.yaml
HomayoonAlimohammadi: all the tests passed with `up_to_weekly`. I'm reverting it back to `pull_request`. 
https://github.com/canonical/k8s-snap/actions/runs/32243983516

docs/canonicalk8s/releases/snap/1.36.md
HomayoonAlimohammadi: Yes, but only under the default CNI. Per `worker_join_config.md` and `control_plane_join_config.md`, these fields are "accepted for backward compatibility and used when the default CNI (Cilium) is not in use". So: no error on set; ignored while Cilium is the network provider (no kube-proxy exists to consume them); still honoured when the default network is disabled and kube-proxy runs.

docs/canonicalk8s/releases/snap/upgrading.md
HomayoonAlimohammadi: This might be bad wording from me. The datapath here means the mechanism to route and handle the packets. With kube-proxy, it's handled via iptables (or IPVS), but Cilium uses eBPF programs.
That sentences basically means "for an upgraded cluster we're keeping kube-proxy enabled to not mess with your networking unnecessarily". I tried to rephrase the sentences to make it clearer.

nhennigan (approved):
Thanks Hue. I think that is much clearer how the kube proxy upgrade will be handled. I am happy from the docs perspective.

docs/canonicalk8s/releases/snap/1.36.md
```diff
@@ -0,0 +1,133 @@
...
+**{{product}} 1.36 LTS - Release notes - 24 August 2026**
+
+Welcome to the second [LTS release] of {{product}}!
+
+## Requirements and compatibility
+
+{{product}} can be installed on a variety of operating systems using
+several methods. For specific requirements, see the [installation guides].
+
+## What's new
```
louiseschmidtgen, line 18: I wonder if it makes sense to have a what's new for the LTS diff and the 1.35 -> 1.36 diff. Wdyt?

.github/workflows/nightly-test.yaml
```diff
@@ -1,13 +1,14 @@
 name: Nightly
 
+# Deliberately not triggered by pull_request. Every job here targets a
```
louiseschmidtgen, line 3: I'm a bit hesitant about this change. We should be testing the nightly's if we change the relevant workflow files. Wdyt?

docs/canonicalk8s/releases/snap/1.36.md
```diff
@@ -0,0 +1,133 @@
...
+  [multi-peer BGP guide](/snap/howto/networking/multi-peer-bgp.md).
+
+- **s390x and ppc64el support** - The snap is now built for the `s390x` and
+  `ppc64el` architectures in addition to `amd64` and `arm64`.
+
+- **Dedicated containerd base directory flag** - `k8s bootstrap` and
+  `k8s join-cluster` accept a `--containerd-base-dir` flag, so a dedicated
+  containerd installation path no longer requires a configuration file.
+
+- **k8sd moved to its own repository** - The `k8sd` daemon is now developed in
```
louiseschmidtgen, line 53: We need to explain what k8sd is and does to a user.

docs/canonicalk8s/releases/snap/1.36.md
```diff
@@ -0,0 +1,133 @@
...
+- Honor `AnnotationDisableSeparateFeatureUpgrades` during node joins
+  ([#2622], [k8sd#53])
+
+## Upgrade notes
+
+See our [upgrade notes] page for instructions on how to upgrade to 1.36.
+
+```{note}
+Clusters using the removed `k8s-dqlite` datastore cannot be upgraded to 1.36.
+The snap refresh is rejected and the cluster stays on its current revision.
```
louiseschmidtgen, line 110: Should we add that there is no migration path and you need to redeploy your cluster entirely? or point to the upgrading docs?

HomayoonAlimohammadi:
I've converted this to draft to decide on keeping Cilium 1.20 vs moving back to 1.19.

louiseschmidtgen (commented):
Very nice work Hue! A few notes/suggestions:

docs/canonicalk8s/releases/snap/1.36.md
HomayoonAlimohammadi: you're right. now that I think about it, since it's an implementation detail and doesn't effect the user experience, maybe we should just remove this part?

.github/workflows/nightly-test.yaml
HomayoonAlimohammadi: the problem that I'm trying to workaround here is that since I've touched both the nightly and weekly test files, an ungodly amount of tests get triggered on this PR, for each and every release (1.32 --> 1.35). We should fix this behaviour in the future.

docs/canonicalk8s/releases/snap/1.36.md
HomayoonAlimohammadi: Yea makes sense. So a 1.32 vs 1.36 change log?
Maybe a question for @nhennigan

docs/canonicalk8s/releases/snap/1.36.md
HomayoonAlimohammadi: I agree either way. Probably a question for @nhennigan

HomayoonAlimohammadi:
we've decided to merge this and address the docs comments on a follow up PR.

PR #2739, opened 2026-08-18
ci: extend integration test matrix to ubuntu 26.04

louiseschmidtgen:
## Summary

Adds Ubuntu 26.04 (Resolute) to the integration test matrices so the 1.36 release is validated on the new LTS.

## Changes

- `nightly-test.yaml`: add `ubuntu:26.04` to the nightly integration and performance test OS matrices
- `weekly-test.yaml`: add `ubuntu:26.04` to the weekly CNCF conformance OS matrix

The FIPS test matrix (`test-on-fips`) is intentionally unchanged — no 26.04 FIPS runners are available yet.

## Acceptance criteria

- [x] Test matrix extended to cover ubuntu:26.04
- [x] FIPS 26.04 excluded (no runners available)

PR #2747, opened 2026-08-21
fix: update cri-containerd AppArmor profile for Ubuntu 26.04

louiseschmidtgen:
## Problem

Adding `ubuntu:26.04` to the CNCF conformance test matrix (#2739) causes all conformance tests to fail with this pattern:

```
kubectl exec execpodgfntx -- /bin/sh -x -c nslookup nodeport-service...
rc: 139
```

## Root cause

The chain is:

1. **Ubuntu 26.04 ships AppArmor 5.x**, which tightens signal mediation for processes in user namespaces (a.k.a. containers).
2. When `kubectl exec` spawns a shell inside a pod, AppArmor 5.x labels those processes `cri-containerd.apparmor.d//&unconfined` (the `//&unconfined` suffix is new in AppArmor 5.x for user-namespace-isolated processes).
3. **BIND 9.18** (bundled in `agnhost:2.53`, used by CNCF tests) introduces an `isc-net-0000` I/O thread that uses POSIX signals for inter-thread coordination.
4. The `cri-containerd.apparmor.d` profile only had `signal (receive)` rules for snap services — **no send rules and no rules for `//&unconfined` peers**.
5. AppArmor denies `SIGTERM`/`SIGABRT` from `isc-net-0000` → `nslookup` crashes with `SIGSEGV` (exit 139).

Evidence from `dmesg` on the failing CI host:
```
apparmor="DENIED" operation="signal" profile="cri-containerd.apparmor.d"
comm="isc-net-0000" requested_mask="send" denied_mask="send" signal=term
peer="cri-containerd.apparmor.d//&unconfined"
```
(82× SIGABRT denied, 79× SIGTERM denied — one per test retry)

## Fix

Two changes:

### 1. `k8s/profiles/containerd`

Add a self-referential signal rule that allows all container processes to signal each other:

```
signal (send, receive) peer=cri-containerd.apparmor.d,
```

This is compatible with AppArmor 3.x (Ubuntu 22.04) and later.

### 2. `k8s/wrappers/services/containerd`

When loading the profile at containerd startup, first try an extended version with the AppArmor 5.x `//&unconfined` wildcard rule:

```
signal (send, receive) peer="cri-containerd.apparmor.d//**",
```

If `apparmor_parser` rejects the `//` namespace syntax (AppArmor 3.x/4.x), fall back to the base profile. This ensures backward compatibility with Ubuntu 22.04 and 24.04.

## Testing

Profile with both rules parses and loads on Ubuntu 26.04 (AppArmor 5.0.2):
```
apparmor_parser -r /tmp/test-updated-profile.d
# → PROFILE WITH ALL RULES: OK
```

Related: #2739

### DIFF
--- a/.github/workflows/lint_and_integration.yaml
+++ b/.github/workflows/lint_and_integration.yaml
@@ -147,6 +147,16 @@ jobs:
       test-collection-branch: ${{ github.head_ref }}
       parallel: true
 
+  test-cncf-ubuntu-2604:
+    name: CNCF conformance (ubuntu:26.04)
+    needs: [build-snap, python-lint]
+    uses: ./.github/workflows/e2e-tests.yaml
+    with:
+      arch: amd64
+      os: ubuntu:26.04
+      test-tags: conformance_tests
+      artifact: k8s-amd64.snap
+
   security-scan:
     name: Security scan
     needs: build-snap
--- a/k8s/profiles/containerd
+++ b/k8s/profiles/containerd
@@ -40,4 +40,7 @@ profile cri-containerd.apparmor.d flags=(attach_disconnected,mediate_deleted) {
   signal (receive) peer=snap.k8s.k8sd,
   signal (receive) peer=snap.k8s.k8s,
   signal (receive) peer=snap.k8s.hook.remove,
+
+  # Allow signals between container processes (e.g. ISC BIND threads in nslookup).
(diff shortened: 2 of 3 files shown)


louiseschmidtgen:
replaced by https://github.com/canonical/k8s-snap/pull/2750

PR #2750, opened 2026-08-24
fix: harden cri-containerd AppArmor profile for kernel 6.17+ stacking

louiseschmidtgen:
## Problem

On kernel 6.17+ with AppArmor profile stacking enabled, processes created via CRI ExecSync (`kubectl exec`, exec-based liveness/readiness probes) run under a **stacked** peer label `cri-containerd.apparmor.d//&unconfined` instead of the plain `cri-containerd.apparmor.d`. The existing AppArmor profile had no rules that matched this stacked variant.

This causes two classes of denial observed in dmesg:

```
apparmor="DENIED" operation="signal" profile="cri-containerd.apparmor.d"
comm="isc-net-0000" requested_mask="send" denied_mask="send"
peer="cri-containerd.apparmor.d//&unconfined"

apparmor="DENIED" operation="ptrace" profile="cri-containerd.apparmor.d"
peer="cri-containerd.apparmor.d//&unconfined"
```

**Signal denial:** ISC BIND 9.18+ introduces an `isc-net-0000` I/O thread that uses POSIX signals for inter-thread coordination. Without a signal-send rule, AppArmor denies these signals, causing `nslookup` inside containers to crash with SIGSEGV (exit 139). This breaks CNCF conformance ExternalName-service tests on Ubuntu 26.04.

**Ptrace denial:** The existing `ptrace` rule only matched the plain profile name, leaving exec'd processes on stacking kernels unmatched.

Upstream reference: https://github.com/containerd/containerd/issues/12886

## Fix

Two changes to `k8s/profiles/containerd`:

### 1. Signal rules
Add self-referential signal rules covering both the plain profile name and the `//&*` stacked variants:
```
signal (send, receive) peer=cri-containerd.apparmor.d,
signal (send, receive) peer=cri-containerd.apparmor.d//&*,
```

### 2. Ptrace rules
Expand the existing ptrace rule to include `tracedby`/`readby` modes and add the `//&*` stacked variant:
```
ptrace (trace,tracedby,read,readby) peer=cri-containerd.apparmor.d,
ptrace (trace,tracedby,read,readby) peer=cri-containerd.apparmor.d//&*,
```

The `//&*` glob is a no-op on older kernels (Ubuntu 22.04/24.04) where AppArmor stacking is inactive — AppArmor 3.x/4.x can parse the syntax but will never match it. No runtime wrapper changes or fallback logic are needed.

## Related
- https://github.com/containerd/containerd/issues/12886
- #2739

### DIFF
--- a/k8s/profiles/containerd
+++ b/k8s/profiles/containerd
@@ -32,12 +32,26 @@ profile cri-containerd.apparmor.d flags=(attach_disconnected,mediate_deleted) {
   deny /sys/kernel/security/** rwklx,
 
 
-  # suppress ptrace denials when using 'docker ps' or using 'ps' inside a container
-  ptrace (trace,read) peer=cri-containerd.apparmor.d,
+  # Allow ptrace between container processes (e.g. 'docker ps', 'ps' inside a container).
+  # On kernel 6.17+ with AppArmor profile stacking enabled, exec'd processes run under
+  # the stacked peer label "cri-containerd.apparmor.d//&unconfined" rather than the plain
+  # profile name; the //&* glob matches all such stacked variants.
+  # See: https://github.com/containerd/containerd/issues/12886
+  ptrace (trace,tracedby,read,readby) peer=cri-containerd.apparmor.d,
+  ptrace (trace,tracedby,read,readby) peer=cri-containerd.apparmor.d//&*,
 
   signal (receive) peer=snap.k8s.containerd,
   signal (receive) peer=snap.k8s.kubelet,
   signal (receive) peer=snap.k8s.k8sd,
   signal (receive) peer=snap.k8s.k8s,
   signal (receive) peer=snap.k8s.hook.remove,
+
+  # Allow signals between container processes (e.g. ISC BIND isc-net-0000 threads in
+  # nslookup). On kernel 6.17+ with AppArmor profile stacking, exec'd processes run under
+  # the stacked peer label "cri-containerd.apparmor.d//&unconfined"; without a send rule
+  # AppArmor denies inter-thread signals, causing nslookup to crash with SIGSEGV (rc:139).
+  # The //&* glob matches all stacked variants; it is a no-op on older kernels where
+  # stacking is inactive. See: https://github.com/containerd/containerd/issues/12886
+  signal (send, receive) peer=cri-containerd.apparmor.d,
+  signal (send, receive) peer=cri-containerd.apparmor.d//&*,
 }
--- a/k8s/profiles/containerd
+++ b/k8s/profiles/containerd
@@ -32,12 +32,8 @@ profile cri-containerd.apparmor.d flags=(attach_disconnected,mediate_deleted) {
   deny /sys/kernel/security/** rwklx,
 
 
-  # Allow ptrace between container processes (e.g. 'docker ps', 'ps' inside a container).
-  # Both trace+read (tracer side) and tracedby+readby (tracee side) are required: AppArmor
-  # checks both the tracer's and the tracee's profile, so omitting either half silently
-  # denies the operation. On kernel 6.17+ with AppArmor stacking, exec'd processes get
-  # a stacked peer label (e.g. "cri-containerd.apparmor.d//&unconfined"); //&* covers
-  # all stacked variants. See: https://github.com/containerd/containerd/issues/12886
+  # suppress ptrace denials when using 'docker ps' or using 'ps' inside a container
+  # //&* covers stacked peer labels on kernel 6.17+ (https://github.com/containerd/containerd/issues/12886)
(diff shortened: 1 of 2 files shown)


k8s/profiles/containerd
```diff
@@ -34,10 +34,21 @@ profile cri-containerd.apparmor.d flags=(attach_disconnected,mediate_deleted) {
...
   signal (receive) peer=snap.k8s.hook.remove,
+
+  # Allow signals between container processes (e.g. ISC BIND isc-net-0000 threads in
+  # nslookup). No self-referential signal rule existed before; without a send rule,
+  # inter-thread signals are denied and nslookup crashes with SIGSEGV (rc:139).
+  # On kernel 6.17+ with AppArmor stacking, exec'd processes get a stacked peer label
+  # (e.g. "cri-containerd.apparmor.d//&unconfined"); //&* covers all stacked variants.
+  # See: https://github.com/containerd/containerd/issues/12886
+  signal (send, receive) peer=cri-containerd.apparmor.d,
+  signal (send, receive) peer=cri-containerd.apparmor.d//&*,
```
louiseschmidtgen: This line covers stacking kernels (26.04+, kernel 6.17+) as per https://github.com/containerd/containerd/issues/12886.

HomayoonAlimohammadi:
> **WARNING: this is an AI generated review.**

### 1. The ptrace revert (ecf386fa) makes the ptrace fix a no-op — `k8s/profiles/containerd:36-38`

AppArmor cross-checks ptrace: the tracer needs `trace`/`read` against the tracee's label, **and the tracee needs `tracedby`/`readby` against the tracer's label** (`apparmor.d.pod`: "the profile of the tracing process and the profile of the target task must both have the correct permissions"; the kernel re-checks the tracee with `request << PTRACE_PERM_SHIFT`). Both sides here are the same profile, so `ptrace (trace,read) peer=cri-containerd.apparmor.d[//&*]` can never permit a same-profile ptrace once the profile mediates the ptrace class — which it does, since a ptrace rule is present.

That is why upstream containerd's `contrib/apparmor/template.go` carries `ptrace (trace,tracedby,read,readby) peer={{.Name}},`. As committed, the `operation="ptrace" peer="cri-containerd.apparmor.d//&unconfined"` denials quoted in the description will still fire. `tracedby,readby` isn't scope creep — it's what makes the rule functional, and it matches upstream. Please restore it on both ptrace lines.

### 2. Signal rules look correct — `:52-53`

All four sender/target combinations (plain↔plain, plain↔stacked, stacked↔stacked) are covered, and it matches upstream's `signal (send,receive) peer={{.Name}}`. The pre-existing `signal (receive) peer=snap.k8s.*` rules are what enable signal mediation in the first place, so the "no self-referential rule existed" note is accurate.

Verified the "no-op on older releases" claim: the profile parses clean under `apparmor_parser` 3.0.4 (22.04) and 5.0.2 (26.04), and the 24.04 integration jobs here are green.

### 3. The CNCF job costs 2h15m per PR and isn't proven to catch this — `lint_and_integration.yaml:150-158`

**It may guard nothing.** The `ubuntu-24.04` runner image already runs kernel `6.17.0-1022-azure`, and the existing `Integration (amd64)` jobs run on that same kernel with a 24.04 guest — the only delta is the guest's AppArmor userspace. Nothing in this run shows it would have failed before the fix. Please run it once against a pre-fix snap and confirm it goes red; that should gate whether the job lands at all.

**It collects tests from `main`, not the PR.** Unlike `test-integration`, it omits `test-collection-repo` / `test-collection-branch`, so `e2e-tests.yaml`'s `prepare` resolves `REF="main"` and both jobs check out `canonical/k8s-snap@main`. The PR's own conformance test and `test_util` harness changes are never exercised, and on a `release-1.3x` PR it would run main's tests against a release snap. Mirror the two lines from `test-integration`.

Even if the pre-fix run is red, I'd gate it rather than run it unconditionally: a path filter on `k8s/profiles/**`, or add `"ubuntu:26.04"` to the weekly `cncf` matrix plus 26.04 to the per-PR integration matrix.

### 4. Stale description

The PR body (and c57c6992's message) still advertise `ptrace (trace,tracedby,read,readby)`, which ecf386fa reverted. If the perms are restored the body is already correct; otherwise it needs updating so a squash-merge doesn't land a misleading message.

### Nits

- `signal (send, receive)` uses a space after the comma, unlike the rest of the file.
- `//&*` won't match multi-level stacks or path-named stack components (`**` would), but upstream uses `*` — probably not worth diverging.
- Worth backporting to `release-1.32`…`release-1.36`; users on stable channels with 26.04 hosts hit this today.

bschimke95 (approved):

bschimke95:
Make sure to remove the CI job before merging this.

PR #2772, opened 2026-08-28
docs: add more references to containerd-base-dir flag

ethandcosta:
This PR adds a couple references to the CLI flag added in https://github.com/canonical/k8sd/pull/51 to other areas in the documentation that reference containerd or the join/bootstrap command.

### DIFF
--- a/docs/canonicalk8s/snap/howto/install/offline.md
+++ b/docs/canonicalk8s/snap/howto/install/offline.md
@@ -146,10 +146,6 @@ file on each node and set the appropriate `http_proxy`, `https_proxy` and
 `no_proxy` variables as described in the
 [adding proxy configuration section][proxy].
 
-```{note}
-The path to the HTTP proxy config file will change if you [changed the containerd base directory][how-to-containerd].
-```
-
 ````
 
 ````{tab-item} Private registry mirrors
@@ -277,7 +273,9 @@ instructions on all nodes. For each upstream registry that needs mirroring,
 create a `hosts.toml` file. Here's an example that configures
 `http://10.10.10.10:5050` as a mirror for `ghcr.io`:
 
-
+```{note}
+The paths below assume the default containerd base directory (`/etc/containerd`). If you [changed the containerd base directory][how-to-containerd], replace `/etc/containerd` with `<containerd-base-dir>/etc/containerd`.
+```
 
 #### HTTP registry mirror
 
--- a/docs/canonicalk8s/snap/howto/install/offline.md
+++ b/docs/canonicalk8s/snap/howto/install/offline.md
@@ -146,6 +146,10 @@ file on each node and set the appropriate `http_proxy`, `https_proxy` and
 `no_proxy` variables as described in the
 [adding proxy configuration section][proxy].
 
+```{note}
+The path to the HTTP proxy config file will change if you [changed the containerd base directory][how-to-containerd].
+```
+
 ````
 
 ````{tab-item} Private registry mirrors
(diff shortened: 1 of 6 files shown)


docs/canonicalk8s/snap/howto/install/offline.md
```diff
@@ -146,6 +146,10 @@ file on each node and set the appropriate `http_proxy`, `https_proxy` and
 `no_proxy` variables as described in the
 [adding proxy configuration section][proxy].
 
+```{note}
```
nhennigan: Is this right? I thought proxy config files were at /etc/systemd/system/snap.k8s.containerd.service.d/http-proxy.conf which is created manually for the k8s snap and are separate to the containerd config files?

docs/canonicalk8s/snap/howto/troubleshooting.md
```diff
@@ -279,17 +279,9 @@ that uses containerd.
 
 ````{dropdown} Solution
 
-We recommend running {{product}} in an isolated environment, for this purpose,
-you can create a LXD VM for your installation. See
-[Install {{product}} in LXD][lxd-install] for instructions.
+{{product}} is built on the assumption that it is deployed in an isolated environment, as is usually the case for production builds. A typical way of isolating {{product}} is to run it on an LXD VM. See [Install {{product}} in LXD][lxd-install] for instructions.
```
nhennigan: I don't like the phrasing that we are assuming users' setup. After all, it can be deployed in many ways and we even provide a guide to install in dev envs. The k8s snap is suitable to everyone, not just production builds. I would suggest switching back to the "we recommend" wording 


```suggestion
We recommend {{product}} is deployed in an isolated environment. A typical way of isolating {{product}} is to run it on an LXD VM. See [Install {{product}} in LXD][lxd-install] for instructions.
```

docs/canonicalk8s/snap/howto/troubleshooting.md
```diff
@@ -279,17 +279,9 @@ that uses containerd.
...
 
-As an alternative, you may specify a custom containerd path like so:
-
-```bash
-cat <<EOF | sudo k8s bootstrap --file -
-containerd-base-dir: $containerdBaseDir
-EOF
-```
+If the cluster is being deployed in a developer environment or cannot have the conflicting containerd bin removed, the flag `--containerd-base-dir` can be added to the `bootstrap` and `join-cluster` commands to specify an alternate absolute path for {{product}} to install containerd. See [How to resolve containerd conflicts][how-to-containerd] for further information.
```
nhennigan:
```suggestion
If the cluster is being deployed in an environment that cannot have the conflicting containerd bin removed, the flag `--containerd-base-dir` can be added to the `bootstrap` and `join-cluster` commands to specify an alternate absolute path for {{product}} to install containerd. See [How to resolve containerd conflicts][how-to-containerd] for further information.
```

nhennigan (commented):
Thanks Ethan. I believe the spread tests are failing due to the 1.36 snap track not existing which is out of this PR's scope. 

Will this need to be backported to release-1.36?

docs/canonicalk8s/snap/howto/install/offline.md
ethandcosta: You're right, I was mistaken here. This file is managed by systemd to pass into the containerd service, so its not affected by the directory change. I moved the note to the `private registry mirrors` section, which does have paths that would be affected by this change

docs/canonicalk8s/snap/howto/troubleshooting.md
ethandcosta: I was under the impression that we made design choices with the snap that assume by default you're running it in a production environment, which is why we had this containerd conflict issue as we didn't have a consideration for a user who already had a containerd installation from a different service. I get what you're saying, though, so I will implement your suggestion.

ethandcosta:
Hey Niamh, thanks for the review! I've implemented your suggestions. The spread tests will probably pass as soon as 1.36-candidate gets promoted and we will need to get this backported into 1.36

nhennigan (approved):
Thanks Ethan 😸

PR #2778, opened 2026-08-31
test: validate --containerd kubelet arg migration across upgrade and downgrade

louiseschmidtgen:
## Summary

The `--containerd` kubelet flag was removed in Kubernetes 1.37 (see canonical/k8sd#76).
Nodes **upgrading from 1.36** have this flag persisted in `$SNAP_COMMON/args/kubelet`.
Kubernetes 1.37 kubelet rejects the flag, preventing the node from coming back up after upgrade.
The inverse is also true on **downgrade**: after a 1.36→1.37 upgrade, downgrading back to 1.36
leaves `--containerd` absent, so 1.36 kubelet cannot locate the containerd socket.

The correct owner of this logic is **k8sd**, which already manages the args file via
`UpdateServiceArguments` in its `hooks_post_refresh.go`. This PR is the test vehicle.

## Changes

- **`build-scripts/components/k8sd/version`**: Pin k8sd to the `fix/remove-containerd-kubelet-flag`
  branch (canonical/k8sd#76), which will be updated to include post-refresh removal (upgrade path)
  and a 1.36 patch release will add post-refresh restoration (downgrade path). Revert to `main`
  once the k8sd branch is merged.
- **`tests/integration/tests/test_version_upgrades.py`**: New NIGHTLY test
  `test_kubelet_containerd_arg_migration` that validates both transitions end-to-end:
  - **Upgrade** (1.36/stable → local snap): `--containerd` must be absent after k8sd 1.37
    post-refresh removes it
  - **Downgrade** (local snap → 1.36/stable): `--containerd` must be restored after k8sd 1.36
    post-refresh re-adds it

## k8sd changes required (tracked separately)

**k8sd 1.37** (`fix/remove-containerd-kubelet-flag` branch, to add to existing PR):
```go
// hooks_post_refresh.go
if err := snaputil.UpdateServiceArguments(s.snap, "kubelet", nil, []string{"--containerd"}); err != nil {
    return fmt.Errorf("failed to remove deprecated --containerd kubelet flag: %w", err)
}
```

**k8sd 1.36 patch** (new PR against 1.36 branch):
```go
// hooks_post_refresh.go
containerdSocket := s.snap.ContainerdSocketPath()
if err := snaputil.UpdateServiceArguments(s.snap, "kubelet", map[string]string{"--containerd": containerdSocket}, nil); err != nil {
    return fmt.Errorf("failed to restore --containerd kubelet flag on downgrade: %w", err)
}
```

## Test Plan

`test_kubelet_containerd_arg_migration` covers the full round-trip:
1. Bootstrap on 1.36/stable → assert `--containerd` is present
2. Upgrade to local snap (1.37) → assert `--containerd` is absent, cluster healthy
3. Downgrade to 1.36/stable → assert `--containerd` is restored, cluster healthy


build-scripts/components/k8sd/version
```diff
@@ -1 +1 @@
-main
+fix/remove-containerd-kubelet-flag
```
louiseschmidtgen: TODO: revert

louiseschmidtgen:
**k8sd pin updated** (`fix/remove-containerd-kubelet-flag` → `agents/k8sd-1.37-release-combined`)

The previous pin had `--containerd` removal (#76) but not the chart updates from k8sd#72 (Cilium 1.20, CoreDNS 1.47, etc.). A 1.37 snap without those chart references would fail to deploy healthy clusters. The combined branch has both.

**Note:** the migration test in this PR (`test_kubelet_containerd_arg_migration`) uses `GH_BASE_REF` to determine `previous_track`. It can only run (not skip) when the base branch is `release-1.37`. That branch will be auto-created when k8s-snap#2764 merges. At that point this PR should be retargeted to `release-1.37` and the k8sd pin reverted to `main` (which will have both k8sd#72 and k8sd#76 merged by then).

louiseschmidtgen:
Only used for testing.

PR #2764, opened 2026-08-27
feat: prep 1.37 release

louiseschmidtgen:
## Summary

Prepares the 1.37 snap release.

### Component versions

| Component | Before | After |
|-----------|--------|-------|
| Kubernetes | v1.36.3 | v1.37.0 |
| Cilium (chart) | 1.19.4 | 1.20.0 |
| CoreDNS (chart) | 1.39.2 | 1.47.0 |
| Gateway API (chart) | 1.4.1 | 1.6.1 |
| Metrics Server (chart) | 3.12.2 | 3.14.0 |
| containerd | v2.3.3 | follow-up PR for v2.4 |
| etcd | v3.6.13 | v3.7.1 |
| runc | v1.4.3 | follow-up PR for v1.5.1 |
| pebble | v1.11.0 | v1.32.1 |

Helm v4.2.4 and CNI v1.9.1 are unchanged.

### Etcd upgrade

According to docs we need to ensure "[b]efore upgrading to 3.7, make sure that all of your 3.6 members are updated to 3.6.11 or later. Earlier 3.6 patch releases may not be compatible with a rolling upgrade to 3.7."
We are on etcd v3.6.13 which fulfills this requirement. [Etcd v3.7 docs](https://etcd.io/docs/v3.7/upgrades/upgrade_3_7/).

-> the weekly's tes_snap_services fails since we downgrade from etcd 3.7 -> previous channel (1.35 release) using v3.6.6. This fails because only 3.6.11+ migrations are possible.

#### New Rocks
This snap requires [k8sd#72](https://github.com/canonical/k8sd/pull/72) to be merged before the snap is functional at runtime. That PR updates chart references and image tags to match this snap:

- Cilium: `cilium-1.20.0.tgz` + images `cilium:1.20.0-ck0`, `cilium-operator-generic:1.20.0-ck0`
- CoreDNS: `coredns-1.47.0.tgz` + image `coredns:1.14.7-ck0`
- Gateway API: `gateway-api-1.6.1.tgz`
- Metrics Server: `metrics-server-3.14.0.tgz`
- LocalPV CSI sidecars: `rawfile-localpv:0.8.3-ck4`, `csi-node-driver-registrar:2.17.0-ck0`, `csi-provisioner:6.3.0-ck0`, `csi-resizer:2.2.1-ck0`, `csi-snapshotter:8.6.0-ck0`
[...]

PR #2795, opened 2026-09-03
fix: revert etcd to v3.6.13 to fix upgrades/downgrades

bschimke95:
Reverts etcd bump from PR #2764. v3.7.1 breaks downgrade tests (3.6 cannot read 3.7 schema) and inplace upgrades (quorum loss).


bschimke95:
Filed the real root cause and fix location: canonical/k8sd#89.

TL;DR: this downgrade path *is* fully supported by etcd (3.6.13 is well above the 3.6.11+ floor for a 3.7→3.6 rolling downgrade) - but only via etcd's own `etcdctl downgrade validate`/`downgrade enable` protocol, which migrates the storage schema down *before* any member's binary is swapped. k8sd never invokes this (confirmed via source search - zero references anywhere), so a plain `snap refresh` to an older etcd leaves the data stamped at the newer schema version, and the older etcd binary correctly refuses to start against it. Verified with a live 3-node repro on the dev VM.

Closing this PR without reverting etcd was the right call - the real fix belongs in k8sd's refresh orchestration, not the version pin.

PR #2797, opened 2026-09-04
test: add tests verifying anonymous node names only allowed for worker nodes

ethandcosta:
## Description
An untested feature of the k8s-snap that surfaced in an issue with CAPI (https://github.com/canonical/k8sd/pull/62) is that when a CP node tries to join the cluster without a name, it is rejected, but when a worker node attempts to do the same, it is accepted as its name is set to the node hostname.

## Solution

add tests verifying this behavior.

### DIFF
--- a/tests/integration/tests/test_clustering.py
+++ b/tests/integration/tests/test_clustering.py
@@ -231,6 +231,73 @@ def test_worker_nodes(instances: List[harness.Instance]):
     ], f"only {cluster_node.id} should be left in cluster"
 
 
+@pytest.mark.node_count(2)
+@pytest.mark.tags(tags.PULL_REQUEST)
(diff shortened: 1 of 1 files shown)


ethandcosta:
Rerunning all jobs to account for k8sd rebase (might fix some issues)

build-scripts/components/k8sd/version
```diff
@@ -1 +1 @@
-main
+KU-6173/fix-anonymous-token-join
```
HomayoonAlimohammadi: make sure to rename this.

HomayoonAlimohammadi (approved):
LGTM, thanks a lot!

ethandcosta:
CI failures are unrelated to this PR and addressed in other PRs.

Issue #2646, opened 2026-07-21
CoreDNS deployment keep restarting indefinitely


kpiyush17:
### Summary

Hi team, 

We are seeing issues where the coredns deployment keeps restarting after a couple of minutes. It is a two-node cluster with 1 control and 1 worker node, running k8s version 1.32. On checking the logs for snap services, we found the logs below in the k8sd service. 

k8sd logs:
```
Jul 21 09:08:22 bm-az3-37 k8s.k8sd[326580]: I0721 09:08:22.694203  326580 dnsrebalancer/reconcile.go:72] "Successfully triggered CoreDNS deployment restart" logger="k8sd.controller-coordinator" controller="node" controllerGroup="" controllerKind="Node" Node="bm-az3-37" namespace="" name="bm-az3-37" reconcileID="43ba7d24-54d3-43b5-a24d-c592cc409800" node="bm-az3-37"
Jul 21 09:08:55 bm-az3-37 k8s.k8sd[326580]: I0721 09:08:55.022796  326580 controllers/service_args.go:69] "checking service arguments for drift" logger="k8sd" controller="service-args"
Jul 21 09:10:20 bm-az3-37 k8s.k8sd[326580]: I0721 09:10:20.540957  326580 dnsrebalancer/reconcile.go:60] "CoreDNS pods need rebalancing, triggering deployment rollout restart" logger="k8sd.controller-coordinator" controller="node" controllerGroup="" controllerKind="Node" Node="bm-az2-83" namespace="" name="bm-az2-83" reconcileID="2f9b3cfa-8b6f-4dfe-86f9-aa70281a56ec" node="bm-az2-83"
Jul 21 09:10:20 bm-az3-37 k8s.k8sd[326580]: I0721 09:10:20.562627  326580 dnsrebalancer/reconcile.go:72] "Successfully triggered CoreDNS deployment restart" logger="k8sd.controller-coordinator" controller="node" controllerGroup="" controllerKind="Node" Node="bm-az2-83" namespace="" name="bm-az2-83" reconcileID="2f9b3cfa-8b6f-4dfe-86f9-aa70281a56ec" node="bm-az2-83"
[...]

PR #2811, opened 2026-09-09
test(dns): add integration test for coredns topology spread rebalancing (#2646)

Greemty:
## Description

This PR addresses issue #2646 where CoreDNS pods fall into an infinite rollout restart loop on single-node clusters, during initial bootstrap, or on late worker joins.

## Solution & Architectural Nuance

The root cause was in the `dnsrebalancer` controller (`k8sd`), not in the CoreDNS scheduling constraints — the CoreDNS TopologySpreadConstraint and pod anti-affinity are **unchanged** from `main`, both remain soft (`ScheduleAnyway` / preferred).

1. **Why we did not make the hostname constraint `DoNotSchedule`**
   - This was considered and rejected. Making `kubernetes.io/hostname` strict caps CoreDNS replicas at one per node, which blocks the HPA from scaling beyond node count and leaves single-node clusters with a permanently `Pending` second replica (`hpa.minReplicas: 2`).
   - It also risks blocking rescheduling during drain/cordon under `maxSkew: 1`, since a drained node counts as 0 pods and a hard constraint can prevent the remaining node from taking a second replica.
   - The zone constraint (`topology.kubernetes.io/zone`) stays `ScheduleAnyway` as before — the label isn't guaranteed to exist on bare-metal/LXD/on-prem clusters, so `DoNotSchedule` there would block scheduling entirely.

2. **Why the `dnsrebalancer` is still needed**
   - Soft constraints don't automatically redistribute pods when topology changes dynamically (e.g., a cluster starts with 1 node → both CoreDNS replicas land on it → a 2nd node joins later, or a cluster runs steady-state and both replicas end up co-located after a rollout).
   - The actual bug was that `dnsrebalancer` reacted to CoreDNS co-location by calling `RestartDeployment` on every reconcile, with no cooldown and a predicate that re-fired on routine kubelet heartbeats — not just real Ready transitions. A full Deployment restart recreates both replicas at once, so the scheduler could place both on the same node again, repeating the loop indefinitely.
   - The fix: the event filter now only reacts to real Node transitions (Ready flip, `Unschedulable` flip, taint changes, deletion — covering downscale too), `Reconcile` skips while a rollout is in progress or CoreDNS pods aren't yet settled (30s), and rebalancing now deletes a single co-located pod instead of restarting the whole Deployment — the surviving pod stays visible to the scheduler so the soft spread constraint has a real signal to act on. `RequeueAfter 30s` keeps retrying at a bounded cadence instead of going idle or looping instantly.

*Reference specification details can be found [here](https://docs.google.com/document/d/1VAaLFzaaDyw4eQ7qQ740VIhMqaLY8fSQRLdsBEPvtps/edit?tab=t.0).*

## Issue

- Resolves/Relates to [#2646](https://github.com/canonical/k8s-snap/issues/2646)

## Backport

- Not needed right now per team discussion (will be handled after merging into `main`).

## Checklist

- [x] PR title formatted as `type: title` (`test(dns): add integration test for coredns rebalancing (#2646)`)
- [ ] Covered by unit tests (not applicable here — controller unit tests live in `k8sd`)
- [ ] Documentation updated
- [x] CLA signed
- [x] Confirm whether the `release-notes` label should be kept or removed

### DIFF
--- a/build-scripts/components/k8sd/version
+++ b/build-scripts/components/k8sd/version
@@ -1,1 +1,1 @@
-main
+KU-6443/fix-coredns-topology-spread
--- a/tests/integration/tests/test_dns.py
+++ b/tests/integration/tests/test_dns.py
@@ -1,7 +1,9 @@
 #
 # Copyright 2026 Canonical, Ltd.
 #
+import json
 import logging
+import time
 from typing import List
 
 import pytest
@@ -84,14 +86,23 @@ def test_dns(instances: List[harness.Instance]):
 @pytest.mark.node_count(2)
 @pytest.mark.tags(tags.PULL_REQUEST)
 def test_dns_ha_rebalancing(instances: List[harness.Instance]):
+    """
+    Verify that CoreDNS replicas end up spread after a late join, and that
+    dnsrebalancer does not storm (it deletes one co-located pod rather than
+    restarting the whole Deployment).
+    """
     initial_node = instances[0]
     joining_cplane_node = instances[1]
 
-    # Wait for initial cluster to be ready
     util.wait_until_k8s_ready(initial_node, [initial_node])
     util.wait_for_dns(initial_node)
 
-    # Verify initial state: all CoreDNS pods should be on the first node
+    rebalance_log_since = initial_node.exec(
+        ["date", "--iso-8601=seconds"],
+        text=True,
+        capture_output=True,
+    ).stdout.strip()
+
     result = initial_node.exec(
         [
             "k8s",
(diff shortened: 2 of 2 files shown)


Greemty:
The integration test now verifies the resulting contract: at most one rollout restart after a late join, no further restart after the rollout settles, and healthy DNS afterward. It records final pod placement for diagnostics but does not require spreading, since that would contradict the current soft scheduling policy.

build-scripts/components/k8sd/version
```diff
@@ -1 +1 @@
-main
+KU-6443/fix-coredns-topology-spread
```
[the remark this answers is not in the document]
louiseschmidtgen: We will revert this before merging.

tests/integration/tests/test_version_upgrades.py
```diff
@@ -467,6 +507,367 @@ def test_feature_upgrades_inplace(instances: List[harness.Instance], tmp_path: P
...
+
+    def trigger_counts():
+        return [
+            node.exec(
+                ["journalctl", "-u", "snap.k8s.k8sd", "--since", since, "--no-pager"],
+                capture_output=True,
+                text=True,
+                timeout=20,
+            ).stdout.count("CoreDNS pods need rebalancing")
+            for node in instances
```
louiseschmidtgen, line 775: This trigger count is duplicated in test_dns_ha_rebalancing. Either move _check_coredns_hpa_scaling into test_dns_ha_rebalancing or find a way to share one of the definitions please

tests/integration/tests/test_version_upgrades.py
```diff
@@ -467,6 +507,367 @@ def test_feature_upgrades_inplace(instances: List[harness.Instance], tmp_path: P
...
+                    "image": "ghcr.io/containerd/busybox:1.28",
+                    "command": [
+                        "sh",
+                        "-c",
+                        "until timeout -t 3 nslookup kubernetes.default.svc.cluster.local "
+                        ">/tmp/dns-lookup.log 2>&1; do sleep 2; done; "
+                        "touch /tmp/ready; "
+                        "while true; do "
+                        "if timeout -t 3 nslookup kubernetes.default.svc.cluster.local >/tmp/dns-lookup.log 2>&1; "
+                        "then echo OK $(date +%s); else echo FAIL $(date +%s); exit 1; fi; sleep 2; done",
```
louiseschmidtgen, line 542:
```suggestion
                        "then echo OK $(date +%s); else echo FAIL $(date +%s); fi; sleep 2; done",
```
With `restartPolicy: Never`, that exit 1 puts the pod in phase Failed permanently on the first failed lookup. It isn't buying stickiness: kubectl logs retains the FAIL line on a still-running pod, and _assert_dns_upgrade_probe already asserts every line starts with OK , so a single FAIL is caught either way. Keeping the pod Running also keeps the <15s freshness check meaningful — right now a dead probe and a healthy-but-stalled one are indistinguishable.

tests/integration/tests/test_dns.py
```diff
@@ -84,73 +86,149 @@ def test_dns(instances: List[harness.Instance]):
...
+
+    trigger_count = _rebalance_trigger_count()
+    LOG.info(f"dnsrebalancer triggered {trigger_count} time(s)")
+    assert trigger_count == 1, (
+        "Expected exactly one dnsrebalancer trigger "
+        f"(triggered {trigger_count} times)"
+    )
+
+    for _ in range(6):
+        time.sleep(10)
```
louiseschmidtgen, line 220: Nit: this sleep statement is a little ugly can we do better?

louiseschmidtgen (commented):
Nice work @Greemty,

Please update the PR description to reflect your latest changes and rebase the PR to include the latest changes from main.

Could we please add one more test case to ensure we don't break existing rebalancing behavior:

Cordon/uncordon rebalance, 2 nodes. drain for maintenance, both replicas consolidate, uncordon, and you need them to spread again. Reuses the existing test_dns_ha_rebalancing fixture: cordon node-2 → delete the pod on node-2 → assert both land on node-1 → uncordon → assert spread with a bounded trigger count. No new cluster.

tests/integration/tests/test_version_upgrades.py
Greemty: Extracted the journalctl counting logic into `util.dnsrebalancer_trigger_counts(instances, since)` in test_util/util.py, and both `test_dns.py` and `test_version_upgrades.py` now call it instead of duplicating the loop.

tests/integration/tests/test_version_upgrades.py
Greemty: fixed

tests/integration/tests/test_dns.py
Greemty: Replaced with a time.monotonic()-based deadline loop (_assert_rebalance_settled), matching the style already used in _check_coredns_hpa_scaling. It's now shared with the new cordon/uncordon test too.

Greemty:
Added `test_dns_ha_rebalancing_cordon_uncordon` reusing the same 2-node fixture pattern: cordon the second node, delete its CoreDNS pod, assert consolidation onto the remaining node, uncordon, and assert it respreads with a bounded trigger count. No extra node join.

Issue #2816, opened 2026-09-10
test: bootstrap/node-ready hangs on cilium image pull — HTTP/2 PROTOCOL_ERROR in CI registry mirror


louiseschmidtgen:
## Symptom

`tests/integration/tests/test_version_upgrades.py::test_version_downgrades_with_rollback` (and its sibling `test_version_upgrades`) fail at the **initial bootstrap readiness check** — before any downgrade/upgrade step runs — when the bootstrap channel needs to pull the `canonical/cilium` / `canonical/cilium-operator-generic` OCI images. `kubectl get node` reports `Ready=False` for ~10-15 minutes until `tenacity` gives up:

```
AssertionError
tenacity.RetryError: RetryError[<Future ... state=finished raised AssertionError>]
```

This is a **different, later-stage failure** than the kube-proxy issue tracked in #2806 (already fixed via PR #2751, backported in #2807-#2810) — it reproduces on PRs that already contain that fix.

## Root cause

The bootstrap node's own `cilium` DaemonSet pod (and, in one case, `cilium-operator`) never finishes pulling its container image because the CI test harness's **local pull-through registry mirror** (`distribution/distribution` v2.8.3, proxying `ghcr.io`, set up in `tests/integration/tests/test_util/registry.py` / `tests/integration/templates/registry/registry-config.yaml`) repeatedly aborts large blob transfers with an HTTP/2 stream error:

```
level=error msg="Error committing to storage: stream error: stream ID 27; PROTOCOL_ERROR; received from peer" ... vars.name="canonical/cilium"
level=error msg="response completed with error" ... http.response.duration=6m5.463s http.response.status=500 vars.name="canonical/cilium"
```

Each failed proxy fetch takes **5-8.5 minutes** before erroring out, forcing containerd to restart the pull from scratch. On the bootstrap (control-plane) node, this loop consumed the entire test readiness-wait budget, so the node's own `cilium-agent` pod stayed stuck in `PodInitializing`/`ContainerCreating` and `kubectl get node` never reported `Ready`.

This is compounded by two separate factors observed in the logs:

[...]

Issue #2806, opened 2026-09-09
test: test_version_downgrades_with_rollback fails on release branches — kube-proxy inactive after bootstrap on >=1.36 channel


louiseschmidtgen:
## Symptom

`tests/integration/tests/test_version_upgrades.py::test_version_downgrades_with_rollback` fails on all release-branch PRs (release-1.32 … release-1.35) with:

```
AssertionError: Service kube-proxy should be active, but it is inactive
tenacity.RetryError: RetryError[<Future at ... state=finished raised AssertionError>]
```

The assertion originates from `check_snap_services_ready` in `tests/integration/tests/test_util/util.py` (called from `wait_until_k8s_ready`).

## Root cause

The test bootstraps the cluster on the newest stable channel (currently `1.36-classic/candidate` or `1.37-classic/edge`, via `TEST_VERSION_DOWNGRADE_CHANNELS: "recent 6"` in `.github/workflows/e2e-tests.yaml`). Since 1.36, the k8s snap uses **Cilium kube-proxy replacement** (`KUBE_PROXY_REPLACEMENT_MIN_VERSION = (1, 36)`), so the `snap.k8s.kube-proxy` systemd service is **inactive by design** on the bootstrap channel.

On release-1.32 … release-1.35, `check_snap_services_ready` still **unconditionally expects `kube-proxy` to be active** on control-plane nodes. The test therefore fails at the **initial bootstrap readiness check** (`util.wait_until_k8s_ready(cp, instances)` right after `k8s bootstrap` + joins) — before any downgrade or rollback step even runs.

Evidence from the logs:

- PR #2760 (release-1.35), job [98446106614](https://github.com/canonical/k8s-snap/actions/runs/33049296251/job/98446106614): fails at `test_version_upgrades.py:231` → `wait_until_k8s_ready` → `check_snap_services_ready` (util.py:1206). The log shows `Node ready: k8s-integration-2-...` repeating for ~22 min while the kube-proxy assertion keeps failing. No `Initiating downgrade` line ever appears.
- PR #2798 (release-1.34), job [101291376404](https://github.com/canonical/k8s-snap/actions/runs/33959696627/job/101291376404): identical failure at `test_version_upgrades.py:225` (same bootstrap readiness check), bootstrap channel `1.37-classic/edge`.
[...]

PR #2817, opened 2026-09-10
test: disable HTTP/2 client on CI registry mirror to avoid GHCR proxy stalls

louiseschmidtgen:
## What

Disables the Go HTTP/2 client (`GODEBUG=http2client=0`) on the local integration-test registry mirror (`tests/integration/templates/registry/registry.service`), so proxy pulls to the upstream registry (`ghcr.io`) fall back to HTTP/1.1.

## Why

Root-caused in #2816: the CI test harness's pull-through registry mirror (`distribution/distribution` v2.8.3 proxying `ghcr.io`) intermittently fails large blob transfers -- specifically the `cilium`/`cilium-operator-generic` images (~170-340MB) -- with:

```
level=error msg="Error committing to storage: stream error: stream ID 27; PROTOCOL_ERROR; received from peer" ... vars.name="canonical/cilium"
http.response.duration=6m5.463s http.response.status=500
```

Each failed pull burns 5-8.5 minutes before erroring, forcing containerd to restart the pull from scratch. When this happens on the bootstrap node right after `k8s bootstrap`, it can exhaust the node-readiness wait budget entirely, causing `test_version_downgrades_with_rollback` (and `test_version_upgrades`) to fail with a `tenacity.RetryError` before the test even reaches its first upgrade/downgrade step.

This pattern reproduced independently across two different bootstrap channels (`1.37-classic/candidate` and `1.32-classic/stable`), confirming it's a registry-mirror/network reliability issue rather than a channel- or build-specific regression.

Disabling the HTTP/2 client for the registry mirror process is a known, low-risk mitigation for this class of GHCR proxy flakiness (large blob transfers over HTTP/2 being reset mid-stream) -- this only affects the mirror's outbound connections, not any snap/cluster runtime behavior.

## Related

- Fixes/mitigates: #2816
- Distinct from the earlier kube-proxy readiness issue (#2806, already fixed via #2751/#2807-#2810)

## Testing

This is a CI-infrastructure-only change (test harness template); should be validated by re-running the affected integration tests and confirming the registry mirror log no longer shows `PROTOCOL_ERROR`/broken-pipe entries during cilium image pulls.

### DIFF
--- a/tests/integration/templates/registry/registry.service
+++ b/tests/integration/templates/registry/registry.service
@@ -8,12 +8,6 @@
     RestartSec=5s
     LimitNOFILE=40000
     TimeoutStartSec=0
-
-    # Disable the Go HTTP/2 client so that proxy pulls of large upstream blobs
-    # (e.g. ghcr.io/canonical/cilium*) fall back to HTTP/1.1. Large blob transfers
-    # over HTTP/2 have been observed to fail intermittently with
-    # "stream error: ...; PROTOCOL_ERROR; received from peer", forcing a full
-    # multi-minute re-pull and, during bootstrap, stalling node readiness.
     Environment=GODEBUG=http2client=0
 
     ExecStart=/bin/registry serve /etc/distribution/$NAME.yaml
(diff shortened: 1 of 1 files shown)


bschimke95 (approved):

bschimke95:
Merging this with failed tests. There are a couple of independent failures that we clean up in follow-up PRs.

PR #2815, opened 2026-09-10
test: cap downgrade channel selection at the branch's own track

louiseschmidtgen:
## Problem

`test_version_downgrades_with_rollback` selects channels via `get_most_stable_channels(reverse=True)` with only a `min_release` floor and **no ceiling**. On release branches the selection floats up to the newest published track, so e.g. on release-1.35 it currently resolves to:

```
1.37-classic/candidate  (bootstrap)
1.36-classic/candidate
1.35-classic/stable
1.34/1.33/1.32-classic/stable
```

It bootstraps on **1.37** and downgrades **1.37 → 1.36 → 1.35**, skipping across the parallel in-flight 1.36/1.37 releases. The 1.37-classic/candidate snap is currently broken (bootstrap node never becomes Ready), which fails the test before any downgrade step runs.

## Fix

Pass the existing (previously unused) `max_release` parameter, set to the branch's own version, so a release branch bootstraps on its own track and downgrades through its history:

- release-1.35 → `1.35 → 1.34 → 1.33 → 1.32`
- release-1.33 → `1.33 → 1.32`
- main → unchanged (uncapped, still tests latest)

Smallest diff: one new `max_release` argument + a 3-line branch-version derivation. Verified by simulating `get_most_stable_channels` against the live store channel-map.

## Follow-up

Once this merges to main, it will be cherry-picked/backported to the release branches (1.32–1.36) so their downgrade CI tests their own track.

Part of the release-branch CI stabilization (see #2806 for the related kube-proxy readiness fix).

PR #2818, opened 2026-09-10
fix: don't skip candidate-only minor versions in upgrade/downgrade channel selection

bschimke95:
## Description

`test_version_downgrades_with_rollback` (and `test_version_upgrades`) can silently skip a minor version in the upgrade/downgrade chain, jumping the etcd (and other component) version further than a real upgrade/downgrade path ever would, which breaks the test.

## Root cause

`get_most_stable_channels()` already picks the best-available channel per minor version in its first pass (e.g. `stable` if published, otherwise `candidate`/`beta`/`edge`), but then had a second pass that **dropped any version whose best channel was riskier than `max_risk`** (default `"stable"`) entirely from the result.

The 1-2 most recent minor versions routinely sit on `candidate` for a while before promotion to `stable` - that's the normal release lifecycle, not an anomaly. Right now (verified live against the Snap Store API), **1.37 and 1.36 both only have a `candidate` channel**; 1.35 and older are `stable`.

`test_version_downgrades_with_rollback` calls `get_most_stable_channels(reverse=True)` to build its downgrade chain. With the old default, both 1.37 and 1.36 were dropped, so `channels[0]` (the bootstrap/"current" version) resolved to **1.35** instead of 1.37, and the first downgrade step went straight **1.35 → 1.34** - the adjacent 1.36 step was skipped entirely.

Note: PR #2815 (open) fixes a *different*, adjacent bug (capping the ceiling so a release branch doesn't float above its own track) - it does **not** fix this; verified locally that even with its `max_release` cap applied, 1.36 still silently disappears.

**No caller anywhere overrides `max_risk`** - it was unused configurability. The property that actually matters here is never skipping an adjacent minor version, not gating on risk tier.

## Fix

Drop the risk-based filtering step entirely. Always use whatever's the best available channel for each version (already computed in the first pass) - no exclusion by risk at all.

## Backport

Yes - this affects every release branch that runs these tests. Will add backport labels once this is reviewed, mirroring #2786/#2789's approach.

## Checklist

- [ ] Covered by unit tests (no existing unit-test scaffolding for `test_util/snap.py`'s pure logic; verified via a local script against the live Snap Store API instead - see above)
- [x] Covered by integration tests (fixes the channel selection used by `test_version_upgrades`/`test_version_downgrades_with_rollback`)
- [ ] Documentation updated
- [x] CLA signed
- [x] Confirm whether the `release-notes` label should be kept or removed (removing - test-only change, not user-facing)

### DIFF
--- a/tests/integration/tests/test_util/snap.py
+++ b/tests/integration/tests/test_util/snap.py
@@ -68,7 +68,6 @@ def get_most_stable_channels(
     include_latest: bool = True,
     min_release: Optional[str] = None,
     max_release: Optional[str] = None,
-    max_risk: str = "candidate",
     reverse: bool = False,
 ) -> List[str]:
     """Get an ascending list of latest channels based on the number of channels
@@ -99,16 +98,6 @@ def get_most_stable_channels(
             risk
         ) < RISK_LEVELS.index(channel_map[version_key][1]):
             channel_map[version_key] = (channel, risk)
-
-    # Only trim the newest contiguous versions below max_risk; never drop
-    # an interior version (e.g. don't skip a candidate-only 1.36 between
-    # 1.37 and stable 1.35).
-    if max_risk:
-        max_risk_index = RISK_LEVELS.index(max_risk)
-        for version_key in sorted(channel_map.keys(), reverse=True):
-            if RISK_LEVELS.index(channel_map[version_key][1]) <= max_risk_index:
-                break
-            del channel_map[version_key]
 
     # Sort channels by major and minor version (ascending order)
     sorted_versions = sorted(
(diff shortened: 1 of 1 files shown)


bschimke95:
Related to #2815 - that PR fixes a different bug (capping the ceiling so a branch doesn't float above its own track), but doesn't fix this one. I verified locally that even with #2815's `max_release` cap applied, this bug still silently drops 1.36 from the downgrade chain (see verification in the PR description above) - these two fixes are complementary, not overlapping.

louiseschmidtgen (approved):

bschimke95:
CI failures are unrelated - merging.

bschimke95:
Removed the `backport release-1.32/1.33/1.34/1.35` labels - verified each is not applicable: `get_most_stable_channels()` on those branches never had the `max_risk` parameter/filtering block this PR removes, so the bug never existed there (it was introduced and fixed entirely within `main`'s post-branch history). Only `release-1.36` actually needed and got this fix (#2829).

PR #2830, opened 2026-09-10
fix: retry Snap Store/dl.k8s.io requests with exponential backoff

bschimke95:
## Description

CI intermittently fails outright on transient Snap Store errors, e.g.:
```
HTTPError ({https://api.snapcraft.io/v2/snaps/info/k8s}): {429} {Too Many Requests}
```
Traced from a failed run of #2812: `test_version_upgrades`/`test_version_downgrades_with_rollback` crash on a single rate-limited request with no chance to recover.

## Root cause

Checked all raw network calls in the integration test suite (`tests/integration/tests/test_util/*.py`) - 4 total:
- `snap.get_snap_info()` - no retry at all (the one that failed in CI)
- `util.tracks_least_risk()` - no retry at all
- `util._find_stable_track()` - no retry at all
- `util._major_minor_from_stable_upstream()` - fixed-delay retry only, no backoff

## Fix

Added generic `fetch_text()`/`fetch_json()` helpers in `test_util/util.py`: exponential backoff on 429/5xx/network-level errors, other 4xx (e.g. 404) still fail fast. Switched all 4 call sites to use them instead of duplicating urllib boilerplate.

## Verification

Ran locally: a live call against the real Snap Store API, a synthetic 429-then-success case, and a synthetic 404 case (confirms it doesn't waste retries on non-transient errors).

## Backport

TBD pending review - will add labels once merged, same as recent related PRs.

## Checklist

- [x] PR title formatted as `type: title`
- [ ] Covered by unit tests (verified locally instead, see above)
- [x] Covered by integration tests (fixes flakiness in the existing suite)
- [ ] Documentation updated
- [x] CLA signed
- [ ] Backport label added if necessary
- [x] release-notes label removed (test-only change)

PR #2832, opened 2026-09-14
ci: resolve snap test channels once per workflow run to avoid Snap Store rate limiting

louiseschmidtgen:
## Description

CI frequently fails with `429 Too Many Requests` from `api.snapcraft.io`. Root cause: the `e2e-tests.yaml` reusable workflow fans out into a large parallel test matrix (`run-tests` over `needs.prepare.outputs.tests`, up to `max-parallel: 100`, and the nightly workflow multiplies that further across os/arch/channel). Every matrix job that runs a version upgrade/downgrade or strict-interface test resolves the dynamic `recent <num> [flavour]` channel specifications (`TEST_VERSION_UPGRADE_CHANNELS`, `TEST_VERSION_DOWNGRADE_CHANNELS`, `TEST_STRICT_INTERFACE_CHANNELS`) **independently**, each calling `api.snapcraft.io/v2/snaps/info/k8s` via `test_util/snap.py` (`get_most_stable_channels`/`get_channels`). Dozens of near-simultaneous identical requests stampede the Snap Store info API and get rate limited.

## Solution

Resolve the dynamic channel specifications **exactly once per workflow run** in the single (non-matrix) `prepare` job, and pass the concrete channel lists down to the matrix jobs via `needs.prepare.outputs` — the same pattern already used for the `tests`, `test-runners`, `repo` and `ref` outputs.

- New script `tests/integration/resolve_test_channels.py` reuses the existing `test_util.snap` resolution helpers (no logic duplication) and mirrors the exact call-site semantics of `test_version_upgrades.py` (upgrade + downgrade, incl. `min_release`/`max_release`/`include_latest` handling for main vs. release branches) and `test_strict_interfaces.py`. It runs with the tox `integration` venv Python that the `prepare` job already creates for test collection.
- New `Resolve test channels` step in `prepare` writes the resolved values to `$GITHUB_OUTPUT`; `run-tests` consumes them with a fallback: `${{ needs.prepare.outputs.version-upgrade-channels || 'recent 6' }}` etc.
- **Skipped when unused**: if no collected test consumes the dynamic specs (e.g. `pull_request`-tagged runs don't include the nightly upgrade/downgrade tests), the step makes zero Snap Store calls.
- **Fails open**: if resolution fails (Snap Store unreachable, script missing on an older test-collection branch, ...), the outputs stay empty and matrix jobs fall back to the previous per-job resolution — no behavior change in the worst case.

Non-`recent` specifications are passed through unchanged, and `extra-test-env` overrides still take precedence in the test step, so existing callers (`weekly-test.yaml`, `nightly-test.yaml`, `lint_and_integration.yaml`) are unaffected.

### Scope note

This fixes the verifiable stampede source: the `recent N` fan-out. `util.previous_track()`/`tracks_least_risk()` also query the Snap Store from individual tests, but those depend on the snap under test (only known inside the matrix job) and are inherently per-test — they can't be hoisted to `prepare`. They remain covered by the retry/backoff hardening in #2830.

### Relation to #2830

This PR **complements, not duplicates** #2830 (`fix: retry Snap Store/dl.k8s.io requests with exponential backoff`): #2830 is a *survival* mechanism (tolerate 429s when they happen), while this PR is the *root-cause* fix (don't stampede the API in the first place — one request per workflow run instead of one per parallel matrix job).

## Issue

N/A — CI reliability improvement.

## Backport

Worth backporting to active release branches so their CI benefits too (the workflow falls back gracefully when the script is absent from an older test-collection ref, but backporting avoids that path).

## Checklist

- [ ] Covered by unit tests — N/A: CI workflow change; the resolution script was validated locally against the live Snap Store API (main + release-branch refs, passthrough, empty specs) using the exact tox `integration` venv the `prepare` job uses
- [ ] Covered by integration tests — N/A: this *is* integration-test infrastructure; will be exercised by the next nightly/weekly run
- [ ] Documentation updated — N/A: behavior is self-documenting via workflow comments and the script docstring
- [x] CLA signed
- [ ] Confirm whether the `release-notes` label should be kept or removed — suggest removing (internal CI change, not user-facing)


louiseschmidtgen:
## CI failure analysis

The failing integration tests on this PR are **pre-existing breakage on `main`**, not caused by this PR's changes.

**Evidence:**
- The same tests fail with identical failure signatures on the `main` nightly runs (e.g. run [34791578541](https://github.com/canonical/k8s-snap/actions/runs/34791578541); nightly has been red since ~Sep 10, last green Aug 14):
  - `test_snap_services` → `assert is_node_ready` RetryError
  - `test_version_upgrades` → `AssertionError: Services have restarted more than 0 time(s): kubelet (2 restarts)`
  - `test_version_downgrades_with_rollback` → `is_node_ready` RetryError
  - `test_feature_upgrades_inplace` → RetryError "Failed to meet condition"
- These tests only run on this PR because it touches `tests/` (tag selection bumps to `up_to_weekly`).

**Correctness of this PR verified out-of-band:** the `prepare` job's channel resolution succeeded in CI, and I reproduced it locally against the live Snap Store API — the script's output is byte-identical to what the tests' inline resolution computes on this ref:

```
TEST_VERSION_UPGRADE_CHANNELS=1.32-classic/stable 1.33-classic/stable 1.34-classic/stable 1.35-classic/stable 1.36-classic/candidate 1.37-classic/candidate latest/edge/classic
TEST_VERSION_DOWNGRADE_CHANNELS=1.37-classic/candidate 1.36-classic/candidate 1.35-classic/stable 1.34-classic/stable 1.33-classic/stable 1.32-classic/stable latest/edge/classic
TEST_STRICT_INTERFACE_CHANNELS=latest/edge/strict
```

So the matrix jobs consumed exactly the channels they would have resolved themselves; the failures are the same node-readiness/restart breakage seen on `main`.

Ready for review.

